Issue #27, opened 2019-05-17
Assign error code for error that occurs when string concatenation fails due to incompatible types.


raganhan:
For example:

```
'hi' || 1
```
Throws an error:

```
code:       null
properties: {
  LINE_NUMBER:1,
  COLUMN_NUMBER:5
}
```

Instead of `null`, the error code should probably be something like `CONCAT_FAILED_DUE_TO_INCOMPATIBLE_TYPE` or somesuch.

This can also be observed in the REPL:

```
sql> 'hi' || 1
   |
software.amazon.sql.eval.EvaluationException: Expected text: 1
	<UNKNOWN>: at line 1, column 6: <UNKNOWN>

...

ERROR! (2 ms)
sql>
```

PR #113, opened 2019-08-02
Assign error code when concatenation fails

pawelduda:
*Issue #27, if available:*

*Description of changes:*
Solves issue #27

### DIFF
--- a/lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
+++ b/lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
@@ -77,8 +77,11 @@ fun ExprValue.numberValue(): Number =
 fun ExprValue.timestampValue(): Timestamp =
     scalar.timestampValue() ?: errNoContext("Expected timestamp: $ionValue", internal = false)
 
-fun ExprValue.stringValue(): String =
-    scalar.stringValue() ?: errNoContext("Expected text: $ionValue", internal = false)
+fun ExprValue.stringValue(): String = scalar.stringValue() ?:
+    err("Expected text: $this",
+        ErrorCode.EVALUATOR_CONCAT_FAILED_DUE_TO_INCOMPATIBLE_TYPE,
+        null,
+        internal = false)
 
 fun ExprValue.bytesValue(): ByteArray =
     scalar.bytesValue() ?: errNoContext("Expected LOB: $ionValue", internal = false)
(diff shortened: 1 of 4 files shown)


pawelduda:
Feel free to give me some feedback, I am not a Kotlin person at all

lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
```diff
@@ -77,8 +77,11 @@ fun ExprValue.numberValue(): Number =
 fun ExprValue.timestampValue(): Timestamp =
     scalar.timestampValue() ?: errNoContext("Expected timestamp: $ionValue", internal = false)
 
-fun ExprValue.stringValue(): String =
-    scalar.stringValue() ?: errNoContext("Expected text: $ionValue", internal = false)
+fun ExprValue.stringValue(): String = scalar.stringValue() ?:
```
therapon: This is a general extension function that is called in multiple use cases that require to obtain a `String` from an `ExprValue`, one such case is when we need to evaluate a string append `||` operator. 

Making this extension function report an error that *always* states it is due to a string append operation might lead to miscommunication. 

I think the logic needs to go [here](https://github.com/partiql/partiql-lang-kotlin/blob/1ad2e93c1192d79029f22f687587d362eb328454/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt#L516) where the code deals with string append. Each argument to `||` needs to be checked and if it is not a `String` value then report the more specific error message that `||` did not receive a value of the appropriate type (`String`)

pawelduda: Makes sense - I'll attempt to fix it before next week.

pawelduda: @therapon Can anyone take this over? A lot going on in my life, I would love to help you out but I cannot do that. Should I just trash this PR?

pawelduda: No problem, I'm glad I could contribute something meaningful.

raganhan:
https://github.com/partiql/partiql-lang-kotlin/pull/131 continuing this work

raganhan:
https://github.com/partiql/partiql-lang-kotlin/pull/131 was pushed

Thanks @pawelduda for your contribution :)

Issue #118, opened 2019-08-05
Publish as library to maven central


bastman:
Are there any plans to publish as library to maven central?

dlurton:
The process to accomplish this is being started today.  Thanks.

raganhan:
related to https://github.com/partiql/partiql-lang-kotlin/issues/3

dlurton:
`org.partiql` has been created with sonatype.  Now in the process of preparing the `partiql-lang-kotlin` artifact.

https://issues.sonatype.org/browse/OSSRH-50768

dlurton:
PartiQL is now appearing in the [maven central search results]( https://search.maven.org/search?q=partiql).

PR #124, opened 2019-08-14
Issue #118, generate pom.xml and sign build artifacts

dlurton:
This is being done in preparation for publishing to maven central.

To ensure the ability to publish development versions of the library to the local maven repository for developers that do not have the GPG keys installed is not broken, we only attempt to sign the build artifacts when we are not building a "-SNAPSHOT" version.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,9 +25,9 @@ allprojects {
     }
 }
 
-
 subprojects {
-    version = '0.1.0'
+    group = 'org.partiql'
+    version = '0.1.0-SNAPSHOT'
 }
 
 buildDir = new File(rootProject.projectDir, "gradle-build/" + project.name)
@@ -48,7 +48,6 @@ configure(subprojects.findAll { it.name != 'example' }) {
         }
     }
 }
-
 
 subprojects {
     plugins.withId('java', { _ ->
--- a/lang/build.gradle
+++ b/lang/build.gradle
@@ -21,10 +21,11 @@ plugins {
     id 'maven-publish'
     id 'org.jetbrains.kotlin.jvm'
     id 'io.gitlab.arturbosch.detekt'
+
+    id "signing"
 }
 
-version = '0.1.0'
-group = 'org.partiql.lang'
+
 
 tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
     kotlinOptions {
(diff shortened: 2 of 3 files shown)


dlurton:
Note to self: remember to add " is not broken" to the commit message when doing the squash and merge.

lang/build.gradle
```diff
@@ -62,16 +62,70 @@ tasks.register('javadocJar', Jar) {
...
+            }
         }
     }
 }
 
+// The below enables developers and CI servers who do not have the GPG keys needed for signing to build locally,
+// while also signing release builds.
+// Signatures are only required when publishing to maven central anyway. We may need a more sophisticated means of
+// determining if we should sign the build artifacts someday, but this should suffice for now.
+if(!version.endsWith("SNAPSHOT")) {
```
raganhan: don't understand why do you need this. AFAIK signing plugin will only try to sign the maven publication. It will ask for the variables you defined in a `gradle.properties` but you can have it in the repo with dummy data, example: https://github.com/amzn/ion-schema-kotlin/blob/master/gradle.properties

lang/build.gradle
```diff
@@ -62,16 +62,70 @@ tasks.register('javadocJar', Jar) {
...
 publishing {
     publications {
-        create('maven', MavenPublication) {
-            artifactId = 'partiql-lang-kotlin'
+        maven(MavenPublication) {
+            // group id and version are set in the root build.gradle for all projects.
+            artifactId = "partiql-lang-kotlin"
+
+            from components.java
+            artifact sourcesJar
```
raganhan: Think you have to configure both this artifacts as gradle tasks, `javadocJar` specially since you need to configure `dokka` to generate javadoc based on the kdoc, example: https://github.com/amzn/ion-schema-kotlin/blob/master/build.gradle#L53

lang/build.gradle
```diff
@@ -62,16 +62,70 @@ tasks.register('javadocJar', Jar) {
...
+            artifact javadocJar
 
-            from(components['java'])
-            artifact(tasks['sourcesJar'])
-            artifact(tasks['javadocJar'])
+            pom {
+                name = "PartiQL-Lang-Kotlin"
```
raganhan: no need for `-` on the name

lang/build.gradle
```diff
@@ -62,16 +62,70 @@ tasks.register('javadocJar', Jar) {
...
+                packaging = "jar"
+                url = "https://github.com/amzn/ion-schema-kotlin"
```
raganhan: wrong url, should be https://partiql.org/

lang/build.gradle
```diff
@@ -62,16 +62,70 @@ tasks.register('javadocJar', Jar) {
...
+                licenses {
+                    license {
+                        name = "The Apache License, Version 2.0"
+                        url = "http://www.apache.org/licenses/LICENSE-2.0.txt"
+                    }
+                }
+                developers {
+                    developer {
+                        name = "Amazon Ion Team"
+                        email = "partiql-community@amazon.com"
```
raganhan: Is the e-mail correct? I don't know the correct one but wanted to double check since this one looks weird, imho it looks more like a discussion list e-mail

raganhan:
Forgot to mention on the review but if you publish to maven local you can inspect that the maven artifacts were generated correctly. This way you can manually verify the pom.xml, the sources and javadoc jars 

Won't really tell you that the pom.xml is correct but it helps

build.gradle
```diff
@@ -25,9 +25,9 @@ allprojects {
     }
 }
 
-
 subprojects {
-    version = '0.1.0'
+    group = 'org.partiql'
+    version = '0.1.0-SNAPSHOT'
```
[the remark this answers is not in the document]
dlurton: Yes.  This is just going to set up the repo for the future.

lang/build.gradle
dlurton: I don't think we should temporarily add secrets to any file where they might accidentally be committed and pushed.  I think we should use an [out-of-tree](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_configuration_properties) `gradle.properties` instead to avoid such accidents.  

Leaving a non-functional template checked into the source tree feels like a rake in the leaves--you can remember its there but you might forget and then step on it.

I discovered the need for this when I had a copy of the exact `gradle.properties` file you linked and I executed:

`./gradlew publishToMavenLocal`

And it was refusing to publish because it couldn't access the signing key.  This works around that.  I'm open to other workarounds, if there are any.  The `ion-schema-kotlin` gradle build has the same problem, BTW.

lang/build.gradle
dlurton: I verified this with @jonwilsdon but I can put any email address here that the team recommends.  I wasn't sure our usual email list was appropriate, but perhaps it is.

lang/build.gradle
```diff
@@ -21,10 +21,10 @@ plugins {
     id 'maven-publish'
     id 'org.jetbrains.kotlin.jvm'
     id 'io.gitlab.arturbosch.detekt'
+    id "signing"
```
pbcornell: the rest of this file uses single-quotes -- i'd suggest doing so here, too

lang/build.gradle
pbcornell: I'd suggest using ion-team@amazon.com here (coincides with "Amazon Ion Team").  If there's an appropriate way to reference https://community.partiql.org/, that might be a nice addition somewhere.

lang/build.gradle
raganhan: 1. Don't see why we would add secrets to the working tree, take a look at the deployment process I linked you it uses a `gradle.properties` out of the working tree. The one in the working tree is there only as a placeholder values for gradle since replacing variables happens independent of the sign plugin.

If we ever want to publish `-SNAPSHOT` versions they would need to be signed, I don't see a reason to generate unsigned maven artifacts tbh so I'm OK with `./gradlew publishToMavenLocal` not working unless you have all things in place for the sign plugin to execute.

dlurton: I think having `publishToMavenLocal` working is important for anyone who wants to test their changes to PartiQL in other projects.  I will look around to see if there's another way avoid signing when publishing to the local repository.

raganhan: Your local m2 shouldn't really be part of your development loop, is too cumbersome. Both maven and gradle let you depend directly on a project or jar in your file system which is much more practical than having to build the jar and push to local m2.

I also think there is value in normalizing all our release process so we don't need to keep specific documentation. Both ion-schema and ion-java-path-extraction use gradle.properties with dummy values

lang/build.gradle
dlurton: will be removed

lang/build.gradle
dlurton: oops--removed in next commit.

lang/build.gradle
dlurton: sure--next commit.

lang/build.gradle
dlurton: Next commit will have `dokka` and the `javadocJar` tasks configured.  I inspected the resulting jars which appear fine AFAIK.

PR #128, opened 2019-09-24
Test script runner fix

dlurton:
While preparing to delete the old testframework I noticed that the new `:pts` project wasn't failing tests that should have been failing.

I traced the problem to the `PtsRunner` class, which was identifying test failures correctly but was incorrectly notifying the `RunNotifier` instance.  It was calling `fireTestStarted` for every test but only calling `fireTestFinished` for tests that passed and only calling `fireTestFailed` instead.  (The [RunNotifier javadoc](http://junit.sourceforge.net/javadoc/org/junit/runner/notification/RunNotifier.html#fireTestFinished(org.junit.runner.Description)) states that it should always be called because runners expect `fireTestStarted` and `fireTestFinished` calls to occur in pairs.

This problem manifested itself in the command-line Gradle build (and therefore CI environments) as well as in the IntelliJ JUnit test runner.  Luckily, once I fixed, none of the tests in `master` which were executed by `PtsRunner` were failing so this problem was caught before it caused a serious issue.

This change will cause the `fireTestFinished` method to *always* be executed regardless of whether or not the test has passed.  

### DIFF
--- a/testscript/src/org/partiql/testscript/junitRunner/PtsRunner.kt
+++ b/testscript/src/org/partiql/testscript/junitRunner/PtsRunner.kt
@@ -43,35 +43,41 @@ class PtsRunner(private val testClass: Class<*>) : Runner() {
             .toList()
 
     private val ast = parser.parse(ionDocs)
-    private val testExpressions = compiler.compile(ast) 
+    private val testExpressions = compiler.compile(ast)
 
-    override fun run(notifier: RunNotifier) = try {
-        val testResults = evaluator.evaluate(testExpressions)
-
-        testResults.forEach {
-            val testDescription = Description.createTestDescription(
-                    testClass, 
+    override fun run(notifier: RunNotifier) {
+        testExpressions.forEach { testExp ->
+            val testResults = evaluator.evaluate(listOf(testExp))
+            testResults.forEach {
+                val testDescription = Description.createTestDescription(
+                    testClass,
                     "${it.test.scriptLocation} ${it.test.id}")
 
-            notifier.fireTestStarted(testDescription)
+                notifier.fireTestStarted(testDescription)
+                try {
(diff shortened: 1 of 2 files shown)


testscript/src/org/partiql/testscript/junitRunner/PtsRunner.kt
```diff
@@ -43,35 +43,41 @@ class PtsRunner(private val testClass: Class<*>) : Runner() {
...
-        val testResults = evaluator.evaluate(testExpressions)
-
-        testResults.forEach {
-            val testDescription = Description.createTestDescription(
-                    testClass, 
+    private val testExpressions = compiler.compile(ast)
+
+    override fun run(notifier: RunNotifier) {
+        testExpressions.forEach { testExp ->
+            val testResults = evaluator.evaluate(listOf(testExp))
```
raganhan: no reason to evaluate them 1 by 1, might as well run the evaluator for all `testExpressions`

testscript/src/org/partiql/testscript/junitRunner/PtsRunner.kt
```diff
@@ -43,35 +43,41 @@ class PtsRunner(private val testClass: Class<*>) : Runner() {
...
-                        |Test failed: ${it.test.scriptLocation} - ${it.test.id} 
-                        |Expected: ${it.test.expectedMessage()}
-                        |  Actual: ${it.actualResult} 
-                    """.trimMargin()
-
-                    notifier.fireTestFailure(Failure(testDescription, AssertionError(errorMessage)))
+                notifier.fireTestStarted(testDescription)
+                try {
+
+                    when (it) {
```
raganhan: replace by `if(it is TestFailure) { ... }`

PR #132, opened 2019-10-09
Adds a root type for all AST nodes

raganhan:
The root type makes AST nodes `Iterable` and adds a `children` value making the AST easily navigable as a tree.


lang/src/org/partiql/lang/ast/ast.kt
```diff
@@ -18,12 +18,33 @@ import com.amazon.ion.*
 import org.partiql.lang.util.*
 import java.util.*
 
+/**
+ * Base type for all AST nodes.  
+ */
+sealed class AstNode : Iterable<AstNode> {
```
dlurton: I'm wondering what's the value of having this implement `Iterable<AstNode>` *and* add a `.children` property?  It strikes me that only one is needed, and I suspect that `.children` slightly better since I may want to use `Iterable<AstNode>` for something else at a later time.

lang/src/org/partiql/lang/ast/ast.kt
```diff
@@ -18,12 +18,33 @@ import com.amazon.ion.*
...
+
+    /**
+     * returns all the children nodes.
+     */
+    abstract val children: List<AstNode>
```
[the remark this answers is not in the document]
raganhan: lists are `Iterable` I don't see the value of making this more generic since AST nodes are immutable

raganhan:
> For visiting all the nodes of all the children, did you also consider adding visitAstNode(node: AstNode) to AstVisitor?

Forgot about that, should be trivial to add. Will modify the `AstWalker` as well. 

Regarding removing the `AstWalker`. If we don't use it I'm ok marking it as deprecated. I'm in favor of having a generic portable way of traversing the tree as anyone can use it to have a baked in traversal that's convenient for some particular task. 

> Consider utilizing some of the existing tests for AstWalker/AstVisitor to more thoroughly test the traversal of AST nodes using .children().

Couldn't find these tests do mind pointing them out?

lang/src/org/partiql/lang/ast/ast.kt
raganhan: the iterator forces a traversal order, depthFirst, while children is more flexible and lets the user define the traversal order. 

Example: `children` is more suitable for finding a subtree of a specific path while the `Iterator` is better for visiting every node applying some operation, or finding all nodes of a certain type

PR #141, opened 2019-11-21
Make ExprValue formatter part of lang

raganhan:
And use it in ExprValue#toString so it generates a consistent string
representation


cli/src/org/partiql/cli/Repl.kt
```diff
@@ -150,7 +150,8 @@ internal class Repl(private val valueFactory: ExprValueFactory,
         val splitIndex = source.indexOfFirst { it == ' ' }.let {
             if (it == -1) {
                 source.length
-            } else {
+            }
```
therapon: nit: keep on same line

PR #142, opened 2019-11-21
(Eval)  Null as false for Join with On condition

therapon:
*Description of changes:*

Expression given as `ON` for `JOIN` treat Unknowns (`null`, `missing`) as `false` removing the row from JOIN's result set. 

### DIFF
--- a/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
+++ b/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
@@ -1267,7 +1267,9 @@ internal class EvaluatingCompiler(
                             .filter { (bindEnv: (Environment) -> Environment, _) ->
                                 // make sure we operate with lexical scoping
                                 val filterEnv = bindEnv(rootEnv).flipToLocals()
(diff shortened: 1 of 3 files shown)


lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -1267,7 +1267,9 @@ internal class EvaluatingCompiler(
                             .filter { (bindEnv: (Environment) -> Environment, _) ->
                                 // make sure we operate with lexical scoping
                                 val filterEnv = bindEnv(rootEnv).flipToLocals()
-                                filter(filterEnv).booleanValue()
+                                val filterResult = filter(filterEnv)
+                                if (filterResult.isUnknown()) { false }
```
raganhan: nit: formatting, I don't think we always format `if/else`

therapon:
> would be good to add some testscript tests as well

You mean here https://github.com/partiql/partiql-lang-kotlin/tree/master/testscript/pts/test-scripts

PR #151, opened 2019-12-18
Fix issue causing the REPL's output stream to be prematurely closed

dlurton:
This was happening any time the REPL attempted to pretty print a `FLOAT`, `TIMESTAMP`, `SYMBOL`, `CLOB`, `BLOB`, or  `SEXP`.

The problem was caused because the instance of `IonWriter` returned from `IonSystemBuilder.standard().build(outputStream)` closes the `OutputStream` when `IonWriter.close()` is called.  (We should also consider if this is a bug in `ion-java` or not and if its possible to fix this since this behavior is now expected in places where it is currently being used.

In any case, this PR implements a simple work-around.

### DIFF
--- a/lang/src/org/partiql/lang/util/ExprValueFormatter.kt
+++ b/lang/src/org/partiql/lang/util/ExprValueFormatter.kt
@@ -107,7 +107,10 @@ class ConfigurableExprValueFormatter(private val config: Configuration) : ExprVa
         private fun prettyPrintIonLiteral(value: ExprValue) {
             val ionValue = value.ionValue
             out.append("`")
-            IonTextWriterBuilder.standard().build(out).use { writer -> ionValue.writeTo(writer) }
+
+            // We intentionally do *not* want to call [IonWriter.close()] on the [IonWriter] here because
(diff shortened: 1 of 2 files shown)


lang/src/org/partiql/lang/util/ExprValueFormatter.kt
```diff
@@ -107,7 +107,10 @@ class ConfigurableExprValueFormatter(private val config: Configuration) : ExprVa
         private fun prettyPrintIonLiteral(value: ExprValue) {
             val ionValue = value.ionValue
             out.append("`")
-            IonTextWriterBuilder.standard().build(out).use { writer -> ionValue.writeTo(writer) }
+
+            // We intentionally do *not* want to call [IonWriter.close()] on the [IonWriter] here because
+            // that will also call 'out.close()`, which is bad because we probably have more stuff to write!
+            ionValue.writeTo(IonTextWriterBuilder.standard().build(out))
             out.append("`")
```
raganhan: we probably want to flush the writer here to make sure it writes any internally buffered value to the output stream

PR #152, opened 2019-12-19
Uses a set to optimize IN clause when right side is all literals

### LINKED ISSUE #25: Optimize `IN` operator when list is only literals
Performance of the `IN` operator is poor when used with many values.  Most often, the values on the right side of `IN` are literals which can sometimes number into the dozens, hundreds and in rare cases, thousands.

We can fairly easily optimize for this case--the compiler must inspect all values in the list on the right side of `IN` and if so, store them in a data structure such as a `TreeSet`.  Then, when the `IN` operator is being evaluated we just need to evaluate the left side of `IN` once per query and check of the result exists in the `TreeSet`.  If a non-literal value appears anywhere on the right side of `IN` we can simply fall back on the existing behavior.

Currently, both sides of the `IN` operator are evaluated for every row in a query and this is very expensive, especially when many values exist on the right side.

*Migrated from IONSQL-447.*

raganhan:
*Issue #, if available:*
resolves #25

*Description of changes:*
Uses a `TreeSet` with the default comparator when the right argument of an IN clause is a list of literals speeding up the predicate. Changes make IN clause for this use case between ~14% to ~30% faster. 

Benchark setup is the query `SELECT COUNT(*) FROM myTable t WHERE t IN <IN_LIST>` where IN list has 10, 100, 1,000 or 5,000 literals and `myTable` is always a list with 10,000 elements. To trigger the query we are forced to call `ionValue` on the resulting `ExprValue` so we use a `COUNT` to minimize the the time spent materializing Ion objects in that operation. 

Benchmark numbers bellow: 
```
InClauseWithLiterals.optimizedWith5000Literals  thrpt   10   0.110 ± 0.027  ops/s
InClauseWithLiterals.with5000Literals           thrpt   10   0.078 ± 0.012  ops/s
InClauseWithLiterals.optimizedWith1000Literals  thrpt   10   0.596 ± 0.136  ops/s
InClauseWithLiterals.with1000Literals           thrpt   10   0.479 ± 0.010  ops/s
InClauseWithLiterals.optimizedWith100Literals   thrpt   10   5.326 ± 1.935  ops/s
InClauseWithLiterals.with100Literals            thrpt   10   5.219 ± 0.364  ops/s
InClauseWithLiterals.optimizedWith10Literals    thrpt   10  61.258 ± 4.269  ops/s
InClauseWithLiterals.with10Literals             thrpt   10  46.861 ± 1.242  ops/s
```

### DIFF
--- a/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
+++ b/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
@@ -452,7 +452,8 @@ internal class EvaluatingCompiler(
 
                 thunkEnv(metas) { env ->
                     val value = leftArg(env)
-                    inSet.any { it.exprEquals(value) }.exprValue()
+                    // we can use contains as exprEquals uses the DEFAULT_COMPARATOR
+                    inSet.contains(value).exprValue()
                 }
             }
             
--- a/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
+++ b/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
@@ -446,9 +446,7 @@ internal class EvaluatingCompiler(
             rightArg is ListExprNode && rightArg.values.all { it is Literal } -> {
                 val inSet = rightArg.values
                     .map { it as Literal }
-                    .fold(TreeSet<ExprValue>(DEFAULT_COMPARATOR)) { set, el ->
-                        set.also { it.add(valueFactory.newFromIonValue(el.ionValue)) }
-                    }
+                    .mapTo(TreeSet<ExprValue>(DEFAULT_COMPARATOR)) { valueFactory.newFromIonValue(it.ionValue) }
 
                 thunkEnv(metas) { env ->
                     val value = leftArg(env)
--- a/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
+++ b/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
@@ -19,9 +19,7 @@ import com.amazon.ion.*
 import org.partiql.lang.ast.*
 import org.partiql.lang.ast.passes.*
 import org.partiql.lang.errors.*
-import org.partiql.lang.eval.ExprValueType.*
 import org.partiql.lang.eval.binding.*
-import org.partiql.lang.eval.isUnknown
 import org.partiql.lang.syntax.SqlParser
 import org.partiql.lang.util.*
 import java.math.*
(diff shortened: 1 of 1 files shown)


lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -437,17 +436,38 @@ internal class EvaluatingCompiler(
...
+            rightArg is ListExprNode && rightArg.values.all { it is Literal } -> {
+                val inSet = rightArg.values
+                    .map { it as Literal }
+                    .fold(TreeSet<ExprValue>(DEFAULT_COMPARATOR)) { set, el ->
+                        set.also { it.add(valueFactory.newFromIonValue(el.ionValue)) }
+                    }
+
+                thunkEnv(metas) { env ->
+                    val value = leftArg(env)
+                    inSet.any { it.exprEquals(value) }.exprValue()
```
dlurton: As we discussed offline, this should be a call to `TreeSet<ExprValue>.contains()`.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -437,17 +436,38 @@ internal class EvaluatingCompiler(
...
-            haystack.any { it.exprEquals(needle) }.exprValue()
+        val leftArg = compileExprNode(args[0])
+        val rightArg = args[1]
+        
+        return when {
+            // When the right arg is a list of literals we use a Set to speed up the predicate
+            rightArg is ListExprNode && rightArg.values.all { it is Literal } -> {
+                val inSet = rightArg.values
+                    .map { it as Literal }
+                    .fold(TreeSet<ExprValue>(DEFAULT_COMPARATOR)) { set, el ->
```
dlurton: Using a `fold` here seems unnecessarily complex, especially because `TreeSet<>` is a mutable data structure.

I would just declare `inSet` above as an empty `TreeSet<>` and then `.forEach` over `rightArg.values` to `.add` each value.

raganhan: forgot that kotlin has `.mapTo` will use that

raganhan:
`contains` made a huge difference as expected 

```
Benchmark                                        Mode  Cnt    Score    Error  Units
InClauseWithLiterals.optimizedWith10Literals    thrpt   10  137.906 ±  5.490  ops/s
InClauseWithLiterals.with10Literals             thrpt   10   43.762 ±  0.420  ops/s
InClauseWithLiterals.with100Literals            thrpt   10    4.727 ±  0.676  ops/s
InClauseWithLiterals.optimizedWith100Literals   thrpt   10   70.497 ± 10.830  ops/s
InClauseWithLiterals.with1000Literals           thrpt   10    0.430 ±  0.062  ops/s
InClauseWithLiterals.optimizedWith1000Literals  thrpt   10   40.973 ±  4.695  ops/s
InClauseWithLiterals.with5000Literals           thrpt   10    0.071 ±  0.012  ops/s
InClauseWithLiterals.optimizedWith5000Literals  thrpt   10   17.995 ±  0.919  ops/s
```

dlurton:
Now, that's more like it.  :)

raganhan:
I won't add the benchmark code here as there it needed a feature flag to trigger the different PRs, and that feature flag should not be in the final PR.

Would be good to facilitate benchmarks in some way though we should plan for it after the holidays

therapon:
> I won't add the benchmark code here as there it needed a feature flag to trigger the different PRs, and that feature flag should not be in the final PR.
> 
> Would be good to facilitate benchmarks in some way though we should plan for it after the holidays

Agreed

PR #167, opened 2020-01-14
Upgrade JUnitParams to 1.1.1

dlurton:
This should (hopefully) fix our pathologically long build times in openjdk8.


raganhan:
If this helps, I think we should consider migrating the comparator test to junit5 as a test drive

dlurton:
Closing this PR as the upgrade did not solve the problem, while removing `JUnitParams` from `NaturalValueComparatorsTest` did.  Also see https://github.com/partiql/partiql-lang-kotlin/issues/168

PR #223, opened 2020-02-21
Preserve negative zero from both decimal and ion literals

### LINKED ISSUE #164: Pretty Printer Drops Sign for Negative Zero
Executing the following test:

```
class ScratchPadKotlin : EvaluatorTestBase() {
    @Test
    fun foo() {
        val value = eval("-0.0")
        println(value.ionValue)
        println(ConfigurableExprValueFormatter.pretty.format(value))
    }
}
```
Results in the following output:

```
0d-1
0.0
```

My expectation is the second line of the output would be `-0.0` and not `0.0`.

The only impact AFAIK is that the REPL displays `0.0` for `-0.0`:

```
PartiQL> -0.0
   | 
===' 
0.0
--- 
OK! (0 ms)
```

raganhan:
resolves #164 

### DIFF
--- a/lang/src/org/partiql/lang/util/ExprValueFormatter.kt
+++ b/lang/src/org/partiql/lang/util/ExprValueFormatter.kt
@@ -43,10 +43,7 @@ class ConfigurableExprValueFormatter(private val config: Configuration) : ExprVa
 
                 BOOL                                       -> out.append(value.scalar.booleanValue().toString())
 
-                INT, DECIMAL                               -> {
-                    //out.append(value.ionValue.toString(IonTextWriterBuilder.standard()))
-                    out.append(value.scalar.numberValue().toString())
-                }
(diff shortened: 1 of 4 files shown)


lang/src/org/partiql/lang/util/ExprValueFormatter.kt
```diff
@@ -43,7 +43,10 @@ class ConfigurableExprValueFormatter(private val config: Configuration) : ExprVa
 
                 BOOL                                       -> out.append(value.scalar.booleanValue().toString())
 
-                INT, DECIMAL                               -> out.append(value.scalar.numberValue().toString())
+                INT, DECIMAL                               -> {
+                    //out.append(value.ionValue.toString(IonTextWriterBuilder.standard()))
```
dlurton: delete me. :)

raganhan: ops :)

Issue #248, opened 2020-05-21
CLI: Execute PartiQL query when a line ending with a semi-colon is read


metadaddy:
Currently the CLI requires the user to hit enter twice to execute a PartiQL query, even if the query is terminated with a semi-colon.

When it reads a line of PartiQL query input ending with a semi-colon `;`, it would be great if the CLI could execute the query immediately, without waiting for an empty line.

PR #249, opened 2020-05-21
Execute PartiQL query when a line ending in a semi-colon is read.

metadaddy:
*Issue #, if available:*

#248 CLI: Execute PartiQL query when a line ending with a semi-colon is read

*Description of changes:*

Add a new state to the REPL state machine, `LAST_PARTIQL_LINE`. Transition to this state from `READY` or `READ_PARTIQL` when the input line ends with a `;`. `LAST_PARTIQL_LINE` simply appends the current line to the buffer and transitions to the `EXECUTE_PARTIQL` state.

### DIFF
--- a/cli/src/org/partiql/cli/Repl.kt
+++ b/cli/src/org/partiql/cli/Repl.kt
@@ -41,6 +41,7 @@ private enum class ReplState {
     /** Reading a PartiQL query. Transitions to execute when one of the execution tokens is found. */
     READ_PARTIQL,
 
+    /** Read a trailing semi-colon. Add the current line to the buffer and transition to execute. */
     LAST_PARTIQL_LINE,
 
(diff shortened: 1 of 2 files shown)


cli/src/org/partiql/cli/Repl.kt
```diff
@@ -41,6 +41,8 @@ private enum class ReplState {
     /** Reading a PartiQL query. Transitions to execute when one of the execution tokens is found. */
     READ_PARTIQL,
 
+    LAST_PARTIQL_LINE,
```
therapon, line 44: Can we please add a comment here explaining the intent of this state, just like the comments on the existing states.

metadaddy: Good call - done!

PR #258, opened 2020-07-09
Migrate to PIG 0.2.0

dlurton:
And a minor issue is discovered with the `ion-element` API as well, which I've created a issue for.


therapon:
> Task :cli:test
```
org.partiql.cli.ReplTest > addToGlobalEnvAndQuery FAILED

    org.junit.ComparisonFailure at ReplTest.kt:273
```

This is the REPL test that fails due to the travis run taking longer to complete leading the test to compare the string representation of the REPL that includes the milliseconds :)

PR #264, opened 2020-07-29
Remove default parameter from EvaluationException

abhikuhikar:
Removed default parameter for the internal flag from EvaluationException secondary constructor. Initially it was set to true. We expect the internal error flag to be set explicitly during the new instance creation of EvaluationException so that there's no confusion if the error is internal or not.


abhikuhikar:
Also made sure that all the new instance creations of EvaluationException explicitly sets the internal error flag and there is no ambiguity.

lang/src/org/partiql/lang/eval/Exceptions.kt
```diff
@@ -39,7 +39,7 @@ open class EvaluationException(message: String,
     constructor(cause: Throwable,
```
dlurton, line 39: Aren't we removing the constructor on line 30 as well?

abhikuhikar: Sure. Thanks, David. I will update the PR.

PR #265, opened 2020-07-30
Use the thread-safe StringBuffer instead of StringBuilder in CLI test

dlurton:
This is somewhat of a shot in the dark to try to fix what appears to be a race condition in some of our test infrastructure for the CLI sub-project that is causing tests to fail intermittently. i.e. like in this [build](https://travis-ci.org/github/partiql/partiql-lang-kotlin/builds/713101679).

I will retry the travis build a several times before requesting review.

### DIFF
--- a/cli/test/org/partiql/cli/ReplTest.kt
+++ b/cli/test/org/partiql/cli/ReplTest.kt
@@ -91,7 +91,9 @@ private class ReplTester(bindings: Bindings<ExprValue> = Bindings.empty()) {
         while (replThread.isAlive || output.size() > 0) {
             Thread.sleep(SLEEP_TIME)
             if (output.size() > 0) {
-                actualReplPrompt.append(output.toString())
+                synchronized(actualReplPrompt) {
+                    actualReplPrompt.append(output.toString())
+                }
                 output.reset()
                 outputPhaser.arrive()
             }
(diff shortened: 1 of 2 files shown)


cli/test/org/partiql/cli/ReplTest.kt
```diff
@@ -91,7 +91,9 @@ private class ReplTester(bindings: Bindings<ExprValue> = Bindings.empty()) {
         while (replThread.isAlive || output.size() > 0) {
             Thread.sleep(SLEEP_TIME)
             if (output.size() > 0) {
-                actualReplPrompt.append(output.toString())
+                synchronized(actualReplPrompt) {
+                    actualReplPrompt.append(output.toString())
+                }
```
therapon, line 96: While this will fix the issue, I am not sure we have a need for parallel tests (unless JUnit does this by default). 
If we can flip our tests to single threaded I think that would be best, till then this is fine.

dlurton: This doesn't work either. It looks like it does but it's still intermittent.

So... there's a few problems here... I will try one more thing before I mark the whole fixture as ignored and file an issue to refactor this thing.

The way the CLI is currently factored doesn't allow it to be easily tested.  Apparently, the multithreading is actually a way of working around that.  It uses a combination of 3 (!) different threads to simulate input, collect the output, and to run the repl itself.  

I've created https://github.com/partiql/partiql-lang-kotlin/issues/266 to refactor and eliminate the multithreading.

dlurton:
I think the right solution here is to `@Ignore` the unit tests and refactor the CLI to make it more testable and not require multithreading.  I filed #266 and submitted #267 for now...

Issue #252, opened 2020-06-19
Use of MIN function causes unbound binding exception 


therapon:
version 0.2.1

Given the following data 

```
{ 
    'products': [ 
    { 
        'title': 'The unbearable lightness of being', 
        'author': 'Milan Kundera', 
        'price' : [12, 11, 9.99]
    } , 
    { 

        'title': 'How to travel with a salmon',
        'author': 'Umberto Eco', 
        'price' : [10, 5.99]
    } , 
    { 

        'title': 'And the ass saw the angel',
        'author': 'Nick Cave', 
        'price' : [8, 5.99]
    }]
} 
```

The following query errors out 

``` 
SELECT x.title, MIN(money) as price 
FROM products AS x, x.price AS money
WHERE author like '%Milan%'; 
   | 
org.partiql.lang.eval.EvaluationException: No such binding: x
	Evaluator Error: at line 1, column 8: Binding does not exist, <UNKNOWN> : <UNKNOWN>


```

While this one succeeds  

```
PartiQL> SELECT x.title, money as price  FROM products AS x, x.price AS money WHERE author like '%Milan%'; 
   | 
===' 
<<
  {
    'title': 'The unbearable lightness of being',
    'price': 12
  },
  {
    'title': 'The unbearable lightness of being',
    'price': 11
  },
  {
    'title': 'The unbearable lightness of being',
    'price': 9.99
  }
>>
--- 


```

dlurton:
If there's a bug here I believe it to be the that the error message is confusing.

The origin of this error makes more sense if you consider what's going on behind the scenes when any aggregate functions are being used in the select list.  The compiler is essentially treating this query as if it had an empty `GROUP BY` clause, i.e.

```
SELECT x.title, MIN(money) as price 
FROM products AS x, x.price AS money
WHERE author like '%Milan%'; 
GROUP BY /* none */
```

This means that the result of the empty `GROUP BY` clause is fed into the select projection.

To make this query work as expected, you need to include `x.title` in the `GROUP BY` clause and it should just work: `GROUP BY x.title`

dlurton:
While writing the following, a practical way of providing the right error message became obvious:

[...]

PR #270, opened 2020-08-10
Add more descriptive error message for missing GROUP BY

alancai98:
Addresses Issue #252 

- Implements fix for better error message when a given query is missing a `GROUP BY`
- Uses approach outlined in issue [comment](https://github.com/partiql/partiql-lang-kotlin/issues/252#issuecomment-662591844)
- Corresponding tests in EvaluatingCompilerGroupByTest.kt


lang/test/org/partiql/lang/eval/EvaluatingCompilerGroupByTest.kt
```diff
@@ -932,4 +953,96 @@ EvaluatingCompilerGroupByTest : EvaluatorTestBase() {
...
+    }
+
+    @Test
+    fun missingGroupBySFWTest() {
+        checkInputThrowingEvaluationException(
+            """
+            SELECT x.title, MIN(money) as price
+            FROM books AS x, x.price AS money
+            WHERE author like '%Milan%'
+            """.trimIndent(),
```
dlurton, line 977: `.trimIndent()` makes the column number appear to be different than what shows in the IDE's cursor location.  I'd suggest removing it.  Same for the others below.

lang/test/org/partiql/lang/eval/EvaluatingCompilerGroupByTest.kt
```diff
@@ -932,4 +953,96 @@ EvaluatingCompilerGroupByTest : EvaluatorTestBase() {
...
+            null)
+    }
+
+    @Test
+    fun missingGroupBySFWTest() {
+        checkInputThrowingEvaluationException(
+            """
+            SELECT x.title, MIN(money) as price
+            FROM books AS x, x.price AS money
+            WHERE author like '%Milan%'
```
dlurton, line 976: The test cases here a more complex than they need to be.  

Consider removing parts of the SQL which aren't material to the scenario being tested such as `WHERE author like '%Milan%'` and un-needed select-list aliases.  The same goes for the other tests below.  This can greatly improve the readability of the test cases and reduces the likelihood that they will break during future changes.

lang/test/org/partiql/lang/eval/EvaluatingCompilerGroupByTest.kt
```diff
@@ -932,4 +953,96 @@ EvaluatingCompilerGroupByTest : EvaluatorTestBase() {
...
+            """.trimIndent(),
+            session,
+            ErrorCode.EVALUATOR_VARIABLE_NOT_INCLUDED_IN_GROUP_BY,
+            mapOf(
+                    Property.LINE_NUMBER to 1L,
+                    Property.COLUMN_NUMBER to 8L,
+                    Property.BINDING_NAME to "x"
+            ),
+            null)
```
dlurton, line 985: Explicitly passing `null` here isn't required since the parameter has a default value of `null`.

lang/test/org/partiql/lang/eval/EvaluatingCompilerGroupByTest.kt
```diff
@@ -932,4 +953,96 @@ EvaluatingCompilerGroupByTest : EvaluatorTestBase() {
...
+            session,
+            ErrorCode.EVALUATOR_VARIABLE_NOT_INCLUDED_IN_GROUP_BY,
+            mapOf(
+                    Property.LINE_NUMBER to 1L,
+                    Property.COLUMN_NUMBER to 33L,
+                    Property.BINDING_NAME to "o"
+            ),
+            null)
+    }
 }
```
dlurton, line 1048: Need to include tests with varying cases. i.e. `SELECT F.name, MAX(f.id) FROM foo AS f` -- for instance, this should still throw the `EVALUATOR_VARIABLE_NOT_INCLUDED_IN_GROUP_BY` error.  

Also add a test that proves `EVALUATOR_BINDING_DOES_NOT_EXIST` is not thrown for `SELECT "F".name, MAX(f.id) FROM foo AS f`.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -686,17 +689,32 @@ internal class EvaluatingCompiler(
...
                 val evalVariableReference = when (compileOptions.undefinedVariable) {
                     UndefinedVariableBehavior.ERROR   ->
                         thunkEnv(metas) { env ->
-                            env.current[bindingName] ?: throw EvaluationException(
-                                "No such binding: ${bindingName.name}",
-                                ErrorCode.EVALUATOR_BINDING_DOES_NOT_EXIST,
-                                errorContextFrom(metas).also { it[Property.BINDING_NAME] = bindingName.name },
-                                internal = false)
+                            val value = env.current[bindingName]
+                            if(value == null) {
```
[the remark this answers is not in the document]
alancai98: Is `when` preferred even in the binary case? The general Kotlin [style guide](https://kotlinlang.org/docs/reference/coding-conventions.html#if-versus-when) says "Prefer using `if` for binary conditions instead of `when`. And "Prefer using `when` if there are three or more options."

The PartiQL specific [style guide](https://github.com/partiql/partiql-lang-kotlin/blob/master/docs/dev/CODE-STYLE.md#control-flow) says to "Use `when` instead of `if else if` when possible..."

lang/test/org/partiql/lang/eval/EvaluatingCompilerGroupByTest.kt
```diff
@@ -60,6 +60,28 @@ EvaluatingCompilerGroupByTest : EvaluatorTestBase() {
         ]""",
         "widgets_b" to """[
             { categoryId: 2, name: "Thingy" }
+        ]""",
+        "books" to """[
+            { 'title': 'The unbearable lightness of being', 'author': 'Milan Kundera', 'price' : [12, 11, 9.99]},
+            { 'title': 'How to travel with a salmon', 'author': 'Umberto Eco', 'price' : [10, 5.99]},
+            { 'title': 'And the ass saw the angel', 'author': 'Nick Cave', 'price' : [8, 5.99]}
```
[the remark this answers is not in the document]
alancai98: Will replace in the code.

This was taken from the original Github Issue #252. Should I change it there as well?

lang/test/org/partiql/lang/eval/EvaluatingCompilerGroupByTest.kt
alancai98: Got it. Will remove the unnecessary parameter.

lang/test/org/partiql/lang/eval/EvaluatingCompilerGroupByTest.kt
alancai98: Will remove and update the tests.

lang/test/org/partiql/lang/eval/EvaluatingCompilerGroupByTest.kt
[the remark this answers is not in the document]
therapon: That was probably my doing. These are 3 books that are my foo bar baz examples. Sorry.

lang/test/org/partiql/lang/eval/EvaluatingCompilerGroupByTest.kt
```diff
@@ -932,4 +953,96 @@ EvaluatingCompilerGroupByTest : EvaluatorTestBase() {
...
+            mapOf(
+                    Property.LINE_NUMBER to 1L,
+                    Property.COLUMN_NUMBER to 35L,
+                    Property.BINDING_NAME to "o"
+            ),
+            null)
+    }
+
+    @Test
+    fun missingGroupByOuterQueryTest() {
```
therapon, line 1026: I would add a test to make sure the error and its message is correct when the inner query uses GROUP BY and the inner queries SELECT uses a variable that is not bound by the GROUP BY clause

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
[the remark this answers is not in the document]
dlurton: The PartiQL specific style guide should be probably be reviewed and possibly updated.  It was written a while back and should be taken with a grain of salt.

In this case I prefer `when` because it allows inlining the variable declaration inside the `when(...)` which reduces syntactic overhead.  This is however just my opinion which is not any kind of law, of course.

I prefer to use `when` any time the condition is not a boolean value.  i.e. I wouldn't use it for `when(foo == 1) { true -> ... false -> }`.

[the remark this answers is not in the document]
alancai98: Didn't know about the inlining. Changed to use `when`

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -686,17 +689,32 @@ internal class EvaluatingCompiler(
...
+                                val isFromSourceVariable = fromSourceNames.contains(bindingName.name)
+                                if (!isFromSourceVariable) {
```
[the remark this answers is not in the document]
alancai98: Updated to do some case-(in)sensitivity checks

lang/test/org/partiql/lang/eval/EvaluatingCompilerGroupByTest.kt
alancai98: Simplified to remove some untested clauses (e.g. `LIKE...` and the select-list aliases)

lang/test/org/partiql/lang/eval/EvaluatingCompilerGroupByTest.kt
alancai98: Added some tests just for case-(in)sensitivity. Let me know if they aren't sufficient

lang/test/org/partiql/lang/eval/EvaluatingCompilerGroupByTest.kt
alancai98: Good idea. Added a new test for the inner clause

lang/test/org/partiql/lang/eval/EvaluatingCompilerGroupByTest.kt
```diff
@@ -932,4 +953,149 @@ EvaluatingCompilerGroupByTest : EvaluatorTestBase() {
...
+        )
+    }
+
+    @Test
+    fun missingGroupBySFWTest() {
+        checkInputThrowingEvaluationException(
+            """
+            SELECT x.title, MIN(money)
+            FROM books AS x, x.price AS money
+            WHERE author = 'Milan Kundera'
```
dlurton, line 975: This `WHERE` clause isn't important to this test, is it?

alancai98: The test was for SFW queries. I can remove the `WHERE` clause and/or the test if it seems too redundant.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -681,22 +684,42 @@ internal class EvaluatingCompiler(
         return thunkEnv(metas) { _ -> valueFactory.missingValue }
     }
 
+    private fun isFromSourceVariable(bindingName: BindingName, fromSourceNames: Set<String>): Boolean {
+        val bindingInSourcesCaseInsensitive = fromSourceNames.any { it.toLowerCase() == bindingName.name.toLowerCase() }
```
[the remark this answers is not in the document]
alancai98: Oh okay, didn't know that was there. Will change to use `isEquivalentTo`.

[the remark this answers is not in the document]
alancai98: Guess I can also get rid of that helper function.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -869,8 +886,21 @@ internal class EvaluatingCompiler(
...
 
     private fun compileSelect(selectExpr: Select): ThunkEnv {
+        // Get all the FROM source aliases for binding error checks
+        val fold = object : PartiqlAst.VisitorFold<Set<String>>() {
+            override fun visitFromSourceScan(node: PartiqlAst.FromSource.Scan, accumulator: Set<String>): Set<String> {
+                val aliases = listOfNotNull(node.asAlias?.text, node.atAlias?.text, node.byAlias?.text)
+                return accumulator + aliases.toSet()
+            }
+
+            override fun walkExprSelect(node: PartiqlAst.Expr.Select, accumulator: Set<String>): Set<String> {
```
dlurton, line 896: Add `kdoc` that explains the purpose of this override.

alancai98: Added. Let me know if it doesn't provides sufficient information.

alancai98:
> * a `LIMIT` clause, i.e. ` select max(1) from << { 'a': 'a' } >> as f limit f.id`

Talked offline. Doesn't make sense to have a variable in a `LIMIT` clause. May need to provide a better error message in the future.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
dlurton: Hrm.  I would have expected it to convey that the purpose was to prevent it from recursing into nested select statements.

alancai98: Will change to something like "Prevents visitor from recursing into nested select statements"

Issue #179, opened 2020-02-13
Add support for `LET` 


therapon:
Add support for `LET` clause

dlurton:
Details to be added to the specification.

Issue #44, opened 2019-05-17
Don't throw IllegalStateException


raganhan:
In a couple dozen places within the interpreter we throw `IllegalStateException`.  We should probably be throwing a descendant of `SqlException` instead.

PR #278, opened 2020-09-12
LET PIG-domain changes (from_let)

alancai98:
Initial changes to PIG-domain to add LET (#179)

### DIFF
--- a/lang/resources/org/partiql/type-domains/partiql.ion
+++ b/lang/resources/org/partiql/type-domains/partiql.ion
@@ -103,11 +103,17 @@
             (select
                 (setq (? set_quantifier))
                 (project projection)
+                (project_let (? let))
                 (from from_source)
+                (from_let (? let))
                 (where (? expr))
+                (where_let (? let))
                 (group (? group_by))
+                (group_let (? let))
                 (having (? expr))
-                (limit (? expr))))
+                (having_let (? let))
+                (limit (? expr))
+                (limit_let (? let))))
         // end of sum expr
 
         // A "step" within a path expression; that is the components of the expression following the root.
@@ -136,6 +142,12 @@
             (project_all expr::expr)
             // For `<expr> [AS <id>]`
             (project_expr expr::expr as_alias::(? symbol)))
+
+        // A list of LET bindings
+        (product let let_bindings::(* let_binding 1))
+
+        // A LET binding
+        (product let_binding expr::expr name::symbol)
 
         // Models the FROM clause of an SFW query
         // Note that modeling `scan` and `unpivot` separately is effectively re-introducing the same problem described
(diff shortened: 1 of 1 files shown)


lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -137,6 +143,12 @@
             // For `<expr> [AS <id>]`
             (project_expr expr::expr as_alias::(? symbol)))
 
+        // A list of LET bindings
+        (product let let_bindings::(* let_binding 1))
+
+        // A LET binding
+        (product let_binding expr::expr name::symbol)
```
abhikuhikar, line 150: I would name this `as_alias::symbol` instead of `name::symbol` as it resembles the grammar more closely.

alancai98: What grammar are you referring to? The general [spec](https://github.com/partiql/partiql-spec/issues/15) uses the term `name`.

I kind of like `name` because the meaning is slightly different than `as_alias`, which is an optional parameter.

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -103,11 +103,17 @@
...
+                (from_let (? let))
                 (where (? expr))
+                (where_let (? let))
                 (group (? group_by))
+                (group_let (? let))
                 (having (? expr))
-                (limit (? expr))))
+                (having_let (? let))
+                (limit (? expr))
+                (limit_let (? let))))
```
dlurton, line 116: This is a reasonable way to model this now, just add a comment referencing [partiql-ir-generator #44](https://github.com/partiql/partiql-ir-generator/issues/44).

When that issue is completed I'd like to model this with a variadic record element instead, i.e.

```
(select
    ... 
    (project_let (* let_binding 0))
    ...)
```

dlurton (commented):
This is a reasonable starting point, however be aware that the [RFC](https://github.com/partiql/partiql-spec/issues/15) for this is still in flux and may change significantly.

Probably, I'd start implementation with `from_let` since that one has less comments surrounding it.

lang/resources/org/partiql/type-domains/partiql.ion
alancai98: Added.

dlurton:
I guess if we're only focusing on `from_let`, then for now the others should be removed.

Otherwise LGTM.

dlurton (approved):
I guess if we're only focusing on `from_let`, then for now the others should be removed.

dlurton (approved):

lang/resources/org/partiql/type-domains/partiql.ion
alancai98: Will leave as `name` for now. This could change as the implementation and spec are more complete.

PR #279, opened 2020-09-14
Optimize LIKE pattern compilation and make LIKE pattern compilation interruptible.

dlurton:
On `master` this expression takes on average 37 seconds to compile:

`'foo' like '%<n>%'` (`<n>` is 1500 `!` characters)

With this change, on average it takes ~2.6 seconds to compile.

This is still an order of magnitude slower than it really needs be, but is (comparatively speaking) much better and all that I can do in the short term.

### DIFF
--- a/lang/src/org/partiql/lang/eval/LikeMatchingAutomata.kt
+++ b/lang/src/org/partiql/lang/eval/LikeMatchingAutomata.kt
@@ -18,6 +18,7 @@ import org.partiql.lang.util.codePointSequence
 import java.util.ArrayList
 import java.util.HashSet
 
+
 @Suppress("UNCHECKED_CAST")
 private fun <T> Iterable<T>.fastToMutableHashSet(): MutableSet<T> =
     when(this) {
@@ -102,10 +103,11 @@ interface IDFAState {
  * - [accepting] true if this is a Final state, false otherwise
  * - [start] true if this is a Start state, false otherwise
  */
-private open class DFAState(val nfaStates: MutableSet<NFAState>,
-                    val outgoing: MutableMap<Alphabet, DFAState>,
-                    var accepting: Boolean = nfaStates.any { it.isAccepting },
-                    var start: Boolean = nfaStates.any { it.isStartState }
+private open class DFAState(
+    val nfaStates: MutableSet<NFAState>,
+    val outgoing: MutableMap<Alphabet, DFAState>,
+    var accepting: Boolean = nfaStates.any { it.isAccepting },
+    var start: Boolean = nfaStates.any { it.isStartState }
 ) : IDFAState {
 
     fun addTransition(transition: Alphabet, target: DFAState) {
@@ -180,10 +182,11 @@ private open class DFAState(val nfaStates: MutableSet<NFAState>,
  * - [isStartState] true when this State is a Start state, false othewise
  * - [outgoing] map of alphabet letter to NFA State
  */
-private class NFAState(val stateNumber: Int,
-               val isAccepting: Boolean,
-               val isStartState: Boolean,
-               val outgoing: MutableMap<Alphabet, MutableSet<NFAState>> = HashMap()) {
+private class NFAState(
+    val stateNumber: Int,
+    val isAccepting: Boolean,
+    val isStartState: Boolean,
+    val outgoing: MutableMap<Alphabet, MutableSet<NFAState>> = HashMap()) {
 
     fun get(transition: Alphabet): Set<NFAState> =
         outgoing[transition] ?: hashSetOf()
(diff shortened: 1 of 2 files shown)


lang/src/org/partiql/lang/eval/LikeMatchingAutomata.kt
```diff
@@ -165,18 +182,19 @@ private open class DFAState(val nfaStates: MutableSet<NFAState>,
...
-               val outgoing: MutableMap<Alphabet, MutableSet<NFAState>> = mutableMapOf<Alphabet, MutableSet<NFAState>>()) {
+private class NFAState(
+    val stateNumber: Int,
+    val isAccepting: Boolean,
+    val isStartState: Boolean,
+    val outgoing: MutableMap<Alphabet, MutableSet<NFAState>> = HashMap()) {
 
     fun get(transition: Alphabet): Set<NFAState> =
-        outgoing[transition]?.let { it } ?: setOf<NFAState>()
+        outgoing[transition] ?: hashSetOf()
```
dlurton, line 192: We're using *unordered* sets which are `java.util.HashSet` instead of `java.util.LinkedHashSet` which should reduce memory and CPU overhead a little.

lang/src/org/partiql/lang/eval/LikeMatchingAutomata.kt
```diff
@@ -401,8 +412,14 @@ private fun buildDFA(dfaAlphabet: Set<Alphabet>,
...
-                    nfaStates.fold(setOf<NFAState>()) { acc, state ->
-                        acc.union(state.getOutgoingStates(it))
+                    HashSet<NFAState>().apply {
+                        nfaStates.forEach { state ->
+                            addAll(state.getOutgoingStates(it))
+
+                            if (Thread.interrupted()) {
+                                throw InterruptedException()
+                            }
+                        }
```
dlurton, line 422: This is the change in this PR that results in the biggest performance improvement.  The other stuff is incremental.

lang/test/org/partiql/lang/eval/LikePredicateTest.kt
```diff
@@ -629,5 +630,33 @@ class LikePredicateTest : EvaluatorTestBase() {
     }
 
 
+    @Test
+    fun interruptedThreadThrowsInterruptedException() {
+        val sb = StringBuilder()
+        sb.append("foo LIKE '%")
+        repeat(5000) {
+            sb.append("!")
+        }
```
alancai98, line 639: nit: Could replace with [repeat](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/repeat.html).

`sb.append("!".repeat(5000))`

lang/src/org/partiql/lang/eval/LikeMatchingAutomata.kt
abhikuhikar: Can there be a use-case where the order of elements is important? Also, as far as I know, the `LinkedHashSet` is almost as fast and efficient as `HashSet` asymptotically speaking.

lang/src/org/partiql/lang/eval/LikeMatchingAutomata.kt
```diff
@@ -195,17 +213,11 @@ private class NFAState(val stateNumber: Int,
...
+                startSet.fold(get(alpha)) { acc, nfaState -> acc.fastUnion(nfaState.epsilonClosure()) }
             }
         }
 
-    /**
-     * Convineance method for 1-character closure when given a [Char].
-     */
-    fun getOutgoingStates(letter: Char): Set<NFAState> =
-        closure(letter.toInt())
-
```
therapon, line 208: was this unused?

lang/test/org/partiql/lang/eval/LikePredicateTest.kt
```diff
@@ -629,5 +630,33 @@ class LikePredicateTest : EvaluatorTestBase() {
...
+        sb.append("%'")
+
+        // '%!!!!....%' should take a very long time to compile until we make some major refactorings
+        // of the LIKE pattern matching implementations.
+        val heavyPayload = sb.toString()
+
+        var wasIntertupted = false
```
jobarr-amzn, line 646: Typo: `wasInterrupted`

lang/src/org/partiql/lang/eval/LikeMatchingAutomata.kt
dlurton: Well, all of the tests pass and functionality seems to be unaffected.  I don't think order is important.

lang/src/org/partiql/lang/eval/LikeMatchingAutomata.kt
dlurton: It was.

lang/test/org/partiql/lang/eval/LikePredicateTest.kt
dlurton: Actually, `"!"` enables removal of the `StringBuilder` too.  thanks.

lang/test/org/partiql/lang/eval/LikePredicateTest.kt
dlurton: fixed.

alancai98 (approved):

therapon (approved):

therapon (approved):

dlurton:
Superseded by #286.

Issue #281, opened 2020-09-15
Incorrect execution order for `LIMIT` clause during query execution 


dlurton:
Currently, the execution order is:

- `FROM`
- `WHERE`
- `LIMIT`
- `GROUP BY` (if present)
- `HAVING` (if present)
- ...`SELECT`-list or `SELECT VALUE` projection..

In reality it should probably be one of two options:

- `FROM`
- `WHERE`
- `GROUP BY` (if present)
- `HAVING` (if present)
- `LIMIT` (option 1)
- ...`SELECT`-list or `SELECT VALUE` projection..
- `LIMIT` (option 2)

The impact of this bug can be seen when it's used with `GROUP BY`:

```
PartiQL> select * from `[{foo: 1, bar: 10}, {foo: 1, bar: 11}]` AS f GROUP BY f.foo GROUP AS g LIMIT 1
   | ===' 
<<
  {
    'foo': 1,
    'g': <<
      {
        'f': {
          'foo': 1,
          'bar': 10
        }
      }
    >>
  }
>>
```

`LIMIT` should applied *after* the group has occurred and not before.  Since this query is has only one group, the result should be unaffected by `LIMIT 1`.   The expected result is:

```
<<
  {
    'foo': 1,
    'g': <<
      {
        'f': {
          'foo': 1,
          'bar': 10
        }
      },
      {
        'f': {
          'foo': 1,
          'bar': 11
        }
      }
    >>
  }
>>
```

PR #285, opened 2020-09-18
LET (from `FROM` clauses) implementation - Parsing and ExprNode conversions

alancai98:
Part 1 of implementation for `LET` clause (#179) as defined in the high-level [spec](https://github.com/partiql/partiql-spec/issues/15).

Currently just implementing `LET` following `FROM` clauses as they are what's most clearly defined in the high-level spec and use cases.

Future supported clauses will be added later (especially because of #281, _Incorrect execution order for_ `LIMIT` _clause during query execution_)

Current status relating to `LET` following `FROM` clauses (from_let):
- [x] PIG domain model of `LET` clauses (separate [PR](https://github.com/partiql/partiql-lang-kotlin/pull/278))
- [x] `ExprNode` representation of `LET` clauses
- [x] Conversion between `PartiqlAst` and `ExprNode` for `LET` clauses
- [x] V0 Ast Serliazation for `LET` clauses
- [x] Parsing of `LET` clauses
- [x] `LET` parsing tests
(Evaluation to be done in a separate PR before merging into `from-let-final` feature branch)

Possible additions:
- test conversion between `PartiqlAst` and `ExprNode`
- test V0 ast (de)serializer outputs intended error when parsing `LET` clauses

### DIFF
--- a/lang/src/org/partiql/lang/ast/StatementToExprNode.kt
+++ b/lang/src/org/partiql/lang/ast/StatementToExprNode.kt
@@ -239,14 +239,13 @@ private class StatementTransformer(val ion: IonSystem) {
             this.letBindings.map {
                 LetBinding(
                     it.expr.toExprNode(),
-                    it.name.toSymbolicName()!!
-                    //SymbolicName(it.name.text, it.metas.toPartiQlMetaContainer())
+                    it.name.toSymbolicName()
                 )
             }
         )
     }
 
-    private fun SymbolPrimitive?.toSymbolicName() = this?.let { SymbolicName(it.text, it.metas.toPartiQlMetaContainer()) }
+    private fun SymbolPrimitive.toSymbolicName() = SymbolicName(this.text, this.metas.toPartiQlMetaContainer())
 
     private fun GroupBy.toGroupBy(): org.partiql.lang.ast.GroupBy =
         GroupBy(
--- a/lang/src/org/partiql/lang/ast/StatementToExprNode.kt
+++ b/lang/src/org/partiql/lang/ast/StatementToExprNode.kt
@@ -10,6 +10,7 @@ import org.partiql.lang.domains.PartiqlAst.FromSource
 import org.partiql.lang.domains.PartiqlAst.GroupBy
 import org.partiql.lang.domains.PartiqlAst.GroupingStrategy
 import org.partiql.lang.domains.PartiqlAst.JoinType
+import org.partiql.lang.domains.PartiqlAst.Let
 import org.partiql.lang.domains.PartiqlAst.PathStep
 import org.partiql.lang.domains.PartiqlAst.ProjectItem
 import org.partiql.lang.domains.PartiqlAst.Projection
@@ -167,6 +168,7 @@ private class StatementTransformer(val ion: IonSystem) {
                     setQuantifier = setq?.toSetQuantifier() ?: org.partiql.lang.ast.SetQuantifier.ALL,
                     projection = project.toSelectProjection(),
                     from = from.toFromSource(),
+                    fromLet = fromLet?.toLetSource(),
                     where = where?.toExprNode(),
                     groupBy = group?.toGroupBy(),
                     having = having?.toExprNode(),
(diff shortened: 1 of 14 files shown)


lang/src/org/partiql/lang/ast/StatementToExprNode.kt
```diff
@@ -232,6 +234,18 @@ private class StatementTransformer(val ion: IonSystem) {
             is JoinType.Full -> JoinOp.OUTER
         }
 
+    private fun Let.toLetSource(): LetSource {
+        return LetSource(
+            this.letBindings.map {
+                LetBinding(
+                    it.expr.toExprNode(),
+                    it.name.toSymbolicName()!!
```
dlurton, line 242: Pretty sure that `!!` is not needed here.

lang/test/org/partiql/lang/syntax/SqlParserTest.kt
```diff
@@ -3092,4 +3096,96 @@ class SqlParserTest : SqlParserTestBase() {
...
+    @Test
+    fun selectFromLetTest() = assertExpression("SELECT x FROM table1 LET 1 AS A") {
+        select(
+            project = projectX,
+            from = scan(id("table1")),
+            fromLet = let(
+                letBinding(
+                    lit(ionInt(1)),
+                    "A")
+            )
```
dlurton, line 3115: Nit:  put all on one line.  same for others below.

dlurton:
Off to a good start--evaluation might be the interesting part.

lang/src/org/partiql/lang/ast/StatementToExprNode.kt
alancai98: Removed the not-null assertion op. Revised `toSymbolicName` to not use nullable guards (`?`)

lang/test/org/partiql/lang/syntax/SqlParserTest.kt
alancai98: Revised formatting

alancai98:
Updated to resolve David's comments. 

Changed this PR to focus on just parsing and `ExprNode` conversion for `LET`. The evaluation-side implementation of `LET` for `FROM` clauses will have a separate PR. Both of these branches will be merged into the `from-let-clause` base branch before merging into master.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -1815,6 +1845,48 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
+        rem = child.remaining
+
+        if (rem.head?.type != AS) {
+            rem.head.err("Expected AS following LET expr", PARSE_EXPECTED_AS_FOR_LET)
+        }
+
+        rem = rem.tail
+
+        if (rem.head?.type?.isIdentifier() != true) {
+            rem.head.err("Expected identifier for $AS-alias", PARSE_EXPECTED_IDENT_FOR_ALIAS)
```
therapon, line 1861: nit: This error message uses `$AS` to interpolate, can we do the same for `LET` on line 1855? 
I do not mind either way as long as we are consistent

therapon (approved):
minor nit, otherwise looks good

lang/src/org/partiql/lang/syntax/SqlParser.kt
alancai98: Sure, also added to the other error message in the while loop.

therapon (approved):

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -987,6 +1010,7 @@ class SqlParser(private val ion: IonSystem) : Parser {
             "substring" -> tail.parseSubstring(head!!)
             "trim" -> tail.parseTrim(head!!)
             "extract" -> tail.parseExtract(head!!)
+            "let" -> tail.parseLet()
```
dlurton, line 1013: I just bumped into this by accident--this needs to be removed.  It's not helping anything since there are not test cases that cover this.  This might be left over from when we considering allowing `LET` to be a first-class expression.  In any case, this should be removed here...

alancai98: Will remove when merging `from-let-final` into `master`.

alancai98: Removed in this [commit](https://github.com/partiql/partiql-lang-kotlin/pull/303/commits/07b654aca2c75e4fb57dce77712b93a9509b7d6e).

dlurton (commented):
LGTM. Don't forget to squash when merging.  Thanks.

PR #291, opened 2020-09-28
LET (from `FROM` clauses) implementation - Evaluation

alancai98:
Current status:
- [x] Rewriter inclusion of `LET` clauses from `FROM`
- [x] `LET` bindings included in environment
- [x] `LET` clause tests

TODO:
- Could do some further refactoring of `LET` evaluation in `compileQueryWithoutProjection` to its own method


lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -1435,6 +1461,24 @@ internal class EvaluatingCompiler(
                     // bind the joined value to the bindings for the filter/project
                     FromProduction(joinedValues, fromEnv.nest(localsBinder.bindLocals(joinedValues)))
                 }
+            // Nest LET bindings in the FROM environment
+            if (compiledLetSources != null) {
+                seq = seq.map { fromProduction ->
```
therapon, line 1466: Does kotlin `map` process a collection from left to right? 
The `LET` clause can have more than one binding and in that case we want bindings to have access to their preceding `LET` bindings.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -1435,6 +1461,24 @@ internal class EvaluatingCompiler(
...
+                    val parentEnv = fromProduction.env
+
+                    val letEnv: Environment = compiledLetSources.fold(parentEnv) { accEnvironment, curLetSource ->
```
therapon, line 1469: Similar question, is kotlin `fold` a *left* fold?

lang/test/org/partiql/lang/eval/EvaluatingCompilerFromLetTests.kt
```diff
@@ -0,0 +1,204 @@
...
+                expectedSql = """<< {'_1': 100} >>"""),
+                session = session)
+    }
+
+    @Test
+    fun `LET clause using other variables` () {
+        runTestCase(
+            EvaluatorTestCase(
+                query = "SELECT X, Y FROM A LET 1 AS X, X + 1 AS Y",
+                expectedSql = """<< {'X': 1, 'Y': 2} >>"""),
```
therapon, line 85: This is the use case. Great.

lang/test/org/partiql/lang/eval/EvaluatingCompilerFromLetTests.kt
```diff
@@ -0,0 +1,204 @@
...
+            "SELECT foo FROM B LET 100 AS foo GROUP BY B.id",
+            session,
+            ErrorCode.EVALUATOR_VARIABLE_NOT_INCLUDED_IN_GROUP_BY,
+            mapOf(
+                    Property.LINE_NUMBER to 1L,
+                    Property.COLUMN_NUMBER to 8L,
+                    Property.BINDING_NAME to "foo"
+            )
+        )
+    }
```
therapon, line 203: A case where `LET` has multiple bindings and one of them depends on a binding that appears later in the `LET` clause, e.g., 

```
SELECT * FROM X 
LET 1 AS X, 
    Y AS Z, 
    3 AS Y   
```

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
alancai98: In Kotlin, `fold` is a _left_ fold.

Per the [docs](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/fold.html):
"Accumulates value starting with initial value and applying operation from **left** to **right** to current accumulator value and each element."

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
alancai98: I think because `seq` is a Kotlin [Sequence](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.sequences/-sequence/), which returns values through its iterator, the `map` will process the sequence from left to right.

lang/test/org/partiql/lang/eval/EvaluatingCompilerFromLetTests.kt
alancai98: Got it will add.

therapon (approved):

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -887,27 +887,42 @@ internal class EvaluatingCompiler(
...
         return nestCompilationContext(ExpressionContext.NORMAL, emptySet()) {
             val (setQuantifier, projection, from, fromLet, _, groupBy, having, _, metas: MetaContainer) = selectExpr
 
             val fromSourceThunks = compileFromSources(from)
-            val sourceThunks = compileQueryWithoutProjection(selectExpr, fromSourceThunks)
+
+            val letSourceThunks = when(fromLet) {
+                null -> null
+                else -> compileLetSources(fromLet)
+            }
```
dlurton, line 923:
```Kotlin
val letSourcethunks = fromLet?.let { compileLetSources(it) }
```

dlurton (changes requested):

lang/test/org/partiql/lang/eval/EvaluatingCompilerFromLetTests.kt
```diff
@@ -0,0 +1,218 @@
...
+
+    private val session = mapOf("A" to "[ { id : 1 } ]",
+        "B" to "[ { id : 100 }, { id : 200 } ]",
+        "C" to """[ { name: 'foo', region: 'NA' },
+                    { name: 'foobar', region: 'EU' },
+                    { name: 'foobarbaz', region: 'NA' } ]""").toSession()
+
+    // Valid test cases
+    @Test
+    fun `LET clause binding in WHERE` () {
```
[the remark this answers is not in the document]
dlurton: Also note that the example I linked there uses JUnit5.  When that #297 is merged, you can re-base this PR and use parameterization the same way.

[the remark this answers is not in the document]
alancai98: Changed. Also played around with error cases by adding [`EvaluatorErrorTestCase`](https://github.com/partiql/partiql-lang-kotlin/blob/d060f83576c59c16e56420bde677bde6337fc596/lang/test/org/partiql/lang/eval/EvaluatorErrorTestCase.kt) ([method](https://github.com/partiql/partiql-lang-kotlin/blob/d060f83576c59c16e56420bde677bde6337fc596/lang/test/org/partiql/lang/eval/EvaluatorTestBase.kt#L277) called in test) for error cases. Can revise as needed.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -887,27 +887,42 @@ internal class EvaluatingCompiler(
...
                 return accumulator
             }
         }
         val pigGeneratedAst = selectExpr.toAstExpr() as PartiqlAst.Expr.Select
-        val allFromSourceAliases = fold.walkFromSource(pigGeneratedAst.from, emptySet())
+        val allFromSourceAliases = fold.walkFromSource(pigGeneratedAst.from, emptySet()).toMutableSet()
+        if (pigGeneratedAst.fromLet != null) {
+            val letBindingNames = fold.walkLet(pigGeneratedAst.fromLet, emptySet())
+            allFromSourceAliases += letBindingNames
+        }
```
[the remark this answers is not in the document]
alancai98: Agree mutation can be avoided here. Changed.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -1355,13 +1370,24 @@ internal class EvaluatingCompiler(
...
     }
+
+    private fun compileLetSources(letSource: LetSource): List<CompiledLetSource> {
+        val sources = ArrayList<CompiledLetSource>()
+        letSource.bindings.forEach {
+            val thunk = compileExprNode(it.expr)
+            sources.add(CompiledLetSource(name = it.name.name, thunk = thunk))
+        }
+        return sources
+    }
```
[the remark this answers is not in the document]
alancai98: That's definitely a lot cleaner. Changed to your suggestion.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
alancai98: Changed.

dlurton (approved):
LGTM.  Thanks.

PR #298, opened 2020-10-01
Migrate SelectListItemAliasRewriter to PartiqlAst.VisitorTransform

dlurton:
Also:

- Includes some infrastructure needed to allow all of the `AstRewriterBase` implementations to be migrated one by one, namely: `RewriterTransformBridge`.
- Includes some bug-fixes that were related to the `.toAstStatement()` extension function not copying metas over correctly.  This caused some negative unit test failures because line & column numbers of the errors were not being included.  


lang/src/org/partiql/lang/eval/PartiqlAstExtensions.kt
```diff
@@ -0,0 +1,44 @@
+package org.partiql.lang.eval
+
+import com.amazon.ionelement.api.TextElement
+import org.partiql.lang.ast.ExprNode
+import org.partiql.lang.ast.Path
+import org.partiql.lang.ast.VariableReference
+import org.partiql.lang.domains.PartiqlAst
+
```
dlurton, line 8: You may have noticed the `@Deprecated` tag added to `lang/src/org/partiql/lang/eval/ExprNodeExtensions.kt` above... This code essentially duplicates that... I would have deleted that file but it's used by a couple of other `AstRewriterBase` implementations that haven't been migrated yet.  We'll have to remember to delete it when they've been migrated.

lang/src/org/partiql/lang/eval/visitors/SelectListItemAliasVisitorTransform.kt
```diff
@@ -0,0 +1,63 @@
+package org.partiql.lang.eval.visitors
+
+import org.partiql.lang.domains.extractSourceLocation
+
+import org.partiql.lang.domains.PartiqlAst
+import org.partiql.lang.eval.extractColumnAlias
+import org.partiql.pig.runtime.SymbolPrimitive
+
+
+/**
```
dlurton, line 10: This class performs exactly the same function as `SelectListItemAliasRewriter` but works on the PIG-generated `PartiqlAst` instead of `ExprNode`.

lang/test/org/partiql/lang/eval/visitors/SelectListItemAliasVisitorTransformTests.kt
```diff
@@ -0,0 +1,166 @@
...
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ArgumentsSource
+import org.partiql.lang.util.ArgumentsProviderBase
+
+class SelectListItemAliasVisitorTransformTests : VisitorTransformTestBase() {
+
+    class ArgsProvider : ArgumentsProviderBase() {
+        override fun getParameters(): List<Any> = listOf(
+            // singular variable references
+            TransformTestCase(
```
dlurton, line 12: With these tests, it is easy to see visually what the rewriter does.

lang/test/org/partiql/lang/eval/visitors/VisitorTransformTestBase.kt
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+package org.partiql.lang.eval.visitors
+
+import org.junit.jupiter.api.assertDoesNotThrow
+import org.partiql.lang.ast.toAstStatement
+import org.partiql.lang.domains.PartiqlAst
+import org.partiql.lang.syntax.SqlParserTestBase
+
+abstract class VisitorTransformTestBase : SqlParserTestBase() {
```
dlurton, line 22: This class performs exactly the same function as `RewriterTestBase` but is shorter since PIG-generated domain models are nicer to work with than `ExprNode`.

lang/src/org/partiql/lang/eval/visitors/SelectListItemAliasVisitorTransform.kt
```diff
@@ -0,0 +1,63 @@
+package org.partiql.lang.eval.visitors
```
dlurton, line 1: Note that all of the new `VistiorTransform` implementations live in a different package than the `AstRewriterBase` implementations.

lang/test/org/partiql/lang/eval/visitors/VisitorTransformTestBase.kt
dlurton: Note to self: add `kdoc`.

lang/test/org/partiql/lang/eval/visitors/SelectListItemAliasVisitorTransformTests.kt
```diff
@@ -0,0 +1,166 @@
...
+                FROM foo
+                """,
+                """
+                SELECT 
+                    x.a AS a,
+                    x.B AS B,
+                    x.c AS c 
+                FROM foo
+                """
+            ),
```
alancai98, line 96: Could add a multi-nested test. Something like:

```
TransformTestCase(
	"""
	SELECT 
	    x.foo.a,
	    x.bar.B
	FROM foo
	""",
	"""
	SELECT 
	    x.foo.a AS a,
	    x.bar.B AS B
	FROM foo
	"""
```

lang/src/org/partiql/lang/ast/ExprNodeToStatement.kt
```diff
@@ -19,12 +20,12 @@ fun ExprNode.toAstStatement(): PartiqlAst.Statement {
     }
 }
 
-private fun PartiQlMetaContainer.toElectrolyteMetaContainer(): ElectrolyteMetaContainer =
+internal fun org.partiql.lang.ast.MetaContainer.toIonElementMetaContainer(): IonElementMetaContainer =
```
alancai98, line 23: Is the `org.partiql.lang.ast.` part necessary?

lang/src/org/partiql/lang/eval/visitors/SelectListItemAliasVisitorTransform.kt
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * ```
+ *  SELECT
+ *      foo AS foo,
+ *      b.bat AS bat,
+ *      baz + 1 AS _3
+ * FROM bar AS b
+ * ```
+ *
+ * If provided with a query with all of the select list aliases are already specified, an exact clone is returned.
```
therapon, line 33: let's open an issue to opt-out when we have a query that already has AS aliases defined.

lang/test/org/partiql/lang/eval/visitors/SelectListItemAliasVisitorTransformTests.kt
therapon: I would also add a nested query. Do we deal with already aliased names in the outer query that appear in the nested query? Do we also make sure we do not accidentally capture (shadow) names?

dlurton: done

lang/src/org/partiql/lang/ast/ExprNodeToStatement.kt
dlurton: No, I was trying to fix a build error and left this behind by accident.  Removing...

lang/src/org/partiql/lang/eval/visitors/SelectListItemAliasVisitorTransform.kt
dlurton: I'm not clear to me who is doing the opting out.  The partiql service?  Or should the `VisitorTransform` implementation itself detect if there is work to be done?  Detecting if there is work to be done for all input queries may be more expensive than just doing the rewrite... However, if the user knows they've supplied a query with all of the `AS` aliases specified then is your suggestion to provide a way for the user to cause this transform to be skipped?

alancai98 (approved):

lang/src/org/partiql/lang/eval/visitors/SelectListItemAliasVisitorTransform.kt
therapon: Not blocking. I am not sure about your statement "Detecting if there is work to be done for all input queries may be more expensive than just doing the rewrite". Wouldn't it be cheaper to grab all FROM nodes and check that each input has an `AS` name rather than copying trees?

Another idea is to add to our parser the task to keep track of aliases in FROM clauses and pass a flag as metadata on the root node.

therapon (approved):

PR #300, opened 2020-10-01
Fix `LIMIT` clause execution order

### LINKED ISSUE #281: Incorrect execution order for `LIMIT` clause during query execution

alancai98:
Fixes #281

### DIFF
--- a/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
+++ b/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
@@ -889,6 +889,49 @@ internal class EvaluatingCompiler(
         }
     }
 
+    private fun compileLimit(limitClause: ExprNode, env: Environment) : Long {
+        val limitThunk = compileExprNode(limitClause)
+        val limitLocationMeta = limitClause.metas.sourceLocationMeta
+        val limitExprValue = limitThunk(env)
+
+        if(limitExprValue.type != ExprValueType.INT) {
+            err("LIMIT value was not an integer",
+                ErrorCode.EVALUATOR_NON_INT_LIMIT_VALUE,
+                errorContextFrom(limitLocationMeta).also {
+                    it[Property.ACTUAL_TYPE] = limitExprValue.type.toString()
+                },
+                internal = false)
+        }
+
+        // `Number.toLong()` (used below) does *not* cause an overflow exception if the underlying [Number]
+        // implementation (i.e. Decimal or BigInteger) exceeds the range that can be represented by Longs.
+        // This can cause very confusing behavior if the user specifies a LIMIT value that exceeds
+        // Long.MAX_VALUE, because no results will be returned from their query.  That no overflow exception
+        // is thrown is not a problem as long as PartiQL's restriction of integer values to +/- 2^63 remains.
+        // We throw an exception here if the value exceeds the supported range (say if we change that
+        // restriction or if a custom [ExprValue] is provided which exceeds that value).
+        val limitIonValue = limitExprValue.ionValue as IonInt
+        if(limitIonValue.integerSize == IntegerSize.BIG_INTEGER) {
+            err("IntegerSize.BIG_INTEGER not supported for LIMIT values",
+                errorContextFrom(limitLocationMeta),
+                internal = true)
+        }
+
+        val limitValue = limitExprValue.numberValue().toLong()
+
+        if (limitValue < 0) {
+            err("negative LIMIT",
+                ErrorCode.EVALUATOR_NEGATIVE_LIMIT,
+                errorContextFrom(limitLocationMeta),
+                internal = false)
+        }
+
+        // we can't use the Kotlin's Sequence<T>.take(n) for this since it accepts only an integer.
+        // this references [Sequence<T>.take(count: Long): Sequence<T>] defined in [org.partiql.util].
+        return limitValue
+    }
+
+
     private fun compileSelect(selectExpr: Select): ThunkEnv {
         // Get all the FROM source aliases for binding error checks
         val fold = object : PartiqlAst.VisitorFold<Set<String>>() {
@@ -907,7 +950,7 @@ internal class EvaluatingCompiler(
         val allFromSourceAliases = fold.walkFromSource(pigGeneratedAst.from, emptySet())
 
         return nestCompilationContext(ExpressionContext.NORMAL, emptySet()) {
-            val (setQuantifier, projection, from, _, groupBy, having, _, metas: MetaContainer) = selectExpr
+            val (setQuantifier, projection, from, _, groupBy, having, limit, metas: MetaContainer) = selectExpr
 
             val fromSourceThunks = compileFromSources(from)
             val sourceThunks = compileQueryWithoutProjection(selectExpr, fromSourceThunks)
@@ -928,10 +971,14 @@ internal class EvaluatingCompiler(
                                 selectProjectionThunk(projectEnv, joinedValues)
                             }
 
-                            val quantifiedRows = when(setQuantifier) {
+                            var quantifiedRows = when(setQuantifier) {
                                 // wrap the ExprValue to use ExprValue.equals as the equality
                                 SetQuantifier.DISTINCT -> projectedRows.filter(createUniqueExprValueFilter())
                                 SetQuantifier.ALL -> projectedRows
+                            }
+
+                            if (limit != null) {
+                                quantifiedRows = quantifiedRows.take(compileLimit(limit, env))
                             }
 
                             valueFactory.newBag(quantifiedRows.map {
@@ -974,13 +1021,16 @@ internal class EvaluatingCompiler(
                                 // Create a closure that groups all the rows in the FROM source into a single group.
                                 thunkFactory.thunkEnv(metas) { env ->
                                     // Evaluate the FROM clause
-                                    val fromProductions: Sequence<FromProduction> = sourceThunks(env)
+                                    var fromProductions: Sequence<FromProduction> = sourceThunks(env)
                                     val registers = createRegisterBank()
 
                                     // note: the group key can be anything here because we only ever have a single
                                     // group when aggregates are used without GROUP BY expression
                                     val syntheticGroup = Group(valueFactory.nullValue, registers)
 
+                                    if (limit != null) {
+                                        fromProductions = fromProductions.take(compileLimit(limit, env))
+                                    }
                                     // iterate over the values from the FROM clause and populate our
                                     // aggregate register values.
                                     fromProductions.forEach { fromProduction ->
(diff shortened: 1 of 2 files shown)


lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -889,6 +889,49 @@ internal class EvaluatingCompiler(
         }
     }
 
+    private fun compileLimit(limitClause: ExprNode, env: Environment) : Long {
```
alancai98, line 892: This was mostly moved from [`compileQueryWithoutProjection `](https://github.com/partiql/partiql-lang-kotlin/blob/master/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt#L1365).

therapon (approved):

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -928,12 +971,16 @@ internal class EvaluatingCompiler(
...
-                            val quantifiedRows = when(setQuantifier) {
+                            var quantifiedRows = when(setQuantifier) {
                                 // wrap the ExprValue to use ExprValue.equals as the equality
                                 SetQuantifier.DISTINCT -> projectedRows.filter(createUniqueExprValueFilter())
                                 SetQuantifier.ALL -> projectedRows
                             }
 
+                            if (limit != null) {
+                                quantifiedRows = quantifiedRows.take(compileLimit(limit, env))
+                            }
```
abhikuhikar, line 982: Can it be done using `let` to avoid using `var`?

```suggestion
                            val quantifiedRows = when(setQuantifier) {
                                // wrap the ExprValue to use ExprValue.equals as the equality
                                SetQuantifier.DISTINCT -> projectedRows.filter(createUniqueExprValueFilter())
                                SetQuantifier.ALL -> projectedRows
                            }.let { rows ->
                                when (limit) {
                                    null -> rows
                                    else -> rows.take(compileLimit(limit, env))
                                }
                            }
```

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -974,13 +1021,16 @@ internal class EvaluatingCompiler(
...
+                                    var fromProductions: Sequence<FromProduction> = sourceThunks(env)
                                     val registers = createRegisterBank()
 
                                     // note: the group key can be anything here because we only ever have a single
                                     // group when aggregates are used without GROUP BY expression
                                     val syntheticGroup = Group(valueFactory.nullValue, registers)
 
+                                    if (limit != null) {
+                                        fromProductions = fromProductions.take(compileLimit(limit, env))
+                                    }
```
abhikuhikar, line 1033: Can it be done using `let` to avoid using `var`?

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -1085,9 +1139,13 @@ internal class EvaluatingCompiler(
                         val asThunk = compileExprNode(asExpr)
                         val atThunk = compileExprNode(atExpr)
                         thunkFactory.thunkEnv(metas) { env ->
-                            val sourceValue = sourceThunks(env)
+                            var sourceValue = sourceThunks(env).asSequence()
+
+                            if (limit != null) {
+                                sourceValue = sourceValue.take(compileLimit(limit, env))
+                            }
+
```
abhikuhikar, line 1147: Here as well.

abhikuhikar (changes requested):

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
dlurton: +1

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
dlurton: +1

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
dlurton: +1

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -928,12 +971,16 @@ internal class EvaluatingCompiler(
...
 
-                            val quantifiedRows = when(setQuantifier) {
+                            var quantifiedRows = when(setQuantifier) {
                                 // wrap the ExprValue to use ExprValue.equals as the equality
                                 SetQuantifier.DISTINCT -> projectedRows.filter(createUniqueExprValueFilter())
                                 SetQuantifier.ALL -> projectedRows
                             }
 
+                            if (limit != null) {
+                                quantifiedRows = quantifiedRows.take(compileLimit(limit, env))
```
dlurton, line 981: Since this is inside a call to `thunkFactory.thunkEnv(metas)` (line 968), the call to `compileLimit` is being invoked at *evaluation-time*, which means the limit clause is being re-compiled for every row.  To fix this move the `compileLimit` call outside of the closure, i.e. to 964.

dlurton: This is an important aspect of how `EvaluatingCompiler` works: a significant portion of the the code in it  executes at evaluation-time and not-compile time because it resides in closures (thunks) such as this one.  It is important to avoid re-compiling parts of expressions within the body of a thunk.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -974,13 +1021,16 @@ internal class EvaluatingCompiler(
...
-                                    val fromProductions: Sequence<FromProduction> = sourceThunks(env)
+                                    var fromProductions: Sequence<FromProduction> = sourceThunks(env)
                                     val registers = createRegisterBank()
 
                                     // note: the group key can be anything here because we only ever have a single
                                     // group when aggregates are used without GROUP BY expression
                                     val syntheticGroup = Group(valueFactory.nullValue, registers)
 
+                                    if (limit != null) {
+                                        fromProductions = fromProductions.take(compileLimit(limit, env))
```
dlurton, line 1032: Move `compileLimit` outside of thunk body.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
dlurton, line 892: In general, any `compile*` function should return a thunk.  This appears to be compiling and evaluating.  Compilation and evaluation should be separate steps.

dlurton (changes requested):
Need to avoid re-compiling the LIMIT clause by moving it's thunk creations outside of the body of all of its thunks.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
alancai98: Thanks, updated w/ the suggestion.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
alancai98: Updated

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
alancai98: Updated

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
alancai98: I separated out the compilation ([L955](https://github.com/partiql/partiql-lang-kotlin/blob/ad2aa3e73f549c3ff66605d254c2c5b7dfdbc277/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt#L955)) from the evaluation ([L892](https://github.com/partiql/partiql-lang-kotlin/blob/ad2aa3e73f549c3ff66605d254c2c5b7dfdbc277/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt#L892)) so the compilation isn't done for every row.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
alancai98: In the new commit, I separated out the compilation ([L955](https://github.com/partiql/partiql-lang-kotlin/blob/ad2aa3e73f549c3ff66605d254c2c5b7dfdbc277/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt#L955)) from the evaluation ([L892](https://github.com/partiql/partiql-lang-kotlin/blob/ad2aa3e73f549c3ff66605d254c2c5b7dfdbc277/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt#L892)) so the compilation isn't done from within the thunk evaluation.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
alancai98: In the new commit, I separated out the compilation ([L955](https://github.com/partiql/partiql-lang-kotlin/blob/ad2aa3e73f549c3ff66605d254c2c5b7dfdbc277/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt#L955)) from the evaluation ([L892](https://github.com/partiql/partiql-lang-kotlin/blob/ad2aa3e73f549c3ff66605d254c2c5b7dfdbc277/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt#L892)).

abhikuhikar (dismissed):

therapon (dismissed):

therapon (approved):

therapon (approved):

dlurton (approved):

PR #302, opened 2020-10-05
Migrate FromSourceAliasRewriter to PartiqlAst.VisitorTransform

alancai98:
~#298 PR should be merged into `visitor-transforms` first (contains changes to metas and adds the `VisitorTransform`-related classes).~ (has already been merged, so this can be looked at)

Migrates `FromSourceAliasRewriter` to `PartiqlAst.VisitorTransform`. Conveniently, there were previous tests for `FromSourceAliasRewriter`.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #304, opened 2020-10-06
Migrate GroupByItemAliasRewriter to PartiqlAst.VisitorTransform

alancai98:
~Before this can be reviewed, will need #298 to be merged into the `visitor-transforms` feature branch first. (Should also merge #302 before).~

Written tests (`GroupByPathExpressionRewriterTest`) are new, so will need to be reviewed.

This PR should be merged after the `FromSourceAlias` PR (#302). It can still be looked at though.

TODO:
- [x] Remove `GroupByItemAliasRewriter` and mentions in code


lang/test/org/partiql/lang/eval/visitors/GroupByItemAliasVisitorTransformTests.kt
```diff
@@ -0,0 +1,210 @@
+package org.partiql.lang.eval.visitors
+
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ArgumentsSource
+import org.partiql.lang.util.ArgumentsProviderBase
+
+class GroupByItemAliasVisitorTransformTests : VisitorTransformTestBase() {
```
alancai98, line 7: These are new tests, largely based off the tests from #298.

lang/src/org/partiql/lang/eval/visitors/GroupByItemAliasVisitorTransform.kt
```diff
@@ -0,0 +1,84 @@
...
+    override fun transformGroupBy(node: PartiqlAst.GroupBy): PartiqlAst.GroupBy {
+        return PartiqlAst.build {
+            groupBy_(
+                strategy = node.strategy,
+                keyList = PartiqlAst.GroupKeyList(node.keyList.keys.mapIndexed { index, it ->
+                    val aliasText = it.asAlias?.text ?: it.expr.extractColumnAlias(index)
+                    var metas = it.expr.metas + metaContainerOf(UniqueNameMeta("\$__partiql__group_by_${nestLevel}_item_$index")).toIonElementMetaContainer()
+                    if (it.asAlias == null) {
+                        metas = metas + metaContainerOf(IsSyntheticNameMeta.instance).toIonElementMetaContainer()
+                    }
```
alancai98, line 48: Not sure of the best way to add metas using the PIG/IonElement `MetaContainer`. The approach directly using IonElement's `MetaContainer`'s`metaContainerOf` caused some errors.

alancai98: Nvm found a way. Was previously trying to pass the value as a string rather than an actual meta. Changing that fixed the previous errors. Will change in the other PRs.

lang/src/org/partiql/lang/eval/visitors/GroupByItemAliasVisitorTransform.kt
```diff
@@ -0,0 +1,85 @@
...
+                from = new_from,
+                fromLet = new_fromLet,
+                where = new_where,
+                group = new_group,
+                having = new_having,
+                limit = new_limit,
+                metas = new_metas
+            )
+        }
+    }
```
dlurton, line 84: Would it not be possible to this instead:

```Kotlin
override fun transformExprSelect(node: PartiqlAst.Expr.Select): PartiqlAst.Expr {
    nestLevel++
    return super.transformExprSelect(node).also { nestLevel-- }
}
```
?

lang/test/org/partiql/lang/eval/visitors/GroupByItemAliasVisitorTransformTests.kt
dlurton: Great!

dlurton (changes requested):
One possible thing to change, otherwise LGTM.

lang/src/org/partiql/lang/eval/visitors/GroupByItemAliasVisitorTransform.kt
alancai98: Thanks, that works. I will see if that can also work for similar cases in #318.

dlurton (approved):

PR #305, opened 2020-10-08
Migrate AggregateSupportRewriter to PartiqlAst.VisitorTransform and ExprNode `GroupBy`-> PIG `GroupBy` bug

alancai98:
Migrate AggregateSupportRewriter to PartiqlAst.VisitorTransform.

Also fixes bug in `ExprNodeToStatement` when converting ExprNode `GroupBy` into PIG `GroupBy`. Previously, metas part of a given `groupKey` were not included, causing some missing meta errors when calling `.toAstStatement` with an ExprNode that had a GroupBy clause.

Need to resolve whether a `CallAgg` used in the evaluator should be stored as an ExprNode or PIG version when doing the VisitorTransform (i.e. when should the `CallAgg` be converted into an ExprNode `CallAgg`, in the VisitorTransform, evaluator, or somewhere else?). This current PR does the conversion in the evaluator and leaves the ExprNode concept out of the VisitorTransform. However, it may actually be better to do it in the VisitorTransform (still need to try it out).

Also need to discuss if there's a better way to convert parts of the `CallAgg` from the PIG version to the ExprNode version. In the current approach, I had to expose some methods from StatementToExprNode.kt to get this working.

Lastly, ran into an issue where an existing [test](https://github.com/partiql/partiql-lang-kotlin/blob/774863f73d06f7065166801c62bff4bff46c56e0/lang/test/org/partiql/lang/eval/EvaluatingCompilerExceptionsTest.kt#L62), `selectListNestedAggregateCall`, may have incorrect behavior? It would make more sense to me that the error's source location should be column 12, which is where the nested aggregate starts. This could be a potential bug in using the wrong metas in the [evaluator](https://github.com/partiql/partiql-lang-kotlin/blob/774863f73d06f7065166801c62bff4bff46c56e0/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt#L1205).


lang/src/org/partiql/lang/ast/StatementToExprNode.kt
```diff
@@ -27,6 +27,28 @@ internal typealias IonElementMetaContainer = com.amazon.ionelement.api.MetaConta
...
+        // We may need to account for this in the future, but for now we require that all metas placed
+        // on any `partiql_ast` instances to implement Meta.  It's not clear how to deal with that now
+        // so we should wait until it's needed.
+        val partiQlMeta = it as? Meta ?: error("The meta was not an instance of Meta.")
+        partiQlMeta
+    }
+
+    return org.partiql.lang.ast.metaContainerOf(nonLocationMetas)
+}
+
```
alancai98, line 51: May need some guidance on a better approach to convert a PIG `CallAgg` to an ExprNode `CallAgg` without exposing these methods.

lang/src/org/partiql/lang/ast/AggregateCallSiteListMeta.kt
```diff
@@ -15,11 +15,12 @@
 package org.partiql.lang.ast
 
 import com.amazon.ion.*
+import org.partiql.lang.domains.PartiqlAst
 
 /**
  * Contains references to each of the aggregate call-sites in a given [Select].
  */
-data class AggregateCallSiteListMeta(val aggregateCallSites: List<CallAgg>): Meta {
+data class AggregateCallSiteListMeta(val aggregateCallSites: List<PartiqlAst.Expr.CallAgg>): Meta {
```
alancai98, line 23: This is the crux of the main issue. Need to determine if this `Meta` should store a `CallAgg` in the PIG/PartiqlAst form or ExprNode form. Depending on what we decide, we would have to do the conversion in the evaluator (before [L45](https://github.com/partiql/partiql-lang-kotlin/blob/d5b2152cb04ba65636c66294d4dd447187cf3f81/lang/src/org/partiql/lang/eval/visitors/AggregateSupportVisitorTransform.kt#L45)) or in the VisitorTransform (current version).

alancai98:
Alternate version converting the `CallAgg` from the VisitorTransform can be found on the `visitor-transforms-aggregate-support-alternate` [branch](https://github.com/partiql/partiql-lang-kotlin/tree/visitor-transforms-aggregate-support-alternate). 

This [commit](https://github.com/partiql/partiql-lang-kotlin/commit/6abf9249d5c5e45df747dc43a9460fd460d00fb0) shows how converting may look from the VisitorTransform, leaving the evaluator the same (i.e. using the ExprNode type for `CallAgg`).

Wasn't sure how to get around passing the `IonSystem` to the VisitorTransform since the `valueFactory` must have the same instance of `IonSystem` as `ionValue`

lang/src/org/partiql/lang/ast/ExprNodeToStatement.kt
```diff
@@ -159,7 +160,10 @@ private fun GroupBy.toAstGroupSpec(): PartiqlAst.GroupBy =
     PartiqlAst.build {
         groupBy_(
             this@toAstGroupSpec.grouping.toAstGroupStrategy(),
-            groupKeyList(this@toAstGroupSpec.groupByItems.map { groupKey_(it.expr.toAstExpr(), it.asName?.name?.asPrimitive()) }),
+            groupKeyList(this@toAstGroupSpec.groupByItems.map {
+                val keyMetas = it.asName?.metas?.toIonElementMetaContainer() ?: emptyMetaContainer()
+                groupKey_(it.expr.toAstExpr(), it.asName?.name?.asPrimitive(keyMetas) )
+            }),
```
alancai98, line 166: This is a bug fix where the metas were not being passed properly to the `groupBy_`

lang/src/org/partiql/lang/eval/visitors/AggregateSupportVisitorTransform.kt
```diff
@@ -0,0 +1,90 @@
...
+            projectExpr_(
+                expr = registerIdAdder.transformExpr(node.expr),
+                asAlias = node.asAlias,
+                metas = node.metas
+            )
+        }
+
+    /**
+     * Applies a new instance of [AggregateSupportVisitorTransform]
+     * to [SelectProjectionValue] nodes so that a new they different instance of [aggregateCallSites].
```
dlurton, line 73: "...to [SelectProjectionValue] nodes so that a different instance of [aggregateCallSites] is used."

lang/src/org/partiql/lang/eval/visitors/AggregateSupportVisitorTransform.kt
```diff
@@ -0,0 +1,90 @@
...
+ * and stores them in an instance of [AggregateCallSiteListMeta] on the [PartiqlAst.Expr.Select] instance.
+ */
+
+class AggregateSupportVisitorTransform : PartiqlAst.VisitorTransform() {
+    private val aggregateCallSites = ArrayList<PartiqlAst.Expr.CallAgg>()
+
+    inner class RegisterIdAdderSubVisitorTransform : PartiqlAst.VisitorTransform() {
+        /**
+         * Nests another [AggregateSupportVisitorTransform] within this transform
+         * in order to avoid operating on [aggregateCallSites] of a nested SELECT.
```
dlurton, line 38: "...in order to avoid exposing [aggregateCallSites] to subqueries."

lang/src/org/partiql/lang/ast/ExprNodeToStatement.kt
dlurton: :thumbsup:

lang/src/org/partiql/lang/ast/AggregateCallSiteListMeta.kt
dlurton: The `PartiqlAst` form here is desirable.

dlurton (approved):
LGTM in general, the only changes I recommend are to my own poor wording in the `kdoc` I originally wrote.

dlurton:
> Lastly, ran into an issue where an existing [test](https://github.com/partiql/partiql-lang-kotlin/blob/774863f73d06f7065166801c62bff4bff46c56e0/lang/test/org/partiql/lang/eval/EvaluatingCompilerExceptionsTest.kt#L62), `selectListNestedAggregateCall`, may have incorrect behavior? It would make more sense to me that the error's source location should be column 12, which is where the nested aggregate starts. This could be a potential bug in using the wrong metas in the [evaluator](https://github.com/partiql/partiql-lang-kotlin/blob/774863f73d06f7065166801c62bff4bff46c56e0/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt#L1205).

Yes, column 12 seems to be the correct location for that error.

lang/src/org/partiql/lang/eval/visitors/AggregateSupportVisitorTransform.kt
alancai98: Fixed.

lang/src/org/partiql/lang/eval/visitors/AggregateSupportVisitorTransform.kt
alancai98: Fixed.

lang/src/org/partiql/lang/eval/visitors/AggregateSupportVisitorTransform.kt
```diff
@@ -0,0 +1,90 @@
...
+
+        override fun transformExprCallAgg(node: PartiqlAst.Expr.CallAgg): PartiqlAst.Expr {
+            aggregateCallSites.add(node)
+
+            return PartiqlAst.build {
+                callAgg_(
+                    setq = node.setq,
+                    funcName = node.funcName,
+                    arg = transformExpr(node.arg),
+                    metas = transformMetas(node.metas) + metaContainerOf(AggregateRegisterIdMeta.TAG to AggregateRegisterIdMeta(aggregateCallSites.size - 1))
```
alancai98, line 51: To fix this [test](https://github.com/partiql/partiql-lang-kotlin/blob/774863f73d06f7065166801c62bff4bff46c56e0/lang/test/org/partiql/lang/eval/EvaluatingCompilerExceptionsTest.kt#L62), had to change to use `node.metas` rather than `node.arg.metas`. It makes more sense that the `CallAgg`'s source location meta use the source location of where the aggregation function starts rather than where the argument to the function starts.

PR #308, opened 2020-10-10
prepare 0.2.4-SNAPSHOT

therapon:
*Description of changes:*

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,7 +38,7 @@ allprojects {
 
 subprojects {
     group = 'org.partiql'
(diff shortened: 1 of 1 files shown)


alancai98 (commented):
Does the README.md version also need to change?

therapon:
> Does the README.md version also need to change?

We do not typically. The README is what people read to use the released version. The code on `master` on the other hand is not realsed yet and to avoid accidentally publishing to maven we add the snapshot suffix

alancai98 (approved):

PR #310, opened 2020-10-13
Migrate SelectStarRewriter to PartiqlAst.VisitorTransform

alancai98:
Migrates SelectStarRewriter to PartiqlAst.VisitorTransform.

Also had to fix the ExprNode GroupBy -> PIG GroupBy conversion bug mentioned in #305 


lang/src/org/partiql/lang/ast/ExprNodeToStatement.kt
```diff
@@ -159,7 +160,10 @@ private fun GroupBy.toAstGroupSpec(): PartiqlAst.GroupBy =
     PartiqlAst.build {
         groupBy_(
             this@toAstGroupSpec.grouping.toAstGroupStrategy(),
-            groupKeyList(this@toAstGroupSpec.groupByItems.map { groupKey_(it.expr.toAstExpr(), it.asName?.name?.asPrimitive()) }),
+            groupKeyList(this@toAstGroupSpec.groupByItems.map {
+                val keyMetas = it.asName?.metas?.toIonElementMetaContainer() ?: emptyMetaContainer()
+                groupKey_(it.expr.toAstExpr(), it.asName?.name?.asPrimitive(keyMetas) )
+            }),
```
alancai98, line 166: Copied from #305.

lang/src/org/partiql/lang/eval/visitors/SelectStarVisitorTransform.kt
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    override fun transformExprSelect(node: PartiqlAst.Expr.Select): PartiqlAst.Expr {
+        val rewritten = super.transformExprSelect(node) as PartiqlAst.Expr.Select
+
+        val projection = rewritten.project
+
+        // Check if SELECT * is being used.
+        if (projection is PartiqlAst.Projection.ProjectStar) {
```
alancai98, line 34: This check is slightly different than the ExprNode rewriter version. In that version, a `SelectProjection` could be a `SelectProjectionList`, which has a list of `SelectListItem`. Of these items, it can be a `SelectListItemExpr`, `SelectListItemProjectAll`, or **`SelectListItemStar`**.

While in PIG, a `ProjectStar` (~ `SelectListItemStar`) can directly be a `Projection` (~ `SelectProjection`). So not sure if we need the checks from the original [code](https://github.com/partiql/partiql-lang-kotlin/blob/80e8665bd167e98ad7967eb4fda096948c18a028/lang/src/org/partiql/lang/ast/passes/SelectStarRewriter.kt#L27).

lang/src/org/partiql/lang/eval/visitors/SelectStarVisitorTransform.kt
```diff
@@ -0,0 +1,117 @@
...
+                project = newProjection,
+                from = node.from,
+                fromLet = node.fromLet,
+                where = node.where,
+                group = node.group,
+                having = node.having,
+                limit = node.limit,
+                metas = node.metas)
+        }
+    }
```
alancai98, line 26: Again, not sure if there's a better way to make a copy of a PIG type with one parameter changed.

lang/test/org/partiql/lang/eval/visitors/SelectStarVisitorTransformTests.kt
```diff
@@ -1,12 +1,16 @@
...
+import org.partiql.lang.ast.passes.RewriterTestBase
+import org.partiql.lang.ast.passes.RewriterTransformBridge
 
-class SelectStarRewriterTests : RewriterTestBase() {
+class SelectStarVisitorTransformTests : RewriterTestBase() {
 
-    // SelectStar rewriter depends on UniqueNameMetas injected by GroupByItemAliasRewriter.
-    private val rewriters = PipelinedRewriter(GroupByItemAliasRewriter(), SelectStarRewriter())
+    // SelectStar visitor transform depends on UniqueNameMetas injected by GroupByItemAliasVisitorTransform.
+    private val rewriters = PipelinedRewriter(GroupByItemAliasRewriter(), RewriterTransformBridge(SelectStarVisitorTransform(), ion))
```
alancai98, line 13: After `GroupByItemAliasRewriter`'s PR (#304) is merged into the `visitor-transform` branch along with a VisitorTransform version of `PipelinedRewriter`, will update these tests to use `TransformTestCase` rather than `RewriterTestCase`.

lang/src/org/partiql/lang/eval/visitors/SelectStarVisitorTransform.kt
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+    private fun createProjectExpr(variableName: String, asAlias: String = variableName) =
+        PartiqlAst.build {
+            projectExpr(id(variableName, caseSensitive(), unqualified(), emptyMetaContainer()), asAlias)
+        }
+
+    private class FromSourceAliases(val asAlias: String, val atAlias: String?, val byAlias: String?)
+
+    private fun extractAliases(fromSource: PartiqlAst.FromSource): List<FromSourceAliases> =
```
alancai98, line 92: This follows what the previous rewriter did. Perhaps, we could use a PIG visitor to perform this like what is already used in EvaluatingCompiler ([L894](https://github.com/partiql/partiql-lang-kotlin/blob/36fe55b6457f0dc82c02da6975358c01dd13aa3b/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt#L894)).

lang/src/org/partiql/lang/eval/visitors/SelectStarVisitorTransform.kt
dlurton: For the time being, this is a reasonable work-around.  Please add a link to https://github.com/partiql/partiql-ir-generator/issues/52 on line 14 to remind us to change this to use the generated `.copy` function.

lang/src/org/partiql/lang/eval/visitors/SelectStarVisitorTransform.kt
dlurton: Yes, the `PartiqlAst` models `SELECT *` better than `ExprNode` does...  The result of the better modeling is that those checks are in fact not needed here.

lang/test/org/partiql/lang/eval/visitors/SelectStarVisitorTransformTests.kt
dlurton: Cool.

lang/src/org/partiql/lang/eval/visitors/SelectStarVisitorTransform.kt
dlurton: I would encourage you to do that as part of this PR, but it's not *strictly* required.

dlurton (changes requested):

lang/src/org/partiql/lang/eval/visitors/SelectStarVisitorTransform.kt
alancai98: Added.

lang/src/org/partiql/lang/eval/visitors/SelectStarVisitorTransform.kt
alancai98: Changed to use PIG visitor.

lang/test/org/partiql/lang/eval/visitors/SelectStarVisitorTransformTests.kt
```diff
@@ -1,13 +1,14 @@
...
+import org.partiql.lang.ast.passes.RewriterTestBase
+import org.partiql.lang.ast.passes.RewriterTransformBridge
 
-class SelectStarRewriterTests : RewriterTestBase() {
-
-    // SelectStar rewriter depends on UniqueNameMetas injected by GroupByItemAliasVisitorTransform.
-    private val rewriters = PipelinedRewriter(RewriterTransformBridge(GroupByItemAliasVisitorTransform(), ion), SelectStarRewriter())
+class SelectStarVisitorTransformTests : RewriterTestBase() {
+    // SelectStar visitor transform depends on UniqueNameMetas injected by GroupByItemAliasVisitorTransform.
+    private val rewriters = PipelinedRewriter(RewriterTransformBridge(GroupByItemAliasVisitorTransform(), ion), RewriterTransformBridge(SelectStarVisitorTransform(), ion))
```
alancai98, line 11: These tests will still use `Rewriter` and `RewriterTestBase` until `PipelinedRewriter` is migrated to PIG visitor transforms. These tests will use visitor transforms as part of the `PipelinedRewriter` migration PR.

lang/test/org/partiql/lang/eval/visitors/SelectStarVisitorTransformTests.kt
dlurton: Have you done this yet?

dlurton (approved):

lang/test/org/partiql/lang/eval/visitors/SelectStarVisitorTransformTests.kt
alancai98: Not yet, but now that this PR is approved, I can finish up the `PipelinedRewriter` migration [PR](https://github.com/partiql/partiql-lang-kotlin/pull/320), which will update these tests and remove other rewriter references.

PR #313, opened 2020-10-14
Strictness "handler" WIP

dlurton:
Preliminary version for early feedback.

The only API change this represents is the new property to `CompileOptions`.

### DIFF
--- a/lang/src/org/partiql/lang/eval/CompileOptions.kt
+++ b/lang/src/org/partiql/lang/eval/CompileOptions.kt
@@ -14,7 +14,6 @@
 
 package org.partiql.lang.eval
 
-import org.partiql.lang.ast.SourceLocationMeta
 import org.partiql.lang.ast.passes.AstRewriter
 import org.partiql.lang.ast.passes.IDENTITY_REWRITER
 import org.partiql.lang.ast.passes.basicRewriters
@@ -41,6 +40,12 @@ enum class ProjectionIterationBehavior {
     FILTER_MISSING, UNFILTERED
 }
 
+enum class StrictnessMode {
+    STANDARD,
+    PERMISSIVE,
+    //TODO: STRICT
+}
+
 /**
  * Controls the behavior of intrinsic AST rewriting with [EvaluatingCompiler.compile].
  *
@@ -65,6 +70,7 @@ data class CompileOptions private constructor (
         val undefinedVariable: UndefinedVariableBehavior,
         val projectionIteration: ProjectionIterationBehavior = ProjectionIterationBehavior.FILTER_MISSING,
         val rewritingMode: RewritingMode = RewritingMode.DEFAULT,
+        val strictnessMode: StrictnessMode = StrictnessMode.STANDARD,
         val thunkOptions: ThunkOptions = ThunkOptions.standard()
 ) {
 
@@ -98,6 +104,7 @@ data class CompileOptions private constructor (
         fun projectionIteration(value: ProjectionIterationBehavior) = set { copy(projectionIteration = value) }
         fun rewriterMode(value: RewritingMode) = set { copy(rewritingMode = value) }
         fun thunkOptions(value: ThunkOptions) = set { copy(thunkOptions = value)}
+        fun strictness(value: StrictnessMode) = set { copy(strictnessMode = value)}
         private inline fun set(block: CompileOptions.() -> CompileOptions) : Builder {
             options = block(options)
             return this
--- a/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
+++ b/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
@@ -57,6 +57,7 @@ internal class EvaluatingCompiler(
     private val compileOptions: CompileOptions = CompileOptions.standard()
 ) {
     private val thunkFactory = ThunkFactory(compileOptions.thunkOptions)
+    private val strictnessContext = StrictnessContext(compileOptions.strictnessMode, valueFactory)
     private val compilationContextStack = Stack<CompilationContext>()
 
     private val currentCompilationContext: CompilationContext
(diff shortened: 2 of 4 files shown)


lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -429,16 +430,18 @@ internal class EvaluatingCompiler(
...
+                test = denominator.isZero(),
+                detailsConstructor = { ErrorDetails(metas, "/ by zero", ErrorCode.EVALUATOR_DIVIDE_BY_ZERO) }
+            ) {
+                try {
+                    (lValue.numberValue() / denominator).exprValue()
+                } catch (e: ArithmeticException) {
+                    // Setting the internal flag as true as it is not clear what
+                    // ArithmeticException may be thrown by the above
+                    throw EvaluationException(cause = e, internal = true)
+                }
```
dlurton, line 444: Demonstrates intended usage.

lang/test/org/partiql/lang/eval/StrictnessTests.kt
```diff
@@ -0,0 +1,11 @@
+package org.partiql.lang.eval
+
+class StrictnessTests {
```
dlurton, line 3: I would like to put all permissive vs strict mode-related tests here...

lang/src/org/partiql/lang/eval/StrictnessContext.kt
```diff
@@ -0,0 +1,51 @@
...
+            test ->
+                when(strictnessMode) {
+                    StrictnessMode.STANDARD ->
+                        with(detailsConstructor()) {
+                            throw EvaluationException(
+                                message = message,
+                                errorCode = errorCode,
+                                errorContext = errorContext,
+                                cause = null,
+                                internal = false)
```
dlurton, line 32: Yes, this *stongly* implies that exceptions marked as `internal` will *never* result in `MISSING`. We will instead throw them as per usual.

lang/src/org/partiql/lang/eval/StrictnessContext.kt
```diff
@@ -0,0 +1,51 @@
...
+                        }
+                    StrictnessMode.PERMISSIVE ->
+                        valueFactory.missingValue
+                }
+            else -> otherwise()
+        }
```
dlurton, line 38: I also intend to pull the line & column number out of the metas and include them in the `errorContext` if they are not already specified.  I will do that in the next couple of commits to this PR.

lang/src/org/partiql/lang/eval/CompileOptions.kt
```diff
@@ -41,6 +40,12 @@ enum class ProjectionIterationBehavior {
     FILTER_MISSING, UNFILTERED
 }
 
+enum class StrictnessMode {
+    STANDARD,
+    PERMISSIVE,
+    //TODO: STRICT
+}
```
almann, line 47: Really, this is `ErrorMode` more than `StrictnessMode`.  And what is the difference between `STANDARD` and `PERMISSIVE` (needs API documentation to indicate intent).  It seems to me we only have `PERMISSIVE` right now...

lang/src/org/partiql/lang/eval/StrictnessContext.kt
```diff
@@ -0,0 +1,51 @@
...
+internal class ErrorDetails(
+    val metas: MetaContainer,
+    val message: String,
+    val errorCode: ErrorCode,
+    val errorContext: PropertyValueMap? = null
+)
+
+internal class StrictnessContext(val strictnessMode: StrictnessMode, val valueFactory: ExprValueFactory) {
+
+    inline fun errorIf(
```
almann, line 17: API documentation please, intent and exceptions in the `otherwise` thunk is not well defined...

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -429,16 +430,18 @@ internal class EvaluatingCompiler(
...
+            strictnessContext.errorIf(
+                test = denominator.isZero(),
+                detailsConstructor = { ErrorDetails(metas, "/ by zero", ErrorCode.EVALUATOR_DIVIDE_BY_ZERO) }
+            ) {
+                try {
+                    (lValue.numberValue() / denominator).exprValue()
+                } catch (e: ArithmeticException) {
+                    // Setting the internal flag as true as it is not clear what
+                    // ArithmeticException may be thrown by the above
+                    throw EvaluationException(cause = e, internal = true)
```
almann, line 443: Hmm, shouldn't this be captured somehow in the error handling?

lang/src/org/partiql/lang/eval/StrictnessContext.kt
```diff
@@ -0,0 +1,51 @@
...
+                            throw EvaluationException(
+                                message = message,
+                                errorCode = errorCode,
+                                errorContext = errorContext,
+                                cause = null,
+                                internal = false)
+                        }
+                    StrictnessMode.PERMISSIVE ->
+                        valueFactory.missingValue
+                }
```
almann, line 36: Stylistically, should we just put this as a method on the `enum`, then we don't need an if-bridge clause here, this smells like a factoring that needs some kind of polymorphism either in the `enum` or an algebraic type.  Really the context is just for capturing the evaluation context.  It almost seems like the `enum` is getting in our way in some regards.

lang/src/org/partiql/lang/eval/StrictnessContext.kt
```diff
@@ -0,0 +1,51 @@
...
+        test: Boolean,
+        crossinline detailsConstructor: () -> ErrorDetails,
+        crossinline otherwise: () -> ExprValue
+    ): ExprValue =
+        errorIf(
+            !test,
+            detailsConstructor,
+            otherwise)
+
+}
```
almann, line 50: Do we really need this?

lang/src/org/partiql/lang/eval/CompileOptions.kt
dlurton: `STANDARD` will be the error behaviour that we currently have.  This is intended for existing customers and will be the default setting. `STRICT` and `PERMISSIVE` are the modes outlined in the specification..

lang/src/org/partiql/lang/eval/StrictnessContext.kt
dlurton: Adding...

lang/src/org/partiql/lang/eval/StrictnessContext.kt
dlurton: Not, ahem, *strictly* speaking.  Will remove.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
dlurton: So, my intention here was just to capture the div 0 case, not to change anything else about how the error handling works, and this try/catch block preserves that.

To be clear, I am strongly leaning against catching any exception and returning `MISSING` values in the case of internal exceptions or other unhandled exceptions since that can have negative consequences to runtime performance.

lang/src/org/partiql/lang/eval/StrictnessContext.kt
dlurton: I'll try again to put these functions as methods on the `ErrorMode` enum but there was a reason I didn't that I can't recall right now...

dlurton: I remember now.  It's because the `ErrorMode` enum is a compiler option, which doesn't seem like the right place for implementation of runtime behavior.  I could split this into two different classes that implement an `ErrorBehavior` interface instead.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
almann: There is still a problem for dealing with this case regardless--how do you intend to handle it, as an exception handler at every call-site explicitly?  Same performance cost, more code...

PR #318, opened 2020-10-16
Migrate StaticTypeRewriter to PartiqlAst.VisitorTransform

alancai98:
Migrates `StaticTypeRewriter` to `PartiqlAst.VisitorTransform`.

Possible TODOs:
- Change `StaticTypeVisitorTransformTests` to use `ParameterizedTest`. This would involve changing many of the comments showing the column


lang/src/org/partiql/lang/eval/visitors/StaticTypeVisitorTransform.kt
```diff
@@ -0,0 +1,532 @@
...
+                    from = new_from,
+                    fromLet = new_fromLet,
+                    where = new_where,
+                    group = new_group,
+                    having = new_having,
+                    limit = new_limit,
+                    metas = new_metas
+                )
+            }
+        }
```
alancai98, line 461: Would be able to change this to
```kotlin
return super.transformExprSelect(node)
```
but the order the clauses are transformed affects the end result (causing tests to fail).

The original rewriter code called [`innerRewriteSelect`](https://github.com/partiql/partiql-lang-kotlin/blob/master/lang/src/org/partiql/lang/ast/passes/AstRewriterBase.kt#L151) which transforms in the execution order (i.e. transforms `FROM` first). `transformExprSelect` transforms `SELECT` first. Should the PIG version (`transformExprSelect`) follow the original rewriter transform order?

lang/src/org/partiql/lang/eval/visitors/StaticTypeVisitorTransform.kt
```diff
@@ -0,0 +1,532 @@
...
+        private fun innerTransformDataManipulation(node: PartiqlAst.Statement.Dml): PartiqlAst.Statement {
+            val from = node.from?.let { transformFromSource(it) }
+            val where = node.where?.let { transformStatementDml_where(node) }
+            val dmlOperation = transformDmlOp(node.operation)
+            val metas = transformMetas(node.metas)
+
+            return PartiqlAst.build {
+                dml(dmlOperation, from, where, metas)
+            }
+        }
```
alancai98, line 472: Similar to the other comment regarding `transformExprSelect`. Tried changing this to
```kotlin
return super.transformStatementDml(node)
```
but the order affects the end result, causing some tests for fail.

The original rewriter code called [`innerRewriteDataManipulation `](https://github.com/partiql/partiql-lang-kotlin/blob/master/lang/src/org/partiql/lang/ast/passes/AstRewriterBase.kt#L334) which transforms `dmlOperation` after transforming `WHERE`. The PIG code transforms the operation first. Should the PIG version follow the original rewriter transform order?

lang/src/org/partiql/lang/eval/visitors/StaticTypeVisitorTransform.kt
```diff
@@ -0,0 +1,532 @@
...
+                    node.setq,
+                    node.funcName,
+                    transformExpr(node.arg),
+                    transformMetas(node.metas))
+            }
+        }
+
+        // TODO: find if this override shouldn't be necessary because PIG's Call doesn't store the function call name as the first arg
+        override fun transformExprCall(node: PartiqlAst.Expr.Call): PartiqlAst.Expr {
+            // do not write the name of the call--this should be a symbolic name in another namespace (AST is over generalized here)
```
dlurton, line 215: This comment about overgeneralization is no longer valid.

lang/src/org/partiql/lang/eval/visitors/StaticTypeVisitorTransform.kt
```diff
@@ -0,0 +1,532 @@
...
+                callAgg_(
+                    // do not transform the funcExpr--as this is a symbolic name in another namespace (AST is over generalized here)
+                    node.setq,
+                    node.funcName,
+                    transformExpr(node.arg),
+                    transformMetas(node.metas))
+            }
+        }
+
+        // TODO: find if this override shouldn't be necessary because PIG's Call doesn't store the function call name as the first arg
```
dlurton, line 213: yeah, this is basically the same as the method in the super class now right?  If so, just delete this.

lang/src/org/partiql/lang/eval/visitors/StaticTypeVisitorTransform.kt
dlurton: - Add `kdoc` that says this is needed to adjust the transformation order.
- The `inner` prefix I came up with this is not great.  Can you think of a better name for this function?
- Consdier moving this to an abstract class that this `StaticTypeRewriter` derives from.  I know for a fact that this will be helpful in the near future.

lang/src/org/partiql/lang/eval/visitors/StaticTypeVisitorTransform.kt
dlurton: - Add `kdoc` that says this is needed to adjust the transformation order.
- The `inner` prefix I came up with this is not great.  Can you think of a better name for this function?
- Consider moving this to an abstract class that this `StaticTypeRewriter` derives from.  I know for a fact that this will be helpful in the near future.  (The same abstract class as `innerRewriteSelect`?)

dlurton (changes requested):
Overall LGTM--only minor changes needed.

lang/src/org/partiql/lang/eval/visitors/StaticTypeVisitorTransform.kt
alancai98: Removed.

lang/src/org/partiql/lang/eval/visitors/StaticTypeVisitorTransform.kt
alancai98: Got it. Removed the comment + override.

lang/src/org/partiql/lang/eval/visitors/StaticTypeVisitorTransform.kt
alancai98: - Added `kdoc`
- Changed name to `transformExprSelectEvaluationOrder`
- Moved to an abstract class

lang/src/org/partiql/lang/eval/visitors/StaticTypeVisitorTransform.kt
alancai98: - Added `kdoc`
- Changed name to `transformDataManipulationEvaluationOrder`
- Moved to an abstract class

alancai98:
Latest commits address David's comments regarding removing comments, removing unneeded override to `call`, and changes to `transformExprSelectEvaluationOrder`/`transformDataManipulationEvaluationOrder`.

dlurton (approved):

PR #319, opened 2020-10-19
Migrate AstSanityValidator to PIG

alancai98:
Migrates AstSanityValidator to PIG. New sanity validator is called `PartiqlAstSanityValidator` and is moved to the `visitors` folder.

Migrating to use PIG removes some checks from the existing `AstSanityValidator`. Notably, arity checking for `NAry` operations and `DataType` shouldn't be necessary. Also select star checking in a `ProjectList` isn't necessary for PIG since `ProjectStar` itself is a `Projection`.

TODO:
- ~Decide if we want to also delete the deprecated `AstSanityVisitor` (and `AstSanityVisitorTests`), which is reliant on the original `AstSanityValidator`. If we decide to delete `AstSanityVisitor`, then we can more easily delete `AstSanityValidator` and its tests~ decided to remove


lang/src/org/partiql/lang/ast/passes/AstSanityValidator.kt
```diff
@@ -1,131 +0,0 @@
...
- *
- * - [org.partiql.lang.syntax.SqlParser]
- * - A rewrite pass (internal or external)
- *
- * At the time of this writing there are only 3 checks performed, with the idea that more may be added later.  Checks
- * that:
- *
- * - [NAry] arity is correct
- * - [DataType] arity is correct.
- * - `*` is not used in conjunction with other select list items.
```
alancai98, line 39: These 3 checks were removed in the migration to use PIG.

Arity checks shouldn't be necessary b/c of how PIG defines types. Select star is a PIG `Projection` so I believe it can't be used from a PIG `ProjectList`.

lang/src/org/partiql/lang/eval/visitors/PartiqlAstSanityValidator.kt
```diff
@@ -0,0 +1,105 @@
...
+ * This is provided as a distinct visitor so that all other visitors may assume that the AST at least passed the
+ * checking performed here.
+ *
+ * Any exception thrown by this class should always be considered an indication of a bug in one of the following places:
+ *
+ * - [org.partiql.lang.syntax.SqlParser]
+ * - A visitor transform pass (internal or external)
+ *
+ */
+object PartiqlAstSanityValidator {
```
dlurton, line 42: This should be implementing, `PartiqlAst.Visitor`.

What you have here doesn't catch errors in child nodes.  

Please add some tests which with bad non-root nodes.

dlurton (changes requested):

lang/src/org/partiql/lang/eval/visitors/PartiqlAstSanityValidator.kt
alancai98: Good catch. Should be fixed in the newest revision. Also added new subquery tests. Thanks

alancai98:
New revision [commit](https://github.com/partiql/partiql-lang-kotlin/pull/319/commits/73a315062dfc82472415b1c1a0fae17c709c1b0f) makes `PartiqlAstSanityValidator` implement `PartiqlAst.Visitor` and walk the `PartiqlAst.Statement` to check the subqueries. Also added new subquery tests.

dlurton (approved):

PR #330, opened 2020-11-20
Add UNIX epoch <-> TIMESTAMP conversion functions

alancai98:
Adds functions to convert from UNIX epoch to TIMESTAMP and TIMESTAMP to UNIX epoch. Largely follows MySQL's variants: [FROM_UNIXTIME](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_from-unixtime) and [UNIX_TIMESTAMP](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_unix-timestamp).

Still might need discussion on:
- Better names for the functions. I just followed MySQL's convention given that I couldn't find similar functions in other DBMS and SQL-92 standard. Perhaps `UNIX_TO_TIMESTAMP` and `TIMSTAMP_TO_UNIX` would be better?
- Supporting a format argument when converting UNIX epoch -> TIMESTAMP. This is an additional argument in MySQL's version, but this may be better served as its own function.
- Datatype input for UNIX epoch -> TIMESTAMP currently is just a numeric value. Do we want to allow for strings (e.g. `from_unixtime("1234567890")`)?
- Datatype to support for UNIX epochs? I largely followed MySQL's [FROM_UNIXTIME](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_from-unixtime) which uses decimal depending on if the timestamp had fractional seconds, integer otherwise.
- Some edge cases, namely: 1. What should we do when given a UNIX epoch to convert less than 0? 2. What do we do when given a timestamp before '1970-01-01 00:00:01.000000' UTC? (currently return `0`)

### DIFF
--- a/lang/src/org/partiql/lang/eval/builtins/FromUnixTimeFunction.kt
+++ b/lang/src/org/partiql/lang/eval/builtins/FromUnixTimeFunction.kt
@@ -1,0 +1,37 @@
+package org.partiql.lang.eval.builtins
+
+import com.amazon.ion.Timestamp
+import org.partiql.lang.eval.Environment
+import org.partiql.lang.eval.ExprValue
+import org.partiql.lang.eval.ExprValueFactory
+import org.partiql.lang.eval.NullPropagatingExprFunction
+import org.partiql.lang.eval.bigDecimalValue
+import java.math.BigDecimal
+
+/**
+ * Builtin function to convert the given unix epoch into a PartiQL `TIMESTAMP` [ExprValue]. A unix epoch represents
+ * the seconds since '1970-01-01 00:00:00' UTC. Largely based off MySQL's FROM_UNIXTIME.
+ *
+ * Syntax: `FROM_UNIXTIME(unix_timestamp)`
+ * Where unix_timestamp is a non-negative (potentially decimal) numeric value.
+ *
+ * When given a negative numeric value, this function returns a PartiQL `NULL` [ExprValue].
+ * When given a non-negative numeric value, this function returns a PartiQL `TIMESTAMP` [ExprValue] that has fractional
+ * seconds depending on if the value is a decimal.
+ */
+internal class FromUnixTimeFunction(valueFactory: ExprValueFactory) : NullPropagatingExprFunction("from_unixtime", 1, valueFactory) {
+    private val millisPerSecond = BigDecimal(1000)
+
+    override fun eval(env: Environment, args: List<ExprValue>): ExprValue {
+        val unixTimestamp = args[0].bigDecimalValue()
+
+        val numMillis = unixTimestamp.times(millisPerSecond).stripTrailingZeros()
+
+        if (unixTimestamp < BigDecimal.ZERO) {
+            return valueFactory.nullValue
+        }
+
+        val timestamp = Timestamp.forMillis(numMillis, null)
+        return valueFactory.newTimestamp(timestamp)
+    }
+}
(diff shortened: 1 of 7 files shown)


lang/src/org/partiql/lang/eval/builtins/UnixTimestampFunction.kt
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * If UNIX_TIMESTAMP() is called with no [timestamp] argument, it returns the number of whole seconds since
+ * '1970-01-01 00:00:00' UTC as a PartiQL `INT` [ExprValue]
+ *
+ * If UNIX_TIMESTAMP() is called with a [timestamp] argument, it returns the number of seconds from
+ * '1970-01-01 00:00:00' UTC to the given [timestamp] argument. The return value will be a decimal if and only if
+ * the given [timestamp] has a fractional seconds part.
+ *
+ * The valid range of argument values is from '1970-01-01 00:00:01.000000' UTC up to the maximum PartiQL `TIMESTAMP`
+ * value. If given a [timestamp] out of this range, UNIX_TIMESTAMP() will return 0.
```
therapon, line 24: Should we retun `0` here or `NULL`?

alancai98: I was following MySQL's version to convert timestamps -> unix epoch ([UNIX_TIMESTAMP](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_unix-timestamp)) which:
> ... If you pass an out-of-range date to UNIX_TIMESTAMP(), it returns 0.

Perhaps it would be better to keep it consistent with the other function to convert unix epoch -> timestamp. Currently that function will return a `NULL` `ExprValue` when given a unix epoch out of range.

therapon: That is my dilemma as well. Personally I like the symmetry between the 2 functions and when the passed inputs result in an epoch that does not make sense, `NULL` seems more appropriate to me than `0`. Being "consistent" with other DBs will be less of a surprise to users. 

The alternative is to allow for "negative" values that denote time *before* 1970 which some epoch-functions tend to do (I think PostgreSQL does that).

alancai98: I like the alternative you suggested for allowing "negative" values to denote time _before_ 1970. Would be a pretty simple change to both functions.

alancai98:
New commit removes the negative unix epoch / timestamp before the last epoch restriction and updates tests + docs accordingly. Also adds `UNIX_TIMESTAMP` and `FROM_UNIXTIME` entries to `BuiltinFunctions.md`.

lang/src/org/partiql/lang/eval/builtins/FromUnixTimeFunction.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.partiql.lang.eval.NullPropagatingExprFunction
+import org.partiql.lang.eval.bigDecimalValue
+import java.math.BigDecimal
+
+/**
+ * Builtin function to convert the given unix epoch into a PartiQL `TIMESTAMP` [ExprValue]. A unix epoch represents
+ * the seconds since '1970-01-01 00:00:00' UTC. Largely based off MySQL's FROM_UNIXTIME.
+ *
+ * Syntax: `FROM_UNIXTIME(unix_timestamp)`
+ * Where unix_timestamp is a non-negative (potentially decimal) numeric value.
```
therapon, line 16: This seems to no longer be true, we can pass a negative numeric value. 
I like that you added the syntax in the comment. I think it would be nice to also add the return type or at least a description of the expected function's return values

alancai98: Good catch, thanks! Also edited to make it more clear on the expected return values.

therapon (approved):

PR #333, opened 2020-11-25
Stored Procedures - call with unnamed args (parsing + PIG definition)

alancai98:
Adds parsing and PIG definition of stored procedure calls with the unnamed argument syntax defined in PartiQL spec issue #[17](https://github.com/partiql/partiql-spec/issues/17).

A valid call to a stored procedure as part of this PR is defined by `EXEC <symbol> [<expr>.*]` and is only a valid PartiQL statement when in the top level query.

### DIFF
--- a/lang/src/org/partiql/lang/ast/passes/AstWalker.kt
+++ b/lang/src/org/partiql/lang/ast/passes/AstWalker.kt
@@ -117,7 +117,7 @@ open class AstWalker(private val visitor: AstVisitor) {
                     }
                 }
                 is CreateTable, is DropTable, is DropIndex -> case { }
-                is Exec         -> throw UnsupportedOperationException("EXEC clause is not supported by the V0 AST")
+                is Exec -> case { }
             }.toUnit()
         }
     }
--- a/lang/src/org/partiql/lang/ast/passes/AstWalker.kt
+++ b/lang/src/org/partiql/lang/ast/passes/AstWalker.kt
@@ -16,6 +16,7 @@ package org.partiql.lang.ast.passes
 
 import org.partiql.lang.ast.*
 import org.partiql.lang.util.*
+import java.lang.UnsupportedOperationException
 
 /**
  * Contains the logic necessary to walk every node in the AST and invokes methods of [AstVisitor] along the way.
@@ -116,6 +117,7 @@ open class AstWalker(private val visitor: AstVisitor) {
                     }
                 }
                 is CreateTable, is DropTable, is DropIndex -> case { }
+                is Exec         -> throw UnsupportedOperationException("EXEC clause is not supported by the V0 AST")
             }.toUnit()
         }
     }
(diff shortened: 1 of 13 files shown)


lang/src/org/partiql/lang/ast/passes/AstWalker.kt
```diff
@@ -116,6 +117,7 @@ open class AstWalker(private val visitor: AstVisitor) {
                     }
                 }
                 is CreateTable, is DropTable, is DropIndex -> case { }
+                is Exec         -> throw UnsupportedOperationException("EXEC clause is not supported by the V0 AST")
```
dlurton, line 120: 1.  Copy-pasta from the `V0` serializer?
1. We do need to support this in the `AstWalker` for time being, so we shouldn't be throwing an exception here.  Probably, an empty `case { }` like the previous line is appropriate.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -278,6 +278,8 @@ internal class EvaluatingCompiler(
             is CreateIndex,
             is DropIndex,
             is DropTable -> compileDdl(expr)
+            is Exec              -> compileExec(expr)
```
dlurton, line 281: Nit: formatting.

dlurton (changes requested):

lang/src/org/partiql/lang/ast/passes/AstWalker.kt
alancai98: 1. Yeah that was copied from the `AstSerialization` code.
2. Changed to `case { }`.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
alancai98: Fixed. Thanks

alancai98:
New revision fixes previous feedback. Also changes `func_name`/`funcName` usages to `procedure_name`/`procedureName`. And adds check that the args aren't surrounded in parens (i.e. calling `EXEC sproc_foo(arg0, arg1)` now throws an error).

lang/src/org/partiql/lang/ast/passes/AstWalker.kt
```diff
@@ -16,6 +16,7 @@ package org.partiql.lang.ast.passes
 
 import org.partiql.lang.ast.*
 import org.partiql.lang.util.*
+import java.lang.UnsupportedOperationException
```
therapon, line 19: is this import needed? looks like a left-over

alancai98: Yes, that is leftover. Removed in the new commit. Thanks!

dlurton (approved):

PR #334, opened 2020-11-25
Adds documentation regarding thread safety of `CompilerPipeline`

alancai98:
Adds to the kdoc, detailing how the builder instance of `CompilerPipeline` is not thread safe.

Possible TODOs:
- Should we also document this for the `CompilerPipeline` interface (not just the builder instance)?
- Should we add annotations of non-thread safety?
- Should we also document that `EvaluatingCompiler` is not thread safe?


lang/src/org/partiql/lang/CompilerPipeline.kt
```diff
@@ -93,7 +93,10 @@ interface CompilerPipeline  {
             builder(valueFactory).build()
     }
 
-    /** An implementation of the builder pattern for instances of [CompilerPipeline]. */
+    /**
+     * An implementation of the builder pattern for instances of [CompilerPipeline]. The created instance of
+     * [CompilerPipeline] is NOT thread safe and should NOT be used to compile queries concurrently.
+     */
```
dlurton, line 99: Add:

> If used in a multithreaded application, use one instance of [CompilerPipeline] per thread.

dlurton (changes requested):
- Should we also document this for the CompilerPipeline interface (not just the builder instance)?
    - I would, yes.
- Should we add annotations of non-thread safety?
    - Of what annotations specifically do you speak of?  I am not aware of any that aren't part of another library and I'd rather avoid taking a dependency on a new library just for that.
- Should we also document that EvaluatingCompiler is not thread safe?
   - `EvaluatingCompiler` is `internal`, but you can add a quick note to it if you'd like.

lang/src/org/partiql/lang/CompilerPipeline.kt
alancai98: Added.

alancai98:
New revision includes updated kdoc to `CompilerPipeline` builder and interface. Also adds a note about thread safety to `EvaluatingCompiler`.

Regarding
>Of what annotations specifically do you speak of? I am not aware of any that aren't part of another library and I'd rather avoid taking a dependency on a new library just for that.

None in particular. I wasn't sure if there was any standard annotation to denote a class was not thread safe.

dlurton (approved):

PR #335, opened 2020-12-01
Stored Procedures - call with unnamed args (evaluation)

alancai98:
Implements evaluation of stored procedure calls with unnamed argument syntax defined in PartiQL spec issue #[17](https://github.com/partiql/partiql-spec/issues/17).

This should be reviewed **after** the parsing of stored procedure calls (#333) is merged into the `stored-procedures` feature branch.

### DIFF
--- a/examples/src/kotlin/org/partiql/examples/CustomProceduresExample.kt
+++ b/examples/src/kotlin/org/partiql/examples/CustomProceduresExample.kt
@@ -1,0 +1,128 @@
+package org.partiql.examples
+
+import com.amazon.ion.IonDecimal
+import com.amazon.ion.IonStruct
+import com.amazon.ion.system.IonSystemBuilder
+import org.partiql.examples.util.Example
+import org.partiql.lang.CompilerPipeline
+import org.partiql.lang.errors.ErrorCode
+import org.partiql.lang.errors.Property
+import org.partiql.lang.errors.PropertyValueMap
+import org.partiql.lang.eval.BindingCase
+import org.partiql.lang.eval.BindingName
+import org.partiql.lang.eval.Bindings
+import org.partiql.lang.eval.EvaluationException
+import org.partiql.lang.eval.EvaluationSession
+import org.partiql.lang.eval.ExprValue
+import org.partiql.lang.eval.ExprValueFactory
+import org.partiql.lang.eval.ExprValueType
+import org.partiql.lang.eval.builtins.storedprocedure.StoredProcedure
+import org.partiql.lang.eval.builtins.storedprocedure.StoredProcedureSignature
+import org.partiql.lang.eval.stringValue
+import java.io.PrintStream
+import java.math.BigDecimal
+import java.math.RoundingMode
+
+private val ion = IonSystemBuilder.standard().build()
+
+/**
+ * A simple custom stored procedure that calculates the moon weight for each crewmate of the given crew, storing the
+ * moon weight in the [EvaluationSession] global bindings. This procedure also returns the number of crewmates we
+ * calculated the moon weight for, returning -1 if no crew is found.
+ *
+ * This example demonstrates how to create a custom stored procedure, check argument types, and modify the
+ * [EvaluationSession].
+ */
+class CalculateCrewMoonWeight(private val valueFactory: ExprValueFactory): StoredProcedure {
+    private val MOON_GRAVITATIONAL_CONSTANT = BigDecimal(1.622 / 9.81)
+
+    // [StoredProcedureSignature] takes two arguments:
+    //   1. the name of the stored procedure
+    //   2. the arity of this stored procedure. Checks to arity are taken care of by the evaluator. However, we must
+    //      still check that the passed arguments are of the right type in our implementation of the procedure.
+    override val signature = StoredProcedureSignature(name = "calculate_crew_moon_weight", arity = 1)
+
+    // `call` is where you define the logic of the stored procedure given an [EvaluationSession] and a list of
+    // arguments
+    override fun call(session: EvaluationSession, args: List<ExprValue>): ExprValue {
+        // We first check that the first argument is a string
+        val crewName = args.first()
+        if (crewName.type != ExprValueType.STRING) {
+            val errorContext = PropertyValueMap().also {
+                it[Property.EXPECTED_ARGUMENT_TYPES] = "STRING"
+                it[Property.ACTUAL_ARGUMENT_TYPES] = crewName.type.name
+                it[Property.FUNCTION_NAME] = signature.name
+            }
+            throw EvaluationException("First argument to ${signature.name} was not a string",
+                ErrorCode.EVALUATOR_INCORRECT_TYPE_OF_ARGUMENTS_TO_PROCEDURE_CALL,
+                errorContext,
+                internal = false)
+        }
+
+        // Next we check if the given `crewName` is in the [EvaluationSession]'s global bindings. If not, we return 0.
+        val sessionGlobals = session.globals
+        val crewBindings = sessionGlobals[BindingName(crewName.stringValue(), BindingCase.INSENSITIVE)]
+            ?: return valueFactory.newInt(-1)
+
+        // Now that we've confirmed the given `crewName` is in the session's global bindings, we calculate and store
+        // the moon weight for each crewmate in the crew.
+        // In addition, we keep a running a tally of how many crewmates we do this for.
+        var numCalculated = 0
+        for (crewmateBinding in crewBindings) {
+            val crewmate = crewmateBinding.ionValue as IonStruct
+            val mass = crewmate["mass"] as IonDecimal
+            val moonWeight = (mass.decimalValue() * MOON_GRAVITATIONAL_CONSTANT).setScale(1, RoundingMode.HALF_UP)
+            crewmate.add("moonWeight", ion.newDecimal(moonWeight))
+
+            numCalculated++
+        }
+        return valueFactory.newInt(numCalculated)
+    }
+}
+
+/**
+ * Demonstrates the use of custom stored procedure [CalculateCrewMoonWeight] in PartiQL queries.
+ */
+class CustomProceduresExample(out: PrintStream) : Example(out) {
+    override fun run() {
+        /**
+         * To make custom stored procedures available to the PartiQL query being executed, they must be passed to
+         * [CompilerPipeline.Builder.addProcedure].
+         */
+        val pipeline = CompilerPipeline.build(ion) {
+            addProcedure(CalculateCrewMoonWeight(valueFactory))
+        }
+
+        // Here, we initialize the crews to be stored in our global session bindings
+        val initialCrews = Bindings.ofMap(
+            mapOf(
+                "crew1" to pipeline.valueFactory.newFromIonValue(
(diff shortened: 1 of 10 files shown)


lang/src/org/partiql/lang/eval/builtins/storedprocedure/BuiltinProcedures.kt
```diff
@@ -0,0 +1,7 @@
+package org.partiql.lang.eval.builtins.storedprocedure
+
+import org.partiql.lang.eval.ExprFunction
+import org.partiql.lang.eval.ExprValueFactory
+
+internal fun createBuiltinProcedures(valueFactory: ExprValueFactory) : List<ExprFunction> =
+    listOf()
```
alancai98, line 7: Not sure if we want to define any default builtin stored-procedures?

lang/test/org/partiql/lang/eval/EvaluatingCompilerExecTests.kt
```diff
@@ -0,0 +1,302 @@
+package org.partiql.lang.eval
```
alancai98, line 1: I decided to put the tests in the `eval` sub-package rather than `examples` for the better flexibility in testing. I can also add a smaller example of adding a stored procedure similar to the [`CustomFunctionsExample`](https://github.com/partiql/partiql-lang-kotlin/blob/master/examples/src/kotlin/org/partiql/examples/CustomFunctionsExample.kt) if that would be useful.

lang/test/org/partiql/lang/eval/EvaluatingCompilerExecTests.kt
```diff
@@ -0,0 +1,302 @@
...
+import org.partiql.lang.errors.PropertyValueMap
+import org.partiql.lang.util.ArgumentsProviderBase
+import org.partiql.lang.util.isAnyUnknown
+import org.partiql.lang.util.softAssert
+import org.partiql.lang.util.to
+
+/**
+ * Similar class to [NullPropagatingExprFunction], but will output a stored procedure related error for invalid arity.
+ */
+abstract class NullPropagatingProcedure(
```
alancai98, line 17: Created this class because `NullPropagatingExprFunction` outputs `EVALUATOR_INCORRECT_NUMBER_OF_ARGUMENTS_TO_FUNCTION_CALL` for invalid argument arity. Other alternatives could be:
- Use `NullPropagatingExprFunction`, which would output a function call-related error for invalid arity
- Change `EVALUATOR_INCORRECT_NUMBER_OF_ARGUMENTS_TO_FUNCTION_CALL` to a more general form for function calls and procedure calls
- Create a new simple stored procedure abstract class (which may have a lot of repeated code as `NullPropagatingExprFunction`)

dlurton: Hrm.  Ok.  `NULL` propagation does not actually apply to stored procedures, i.e. `NULL` values should always pass thru to the `ExprFunction` implementation.

I think what we need is a new class named something like `StoredProcedureBase` that does the arity check but not the `NULL` propagation.

dlurton: Also, go ahead and put `StoredProcedureBase` under  `lang/src/org/partiql/lang/eval` so as to make it a part of the public API.  I'd like to avoid having customers try to implement their own arity check since throwing an `EvaluationException` with the correct error code and error context is knowledge they are not likely to have.

lang/test/org/partiql/lang/eval/EvaluatingCompilerExecTests.kt
```diff
@@ -0,0 +1,302 @@
...
+                                      errorContext,
+                                      internal = false)
+        }
+    }
+
+    override fun call(env: Environment, args: List<ExprValue>): ExprValue {
+        checkSProcArity(args)
+
+        return when {
+            args.isAnyUnknown() -> valueFactory.nullValue
```
dlurton, line 47: No null propagation for `StoredProcedureBase`.

lang/test/org/partiql/lang/eval/EvaluatingCompilerExecTests.kt
dlurton: > I can also add a smaller example of adding a stored procedure similar to the CustomFunctionsExample if that would be useful.
 
I do think having one implementation of `StoredProcedureBase` there would be helpful.  Thanks.

lang/test/org/partiql/lang/eval/EvaluatingCompilerExecTests.kt
```diff
@@ -0,0 +1,302 @@
...
+                errorContext,
+                internal = false)
+        }
+        val bindingName = BindingName(arg.stringValue(), BindingCase.INSENSITIVE)
+        return when(val value = env.current[bindingName]) {
+            null -> valueFactory.missingValue
+            else -> value
+        }
+    }
+}
```
dlurton, line 144: This sproc is super interesting, however I don't think that we should support accessing the local variables of the current context this in this fashion because `Environment` is full of implementation details of the interpreter and that might be hard to support if we later want to change how the interpreter works in some substantive way, or say, replace the interpreter entirely (likely).

In looking at how the `env` parameter is actually used in all the built-in `ExprFunction` implementations, it's only ever used by `utcnow` to obtain the current `EvaluationSession`.  So I think we should change the `env: Environment` argument of `ExprFunction.call` to `session: EvaluationSession`, but that is out of scope for now.

Anyway, this makes me think we're better off creating a new interface so we can avoid spreading that problem around for now.  There will be a little duplication here but I think that's better than 

Possibly, something like this:

```
data class StoredProcedureSignature(val name: String, val arity: IntRange)
interface StoredProcedure {
    val signature: StoredProcedureSignature
    fun call(args: List<ExprValue>, session: EvaluationSession, valueFactory: ExprValueFactory) 
}
```

We throw `valueFactory: ExprValueFactory` in there so `StoredProcedure` implementations can avoid having to inject their own `ExprValueFactory` instances.  Let me know if you see any downsides to this.

dlurton (changes requested):

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -1926,8 +1927,21 @@ internal class EvaluatingCompiler(
...
+            "No such stored procedure: ${procedureName.name}",
+            ErrorCode.EVALUATOR_NO_SUCH_PROCEDURE,
+            errorContextFrom(metas).also {
+                it[Property.PROCEDURE_NAME] = procedureName.name
+            },
+            internal = false)
+
+        return thunkFactory.thunkEnv(metas) { env ->
+            val procedureArgValues = argThunks.map { it(env) }
+            procedure.call(env, procedureArgValues)
```
dlurton, line 1943: Validate the arity & throw here!

lang/src/org/partiql/lang/eval/builtins/storedprocedure/BuiltinProcedures.kt
alancai98: Ok, took out in the new revision.

lang/test/org/partiql/lang/eval/EvaluatingCompilerExecTests.kt
alancai98: Added an example to the `examples` sub-package.

lang/test/org/partiql/lang/eval/EvaluatingCompilerExecTests.kt
alancai98: Deleted `NullPropagatingProcedure` and performed the arity checks in the `EvaluatingCompiler` thunk.

lang/test/org/partiql/lang/eval/EvaluatingCompilerExecTests.kt
alancai98: Added the `StoredProcedure` interface that takes in an `EvaluationSession` rather than an `Environment`. 

In the latest revision, I decided to leave out `ExprValueFactory` in the `call()`. I felt this is more in line with what `ExprFunction` currently does, where function implementations are passed an `ExprValueFactory` when created. Let me know if we shouldn't follow `ExprFunction`'s way of passing an `ExprValueFactory`.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
alancai98: I put the arity validation and throwing after checking if the given stored procedure exists ([L1941](https://github.com/partiql/partiql-lang-kotlin/blob/d618ba5239f05d59112679400b92f2f1b8ba19bc/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt#L1941)). Would it be better to put it in the thunk?

alancai98:
New revision addresses David's comments. To summarize:
- Adds stored procedure interface
- Moves arity checking and throwing to `EvaluatingCompiler`
- Adds an example to the `examples` package

examples/src/kotlin/org/partiql/examples/CustomProceduresExample.kt
```diff
@@ -0,0 +1,128 @@
...
+            val errorContext = PropertyValueMap().also {
+                it[Property.EXPECTED_ARGUMENT_TYPES] = "STRING"
+                it[Property.ACTUAL_ARGUMENT_TYPES] = crewName.type.name
+                it[Property.FUNCTION_NAME] = signature.name
+            }
+            throw EvaluationException("First argument to ${signature.name} was not a string",
+                ErrorCode.EVALUATOR_INCORRECT_TYPE_OF_ARGUMENTS_TO_PROCEDURE_CALL,
+                errorContext,
+                internal = false)
+        }
```
dlurton, line 60: Add above this line: `// In the future the evaluator will also verify function argument types, but for now we must verify their type manually`

lang/src/org/partiql/lang/CompilerPipeline.kt
```diff
@@ -35,7 +36,12 @@ data class StepContext(
      * Includes built-in functions as well as custom functions added while the [CompilerPipeline]
      * was being built.
      */
-    val functions: @JvmSuppressWildcards Map<String, ExprFunction>
+    val functions: @JvmSuppressWildcards Map<String, ExprFunction>,
+
+    /**
+     * Returns a list of all stored procedures which are available for execution.
+     */
```
dlurton, line 43: did you want to include a note here:

> Includes built-in stored procedures as well as custom stored procedures added while the [CompilerPipeline] was being built.

?

dlurton (approved):
LGTM, only two minor nits about comments and documentation.  Feel free to ignore, but if you don't ping me and I'll quickly follow up with another approval.

examples/src/kotlin/org/partiql/examples/CustomProceduresExample.kt
alancai98: Added.

lang/src/org/partiql/lang/CompilerPipeline.kt
alancai98: Added but omitted the "built-in stored procedures" part as there are no built-in PartiQL stored procedures.

Issue #266, opened 2020-07-31
Refactor the CLI to make it more testable and eliminate multhreading in its unit tests


dlurton:
The unit tests for the REPL are [multithreaded](https://github.com/partiql/partiql-lang-kotlin/blob/18423091a259f651e8478ab1dc995affe494c1ac/cli/test/org/partiql/cli/ReplTest.kt#L87-L88).  I'm guessing this is because the design of the CLI isn't easily testable.  Unfortunately, there's a mysterious race condition in this code that showed up recently that's causing our travis CI builds to fail randomly.  This [build](https://travis-ci.org/github/partiql/partiql-lang-kotlin/builds/713101679), for example

While doing this, we should also consider these other two issues #112, #255, which together strongly suggest to me that the CLI should be refactored to use an `ncurses` type library for the JVM.  (Perhaps [`lanterna`](https://github.com/mabe02/lanterna) could be wrangled for this purpose? This is just an idea, I haven't seriously looked to see if `lanterna` is a good fit for this.  Another alternative is [jline3](https://github.com/jline/jline3) which seems targeted to this purpose.)

EDIT:  #267 disables this test for the time being.

PR #339, opened 2020-12-11
Add version number to CLI output

alancai98:
Adds current PartiQL version number (from gradle) and the most recent commit hash to the CLI output.

Example output shown below:
```
Welcome to the PartiQL REPL!
Using version: 0.2.5-SNAPSHOT-6786c36
PartiQL> 1+1;
==='
2
---
OK!
```

This approach follows from this similar stackoverflow [answer](https://stackoverflow.com/a/50119235). Essentially, the root `build.gradle` file will output the `version` to a file in the Gradle build, which can be accessed as a resource by the CLI/REPL.

Retrieving the most recent commit hash follows this stackoverflow [answer](https://stackoverflow.com/a/35041457). We similarly store the commit hash to a Gradle build file.

Also, some existing REPL tests (previously ignored by #266) were updated to incorporate this change.

### DIFF
--- a/cli/test/org/partiql/cli/ReplTest.kt
+++ b/cli/test/org/partiql/cli/ReplTest.kt
@@ -21,7 +21,6 @@ import org.junit.*
 import org.junit.Assert.*
 import org.partiql.lang.*
 import java.io.*
-import java.util.*
 import java.util.concurrent.*
 import kotlin.concurrent.*
 
--- a/cli/test/org/partiql/cli/ReplTest.kt
+++ b/cli/test/org/partiql/cli/ReplTest.kt
@@ -79,7 +79,7 @@ private class ReplTester(bindings: Bindings<ExprValue> = Bindings.empty()) {
 
     private val repl = Repl(valueFactory, input, output, parser, compiler, bindings, zeroTimer)
 
-    val partiqlVersion = repl.retrievePartiQLVersion()
+    val partiqlVersionAndHash = repl.retrievePartiQLVersionAndHash()
 
     private val actualReplPrompt = StringBuilder()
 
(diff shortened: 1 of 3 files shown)


cli/test/org/partiql/cli/ReplTest.kt
```diff
@@ -201,66 +207,24 @@ class ReplTest {
...
-            #PartiQL> 
-        """.trimMargin("#"))
-    }
-
-    @Test
-    fun astWithMetas() {
-        ReplTester().assertReplPrompt("""
-            #Welcome to the PartiQL REPL!
-            #PartiQL> 1 + 1
-            #   | !?
```
alancai98, line 230: This test was removed due to removal of printing ASTs without metas in #272.

cli/test/org/partiql/cli/ReplTest.kt
```diff
@@ -201,66 +207,24 @@ class ReplTest {
...
-            #===' 
-            #
-            #(
-            #  meta
+            #  query
```
alancai98, line 216: Fix to the existing test's AST.

dlurton (approved):
Overall LGTM, however is it useful/possible/feasible to include the short git hash in the version string?  i.e. 

```
Welcome to the PartiQL REPL!
Using version: 0.2.5-SNAPSHOT-c6a3b073
PartiQL>
```

alancai98:
I think providing the commit hash would be pretty useful. New revision adds the short commit hash to the REPL output following this [approach](https://stackoverflow.com/a/35041457) to get the hash and storing it as another property.

New REPL output looks like:
```
Welcome to the PartiQL REPL!
Using version: 0.2.5-SNAPSHOT-6786c36
PartiQL>
```

dlurton (approved):

PR #342, opened 2020-12-15
Moves `MetaStrippingRewriter` and `AstRewriterBase` to test dir

alancai98:
Moves `MetaStrippingRewriter` and `AstRewriterBase` to test directory. Also deletes `RewriterTestBase`.


lang/test/org/partiql/lang/TestBase.kt
```diff
@@ -14,20 +14,18 @@
...
 import org.junit.Assert
 import org.junit.runner.RunWith
 import java.util.*
-
-import org.partiql.lang.errors.Property.*
-import org.partiql.lang.eval.*
-import org.partiql.lang.util.*
-import org.assertj.core.api.*
-import com.amazon.ion.*
+import junitparams.JUnitParamsRunner
```
alancai98, line 28: Fix to ordering of imports.

lang/test/org/partiql/lang/TestBase.kt
```diff
@@ -73,10 +71,10 @@ abstract class TestBase : Assert() {
...
 
-    protected fun assertBaseRewrite(originalSql: String, exprNode: ExprNode) {
-        val clonedAst = defaultRewriter.rewriteExprNode(exprNode)
+    protected fun assertBaseTransform(originalSql: String, exprNode: ExprNode) {
+        val clonedAst = defaultTransform.transformStatement(exprNode.toAstStatement()).toExprNode(ion)
         assertEquals(
-            "AST returned from default AstRewriterBase should match the original AST. SQL was: $originalSql",
+            "AST returned from default PartiqlAst.VisitorTransform should match the original AST. SQL was: $originalSql",
             exprNode, clonedAst)
     }
```
alancai98, line 79: I changed the call from `rewriteExprNode` to `transformStatement`. To get this working, I had to fix some of the passing of metas in [`ExprNodeToStatement`](https://github.com/partiql/partiql-lang-kotlin/pull/342/files#diff-adbd0081150155dc39c0df60f859fa7cf962e821a67b78395993ce837424e0ad). An alternative to comparing `ExprNode`s could be to compare `PartiqlAst`s i.e.:
```kotlin
protected fun assertBaseTransform(originalSql: String, exprNode: ExprNode) {
    val partiqlAst = exprNode.toAstStatement()
    val clonedAst = defaultTransform.transformStatement(partiqlAst)
    assertEquals("...", partiqlAst, clonedAst)
}
```
Should I be comparing `ExprNode`s or `PartiqlAst`s here?

lang/src/org/partiql/lang/ast/ExprNodeToStatement.kt
```diff
@@ -99,10 +99,11 @@ fun ExprNode.toAstExpr(): PartiqlAst.Expr {
                 }
             }
             is CallAgg -> {
-                val symbol1 = (node.funcExpr as? VariableReference)?.id
+                val symbol1 = (node.funcExpr as? VariableReference)
```
alancai98, line 102: When working on a change to [`TestBase.kt`](https://github.com/partiql/partiql-lang-kotlin/pull/342/files#diff-32f794c776727f8be85a860d0f8157e373f1ba9951d2b261d9898e396701d359), some of the tests were failing because of missing metas. Changes to this file adds `metas` that were previously not passed.

lang/test/org/partiql/lang/TestBase.kt
```diff
@@ -36,7 +34,7 @@ abstract class TestBase : Assert() {
     val ion: IonSystem = IonSystemBuilder.standard().build()
     val valueFactory = ExprValueFactory.standard(ion)
 
-    val defaultRewriter = AstRewriterBase()
+    val defaultTransform = PartiqlAst.VisitorTransform()
```
dlurton, line 37: Probably aught to undo this particular change as this is one of the main ways that the `AstRewriterBase` is itself tested.  See other comment on `assertBaseTransform`.

lang/test/org/partiql/lang/TestBase.kt
dlurton: Yeah, this function is intended to test `AstRewriterBase` and should continue to test it.  This change should probably be undone before you merge this.

For the PIG-generated equivalent its not necessary since we can assume that it functions correctly.  (PIG itself should have a test like this.  If it doesn't we should create one.)

dlurton (changes requested):
Mostly LGTM, only one issue about still needing to test `AstRewriterBase` but not the `PartiqlAst.VistiorTransform`.

lang/test/org/partiql/lang/TestBase.kt
alancai98: Got it. Reverted the function back to test `AstRewriterBase`

dlurton (approved):

PR #345, opened 2020-12-17
Add stored procedure calls with unnamed args

alancai98:
Adds the ability to create custom stored procedures and call them with the `EXEC` clause, following the description given in this [issue](https://github.com/partiql/partiql-spec/issues/17).

Also provides a simple custom stored procedure [example](https://github.com/partiql/partiql-lang-kotlin/blob/stored-procedure/examples/src/kotlin/org/partiql/examples/CustomProceduresExample.kt).

Previously approved PRs part of this PR:
- parsing + PIG definition https://github.com/partiql/partiql-lang-kotlin/pull/333
- evaluation + example https://github.com/partiql/partiql-lang-kotlin/pull/335

### DIFF
--- a/lang/src/org/partiql/lang/ast/ExprNodeToStatement.kt
+++ b/lang/src/org/partiql/lang/ast/ExprNodeToStatement.kt
@@ -16,6 +16,7 @@ fun ExprNode.toAstStatement(): PartiqlAst.Statement {
 
         is CreateTable, is CreateIndex, is DropTable, is DropIndex -> toAstDdl()
 
+        is Exec -> toAstExec()
     }
 }
 
@@ -30,7 +31,7 @@ private fun ExprNode.toAstDdl(): PartiqlAst.Statement {
         when(thiz) {
             is Literal, is LiteralMissing, is VariableReference, is Parameter, is NAry, is CallAgg, is Typed,
             is Path, is SimpleCase, is SearchedCase, is Select, is Struct, is Seq,
-            is DataManipulation -> error("Can't convert ${thiz.javaClass} to PartiqlAst.ddl")
+            is DataManipulation, is Exec -> error("Can't convert ${thiz.javaClass} to PartiqlAst.ddl")
 
             is CreateTable -> ddl(createTable(thiz.tableName), metas)
             is CreateIndex -> ddl(createIndex(identifier(thiz.tableName, caseSensitive()), thiz.keys.map { it.toAstExpr() }), metas)
@@ -44,6 +45,18 @@ private fun ExprNode.toAstDdl(): PartiqlAst.Statement {
             is DropTable ->
                 // case-sensitivity of table names cannot be represented with ExprNode.
                 ddl(dropTable(identifier(thiz.tableName, caseSensitive())), metas)
+        }
+    }
+}
+
+private fun ExprNode.toAstExec() : PartiqlAst.Statement {
+    val node = this
+    val metas = metas.toElectrolyteMetaContainer()
+
+    return PartiqlAst.build {
+        when (node) {
+            is Exec -> exec(node.procedureName.name, node.args.map { it.toAstExpr() }, metas)
+            else -> error("Can't convert ${node.javaClass} to PartiqlAst.Statement.Exec")
         }
     }
 }
@@ -147,8 +160,8 @@ fun ExprNode.toAstExpr(): PartiqlAst.Expr {
                     SeqType.BAG -> bag(node.values.map { it.toAstExpr() })
                 }
 
-            // These are handled by `toAstDml()`
-            is DataManipulation, is CreateTable, is CreateIndex, is DropTable, is DropIndex ->
+            // These are handled by `toAstDml()`, `toAstDdl()`, and `toAstExec()`
+            is DataManipulation, is CreateTable, is CreateIndex, is DropTable, is DropIndex, is Exec ->
                 error("Can't transform ${node.javaClass} to a PartiqlAst.expr }")
         }
     }
--- a/lang/src/org/partiql/lang/ast/ExprNodeToStatement.kt
+++ b/lang/src/org/partiql/lang/ast/ExprNodeToStatement.kt
@@ -2,6 +2,7 @@ package org.partiql.lang.ast
 
 import com.amazon.ionelement.api.toIonElement
 import org.partiql.lang.domains.PartiqlAst
+import org.partiql.pig.runtime.SymbolPrimitive
 import org.partiql.pig.runtime.asPrimitive
 
 /** Converts an [ExprNode] to a [PartiqlAst.statement]. */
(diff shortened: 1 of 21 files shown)


lang/src/org/partiql/lang/ast/ExprNodeToStatement.kt
```diff
@@ -48,6 +49,18 @@ private fun ExprNode.toAstDdl(): PartiqlAst.Statement {
     }
 }
 
+private fun ExprNode.toAstExec() : PartiqlAst.Statement {
+    val node = this
+    val metas = metas.toElectrolyteMetaContainer()
+
+    return PartiqlAst.build {
+        when (node) {
+            is Exec -> exec(node.procedureName.name, node.args.map { it.toAstExpr() }, metas)
```
Zhouenze, line 58: `procedureName.meta` is not kept in `PartiqlAst.Statement`? Not to say it's important but just want to confirm.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2214,9 +2242,23 @@ class SqlParser(private val ion: IonSystem) : Parser {
         return ParseNode(ARG_LIST, null, items, rem)
     }
 
+    /**
```
Zhouenze, line 2245: There are other tokens that should only exist at top level right? Why is `EXEC` different and requires a separate check?

lang/test/org/partiql/lang/errors/ParserErrorsTest.kt
```diff
@@ -1546,4 +1546,87 @@ class ParserErrorsTest : TestBase() {
...
+        "EXEC foo EXEC",
+        ErrorCode.PARSE_EXEC_AT_UNEXPECTED_LOCATION,
+        mapOf(
+            Property.LINE_NUMBER to 1L,
+            Property.COLUMN_NUMBER to 10L,
+            Property.TOKEN_TYPE to TokenType.KEYWORD,
+            Property.TOKEN_VALUE to ion.newSymbol("exec")))
+
+    @Test
+    fun execAtUnexpectedLocationInExpression() = checkInputThrowingParserException(
```
Zhouenze, line 1624: What will happen for `SELECT * FROM EXEC` or `SELECT * FROM exec`?

lang/src/org/partiql/lang/eval/builtins/storedprocedure/StoredProcedure.kt
```diff
@@ -0,0 +1,40 @@
...
+interface StoredProcedure {
+    /**
+     * [StoredProcedureSignature] representing the stored procedure's name and arity to be referenced in stored
+     * procedure calls.
+     */
+    val signature: StoredProcedureSignature
+
+    /**
+     * Invokes the stored procedure. Proper arity is checked by the [EvaluatingCompiler], but argument type checking
+     * is left to the implementation.
```
Zhouenze, line 34: Not related to this change but wonder if there could be some utility for user to easily build their type check. Would be useful in many places, not limited to stored procedure. BTW I heard `ExprValue` is going to be deprecated? Maybe it's better to consider utility after that.

lang/src/org/partiql/lang/errors/ErrorCode.kt
```diff
@@ -432,6 +445,16 @@ enum class ErrorCode(private val category: ErrorCategory,
             "got: ${errorContext?.get(Property.ACTUAL_ARGUMENT_TYPES) ?: UNKNOWN}"
     },
 
+    EVALUATOR_INCORRECT_TYPE_OF_ARGUMENTS_TO_PROCEDURE_CALL(
```
Zhouenze, line 448: According to comment, type check is done in procedure implementation, so usage of this error code depends on user's best intention. As mentioned in the other comment, this could be improved by taking over type check or providing utility to help user defining their own type check.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -1926,8 +1928,46 @@ internal class EvaluatingCompiler(
...
+        val procedure = procedures[procedureName.name] ?: err(
+            "No such stored procedure: ${procedureName.name}",
+            ErrorCode.EVALUATOR_NO_SUCH_PROCEDURE,
+            errorContextFrom(metas).also {
+                it[Property.PROCEDURE_NAME] = procedureName.name
+            },
+            internal = false)
+
+        // Check arity
+        if (args.size !in procedure.signature.arity) {
```
Zhouenze, line 1942: Just found something weird regarding `IntRange`: `IntRange(2,1)` is valid, no exception, but internally it assumes start <= end ([link](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/-int-range/)). `1 in IntRange(1,2)=true` but `1 in IntRange(2,1)=false`. Not to say we should do anything about it but maybe good to know.

Zhouenze (approved):

lang/src/org/partiql/lang/ast/ExprNodeToStatement.kt
alancai98: That particular `meta` may not be too important, but we should be copying metas wherever possible. I will add that in the next commit.

I noticed there are quite a few other metas that aren't copied when converting `ExprNode`<-> `PartiqlAst`, so I created an issue (#353) to track.

lang/src/org/partiql/lang/syntax/SqlParser.kt
alancai98: The other tokens that should only exist at the top level (i.e. DML and DDL ops) weren't properly checked in the parser (created an issue #354 to add these checks in the future). I will rename the `checkForUnexpectedExec()` to something more general for all tokens that should only exist in the top level.

lang/test/org/partiql/lang/errors/ParserErrorsTest.kt
alancai98: Currently for both of those queries, the `PARSE_EXEC_AT_UNEXPECTED_LOCATION` will be thrown. Perhaps I can make it more general for other tokens that should only be at the top-level.

lang/src/org/partiql/lang/eval/builtins/storedprocedure/StoredProcedure.kt
alancai98: We have some future plans to add a utility to help type-check stored procedures and user-defined functions. 

`ExprValue` will likely be deprecated sometime in the future, so the utility will probably be implemented with its replacement.

lang/src/org/partiql/lang/errors/ErrorCode.kt
alancai98: As mentioned in the other reply, we do have plans to add a type-checking utility for sprocs and udfs, so once that's added, throwing of this error should be handled by that utility.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
alancai98: Hmm interesting. I don't think that internally `IntRange` assumes start <= end. The Kotlin 1.3 jar even defines an `EMPTY` range as `IntRange(1, 0)`. It seems `in` (i.e. `contains`) only returns true if and only if `start <= value && value <= last`, so the behavior makes some sense.

This leads me to think that nothing should be done unless we decide not to use `IntRange`.

Zhouenze (approved):

dlurton (approved):

PR #351, opened 2020-12-24
gradle: fix test target for cli, examples, pts

evanj:
*Description of changes:*

The following gradle targets were not running any of the unit tests
contained in their directories due to incorrect test dependencies. It
appears these projects all use JUnit 5 "vintage" tests, so need that
engine as a test dependency.

    :cli:test
    :examples:test
    :pts:test

CliTest: Add test for the PARTIQL_PRETTY output format added in commit
91750026f1. This was how I discovered this issue.

examples: Remove tests from the implementation dependencies.

### DIFF
--- a/cli/build.gradle
+++ b/cli/build.gradle
@@ -1,6 +1,5 @@
 plugins {
     id 'org.jetbrains.kotlin.jvm'
-    id 'io.gitlab.arturbosch.detekt'
     id 'application'
 }
 
@@ -35,10 +34,7 @@ dependencies {
     implementation 'net.sf.jopt-simple:jopt-simple:[5.0,6.0)'
     implementation 'org.jetbrains.kotlin:kotlin-stdlib-jdk8'
 
-    testImplementation 'org.jetbrains.kotlin:kotlin-test'
-    testImplementation 'pl.pragmatists:JUnitParams:[1.0.0,1.1.0)'
-
-    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.7.0'
+    testImplementation 'org.junit.vintage:junit-vintage-engine:5.7.0'
 }
 
 jacocoTestReport {
--- a/cli/test/org/partiql/cli/CliTest.kt
+++ b/cli/test/org/partiql/cli/CliTest.kt
@@ -112,4 +112,13 @@ class CliTest {
 
         assertEquals("<<{'a': 1}>>", actual)
     }
+
+    @Test
+    fun withPartiQLPrettyOutput() {
+        val subject = makeCli("SELECT * FROM input_data", "{a: 1, b: 2}",
+            outputFormat = OutputFormat.PARTIQL_PRETTY)
+        val actual = subject.runAndOutput()
+
+        assertEquals("<<\n  {\n    'a': 1,\n    'b': 2\n  }\n>>", actual)
+    }
 }
--- a/cli/test/org/partiql/cli/CliTest.kt
+++ b/cli/test/org/partiql/cli/CliTest.kt
@@ -115,8 +115,7 @@ class CliTest {
 
     @Test
     fun withPartiQLPrettyOutput() {
-        val subject = makeCli("SELECT * FROM input_data", "{a: 1, b: 2}",
-            outputFormat = OutputFormat.PARTIQL_PRETTY)
+        val subject = makeCli("SELECT * FROM input_data", "{a: 1, b: 2}", outputFormat = OutputFormat.PARTIQL_PRETTY)
         val actual = subject.runAndOutput()
 
         assertEquals("<<\n  {\n    'a': 1,\n    'b': 2\n  }\n>>", actual)
--- a/examples/src/kotlin/org/partiql/examples/ParserExample.kt
+++ b/examples/src/kotlin/org/partiql/examples/ParserExample.kt
@@ -6,7 +6,6 @@ import org.partiql.lang.ast.*
 import org.partiql.lang.domains.PartiqlAst
 import org.partiql.lang.syntax.*
 import java.io.PrintStream
-import java.util.Objects
 
 /**
  * Demonstrates the use of [SqlParser] and [PartiqlAst].
--- a/examples/src/kotlin/org/partiql/examples/ParserExample.kt
+++ b/examples/src/kotlin/org/partiql/examples/ParserExample.kt
@@ -1,12 +1,12 @@
 package org.partiql.examples
 
-import kotlin.test.*
 import com.amazon.ion.system.*
 import org.partiql.examples.util.Example
 import org.partiql.lang.ast.*
 import org.partiql.lang.domains.PartiqlAst
 import org.partiql.lang.syntax.*
 import java.io.PrintStream
+import java.util.Objects
 
 /**
  * Demonstrates the use of [SqlParser] and [PartiqlAst].
@@ -45,7 +45,9 @@ class ParserExample(out: PrintStream) : Example(out) {
         // version of the s-expression form to an instance of [ExprNode].
         val deserializedAst = serializedAst.toExprNode(ion)
         // Verify that we have the correct AST.
-        assertEquals(originalAst, deserializedAst)
+        if (originalAst != deserializedAst) {
+            throw Exception("expected AST to be equal")
+        }
 
         // Here we show how to parse a query directly to a PartiqlAst statement
         val statement = parser.parseAstStatement(query)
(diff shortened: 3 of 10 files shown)


examples/src/kotlin/org/partiql/examples/ParserExample.kt
```diff
@@ -45,7 +45,9 @@ class ParserExample(out: PrintStream) : Example(out) {
         // version of the s-expression form to an instance of [ExprNode].
         val deserializedAst = serializedAst.toExprNode(ion)
         // Verify that we have the correct AST.
-        assertEquals(originalAst, deserializedAst)
+        if (originalAst != deserializedAst) {
+            throw Exception("expected AST to be equal")
+        }
```
therapon, line 50: Could we instead use the `assertEquals` with a message parameter for the exception? I am not sure what we are gaining by using `if` and throwing our own exception here, perhaps I am missing something though.

examples/src/kotlin/org/partiql/examples/ParserExample.kt
```diff
@@ -55,6 +57,8 @@ class ParserExample(out: PrintStream) : Example(out) {
         // and back into a PartiqlAst statement
         val roundTrippedStatement = PartiqlAst.transform(elements) as PartiqlAst.Statement
         // Verify that we have the original Partiql Ast statement
-        assertEquals(statement, roundTrippedStatement)
+        if (statement != roundTrippedStatement) {
+            throw Exception("Expected statements to be the same")
+        }
```
therapon, line 62: Same argument here concerning `assertEquals`

lang/build.gradle
```diff
@@ -48,17 +48,12 @@ dependencies {
...
     testImplementation 'org.jetbrains.kotlin:kotlin-test-junit5'
     testImplementation 'pl.pragmatists:JUnitParams:[1.0.0,1.1.0)'
     testImplementation 'org.assertj:assertj-core:[3.11.0,3.12.0)'
 
-    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.7.0'
-    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.7.0'
     testImplementation 'org.junit.vintage:junit-vintage-engine:5.7.0'
     testImplementation 'org.junit.jupiter:junit-jupiter-params:5.7.0'
-    // PTS
-    testImplementation project(':testscript')
```
therapon, line 61: Was this dependency breaking anything? This (I believe) is what hooks into the PTS (included in this repo) system which runs more tests focusing on spec compliance.

cli/test/org/partiql/cli/CliTest.kt
```diff
@@ -112,4 +112,13 @@ class CliTest {
 
         assertEquals("<<{'a': 1}>>", actual)
     }
+
+    @Test
+    fun withPartiQLPrettyOutput() {
+        val subject = makeCli("SELECT * FROM input_data", "{a: 1, b: 2}",
+            outputFormat = OutputFormat.PARTIQL_PRETTY)
```
alancai98, line 119: nit: could move onto the same line as the previous test

```suggestion
        val subject = makeCli("SELECT * FROM input_data", "{a: 1, b: 2}", outputFormat = OutputFormat.PARTIQL_PRETTY)
```

cli/build.gradle
```diff
@@ -1,6 +1,5 @@
 plugins {
     id 'org.jetbrains.kotlin.jvm'
-    id 'io.gitlab.arturbosch.detekt'
```
alancai98, line 3: Why is this plugin removed?

alancai98 (commented):
Good catch. I think my previous PR (#338) to fix Gradle to run on `ParameterizedTest`s introduced this testing target bug.

Only have a few nits/questions.

examples/src/kotlin/org/partiql/examples/ParserExample.kt
evanj: The reason was to avoid requiring the `kotlin-test` JAR from code that is technically not an automated unit test. I will revert that part of the change since you would prefer to use this dependency.

lang/build.gradle
evanj: It was not breaking anything, but it seems to be unnecessary as the code builds without it? I haven't read the code in detail, but it appears to me that the dependency tree is that the `pts` targets/directory imports the code in `testscript`. I was assuming that means those are the targets that actually use this code. Nothing in the `lang` directory references testscript AFAIK?

I barely understand this though; I just was curious if all these dependencies were actually needed. I will revert my changes to this file: the important updates are to `cli`, `examples`, and `pts` so `./gradlew test` actually executes the tests in those directories.

examples/src/kotlin/org/partiql/examples/ParserExample.kt
```diff
@@ -1,12 +1,12 @@
 package org.partiql.examples
 
-import kotlin.test.*
 import com.amazon.ion.system.*
 import org.partiql.examples.util.Example
 import org.partiql.lang.ast.*
 import org.partiql.lang.domains.PartiqlAst
 import org.partiql.lang.syntax.*
 import java.io.PrintStream
+import java.util.Objects
```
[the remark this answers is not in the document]
evanj: Oops thanks sorry; This was a stray import from a previous edit.

cli/build.gradle
evanj: I don't understand gradle. I assumed since this file has no references to it, this is unnecessary?

The top level `build.gradle` references `detekt`, and I assumed the `apply plugin` thing attempts to apply the plugin to all subprojects? If so, doesn't that mean this is unnecessary, since the top level project includes it?

FWIW, with this change, the `./gradlew cli:detekt` task still seems to work?

```
~/partiql-lang-kotlin $ ./gradlew cli:detekt --info
...
> Task :cli:detekt UP-TO-DATE
...
```

evanj (commented):
I will make the suggested changes ... but maybe not for a few days; thanks for the prompt review!

examples/src/kotlin/org/partiql/examples/ParserExample.kt
therapon: Thanks for explaining. In that case, ignore my comments for this file on using `assert`. Let's keep the examples clean of testing libs.

cli/build.gradle
alancai98: I also don't understand gradle too well. You seem to be right regarding the top level references to `detekt`, which is applied to all the subprojects. Since the `detekt` plugin isn't referenced in the other subprojects, could you also remove the `detekt` plugin from
- [lang/build.gradle](https://github.com/partiql/partiql-lang-kotlin/blob/d8fc1691b7335adb4856924f5b113e4698c16d34/lang/build.gradle#L29)
- [pts/build.gradle](https://github.com/partiql/partiql-lang-kotlin/blob/d8fc1691b7335adb4856924f5b113e4698c16d34/pts/build.gradle#L19)
- [testscript/build.gradle](https://github.com/partiql/partiql-lang-kotlin/blob/d8fc1691b7335adb4856924f5b113e4698c16d34/testscript/build.gradle#L3)

evanj: Okay, done, thanks.

evanj:
sorry for the delay. I believe I addressed all the comments. Thanks again for the prompt review!

lang/build.gradle
```diff
@@ -48,17 +47,12 @@ dependencies {
     api 'org.partiql:partiql-ir-generator-runtime:0.3.0'
 
     // test-time dependencies
-    testImplementation 'org.jetbrains.kotlin:kotlin-reflect'
     testImplementation 'org.jetbrains.kotlin:kotlin-test-junit5'
     testImplementation 'pl.pragmatists:JUnitParams:[1.0.0,1.1.0)'
     testImplementation 'org.assertj:assertj-core:[3.11.0,3.12.0)'
 
-    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.7.0'
```
alancai98, line 56: We need this dependency to run  junit `ParameterizedTest`s with `ArgumentsSource` that we are starting to use (e.g. [EvaluatingCompilerFromLetTests.kt](https://github.com/partiql/partiql-lang-kotlin/blob/0ca86ac64984350dd236552de99519f2115a9286/lang/test/org/partiql/lang/eval/EvaluatingCompilerFromLetTests.kt)). Without this dependency, these tests aren't being run by gradle.

alancai98 (commented):
Sorry, I missed a deleted line in the first review. Once that's added back, should be approved and merged. Thanks again for the contribution!

lang/build.gradle
evanj: Oops nice catch thanks! This is probably why I should not make wholesale changes without understanding what I am actually doing :)

alancai98 (approved):

PR #356, opened 2021-01-11
Migrate `AstRewriterBase` to PIG's `VisitorTransform`

alancai98:
Migrates `AstRewriterBase` to PIG's `VisitorTransform`.


dlurton (changes requested):
Overall LGTM (as it should since each commit was a previously reviewed PR!) but I think we need to add a `CompilerPipeline.addPreprocessingStep` overload for `PartiqlAst`.

dlurton (dismissed):
Actually, if you want you can do that `addProcessingStep` overload in a PR and merge this one.  Approving now so you have that option.

alancai98:
New commit
- moves `MetaStrippingRewriter` back to the `src` directory and adds a deprecation warning
- adds back `RewriterTestBase` and adds a deprecation warning
- deletes deprecation warnings from `ExprNodeExtensions`

I felt the `MetaStrippingRewriter` and `RewriterTestBase` additions/changes are needed (at least for now) to not break existing client code that may rely on those public APIs. Deprecation warnings were added to prevent further usage.

I deleted the deprecation warnings in `ExprNodeExtensions` because the functions are still in use by the `EvaluatingCompiler`. In the future when we deprecate `ExprNode`, we should deprecate the `ExprNodeExtensions` functions.

As for the `addProcessingStep` overload, I will make a separate PR to track that change since it may involve adding some additional deprecation warnings.

dlurton (approved):

Issue #363, opened 2021-02-01
Parser accepts statement that doesn't follow type-domain grammar


Zhouenze:
Hi,

I noticed that there's discrepancy between the grammar parser accepts and the grammar in [partiql.ion](https://github.com/partiql/partiql-lang-kotlin/blob/master/lang/resources/org/partiql/type-domains/partiql.ion).

```
    // In org.partiql.lang.syntax SqlParserTest.kt
    @Test
    fun reproError() {
        val parsedExprNode = parse("update test set x = set test.y = 6")  // Succeed
        println(parsedExprNode.toAstStatement())                          // Fail with IllegalStateException
    }
```

Seems like parser should throw `ParserException` in this case?

Thanks!

dlurton:
Seems related to: https://github.com/partiql/partiql-lang-kotlin/issues/354

PR #369, opened 2021-02-26
Fixes parser for the top level tokens

abhikuhikar:
Issue #363 

Fixes parser by traversing and validating the parse tree to ensure that the top level tokens are not present at the level(s) below. 

### DIFF
--- a/lang/src/org/partiql/lang/syntax/SqlParser.kt
+++ b/lang/src/org/partiql/lang/syntax/SqlParser.kt
@@ -997,7 +997,7 @@ class SqlParser(private val ion: IonSystem) : Parser {
         KEYWORD -> when (head?.keywordText) {
             in BASE_DML_KEYWORDS -> parseBaseDml()
             "update" -> tail.parseUpdate()
-            "delete" -> tail.parseDelete()
+            "delete" -> tail.parseDelete(head!!)
             "case" -> when (tail.head?.keywordText) {
                 "when" -> tail.parseCase(isSimple = false)
                 else -> tail.parseCase(isSimple = true)
@@ -1220,12 +1220,12 @@ class SqlParser(private val ion: IonSystem) : Parser {
         copy(type = type)
     }
 
-    private fun List<Token>.parseDelete(): ParseNode {
+    private fun List<Token>.parseDelete(name: Token): ParseNode {
         if (head?.keywordText != "from") {
             err("Expected FROM after DELETE", PARSE_UNEXPECTED_TOKEN)
         }
 
-        return tail.parseLegacyDml { ParseNode(DELETE, null, emptyList(), this) }
+        return tail.parseLegacyDml { ParseNode(DELETE, name, emptyList(), this) }
     }
 
     private fun List<Token>.parseUpdate(): ParseNode = parseLegacyDml {
@@ -1722,6 +1722,12 @@ class SqlParser(private val ion: IonSystem) : Parser {
         var rem = this
         if (rem.head?.type == EOF) {
             rem.err("No stored procedure provided", PARSE_NO_STORED_PROCEDURE_PROVIDED)
+        }
+
+        rem.forEach {
+            if (it.keywordText?.toLowerCase() == "exec") {
+                it.err("EXEC call found at unexpected location", PARSE_UNEXPECTED_TERM)
+            }
         }
 
         val procedureName = rem.head
@@ -2244,25 +2250,11 @@ class SqlParser(private val ion: IonSystem) : Parser {
         return ParseNode(ARG_LIST, null, items, rem)
     }
 
-    /**
-     * Checks that the given [Token] list does not have any top-level tokens outside of the top level query. Throws
-     * an error if so.
-     *
-     * Currently only checks for `EXEC`. DDL and DML along with corresponding error codes to be added in the future
-     * (https://github.com/partiql/partiql-lang-kotlin/issues/354).
-     */
-    private fun List<Token>.checkUnexpectedTopLevelToken() {
-        for ((index, token) in this.withIndex()) {
-            if (token.keywordText == "exec" && index != 0) {
-                token.err("EXEC call found at unexpected location", PARSE_EXEC_AT_UNEXPECTED_LOCATION)
-            }
-        }
-    }
 
     /**
      * Checks if the [ParseNode] requires to be only at top level. DML keyword tokens, for example, requires to be top level.
      */
-    private fun ParseNode.isTopLevelType() = type in listOf(UPDATE, DELETE, INSERT)
+    private fun ParseNode.isTopLevelType() = type in listOf(UPDATE, DELETE, INSERT, INSERT_VALUE, REMOVE, EXEC)
 
     /**
      * Validates tree to make sure that the top level tokens are not found below the top level
@@ -2270,7 +2262,8 @@ class SqlParser(private val ion: IonSystem) : Parser {
     private fun validateTopLevelNodes(node: ParseNode, level: Int) {
         // Note that for DML operations, top level parse node is of type 'FROM'. Hence the check level > 1
         if (level > 1 && node.isTopLevelType()) {
-            node.remaining.err("Type ${node.type} only expected at the top level", PARSE_UNEXPECTED_TERM)
+            node.token?.err("Type ${node.type} only expected at the top level", PARSE_UNEXPECTED_TERM)
+                ?: throw ParserException("Type ${node.type} only expected at the top level", PARSE_UNEXPECTED_TERM, PropertyValueMap())
         }
         node.children.map { validateTopLevelNodes(it, level + 1) }
     }
@@ -2278,7 +2271,6 @@ class SqlParser(private val ion: IonSystem) : Parser {
     /** Entry point into the parser. */
     override fun parseExprNode(source: String): ExprNode {
         val tokens = SqlLexer(ion).tokenize(source)
-        tokens.checkUnexpectedTopLevelToken()
         val node = tokens.parseExpression()
         val rem = node.remaining
         if (!rem.onlyEndOfStatement()) {
--- a/lang/src/org/partiql/lang/syntax/SqlParser.kt
+++ b/lang/src/org/partiql/lang/syntax/SqlParser.kt
@@ -2259,6 +2259,11 @@ class SqlParser(private val ion: IonSystem) : Parser {
         }
     }
 
+    /**
+     * Checks if the [ParseNode] requires to be only at top level. DML keyword tokens, for example, requires to be top level.
+     */
+    private fun ParseNode.isTopLevelType() = type in listOf(UPDATE, DELETE, INSERT)
+
     /** Entry point into the parser. */
     override fun parseExprNode(source: String): ExprNode {
         val tokens = SqlLexer(ion).tokenize(source)
@@ -2272,6 +2277,18 @@ class SqlParser(private val ion: IonSystem) : Parser {
                 else      -> rem.err("Unexpected token after expression", PARSE_UNEXPECTED_TOKEN)
             }
         }
+
+        // validate tree to make sure that the top level types are not found below the top level
+        fun validateTopLevelNodes(node: ParseNode, level: Int) {
+            // Note that for DML operations, top level parse node is of type 'FROM'. Hence the check level > 1
+            if (level > 1 && node.isTopLevelType()) {
+                rem.err("Type ${node.type} only expected at the top level", PARSE_UNEXPECTED_TERM)
+            }
+            node.children.map { validateTopLevelNodes(it, level + 1) }
+        }
+
+        validateTopLevelNodes(node, 0)
+
         return node.toExprNode()
     }
 
--- a/lang/src/org/partiql/lang/syntax/SqlParser.kt
+++ b/lang/src/org/partiql/lang/syntax/SqlParser.kt
@@ -55,7 +55,10 @@ class SqlParser(private val ion: IonSystem) : Parser {
         SIMPLE_PATH
     }
 
-    internal enum class ParseType(val isJoin: Boolean = false) {
+    /** If the property [isTopLevelType] is true, then the parse node of this ParseType will only be valid at top level in the query.
+     *  For example, EXEC, DDL and DML keywords can only be used at the top level in the query.
+     */
+    internal enum class ParseType(val isJoin: Boolean = false, val isTopLevelType: Boolean = false, val isDml: Boolean = false) {
         ATOM,
         CASE_SENSITIVE_ATOM,
         CASE_INSENSITIVE_ATOM,
@@ -100,22 +103,22 @@ class SqlParser(private val ion: IonSystem) : Parser {
         WHEN,
         ELSE,
         BAG,
-        INSERT,
-        INSERT_VALUE,
-        REMOVE,
+        INSERT(isTopLevelType = true, isDml = true),
+        INSERT_VALUE(isTopLevelType = true, isDml = true),
+        REMOVE(isTopLevelType = true, isDml = true),
         SET,
-        UPDATE,
-        DELETE,
+        UPDATE(isTopLevelType = true, isDml = true),
+        DELETE(isTopLevelType = true, isDml = true),
         ASSIGNMENT,
         FROM,
         FROM_CLAUSE,
         FROM_SOURCE_JOIN,
-        CREATE_TABLE,
-        DROP_TABLE,
-        DROP_INDEX,
-        CREATE_INDEX,
+        CREATE_TABLE(isTopLevelType = true),
+        DROP_TABLE(isTopLevelType = true),
+        DROP_INDEX(isTopLevelType = true),
+        CREATE_INDEX(isTopLevelType = true),
         PARAMETER,
-        EXEC;
+        EXEC(isTopLevelType = true);
 
         val identifier = name.toLowerCase()
     }
@@ -2252,16 +2255,11 @@ class SqlParser(private val ion: IonSystem) : Parser {
 
 
     /**
-     * Checks if the [ParseNode] requires to be only at top level. DML keyword tokens, for example, requires to be top level.
-     */
-    private fun ParseNode.isTopLevelType() = type in listOf(UPDATE, DELETE, INSERT, INSERT_VALUE, REMOVE, EXEC)
-
-    /**
      * Validates tree to make sure that the top level tokens are not found below the top level
      */
     private fun validateTopLevelNodes(node: ParseNode, level: Int) {
         // Note that for DML operations, top level parse node is of type 'FROM'. Hence the check level > 1
-        if (level > 1 && node.isTopLevelType()) {
+        if (node.type.isTopLevelType && ((node.type.isDml && level > 1) || (!node.type.isDml && level > 0))) {
             node.token?.err("Type ${node.type} only expected at the top level", PARSE_UNEXPECTED_TERM)
                 ?: throw ParserException("Type ${node.type} only expected at the top level", PARSE_UNEXPECTED_TERM, PropertyValueMap())
         }
--- a/lang/src/org/partiql/lang/syntax/SqlParser.kt
+++ b/lang/src/org/partiql/lang/syntax/SqlParser.kt
@@ -2264,6 +2264,17 @@ class SqlParser(private val ion: IonSystem) : Parser {
      */
     private fun ParseNode.isTopLevelType() = type in listOf(UPDATE, DELETE, INSERT)
 
+    /**
+     * Validates tree to make sure that the top level tokens are not found below the top level
+     */
+    private fun validateTopLevelNodes(node: ParseNode, level: Int) {
+        // Note that for DML operations, top level parse node is of type 'FROM'. Hence the check level > 1
+        if (level > 1 && node.isTopLevelType()) {
+            node.remaining.err("Type ${node.type} only expected at the top level", PARSE_UNEXPECTED_TERM)
+        }
+        node.children.map { validateTopLevelNodes(it, level + 1) }
+    }
+
     /** Entry point into the parser. */
     override fun parseExprNode(source: String): ExprNode {
         val tokens = SqlLexer(ion).tokenize(source)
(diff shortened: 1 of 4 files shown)


lang/test/org/partiql/lang/errors/ParserErrorsTest.kt
```diff
@@ -1246,6 +1246,16 @@ class ParserErrorsTest : TestBase() {
...
 
+    @Test
+    fun updateWithNestedSet() = checkInputThrowingParserException(
+        "UPDATE test SET x = SET test.y = 6",
+        ErrorCode.PARSE_UNEXPECTED_TERM,
+        mapOf(
+            Property.LINE_NUMBER to 1L,
+            Property.COLUMN_NUMBER to 35L,
+            Property.TOKEN_TYPE to TokenType.EOF,
+            Property.TOKEN_VALUE to ion.newSymbol("EOF")))
```
alancai98, line 1257: Shouldn't the unexpected parse term be `SET` here and not `EOF`?

alancai98 (changes requested):
The approach lgtm overall. Could add more top-level token checks to `EXEC` and other DML/DDL ops. Also may need more tests.

lang/test/org/partiql/lang/syntax/SqlParserTest.kt
```diff
@@ -2636,6 +2636,32 @@ class SqlParserTest : SqlParserTestBase() {
         """
     )
 
+    @Test
+    fun fromWhereSetDml() = assertExpression(
```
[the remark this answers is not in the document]
abhikuhikar: Ohh yes. I missed that. Thanks. I will remove it.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2259,6 +2259,22 @@ class SqlParser(private val ion: IonSystem) : Parser {
         }
     }
 
+    /**
+     * Checks if the [ParseNode] requires to be only at top level. DML keyword tokens, for example, requires to be top level.
+     */
+    private fun ParseNode.isTopLevelType() = type in listOf(UPDATE, DELETE, INSERT)
+
+    /**
+     * Validates tree to make sure that the top level tokens are not found below the top level
```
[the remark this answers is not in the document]
abhikuhikar: Since it is a tree representation, "below" makes sense IMO.

lang/test/org/partiql/lang/errors/ParserErrorsTest.kt
```diff
@@ -1246,6 +1246,16 @@ class ParserErrorsTest : TestBase() {
             Property.TOKEN_TYPE to TokenType.EOF,
             Property.TOKEN_VALUE to ion.newSymbol("EOF")))
 
+    @Test
```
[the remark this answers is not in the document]
abhikuhikar: Thanks. Added tests for other top level tokens.

lang/test/org/partiql/lang/errors/ParserErrorsTest.kt
abhikuhikar: You are right. Thanks. I have made changes in the next revision.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2259,6 +2259,22 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
+     */
+    private fun validateTopLevelNodes(node: ParseNode, level: Int) {
+        // Note that for DML operations, top level parse node is of type 'FROM'. Hence the check level > 1
+        if (level > 1 && node.isTopLevelType()) {
+            node.remaining.err("Type ${node.type} only expected at the top level", PARSE_UNEXPECTED_TERM)
+        }
```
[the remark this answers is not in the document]
abhikuhikar: I think the logic remains the same for the top level tokens even if it is DML or not.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2259,6 +2259,22 @@ class SqlParser(private val ion: IonSystem) : Parser {
         }
     }
 
+    /**
+     * Checks if the [ParseNode] requires to be only at top level. DML keyword tokens, for example, requires to be top level.
+     */
+    private fun ParseNode.isTopLevelType() = type in listOf(UPDATE, DELETE, INSERT)
```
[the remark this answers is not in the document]
abhikuhikar: Added.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2259,6 +2259,22 @@ class SqlParser(private val ion: IonSystem) : Parser {
         }
     }
 
+    /**
+     * Checks if the [ParseNode] requires to be only at top level. DML keyword tokens, for example, requires to be top level.
+     */
```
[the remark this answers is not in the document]
abhikuhikar: Added `EXEC`.

lang/test/org/partiql/lang/errors/ParserErrorsTest.kt
```diff
@@ -1623,10 +1673,10 @@ class ParserErrorsTest : TestBase() {
...
-        ErrorCode.PARSE_EXEC_AT_UNEXPECTED_LOCATION,
+        ErrorCode.PARSE_UNEXPECTED_TERM,
         mapOf(
             Property.LINE_NUMBER to 1L,
-            Property.COLUMN_NUMBER to 16L,
-            Property.TOKEN_TYPE to TokenType.KEYWORD,
-            Property.TOKEN_VALUE to ion.newSymbol("exec")))
+            Property.COLUMN_NUMBER to 21L,
+            Property.TOKEN_TYPE to TokenType.IDENTIFIER,
+            Property.TOKEN_VALUE to ion.newSymbol("undrop")))
```
alancai98, line 1681: Shouldn't the unexpected term be `EXEC` here?

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2244,25 +2250,27 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
     }
 
+
     /**
-     * Checks that the given [Token] list does not have any top-level tokens outside of the top level query. Throws
-     * an error if so.
-     *
-     * Currently only checks for `EXEC`. DDL and DML along with corresponding error codes to be added in the future
-     * (https://github.com/partiql/partiql-lang-kotlin/issues/354).
+     * Checks if the [ParseNode] requires to be only at top level. DML keyword tokens, for example, requires to be top level.
```
alancai98, line 2255: super nit: at least for me the kdoc wording is confusing. From the implementation, it checks the ParseNode's parse type is one of the top level types right? The first sentence could be something like:
```kotlin
/**
 * Returns true if and only if this [ParseType] is one of the top level types...
 */
```
For the second sentence, it might be helpful to also mention that DML can have `FROM` as the top-level node.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2244,25 +2250,27 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
-            if (token.keywordText == "exec" && index != 0) {
-                token.err("EXEC call found at unexpected location", PARSE_EXEC_AT_UNEXPECTED_LOCATION)
-            }
+    private fun ParseNode.isTopLevelType() = type in listOf(UPDATE, DELETE, INSERT, INSERT_VALUE, REMOVE, EXEC)
+
+    /**
+     * Validates tree to make sure that the top level tokens are not found below the top level
+     */
+    private fun validateTopLevelNodes(node: ParseNode, level: Int) {
+        // Note that for DML operations, top level parse node is of type 'FROM'. Hence the check level > 1
```
alancai98, line 2263: super nit: the top level parse node **could** be of type `FROM` (not necessary for all DML ops)

alancai98 (commented):
Just a few minor nits for the comments/kdoc I missed from the first review. I probably overloaded [this comment](https://github.com/partiql/partiql-lang-kotlin/pull/369#discussion_r584919948) from the first review, but will DDL ops also be included?

lang/src/org/partiql/lang/syntax/SqlParser.kt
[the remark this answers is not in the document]
alancai98: Sorry, I may have put too much in the first comment, but will [`DDL` keywords](https://github.com/partiql/partiql-lang-kotlin/blob/master/lang/resources/org/partiql/type-domains/partiql.ion#L234-L250) also be added?

lang/test/org/partiql/lang/errors/ParserErrorsTest.kt
abhikuhikar: It should be. But the associated token with the `EXEC` ParseNode is the name of the stored procedure. Hence, the issue. Ti get around with this, I think we can pass the name of the stored procedure in the arguments along with the other `arglist` or some such. But that's worthy of a separate issue.

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: I will update it. Thanks.

lang/src/org/partiql/lang/syntax/SqlParser.kt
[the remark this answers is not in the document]
abhikuhikar: We can always add token types to the top level types list.

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: It is for all DML ops if I am not wrong. I have updated the logic to take this into consideration.

abhikuhikar:
- Updated the logic to check for DDL commands as well. 
- Changed `ParseType` to include properties `isTopLevelType` and `isDml`.
- Added tests to cover DDL and DML top level token types.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2244,25 +2253,22 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
-    private fun List<Token>.checkUnexpectedTopLevelToken() {
-        for ((index, token) in this.withIndex()) {
-            if (token.keywordText == "exec" && index != 0) {
-                token.err("EXEC call found at unexpected location", PARSE_EXEC_AT_UNEXPECTED_LOCATION)
-            }
+    private fun validateTopLevelNodes(node: ParseNode, level: Int) {
+        // Note that for DML operations, top level parse node is of type 'FROM'. Hence the check level > 1
+        if (node.type.isTopLevelType && ((node.type.isDml && level > 1) || (!node.type.isDml && level > 0))) {
+            node.token?.err("Type ${node.type} only expected at the top level", PARSE_UNEXPECTED_TERM)
+                ?: throw ParserException("Type ${node.type} only expected at the top level", PARSE_UNEXPECTED_TERM, PropertyValueMap())
```
alancai98, line 2264: nit: I think the word "Type" shouldn't be included in the error message. Maybe it could be replaced with "Keyword"?

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2244,25 +2253,22 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
+     * Validates tree to make sure that the top level tokens are not found below the top level
      */
-    private fun List<Token>.checkUnexpectedTopLevelToken() {
-        for ((index, token) in this.withIndex()) {
-            if (token.keywordText == "exec" && index != 0) {
-                token.err("EXEC call found at unexpected location", PARSE_EXEC_AT_UNEXPECTED_LOCATION)
-            }
+    private fun validateTopLevelNodes(node: ParseNode, level: Int) {
+        // Note that for DML operations, top level parse node is of type 'FROM'. Hence the check level > 1
+        if (node.type.isTopLevelType && ((node.type.isDml && level > 1) || (!node.type.isDml && level > 0))) {
```
alancai98, line 2262: nit: This can be a little hard to understand with so many logic symbols. Can this be simplified or broken into multiple if-statements?

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -55,7 +55,10 @@ class SqlParser(private val ion: IonSystem) : Parser {
         SIMPLE_PATH
     }
 
-    internal enum class ParseType(val isJoin: Boolean = false) {
+    /** If the property [isTopLevelType] is true, then the parse node of this ParseType will only be valid at top level in the query.
+     *  For example, EXEC, DDL and DML keywords can only be used at the top level in the query.
+     */
+    internal enum class ParseType(val isJoin: Boolean = false, val isTopLevelType: Boolean = false, val isDml: Boolean = false) {
```
alancai98, line 61: How is the `isDml` parameter defined? `SET` is a DML op, yet `ParseType.SET` `isDml` is `false`.

lang/test/org/partiql/lang/errors/ParserErrorsTest.kt
```diff
@@ -1246,6 +1246,96 @@ class ParserErrorsTest : TestBase() {
             Property.TOKEN_TYPE to TokenType.EOF,
             Property.TOKEN_VALUE to ion.newSymbol("EOF")))
 
+    @Test
+    fun updateWithNestedSet() = checkInputThrowingParserException(
+        "UPDATE test SET x = SET test.y = 6",
```
alancai98, line 1251: So when I'm testing this query in the REPL, it gives the following:

```
PartiQL> UPDATE test SET x = SET test.y = 6;
org.partiql.lang.syntax.ParserException: Type UPDATE only expected at the top level
	Parser Error: at line 1, column 21: unexpected term found, KEYWORD : set
```
Shouldn't the 2nd line say `SET` and not `UPDATE`? I haven't had a chance to test w/ the other top-level tokens.

lang/src/org/partiql/lang/syntax/SqlParser.kt
alancai98: From the [partiql.ion](https://github.com/partiql/partiql-lang-kotlin/blob/master/lang/resources/org/partiql/type-domains/partiql.ion#L18-L26) grammar for DML ops, the `FROM` source is optional.

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: So the `parseType` is different from the keyword or token, fortunately or unfortunately. I guess we can remove the error message - `type only expected at top level` altogether and just stay with the `unexpected term found` error  message.

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: This is a legit observation. However, the `SET` `parseType` is not used anywhere. We can actually get rid of it.

lang/test/org/partiql/lang/errors/ParserErrorsTest.kt
abhikuhikar: The associated parse type is `UPDATE`.
I will change this to - `"Keyword ${node.token} only expected at the top level"`.
It makes more sense I guess and would be less confusing.

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: It will make the code redundant. I also thought the conditional statement is complicated initially  but on the second glance it is not that complicated. 

```
if (node.type.isTopLevelType && level > 0) {
    if (node.type.isDml) {
        if (level > 1) {
            // throw error
        }
    } else {
        // throw error
    }
}
```

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: Changed the property `isDml` to `true` for `SET`.

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: Changed as mentioned above.

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: Changed the error msg.

abhikuhikar:
Came across interesting edge case where the DML tokens are at top level. Updated the PR to handle these edge cases.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2272,6 +2295,9 @@ class SqlParser(private val ion: IonSystem) : Parser {
                 else      -> rem.err("Unexpected token after expression", PARSE_UNEXPECTED_TOKEN)
             }
         }
+
+        validateTopLevelNodes(node, 0, 0)
```
alancai98, line 2299: nit: could add names to call args

```kotlin
validateTopLevelNodes(node = node, level = 0, topLevelTokens = 0)
```

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2244,25 +2253,39 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
+            // Note that for DML operations, top level parse node may be of type 'FROM'. Hence the check level > 1
+            if (node.type.isDml) {
+                if (level > 1) {
+                    node.throwTopLevelParserError()
+                }
+            } else {
+                node.throwTopLevelParserError()
             }
         }
+        node.children.map { validateTopLevelNodes(it, level + 1, topTokens) }
```
alancai98, line 2283: nit: could add names to call args:

```kotlin
node.children.map { validateTopLevelNodes(node = it, level = level + 1, topLevelTokens = topTokens) }
```

lang/test/org/partiql/lang/errors/ParserErrorsTest.kt
```diff
@@ -1246,6 +1246,116 @@ class ParserErrorsTest : TestBase() {
...
+        mapOf(
+            Property.LINE_NUMBER to 1L,
+            Property.COLUMN_NUMBER to 21L,
+            Property.TOKEN_TYPE to TokenType.KEYWORD,
+            Property.TOKEN_VALUE to ion.newSymbol("create")))
+
+    @Test
+    fun nestedRemove() = checkInputThrowingParserException(
+        "REMOVE REMOVE y",
+        ErrorCode.PARSE_INVALID_PATH_COMPONENT,
```
alancai98, line 1332: Is there a reason this error is not `PARSE_UNEXPECTED_TERM`?

alancai98 (commented):
Just have a few minor nits and questions. The rest lgtm

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2244,25 +2253,39 @@ class SqlParser(private val ion: IonSystem) : Parser {
         return ParseNode(ARG_LIST, null, items, rem)
     }
 
+    private fun ParseNode.throwTopLevelParserError(): Nothing =
+        token?.err("Keyword $token only expected at the top level in the query", PARSE_UNEXPECTED_TERM)
```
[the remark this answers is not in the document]
abhikuhikar: Makes sense. Thanks. Changed it.

lang/test/org/partiql/lang/errors/ParserErrorsTest.kt
abhikuhikar: This is handled actively by the parser while parsing the `REMOVE` keyword.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2244,25 +2253,39 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
-     *
-     * Currently only checks for `EXEC`. DDL and DML along with corresponding error codes to be added in the future
-     * (https://github.com/partiql/partiql-lang-kotlin/issues/354).
+     * Validates tree to make sure that the top level tokens are not found below the top level
      */
-    private fun List<Token>.checkUnexpectedTopLevelToken() {
-        for ((index, token) in this.withIndex()) {
-            if (token.keywordText == "exec" && index != 0) {
-                token.err("EXEC call found at unexpected location", PARSE_EXEC_AT_UNEXPECTED_LOCATION)
+    private fun validateTopLevelNodes(node: ParseNode, level: Int, topLevelTokens: Int) {
```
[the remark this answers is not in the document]
abhikuhikar: Doesn't make a significant difference but you are right. This can be done with `Boolean`.

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: Added.

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: Added.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2254,19 +2254,19 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
-        token?.err("Keyword $token only expected at the top level in the query", PARSE_UNEXPECTED_TERM)
-        ?: throw ParserException("Keyword $token only expected at the top level in the query", PARSE_UNEXPECTED_TERM, PropertyValueMap())
+        token?.err("Keyword ${token.text} only expected at the top level in the query", PARSE_UNEXPECTED_TERM)
+        ?: throw ParserException("Keyword ${token?.text} only expected at the top level in the query", PARSE_UNEXPECTED_TERM, PropertyValueMap())
 
     /**
-     * Validates tree to make sure that the top level tokens are not found below the top level
+     * Validates tree to make sure that the top level tokens are not found below the top level.
+     * Top level tokens are the tokens or keywords which are valid to be used only at the top level in the query.
+     * i.e. these tokens cannot be used with a mix of other commands. Hence if more than one top level tokens are found
```
alancai98, line 2263: nit: "i.e. these tokens cannot be used with a mix of other ~commands~ top level tokens"? "commands" seems too generalized.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2254,19 +2254,19 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
+     * in the query then it is invalid.
+     * [level] is the current traversal level in the parse tree.
+     * If [topLevelTokenSeen] is true, it means it has been encountered at least once before while traversing the parse tree.
```
alancai98, line 2266: nit: description of `topLevelTokenSeen` could be more more specific. Consider "...it means a top level token has been previously encountered while traversing the parse tree"

alancai98 (approved):
Couple very minor nits for the kdoc of `validateTopLevelNodes`. Rest lgtm

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: I believe the variable name itself tells it all.

PR #377, opened 2021-03-11
Adds support for parsing date literals

abhikuhikar:
*Description of changes:*
Adds support for parsing parsing date literals. Syntax for the date literal is `DATE 'YYYY-MM-DD'`.

### DIFF
--- a/lang/resources/org/partiql/type-domains/partiql.ion
+++ b/lang/resources/org/partiql/type-domains/partiql.ion
@@ -95,7 +95,7 @@
             (sexp values::(* expr 0))
 
             // Constructors for DateTime types
-            (date date_string::symbol)
+            (date year::int month::int day::int)
 
             // Set operators
             (union setq::set_quantifier operands::(* expr 2))
--- a/lang/resources/org/partiql/type-domains/partiql.ion
+++ b/lang/resources/org/partiql/type-domains/partiql.ion
@@ -94,6 +94,9 @@
             (list values::(* expr 0))
             (sexp values::(* expr 0))
 
+            // Constructors for DateTime types
+            (date date_string::symbol)
+
             // Set operators
             (union setq::set_quantifier operands::(* expr 2))
             (except setq::set_quantifier operands::(* expr 2))
@@ -353,6 +356,7 @@
             (symbol_type)
             (blob_type)
             (clob_type)
+            (date_type)
             (struct_type)
             (tuple_type)
             (list_type)
--- a/lang/src/org/partiql/lang/ast/ast.kt
+++ b/lang/src/org/partiql/lang/ast/ast.kt
@@ -969,10 +969,21 @@ enum class OrderingSpec {
 
 /**
  * The sealed class includes all the datetime types such as DATE, TIME, TIMESTAMP
+ * Note that the ast nodes corresponding to the DATE, TIME and TIMESTAMP here are different from the [Literal] nodes.
+ * You can create an Ion literal as [Timestamp] which will correspond to the [Literal] node and will have the type
+ * [SqlDataType.TIMESTAMP]. However that will be different from the
+ * `TIMESTAMP` here.
  * Note: TIME and TIMESTAMP are yet to be added.
  */
 sealed class DateTimeType : ExprNode() {
-    data class Date( val dateString: String, override val metas: MetaContainer ) : DateTimeType() {
+    /**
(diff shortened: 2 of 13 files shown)


lang/src/org/partiql/lang/ast/ast.kt
```diff
@@ -964,6 +967,16 @@ enum class OrderingSpec {
...
 
+/**
+ * The sealed class includes all the datetime types such as DATE, TIME, TIMESTAMP
+ * Note: TIME and TIMESTAMP are yet to be added.
+ */
+sealed class DateTimeType : ExprNode() {
+    data class Date( val dateString: String, override val metas: MetaContainer ) : DateTimeType() {
+        override val children: List<AstNode> = listOf()
+    }
+}
```
almann, line 978: How should one think about this node relative to an Ion literal node that happens to be a timestamp?  I think we should at least have a comment about that.  Is the expectation that we normalize one form, or does the code have to deal with both?

almann: If this is modeled as an algebraic type, we should also consider making the AST form not have a `String` as the representation.  More specifically, why not model the `DATE` fields explicitly and require the parser to fill those in?

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2220,6 +2228,23 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
+    private fun List<Token>.parseDate(): ParseNode =
+        when (head?.value?.isText) {
+            true -> {
+                // validate that the date string follows the format YYYY-MM-DD
+                try {
+                    LocalDate.parse(head?.value?.stringValue())
+                } catch (e: DateTimeParseException) {
+                    err(e.localizedMessage, PARSE_INVALID_DATE_STRING)
+                }
+                ParseNode(DATE, head, listOf(), tail)
```
almann, line 2243: If we did the needful to parse the date, we definitely should put those fields into the node that we generate.

almann (changes requested):
Good start--I really think we should encode the structure of the `DATE` fields into the AST otherwise downstream code needs to reparse it

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -94,6 +94,9 @@
             (list values::(* expr 0))
             (sexp values::(* expr 0))
 
+            // Constructors for DateTime types
+            (date date_string::symbol)
```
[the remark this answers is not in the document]
abhikuhikar: I think it would be better to be consistent with the AST node and include date components here as well.

lang/src/org/partiql/lang/ast/ast.kt
abhikuhikar: This is different from the `Literal` AST node which corresponds to the Ion literals including `timestamp`. I will be introducing an AST node `TIMESTAMP` similar to `DATE` and `TIME`. The code will have to deal with both I guess. One would correspond to the Ion literal with the type `Timestamp` and one would be explicit `TIMESTAMP` ast node. 
I will add comments to mention this explicit difference.

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: I have modeled it to include the date components.

lang/src/org/partiql/lang/ast/ast.kt
almann: I understand that we aren't adding `TIMESTAMP` yet, but you allude to it, and I would like to understand how we explicitly want the code to operate.  It might be worth mentioning that this is orthogonal to the Ion literal nodes.

almann (approved):
:shipit:

lang/src/org/partiql/lang/ast/ast.kt
abhikuhikar: The new `TIMESTAMP` node would be orthogonal to the `Ion literal` node. I have added a comment to mention this explicitly. 
The `TIMESTAMP` node will be instantiated for the timestamp literal `TIMESTAMP '2012-12-02 05:30:00'`, for example.
\`2012-12-02T05:30:00Z\` will correspond to the ion literal with the type `SqlDataType.TIMESTAMP`.
Does that answer the question?

PR #383, opened 2021-04-02
Migrates StatementRedactor to PIG-generated visitor (#375)

### LINKED ISSUE #375: Migrate StatementRedactor to PIG generated visitor
`StatementRedactor` was introduced in https://github.com/partiql/partiql-lang-kotlin/pull/374 and it inherits from `AstRewriterBase`.

We should consider refactoring it to inherit from `PartiqlAst.Visitor`.  (Or `PartiqlAst.VisitorTransform` if that is really needed but I doubt it.)

alancai98:
Closes #375.

Migrates `StatementRedactor` to PIG-generated visitor. Additionally does some initial cleanup and adds tests for `Concat` and `Parameter`.

### DIFF
--- a/lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
+++ b/lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
@@ -24,7 +24,7 @@ private val ion = IonSystemBuilder.standard().build()
 private val parser = SqlParser(ion)
 
 const val INVALID_NUM_ARGS = "Invalid number of args in node"
-const val INPUT_AST_STATEMENT_MISMATCH = "Input ast should be the parsed result of the input statement"
+const val INPUT_AST_STATEMENT_MISMATCH = "Unable to redact the statement. Please check that the input ast is the parsed result of the input statement"
 
 /**
  * Returns true if the given [node] type is to be skipped for redaction or its text is one of the [safeFieldNames].
@@ -108,31 +108,6 @@ private class StatementRedactionVisitor(
     private val sourceLocationMetaForRedaction = arrayListOf<SourceLocationMeta>()
     private val maskPattern = "***(Redacted)"
 
-    override fun visitExprSelect(node: PartiqlAst.Expr.Select) {
-        if (node.where != null) {
-            redactExpr(node.where)
-        }
-    }
-
-    override fun visitStatementDml(node: PartiqlAst.Statement.Dml) {
-        if (node.where != null) {
-            redactExpr(node.where)
-        }
-    }
-
-    override fun visitAssignment(node: PartiqlAst.Assignment) {
-        if (!skipRedaction(node.target, safeFieldNames)) {
-            redactExpr(node.value)
-        }
-    }
-
-    override fun visitDmlOpInsertValue(node: PartiqlAst.DmlOp.InsertValue) {
-        when (node.value) {
-            is PartiqlAst.Expr.Struct -> redactStructInInsertValueOp(node.value)
-            else -> redactExpr(node.value)
-        }
-    }
-
     /**
      * Returns the redacted [statement].
      */
@@ -160,10 +135,29 @@ private class StatementRedactionVisitor(
             redactedStatement.replace(start, start + length, maskPattern)
             offset = offset + maskPattern.length - length
         }
-
         return redactedStatement.toString()
     }
 
+    override fun visitExprSelect(node: PartiqlAst.Expr.Select) {
+        node.where?.let { redactExpr(it) }
+    }
+
+    override fun visitStatementDml(node: PartiqlAst.Statement.Dml) {
+        node.where?.let { redactExpr(it) }
+    }
+
+    override fun visitAssignment(node: PartiqlAst.Assignment) {
+        if (!skipRedaction(node.target, safeFieldNames)) {
+            redactExpr(node.value)
+        }
+    }
+
+    override fun visitDmlOpInsertValue(node: PartiqlAst.DmlOp.InsertValue) {
+        when (node.value) {
+            is PartiqlAst.Expr.Struct -> redactStructInInsertValueOp(node.value)
+            else -> redactExpr(node.value)
+        }
+    }
 
     private fun redactExpr(node: PartiqlAst.Expr) {
         if (node.isNAry()) {
@@ -272,7 +266,6 @@ private class StatementRedactionVisitor(
             null -> node.args.map { redactExpr(it) }
             else -> {
                 redactionLambda(node.args).map { redactExpr(it) }
-
             }
         }
     }
--- a/lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
+++ b/lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
@@ -34,13 +34,7 @@ fun skipRedaction(node: PartiqlAst.Expr, safeFieldNames: Set<String>): Boolean {
         return false
     }
 
-    // Skip redaction for NAry node
-    if (node.isNAry()) {
-        return true
-    }
-
     return when (node) {
-        is PartiqlAst.Expr.Parameter -> true    // Skip redaction for Parameter node
         is PartiqlAst.Expr.Id -> safeFieldNames.contains(node.name.text)
         is PartiqlAst.Expr.Lit -> {
             when (node.value) {
@@ -56,9 +50,7 @@ fun skipRedaction(node: PartiqlAst.Expr, safeFieldNames: Set<String>): Boolean {
                 }
             }
         }
-        else -> {
-            throw IllegalArgumentException("Unexpected PartiqlAst.Expr in StatementRedactor.hasSafeFieldName: ${node.javaClass}")
-        }
+        else -> true // Skip redaction for other nodes
     }
 }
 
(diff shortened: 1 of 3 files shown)


lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
```diff
@@ -82,238 +73,256 @@ fun redact(statement: String,
...
-        private val userDefinedFunctionRedactionConfig: Map<String, UserDefinedFunctionRedactionLambda>
-) : AstRewriterBase() {
-    private val root = node
+private class StatementRedactionVisitor(
+    private val statement: String,
+    private val safeFieldNames: Set<String>,
+    private val userDefinedFunctionRedactionConfig: Map<String, UserDefinedFunctionRedactionLambda>
+) : PartiqlAst.Visitor() {
     private val sourceLocationMetaForRedaction = arrayListOf<SourceLocationMeta>()
     private val maskPattern = "***(Redacted)"
```
dlurton, line 109: `const`?

lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
```diff
@@ -82,238 +73,256 @@ fun redact(statement: String,
...
-    fun rewriteExprNode(): ExprNode {
-        return super.rewriteExprNode(root)
-    }
-
-    override fun rewriteSelectWhere(node: ExprNode): ExprNode {
-        redactOnField(node)
-        return super.rewriteSelectWhere(node)
+    override fun visitExprSelect(node: PartiqlAst.Expr.Select) {
+        if (node.where != null) {
+            redactExpr(node.where)
```
dlurton, line 113: nit: `node.where?.let { redactExpr(it) }` is the same but even shorter.

Could apply this to all similar `if` statements below.

lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
```diff
@@ -82,238 +73,256 @@ fun redact(statement: String,
...
-        val totalCharactersInPreviousLines = IntArray(lines.size)
-        for (lineNum in 1 until lines.size) {
-            totalCharactersInPreviousLines[lineNum] = totalCharactersInPreviousLines[lineNum - 1] + lines[lineNum - 1].length + 1
+    private fun redactCall(node: PartiqlAst.Expr.Call) {
+        val funcName = node.funcName.text
+        when (val redactionLambda = userDefinedFunctionRedactionConfig[funcName]) {
+            null -> node.args.map { redactExpr(it) }
+            else -> {
+                redactionLambda(node.args).map { redactExpr(it) }
+
```
dlurton, line 275: Super-ultra-mega-nit: extra line.

lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
```diff
@@ -82,238 +73,256 @@ fun redact(statement: String,
...
-            is NAry -> redactOnNAry(node)
-            is Literal -> redactOnLiteral(node)
-            is Seq -> redactOnSeq(node)
-            is Struct -> redactOnStruct(node)
-            is Typed -> redactOnTyped(node)
-            else -> { /* intentionally blank */ }
+    /**
+     * Returns the redacted [statement].
+     */
+    fun getRedactedStatement(): String {
```
dlurton, line 139: nit: I'd make this the top of the class so readers unfamiliar with the code can more easily see all of types of the properties defined in the constructor.

lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
```diff
@@ -82,238 +73,256 @@ fun redact(statement: String,
...
-                when (arg) {
-                    is NAry -> redactOnNAry(arg)
-                    is Typed -> redactOnTyped(arg)
-                    else -> { /* intentionally blank */ }
-                }
+        sourceLocationMetaForRedaction.map {
+            val length = it.length.toInt()
+            val lineNum = it.lineNum.toInt()
+            if (lineNum < 1 || lineNum > totalCharactersInPreviousLines.size) {
+                throw IllegalArgumentException("$INPUT_AST_STATEMENT_MISMATCH, line number: $lineNum")
```
dlurton, line 154: The error message "Input ast should be the parsed result of the input statement" doesn't clearly indicate what went wrong here...  Even reading the code I'm not 100% why this is an error condition and what I should do to solve it.

dlurton (changes requested):
Mostly nits, but I do think that error message I mentioned could be improved.  Thanks.

lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
alancai98: `const` is only allowed at top-level or in objects. Would it be better to include the `maskPattern` in a companion object?

lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
alancai98: Thanks. Changed

lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
alancai98: Removed

lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
alancai98: Good idea. Moved to the top of the class

lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
alancai98: To give some background, there was a `redact` [function call](https://github.com/partiql/partiql-lang-kotlin/blob/migrate-statement-redactor-375/lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt#L88) that allows users to pass in the `statement: String` and the `ast: ExprNode` rather than just the `statement`. I think the idea here was if you already had the parsed `ExprNode` from `statement`, you could perform the redaction logic without having to re-parse. 

This check was here if the `statement` and `ast` were mismatched (i.e. `ast` is not the result of parsing `statement`). I slightly changed the wording in the latest commit. Let me know if this should be changed further

lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
dlurton: I'd rather not use a `companion object` if we could avoid it.  `private val` at the top level in the file is sufficient.

lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
```diff
@@ -82,238 +73,249 @@ fun redact(statement: String,
...
-    private fun redactOnCall(args: List<ExprNode>) {
-        if (args.isEmpty()) {
-            return
-        }
-        if (args[0] !is VariableReference) {
-            return
-        }
+    private fun redactLiteral(literal: PartiqlAst.Expr.Lit) {
+        val sourceLocation = literal.metas[SourceLocationMeta.TAG] as SourceLocationMeta?
+            ?: throw NoSuchElementException("No SourceLocation meta data in Literal object $literal")
```
dlurton, line 244: There's an extension property `sourceLocation` which should be used instead, also that doesn't seem like the right exception because [`NoSuchElementException`](https://docs.oracle.com/javase/8/docs/api/java/util/NoSuchElementException.html) is commonly thrown by `Iterator.next()` when there are no remaining elements.  I think a simple call to `error` should suffice here, and the other location I see pattern.

```
val sourceLocation = literal.metas.sourceLocation ?: error("Cannot redact due to missing source location")

File: lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt [hunk 7]
```diff
@@ -82,238 +73,249 @@ fun redact(statement: String,
...
-    private fun redactOnStructInInsertValueOp(struct: Struct) {
-        struct.fields.map {
-            if (it.name is Literal) {
-                if (!validateSafeFieldNames(it.name)) {
-                    redactOnField(it.expr)
-                }
-            } else { /* intentionally blank */ }
+    private fun redactTypes(typed: PartiqlAst.Expr.IsType) {
+        if (typed.value is PartiqlAst.Expr.Id && !skipRedaction(typed.value, safeFieldNames)) {
+            val sourceLocation = typed.type.metas[SourceLocationMeta.TAG] as SourceLocationMeta? ?: throw NoSuchElementException("No SourceLocation meta data in DataType object $typed")
```
dlurton, line 275: As mentioned elsewhere:
```
val sourceLocation = typed.type.metas.sourceLocation ?: error("Cannot redact due to missing source location")
```

dlurton (changes requested):

dlurton:
Requested changes are pretty small--let me know when you've pushed them and I'll approve quickly.  Thanks.

lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
alancai98: Since I moved `maskPattern` to the top-level, I could make it `const`.

lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
alancai98: I agree, the error should be changed. The extension property `sourceLocation` was only for the `partiql.lang.ast.MetaContainer`, so I also added an equivalent one for IonElement MetaContainer in `SourceLocationMeta.kt`.

lang/src/org/partiql/lang/ast/passes/StatementRedactor.kt
alancai98: Thanks. Changed to your suggestion.

dlurton (approved):

PR #384, opened 2021-04-06
Added Parser support for TIME literals

abhikuhikar:
*Issue #, if available:*

*Description of changes:*
* Extends `SqlParser` to support `TIME` literals.
* Supports parsing optional `precision` and optional TIME ZONE specifier `WITH TIME ZONE` for `TIME` literals.
* Adds tests covering positive and negative cases for `TIME` literals.
*  Uses `LocalTime` and `OffsetTime` java classes for parsing and validating `TIME` strings.

### DIFF
--- a/lang/resources/org/partiql/type-domains/partiql.ion
+++ b/lang/resources/org/partiql/type-domains/partiql.ion
@@ -96,6 +96,7 @@
 
             // Constructors for DateTime types
             (date year::int month::int day::int)
+            (time hour::int minute::int second::int nano::int precision::int tz_minutes::(? int))
 
             // Set operators
             (union setq::set_quantifier operands::(* expr 2))
@@ -357,6 +358,7 @@
             (blob_type)
             (clob_type)
             (date_type)
+            (time_type)
             (struct_type)
             (tuple_type)
             (list_type)
--- a/lang/resources/org/partiql/type-domains/partiql.ion
+++ b/lang/resources/org/partiql/type-domains/partiql.ion
@@ -96,7 +96,7 @@
 
             // Constructors for DateTime types
             (date year::int month::int day::int)
-            (time hour::int minute::int second::int nano::int precision::int tz_minutes::(? int))
+            (lit_time value::time_value)
 
             // Set operators
             (union setq::set_quantifier operands::(* expr 2))
@@ -123,6 +123,9 @@
                 (order (? order_by))
                 (limit (? expr))))
         // end of sum expr
+
+        // Time
+        (product time_value hour::int minute::int second::int nano::int precision::int tz_minutes::(? int))
 
         // A "step" within a path expression; that is the components of the expression following the root.
         (sum path_step
--- a/lang/src/org/partiql/lang/ast/ast.kt
+++ b/lang/src/org/partiql/lang/ast/ast.kt
@@ -109,6 +109,9 @@ sealed class ExprNode : AstNode(), HasMetas {
             is DateTimeType.Date -> {
                 copy(metas = metas)
             }
+            is DateTimeType.Time -> {
+                copy(metas = metas)
+            }
         }
     }
 }
@@ -986,6 +989,29 @@ sealed class DateTimeType : ExprNode() {
         override val metas: MetaContainer ) : DateTimeType() {
         override val children: List<AstNode> = listOf()
     }
+
+    /**
+     * AST node representing the TIME literal.
+     *
+     * @param hour represents the hour value.
+     * @param minute represents the minute value.
+     * @param second represents the second value.
+     * @param nano represents the fractional part of the second upto the nano seconds' precision.
+     * @param precision is an optional parameter which, if specified, represents the precision of the fractional second.
+     * The default precision is 9 or nanosecond.
+     * @param tz_minutes is the optional time zone in minutes which can be specified with "WITH TIME ZONE".
+     * If [tz_minutes] is null, that means the time zone is undefined.
+     */
+    data class Time(
+        val hour: Int,
+        val minute: Int,
+        val second: Int,
+        val nano: Int,
+        val precision: Int = 9,
+        val tz_minutes: Int? = null,
+        override val metas: MetaContainer ) : DateTimeType() {
+        override val children: List<AstNode> = listOf()
+    }
 }
 
 /**
@@ -1025,6 +1051,7 @@ enum class SqlDataType(val typeName: String, val arityRange: IntRange) {
     LIST("list", 0..0), // Ion
     SEXP("sexp", 0..0), // Ion
     DATE("date", 0..0), // SQL-92
+    TIME("time", 0..0), // SQL-92
     BAG("bag", 0..0);  // PartiQL
 
     companion object {
--- a/lang/src/org/partiql/lang/ast/ast.kt
+++ b/lang/src/org/partiql/lang/ast/ast.kt
@@ -986,7 +986,8 @@ sealed class DateTimeType : ExprNode() {
         val year: Int,
         val month: Int,
         val day: Int,
-        override val metas: MetaContainer ) : DateTimeType() {
+        override val metas: MetaContainer
+    ) : DateTimeType() {
         override val children: List<AstNode> = listOf()
     }
 
@@ -996,7 +997,7 @@ sealed class DateTimeType : ExprNode() {
      * @param hour represents the hour value.
      * @param minute represents the minute value.
      * @param second represents the second value.
-     * @param nano represents the fractional part of the second upto the nano seconds' precision.
+     * @param nano represents the fractional part of the second up to the nanoseconds' precision.
      * @param precision is an optional parameter which, if specified, represents the precision of the fractional second.
      * The default precision is 9 or nanosecond.
      * @param tz_minutes is the optional time zone in minutes which can be specified with "WITH TIME ZONE".
@@ -1009,7 +1010,8 @@ sealed class DateTimeType : ExprNode() {
         val nano: Int,
         val precision: Int = 9,
         val tz_minutes: Int? = null,
-        override val metas: MetaContainer ) : DateTimeType() {
+        override val metas: MetaContainer
+    ) : DateTimeType() {
         override val children: List<AstNode> = listOf()
     }
 }
--- a/lang/src/org/partiql/lang/syntax/SqlParser.kt
+++ b/lang/src/org/partiql/lang/syntax/SqlParser.kt
@@ -28,8 +28,13 @@ import org.partiql.lang.syntax.SqlParser.ParseType.*
 import org.partiql.lang.syntax.TokenType.*
 import org.partiql.lang.syntax.TokenType.KEYWORD
 import org.partiql.lang.util.*
-import java.time.LocalDate
+import java.lang.Integer.min
+import java.time.*
+import java.time.format.DateTimeFormatter
 import java.time.format.DateTimeParseException
+
+// TODO: (Ques) LocalTime has a field nanosecond and is the precision limit for it. Do we want to use this as the default precision if not specified?
+private const val DEFAULT_PRECISION_FOR_TIME = 9
 
 /**
  * Parses a list of tokens as infix query expression into a prefix s-expression
@@ -87,6 +92,8 @@ class SqlParser(private val ion: IonSystem) : Parser {
         UNPIVOT,
         CALL,
         DATE,
+        TIME,
+        TIME_WITH_TIME_ZONE,
         CALL_AGG,
         CALL_DISTINCT_AGG,
         CALL_AGG_WILDCARD,
@@ -132,7 +139,8 @@ class SqlParser(private val ion: IonSystem) : Parser {
         DROP_INDEX(isTopLevelType = true),
         CREATE_INDEX(isTopLevelType = true),
         PARAMETER,
-        EXEC(isTopLevelType = true);
+        EXEC(isTopLevelType = true),
+        PRECISION;
 
         val identifier = name.toLowerCase()
     }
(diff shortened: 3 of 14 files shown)


lang/test/org/partiql/lang/syntax/SqlParserDateTimeTests.kt
```diff
@@ -25,6 +34,567 @@ class SqlParserDateTimeTests : SqlParserTestBase() {
...
+        ),
+        createErrorCaseForTime(
+            source = "TIME '23:59:59+05:30.00'",
+            line = 1L,
+            col = 6L,
+            errorCode = ErrorCode.PARSE_INVALID_TIME_STRING,
+            tokenType = TokenType.LITERAL,
+            tokenValue = ion.newString("23:59:59+05:30.00")
+        ),
+        // TODO: Investing why the build failed in GH actions for these two tests.
```
abhikuhikar, line 319: typo: investigate* 
This is mostly because `LocalTime.parse` behaves differently for JDK11.

lang/test/org/partiql/lang/syntax/SqlParserDateTimeTests.kt
```diff
@@ -25,6 +34,567 @@ class SqlParserDateTimeTests : SqlParserTestBase() {
...
+        },
+        DateTimeTestCase("TIME (9) '23:59:59.123456789'") {
+            time(23, 59, 59, 123456789, 9, null)
+        },
+        DateTimeTestCase("TIME (0) '23:59:59.123456789'") {
+            time(23, 59, 59, 123456789, 0, null)
+        },
+        DateTimeTestCase("TIME (10) '23:59:59.123456789'") {
+            time(23, 59, 59, 123456789, 9, null)
+        },
```
dlurton, line 58: Since the precision is beyond the maximum value of 9, this should be an error case, no?

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -688,6 +696,18 @@ class SqlParser(private val ion: IonSystem) : Parser {
                 val (year, month, day) = dateString.split("-")
                 DateTimeType.Date(year.toInt(), month.toInt(), day.toInt(), metas)
             }
+            TIME -> {
+                val timeString = token!!.text!!
+                val precision = min(children[0].token?.value?.numberValue()?.toInt() ?: DEFAULT_PRECISION_FOR_TIME, DEFAULT_PRECISION_FOR_TIME)
```
dlurton, line 701: Use of `min` here seems like a bad idea.  Why shouldn't this be an error if the user specifies a precision that is greater than what is supported?  (Postgres does this for precisions beyond 6, BTW.)

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2246,6 +2267,135 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
+     * Otherwise, the function returns the [ParseNode] with the token as null.
+     * It also verifies that the specified precision is an unsigned integer.
+     */
+    private fun List<Token>.parseOptionalPrecision(): ParseNode =
+        // If the optional precision is present
+        if (head?.type == LEFT_PAREN) {
+            var rem = tail
+            // Expected precision token to be unsigned integer
+            if (rem.head == null || rem.head!!.type != LITERAL || !rem.head!!.value!!.isUnsignedInteger) {
+                rem.head.err("Expected integer value for precision", PARSE_INVALID_PRECISION_FOR_TIME)
```
dlurton, line 2283: "Expected integer value for precision" -> "Expected integer value between &lt;min> to &lt;max> for precision"?

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -28,9 +28,14 @@ import org.partiql.lang.syntax.SqlParser.ParseType.*
...
 import org.partiql.lang.syntax.TokenType.KEYWORD
 import org.partiql.lang.util.*
-import java.time.LocalDate
+import java.lang.Integer.min
+import java.time.*
+import java.time.format.DateTimeFormatter
 import java.time.format.DateTimeParseException
 
+// TODO: (Ques) LocalTime has a field nanosecond and is the precision limit for it. Do we want to use this as the default precision if not specified?
+private const val DEFAULT_PRECISION_FOR_TIME = 9
```
dlurton, line 37: This is a good question and I am uncertain I know the correct answer.  May need to bring this up in the team channel.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2246,6 +2267,135 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
+                tailExpectedKeyword("with").
+                tailExpectedKeyword("time").
+                tailExpectedKeyword("zone")
+            return Pair(rem, true)
+        }
+        return Pair(this, false)
+    }
+
+    /**
+     * Parses a time string and verifies that the time string is a string ond is specified in the valid ISO format.
```
dlurton, line 2312: - "ond" -> "and"?
- What ISO number?

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2246,6 +2267,135 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
+     * This behaviour is consistent with most other database languages like SQL92, PostgreSQL etc.
+     * It also makes more sense as the user using the keywords "TIME WITH TIME ZONE .." would expect the
+     * local time zone offset to be used by default.
+     */
+    private fun List<Token>.parseTime(): ParseNode {
+
+        // These are used to validate the generic format of the time string.
+        // The more involved logic such as validating the time is done by LocalTime.parse or OffsetTime.parse
+        val timeWithoutTimeZoneRegex = Regex("\\d\\d:\\d\\d:\\d\\d(\\.\\d*)?")
+        val genericTimeRegex = Regex("\\d\\d:\\d\\d:\\d\\d(\\.\\d*)?([+|-]\\d\\d:\\d\\d)?")
```
dlurton, line 2336: I think you can move these to private top-level `private` variables since `Regex` implementations are thread-safe.

lang/test/org/partiql/lang/syntax/SqlParserDateTimeTests.kt
```diff
@@ -25,6 +34,567 @@ class SqlParserDateTimeTests : SqlParserTestBase() {
...
+        DateTimeTestCase("TIME (9) '23:59:59.123456789-14:50'") {
+            time(23, 59, 59, 123456789, 9, null)
+        },
+        DateTimeTestCase("TIME (0) '23:59:59.123456789-18:00'") {
+            time(23, 59, 59, 123456789, 0, null)
+        },
+        DateTimeTestCase("TIME (10) '23:59:59.123456789+18:00'") {
+            time(23, 59, 59, 123456789, 9, null)
+        },
```
dlurton, line 97: Ack, the precision specified by the user is substituted with the maximum precision supported?  Should be an error.

lang/test/org/partiql/lang/syntax/SqlParserDateTimeTests.kt
```diff
@@ -25,6 +34,567 @@ class SqlParserDateTimeTests : SqlParserTestBase() {
...
+            time(23, 59, 59, 123456789, 7, 60)
+        },
+        DateTimeTestCase("TIME (9) WITH TIME ZONE '23:59:59.123456789-14:50'") {
+            time(23, 59, 59, 123456789, 9, -890)
+        },
+        DateTimeTestCase("TIME (0) WITH TIME ZONE '23:59:59.123456789-18:00'") {
+            time(23, 59, 59, 123456789, 0, -1080)
+        },
+        DateTimeTestCase("TIME (10) WITH TIME ZONE '23:59:59.123456789+18:00'") {
+            time(23, 59, 59, 123456789, 9, 1080)
```
dlurton, line 147: here too...

lang/src/org/partiql/lang/syntax/SqlParser.kt
dlurton: I think also we're going to need to truncate the precision according to the value specified in the precision field--this is what postgres does, i.e.:

```
dlurton=# select time (2) '12:30:23.123456';
    time     
-------------
 12:30:23.12
(1 row)

dlurton=# 

```

dlurton (changes requested):
My main concerns are surrounding precision.  In `TIME (<prec>) '<time>'`:

- If the user specifies a `<prec>` beyond the maximum, the maximum is used instead.  This should be an error.  (postgres issues an error when specifying precision values above 6).
- If the user specifies a precision less than the what's contained within `<time>`, the value is fractional component should be truncated (or possibly rounded?) to that precision. Again, this is what postgres does.

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -96,6 +96,7 @@
 
             // Constructors for DateTime types
             (date year::int month::int day::int)
+            (time hour::int minute::int second::int nano::int precision::int tz_minutes::(? int))
```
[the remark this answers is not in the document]
abhikuhikar: I see. Yeah that would be better. Changed the partiql pig domain.

lang/src/org/partiql/lang/ast/ast.kt
```diff
@@ -986,6 +989,29 @@ sealed class DateTimeType : ExprNode() {
...
+     * If [tz_minutes] is null, that means the time zone is undefined.
+     */
+    data class Time(
+        val hour: Int,
+        val minute: Int,
+        val second: Int,
+        val nano: Int,
+        val precision: Int = 9,
+        val tz_minutes: Int? = null,
+        override val metas: MetaContainer ) : DateTimeType() {
```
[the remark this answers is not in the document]
abhikuhikar: Done.

lang/src/org/partiql/lang/ast/ast.kt
```diff
@@ -986,6 +989,29 @@ sealed class DateTimeType : ExprNode() {
...
         override val children: List<AstNode> = listOf()
     }
+
+    /**
+     * AST node representing the TIME literal.
+     *
+     * @param hour represents the hour value.
+     * @param minute represents the minute value.
+     * @param second represents the second value.
+     * @param nano represents the fractional part of the second upto the nano seconds' precision.
```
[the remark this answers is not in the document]
abhikuhikar: Done.

lang/src/org/partiql/lang/ast/ast.kt
```diff
@@ -986,6 +989,29 @@ sealed class DateTimeType : ExprNode() {
...
+     * @param precision is an optional parameter which, if specified, represents the precision of the fractional second.
+     * The default precision is 9 or nanosecond.
```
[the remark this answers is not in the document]
abhikuhikar: Not sure what `enum` constants `enum class TimePrecision` would contain. `Time` in `java.Time` stores the `nanosecond` value in the field `nano` which happens to be `Int`. So we can create a  `double` second's value at the evaluation or run time taking into account the precision value. It is possible to do it at the `parse` time but I chose to do it at the evaluation time.

lang/test/org/partiql/lang/syntax/SqlParserDateTimeTests.kt
abhikuhikar: Not really. I checked with the Postgres and it does allow specifying the precision beyond the limit. In that case the value is rounded to the default maximum precision. (9 in this case).

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: So I checked with the `4.21` version of `pgAdmin 4` and it allows specifying the precision beyond the maximum. 
For example, this query succeeds. 
`SELECT TIME (19) '12:23:30.1234567898'` -> `12:23:30.123457`

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: Done. Thanks.

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: We will round the value (as Postgres does) taking precision into account but that has been delegated to the evaluation phase. I think we can keep the default precision as 9 (nanoseconds precision) as the `java.time` also has a precision of nano seconds.

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: Changed. ISO 8601

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: Done.

lang/test/org/partiql/lang/syntax/SqlParserDateTimeTests.kt
abhikuhikar: The Postgres I verified with uses the default precision as the maximum precision. It also uses the maximum precision if the specified precision is more than the maximum.

lang/test/org/partiql/lang/syntax/SqlParserDateTimeTests.kt
abhikuhikar: Same comment.

lang/test/org/partiql/lang/syntax/SqlParserDateTimeTests.kt
dlurton: Ok, that's odd. I guess it just issues a warning.

lang/src/org/partiql/lang/ast/ast.kt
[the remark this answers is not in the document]
dlurton: Yeah, on second thought `Int` is better.  Thanks.

lang/src/org/partiql/lang/syntax/SqlParser.kt
dlurton: I'm uncertain what "keep the default precision as 9" means here, since my comment above only applies to `time` literals that includes a precision, thus no default value for precision is in play.

lang/test/org/partiql/lang/syntax/SqlParserDateTimeTests.kt
dlurton: yes, yes, not sure where I got the idea that it was an error from...

dlurton (approved):

Issue #382, opened 2021-03-30
Values of DATE date literals should not be the *extended* ISO-8601 date


dlurton:
[This issue assumes that the values for the year component of `DATE` values should have the same constraints as that of an Ion timestamp.]

In essence:  the Ion timestamp only allows year values between 1 and 9999 while `LocalDate` accepts years in the range of -999999999 thru +999999999, if the year is prefixed with a `+` or a `-` sign.

This is because the `LocalDate.parse` method accepts the *extended* ISO-8601 date which allows for more than 4 digit years if the year is preceded by a `+` or `-` sign.  I doubt we intended to support this, given that the Ion timestamp does not.

I recommend adding some functions for validating and parsing `DATE` values that can be used by `SqlParser` and anything else such as the `CAST` functionality and `ExprValueFactory` implementation.

The following should result in an error instead of a valid AST.

```
Welcome to the PartiQL REPL!
Using version: 0.2.7-SNAPSHOT-ce4f7a0
PartiQL> DATE '+999999-03-01'
   | !!
===' 

(
  query
  (
    date
    999999
    3
    1
  )
)
--- 
OK!
```

abhikuhikar:
Fixed by #385

PR #385, opened 2021-04-07
Fixes parser for extended dates.

abhikuhikar:
*Issue #382 :*

*Description of changes:*
- Fixes parser to filter out the extended dates. The only allowed dates are of the format `yyyy-MM-dd`
- Added tests for random `DATE` literals. 


lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2232,20 +2233,30 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
-            else -> err("Expected date string followed by the keyword DATE, found ${head?.value?.type}", PARSE_UNEXPECTED_TOKEN)
+    private fun List<Token>.parseDate(): ParseNode {
+        val dateStringToken = head
+        if (dateStringToken?.value == null || dateStringToken.type != LITERAL || !dateStringToken.value.isText) {
+            err("Expected date string followed by the keyword DATE, found ${head?.value?.type}", PARSE_UNEXPECTED_TOKEN)
         }
 
+        val dateString = dateStringToken.value.stringValue()
+
+        val datePatternRegex = Regex("\\d\\d\\d\\d-\\d\\d-\\d\\d")
```
dlurton, line 2244: Another one, you can move this definition to a `private val` at the top of the file to avoid re-creating `Regex` instances you don't need.

dlurton (changes requested):
One minor change--let me know when you've submitted the next change and I'll try to give a quick approval.

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: Done.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -38,6 +39,7 @@ import java.time.format.DateTimeParseException
 class SqlParser(private val ion: IonSystem) : Parser {
 
     private val trueValue: IonBool = ion.newBool(true)
+    private val datePatternRegex = Regex("\\d\\d\\d\\d-\\d\\d-\\d\\d")
```
dlurton, line 42: That's better I guess, but I really meant outside of the class too.  A singleton.

dlurton (dismissed):
Ok to merge unless you really want to make that `Regex` a singleton.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -32,14 +32,18 @@ import java.time.LocalDate
 import java.time.format.DateTimeFormatter.*
 import java.time.format.DateTimeParseException
 
+/**
+ * Regex pattern to match date strings of the format yyyy-MM-dd
+ */
+@JvmField val DATE_PATTERN_REGEX = Regex("\\d\\d\\d\\d-\\d\\d-\\d\\d")
```
dlurton, line 38: Not sure why this is `@JvmField`?  That's only needed if you intend this to be accessed by Java clients.  Additionally, this should be `private` since we don't want this to be part of the API.

dlurton (changes requested):

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: If this is to be kept `private`, isn't it better to keep it as a `private` member inside the class?

dlurton (dismissed):

lang/src/org/partiql/lang/syntax/SqlParser.kt
dlurton: I don't know why it would be better.  I suppose that one could make an argument that we should avoid polluting the top-level namespace, but I don't think that applies here since there aren't a lot of top-level definitions in this file, and `private` means its not accessible to code in other files.

dlurton (approved):

PR #391, opened 2021-04-20
Optimizes AST iterator

abhikuhikar:
*Issue #, if available:*

*Description of changes:*
Changed AST `iterator()` to use mutable list for optimizing performance. 


almann (approved):
The change is reasonable--how did you test it to make sure it does what you want (even if said perf testing was manual)?

abhikuhikar:
I observed the "wall-clock" for the multiple runs and the previous `iterator()` took about 119 secs on average and the new `iterator()` took about 1.5 secs on average.

PR #392, opened 2021-04-20
Evaluate time literals

abhikuhikar:
*Issue #, if available:*

*Description of changes:*
* Evaluates `TIME` literals.
* Certain assumptions were made which need attention:
      * Output for the time strings is always represented in the format `HH:mm:ss.nnnnnnnnn[xxx]`
      * Refer [DateTimeFormatter](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html) for more information.
      * `TIME '24:00:00'` is valid in `Postgresql` but invalid in `java.time` and hence in `PartiQL`.
      * For example, `TIME (2) '23:59:59.999'` returns `00:00:00.000000000`
      * The `ionValue` for `time` has the `second` field of type `float`

Revision 2:
- Added wrapper class to consolidate TIME/precision logic and preserve precision value of the `TIME`.
- Changed the type of `second` to `decimal` for `ionValue` for `TIME`.
- Added documentation
- [This commit](https://github.com/partiql/partiql-lang-kotlin/pull/392/commits/8210e7b988856f208221d5c2b19666051c986825) preserves the `precision` of the original `time` value instead of using the maximum default value (9) for the `precision` in case the precision is not specified. 
- `TIME '12:12:12.123'`, for example, will now have a precision of 3 instead of 9.

Revision 3: 
 - Changed inference of precision from time string instead of `nano` int. 
 - Removed the default `precision` as max `precision` for `TIME`.
 - Added checks for range of values in `Time.Of(..., ...)` method.


lang/src/org/partiql/lang/eval/Scalar.kt
```diff
@@ -50,6 +52,20 @@ interface Scalar {
...
+     * To get the time with timezone value, use [timeWithTimezoneValue] which returns an instance of [OffsetTime]
+     */
+    fun timeValue(): LocalTime? = null
+
+    /**
+     * Returns this value as a [OffsetTime] or `null` if not applicable.
+     * This operation is only applicable for [ExprValueType.TIME] with timezone offset.
+     */
+    fun timeWithTimezoneValue(): OffsetTime? = null
+
```
almann, line 68: Would it be more appropriate to return a variant type for both of these modalities?  Or if we have to capture precision, we return our time wrapper that has the capability to return these modes.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -710,13 +709,13 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
+                val precision = min(children[0].token?.value?.numberValue()?.toInt() ?: MAX_PRECISION_FOR_TIME, MAX_PRECISION_FOR_TIME)
                 val time = LocalTime.parse(timeString, DateTimeFormatter.ISO_TIME)
                 DateTimeType.Time(time.hour, time.minute, time.second, time.nano, precision, null, metas)
             }
             TIME_WITH_TIME_ZONE -> {
                 val timeString = token!!.text!!
-                val precision = min(children[0].token?.value?.numberValue()?.toInt() ?: DEFAULT_PRECISION_FOR_TIME, DEFAULT_PRECISION_FOR_TIME)
+                val precision = min(children[0].token?.value?.numberValue()?.toInt() ?: MAX_PRECISION_FOR_TIME, MAX_PRECISION_FOR_TIME)
                 val time = OffsetTime.parse(timeString)
                 DateTimeType.Time(time.hour, time.minute, time.second, time.nano, precision, time.offset.totalSeconds/60, metas)
```
almann, line 720: Some copy/paste that should be avoided--it might be worth capturing a time wrapper of some sort to push this logic outside of the parser or evaluator.

lang/src/org/partiql/lang/util/ExprValueFormatter.kt
```diff
@@ -49,6 +52,9 @@ class ConfigurableExprValueFormatter(private val config: Configuration) : ExprVa
 
                 DATE                                       -> out.append(value.scalar.dateValue().toString())
 
+                TIME                                       -> out.append("${ value.scalar.timeWithTimezoneValue()?.format(DateTimeFormatter.ofPattern("HH:mm:ss.nnnnnnnnnxxx")) 
+                                                                            ?: value.scalar.timeValue()?.format(DateTimeFormatter.ofPattern("HH:mm:ss.nnnnnnnnn")) }")
```
almann, line 56: This looks like something that should be in a utility function (or maybe that wrapper type I mentioned).

lang/src/org/partiql/lang/util/UtilityExtensions.kt
```diff
@@ -0,0 +1,25 @@
...
+package org.partiql.lang.util
+
+import java.time.ZoneOffset
+import kotlin.math.absoluteValue
+
+/**
+ * Returns the string representation of the [ZoneOffset] in HH:mm format.
+ */
+fun ZoneOffset.getOffsetHHmm(): String {
+    return "${if(totalSeconds >= 0) "+" else "-"}${(totalSeconds / 3600).absoluteValue.toString().padStart(2, '0')}:${((totalSeconds / 60) % 60).absoluteValue.toString().padStart(2, '0')}"
```
almann, line 24: Let's break this up into a few lines with variables and less magic numbers (particularly that `60`)

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
```diff
@@ -1,13 +1,42 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates.  All rights reserved.
```
almann, line 2: Remove the copyright date especially since it's not 2019.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
```diff
@@ -34,4 +63,187 @@ class EvaluatingCompilerDateTimeTests : EvaluatorTestBase() {
...
+
+    private fun createIonTimeStruct(value: Time): IonStruct =
+        ion.newEmptyStruct().apply {
+            add("hour", ion.newInt(value.hour))
+            add("minute", ion.newInt(value.minute))
+            add("second", ion.newFloat( value.second + value.nano * 10.0.pow(-9)))
+            add("timezone_hour", ion.newInt(value.tz_minutes?.div(60)))
+            add("timezone_minute", ion.newInt(value.tz_minutes?.rem(60)))
+            addTypeAnnotation("\$partiql_time")
+        }
```
almann, line 75: Is this literally mirroring the equivalent code in `ExprValue` somewhere?

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
```diff
@@ -34,4 +63,187 @@ class EvaluatingCompilerDateTimeTests : EvaluatorTestBase() {
...
+        override fun getParameters() = listOf(
+            case("TIME '00:00:00.000'", "00:00:00.000000000", Time(0, 0, 0)),
+            case("TIME '23:59:59.99999999'", "23:59:59.999999990", Time(23, 59, 59, 999999990)),
+            case("TIME (2) '23:59:59.99999999'", "00:00:00.000000000", Time(0, 0, 0)),
+            case("TIME '00:45:13.840800524'", "00:45:13.840800524", Time(0, 45, 13, 840800524)),
+            case("TIME '05:20:52.015779149'", "05:20:52.015779149", Time(5, 20, 52, 15779149)),
+            case("TIME '23:59:59'", "23:59:59.000000000", Time(23, 59, 59)),
+            case("TIME (12) '12:24:12.123'", "12:24:12.123000000", Time(12, 24, 12, 123000000)),
+            case("TIME (0) '12:59:59.9'", "13:00:00.000000000", Time(13, 0,0)),
+            case("TIME WITH TIME ZONE '00:00:00'", "00:00:00.000000000${LOCAL_TIMEZONE_OFFSET.getOffsetHHmm()}", Time(0,0,0,0,null, LOCAL_TZ_MINUTES)),
```
almann, line 110: This literally changes precision, and I don't think that is right (see earlier comments).

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
```diff
@@ -34,4 +63,187 @@ class EvaluatingCompilerDateTimeTests : EvaluatorTestBase() {
...
+
+    private val RANDOM_GENERATOR = generateRandomSeed()
+
+    private fun generateRandomSeed() : Random {
+        val seed = Random(1).nextInt()
+        println("Randomly generated seed is ${seed}. Use this to reproduce failures in dev environment.")
+        return Random(seed)
+    }
+
+    data class Time(val hour: Int, val minute: Int, val second: Int, val nano: Int = 0, val precision: Int? = null, val tz_minutes: Int? = null) {
```
almann, line 133: The fact that you have this implies that a wrapper type in the library might be appropriate, but I know I sound like a broken record.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
```diff
@@ -34,4 +63,187 @@ class EvaluatingCompilerDateTimeTests : EvaluatorTestBase() {
...
+        val timezoneMinutes = if (withTimezone) {
+            nextInt(-1080, 1081)
+        } else {
+            null
+        }
+        return Time(hour, minute, second, nano, precision, timezoneMinutes)
+    }
+
+    private val RANDOM_TIMES = List(RANDOM_TESTS_SIZE) {
+        RANDOM_GENERATOR.nextTime(
```
almann, line 185: Just to verify, this is deterministic (meaning each run of the tests are repeatable), yes?

lang/test/org/partiql/lang/eval/ExprValueFactoryTest.kt
```diff
@@ -365,4 +369,86 @@ class ExprValueFactoryTest {
...
+
+    private fun createIonTimeStruct(hour: Int, minute: Int, second: Int, nano: Int = 0, totalSeconds: Int? = null): IonStruct =
+        ion.newEmptyStruct().apply {
+            add("hour", ion.newInt(hour))
+            add("minute", ion.newInt(minute))
+            add("second", ion.newFloat( second + nano * 10.0.pow(-9)))
+            add("timezone_hour", ion.newInt(totalSeconds?.div(3600)))
+            add("timezone_minute", ion.newInt(totalSeconds?.div((60))?.rem(60)))
+            addTypeAnnotation("\$partiql_time")
+        }
```
almann, line 381: This looks like copy/paste to me...

lang/test/org/partiql/lang/eval/ExprValueFactoryTest.kt
```diff
@@ -365,4 +369,86 @@ class ExprValueFactoryTest {
...
+            message = "Expected values to be equal."
+        )
+        val timeIonValue = timeExprValue.ionValue
+        val expectedIonValue = createIonTimeStruct(23, 2, 29, 0, -720*60)
+        assertEquals(
+            expected = expectedIonValue,
+            actual = timeIonValue,
+            message = "Expected ionValues to be equal."
+        )
+    }
```
almann, line 453: There is a lot of structurally similar code here--consider making this parameterized or at the least factor out the common code into a method.

almann (changes requested):
Overall, I think we need to centralize the time/precision logic into its own data type.  You have a lot of duplicative logic lurking in various places that leads me to believe this and also I think you are dropping precision which I think is not correct.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
```diff
@@ -205,6 +220,34 @@ private class ExprValueFactoryImpl(override val ion: IonSystem) : ExprValueFacto
     override fun newTimestamp(value: Timestamp): ExprValue =
         TimestampExprValue(ion, value)
 
+    override fun newTime(value: LocalTime, zoneOffset: ZoneOffset?): ExprValue {
+        return TimeExprValue(ion, value, zoneOffset)
+    }
+
+    override fun newTime(hour: Int, minute: Int, second: Int, nano: Int, precision: Int?, tz_minutes: Int?): ExprValue {
```
dlurton, line 227: You can use default values instead of nullable parameters:

```
@JvmOverloads
override fun newTime(
    hour: Int, 
    minute: Int, 
    second: Int, 
    nano: Int, 
    precision: Int = DEFAULT_TIME_PRECISION,
    tz_minutes: Int? = null
): ExprValue { ... }
```
An exception might be if you have a parameter that legitimately needs to be `null`.  From my read `tz_minutes` does need to be nullable, so I have specified a default value of `null` and left it nullable.   When in Kotlin-land, I would definitely avoid using `null` as a way for the caller to indicate "please use the default value".    

We really should be using [`@JvmOverloads`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-overloads/)on all functions that have default parameters.  Java clients will see additional functions that call the one you wrote here, specifying the default values, like so:

```
fun newTime(
    hour: Int, 
    minute: Int, 
    second: Int, 
    nano: Int
) = newTime(hour, minute, second, nano, DEFAULT_TIME_PRECISION, tz_minutes = null)
```
and 
```newTime(
    hour: Int, 
    minute: Int, 
    second: Int, 
    nano: Int, 
    precision: Int
)  = newTime(hour, minute, second, nano, precision, tz_minutes = null)
```

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
```diff
@@ -345,6 +388,28 @@ private class TimestampExprValue(val ion: IonSystem, val value: Timestamp): Scal
     override fun ionValueFun(): IonValue = ion.newTimestamp(value)
 }
 
+// Store LocalTime and ZoneOffset for the ease of Datetime arithmetic operations on the instance.
+private class TimeExprValue(val ion: IonSystem, val value: LocalTime, val zoneOffset: ZoneOffset? = null): ScalarExprValue() {
+    private val PARTIQL_TIME_ANNOTATION = "\$partiql_time"
+    private val SECONDS_IN_ONE_HOUR = 3600
+    private val SECONDS_IN_ONE_MINUTE = 60
```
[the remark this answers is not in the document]
dlurton: Yeah, without `const`, we get a backing field and a getter.  Definitely make them constants.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
```diff
@@ -205,6 +220,34 @@ private class ExprValueFactoryImpl(override val ion: IonSystem) : ExprValueFacto
...
         TimestampExprValue(ion, value)
 
+    override fun newTime(value: LocalTime, zoneOffset: ZoneOffset?): ExprValue {
+        return TimeExprValue(ion, value, zoneOffset)
+    }
+
+    override fun newTime(hour: Int, minute: Int, second: Int, nano: Int, precision: Int?, tz_minutes: Int?): ExprValue {
+        // Round nanoseconds to the given precision.
+        val nanoWithPrecision = when  {
+            precision == null || precision >= 9 -> nano
```
dlurton, line 230: A default value for a non-nullable `precision` parameter avoids the need for the null check here.

dlurton: Also need to consider `precision < 0`, I think, even if the parser rejects the negative sign.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
```diff
@@ -90,7 +94,18 @@ interface ExprValueFactory {
...
+    /** Returns a PartiQL `TIME` [ExprValue] instance representing the specified [LocalTime] and the [ZoneOffset].
+     * If the [zoneOffset] is null, then the instance created is of [LocalTime].
+     * If the [zoneOffset] is not null, then the instance created is of [OffsetTime] with time as [LocalTime] and offset as [zoneOffset]
+     */
+    fun newTime(value: LocalTime, zoneOffset: ZoneOffset? = null): ExprValue
+
+    /** Returns a PartiQL `TIME` [ExprValue] instance for the given hour, minute, second, precision and tz_minutes.
+     * If [tz_minutes] is not null then the time has the timezone offset, not otherwise.
+     */
+    fun newTime(hour: Int, minute: Int, second: Int, nano: Int, precision: Int? = null, tz_minutes: Int? = null): ExprValue
```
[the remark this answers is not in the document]
dlurton: Same thing I mentioned in another comment about default values being preferred over nullable parameters and also please use `@JvmOverloads`.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -1987,6 +1987,13 @@ internal class EvaluatingCompiler(
         return thunkFactory.thunkEnv(metas) { value }
     }
 
+    private fun compileTime(node: DateTimeType.Time) : ThunkEnv {
+        val (hour, minute, second, nano, precision, tz_minutes, metas) = node
+        return thunkFactory.thunkEnv(metas) {
+            valueFactory.newTime(hour, minute, second, nano, precision, tz_minutes)
+        }
+    }
```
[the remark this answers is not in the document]
dlurton: I would make sure that the out of range time literals are checked in `ExprValueFactory.newTime` at least since everything needs to call that to to get a time instance.

dlurton (changes requested):

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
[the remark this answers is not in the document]
abhikuhikar: That's right. The parser catches the out of range values. The evaluator (`java.time` library to be precise) also catches the out of range values when the instance of `LocalTime` is created.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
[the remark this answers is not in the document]
abhikuhikar: @almann As discussed in the design meeting, I have created the wrapper class to encapsulate the `Time`. Added the documentation to make it more clear.
@dlurton That's correct. `Precision` with a default `int` value is better.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
```diff
@@ -205,6 +220,34 @@ private class ExprValueFactoryImpl(override val ion: IonSystem) : ExprValueFacto
...
+    override fun newTime(value: LocalTime, zoneOffset: ZoneOffset?): ExprValue {
+        return TimeExprValue(ion, value, zoneOffset)
+    }
+
+    override fun newTime(hour: Int, minute: Int, second: Int, nano: Int, precision: Int?, tz_minutes: Int?): ExprValue {
+        // Round nanoseconds to the given precision.
+        val nanoWithPrecision = when  {
+            precision == null || precision >= 9 -> nano
+            else -> (BigDecimal(nano * 10.0.pow(-9)).setScale(precision, RoundingMode.HALF_UP) * 10.0.pow(9)).toInt()
+        }
```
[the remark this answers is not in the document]
abhikuhikar: I tried looking up for it but wasn't able to find any better sophisticated way to do the rounding except using `BigDecimal`.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
```diff
@@ -205,6 +220,34 @@ private class ExprValueFactoryImpl(override val ion: IonSystem) : ExprValueFacto
...
+        // If the nanos are added to form up a whole second because of the specified precision, carry over the second all up to the hour
+        // and use the mod values of all the new fields to fit in the valid range.
```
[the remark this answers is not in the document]
abhikuhikar: That is correct. `nano` is the fractional part of the `second` in time value. I have added the documentation to clarify this.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
```diff
@@ -205,6 +220,34 @@ private class ExprValueFactoryImpl(override val ion: IonSystem) : ExprValueFacto
...
+        val newNano = nanoWithPrecision % 10.0.pow(9).toInt()
+        val newSecond = second + (nanoWithPrecision / 10.0.pow(9).toInt())
+        val newMinute = minute + (newSecond / 60)
+        val newHour = (hour + (newMinute / 60))
+        val localTime = LocalTime.of(
+            newHour % 24,
+            newMinute % 60,
+            newSecond % 60,
+            newNano
+        )
```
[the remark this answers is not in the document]
abhikuhikar: I am assuming you mean using the java's `LocalTime` APIs. There are bunch of APIs provided which can achieve this but there's one edge case which is not handled by any of the provided APIs. i.e. when the `nano`s round up to form a whole second. For e.g., `TIME (2) '23:59:59.9999'`. In this case,  `nano`s after being rounded overflows and carries over up to the `hour` value. I was considering using `withNanos` API but it will fail for the above mentioned example as the `nano` value overflows.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
```diff
@@ -345,6 +388,28 @@ private class TimestampExprValue(val ion: IonSystem, val value: Timestamp): Scal
     override fun ionValueFun(): IonValue = ion.newTimestamp(value)
 }
 
+// Store LocalTime and ZoneOffset for the ease of Datetime arithmetic operations on the instance.
+private class TimeExprValue(val ion: IonSystem, val value: LocalTime, val zoneOffset: ZoneOffset? = null): ScalarExprValue() {
```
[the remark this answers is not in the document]
abhikuhikar: New wrapper class preserves the precision value.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
```diff
@@ -345,6 +388,28 @@ private class TimestampExprValue(val ion: IonSystem, val value: Timestamp): Scal
...
+    private val SECONDS_IN_ONE_HOUR = 3600
+    private val SECONDS_IN_ONE_MINUTE = 60
+
+    fun createIonTimeStruct(): IonStruct =
+        ion.newEmptyStruct().apply {
+            add("hour", ion.newInt(value.hour))
+            add("minute", ion.newInt(value.minute))
+            add("second", ion.newFloat( value.second + value.nano * 10.0.pow(-9)))
+            add("timezone_hour", ion.newInt(zoneOffset?.totalSeconds?.div(SECONDS_IN_ONE_HOUR)))
+            add("timezone_minute", ion.newInt(zoneOffset?.totalSeconds?.div((SECONDS_IN_ONE_MINUTE))?.rem(60)))
```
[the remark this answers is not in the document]
abhikuhikar: Changed the code to use constants.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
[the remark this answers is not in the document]
abhikuhikar: Noted. Thanks. Changed the code to use constants.

lang/src/org/partiql/lang/eval/Scalar.kt
abhikuhikar: Included a wrapper class that wraps the entire TIME value including `localTime`, `zoneOffset` and `precision`.

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: Even though this is a little redundant code, it is cleaner this way in my opinion. Moreover, as this code is converting the string token back to the `AST` node , I don't think including this logic in the wrapper class is a great idea.

lang/src/org/partiql/lang/util/ExprValueFormatter.kt
abhikuhikar: Moved this logic to the `toString` method of the wrapper class.

lang/src/org/partiql/lang/util/UtilityExtensions.kt
abhikuhikar: Done. Thanks.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
abhikuhikar: Yes. I might as well get rid of this assertion as it is mirroring the code.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
abhikuhikar: The new wrapper class preserves the originally specified precision.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
abhikuhikar: I have included the wrapper class. On an orthogonal note, the class is only for the test purpose. I have renamed it to `TimeForTest` for more clarity.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
abhikuhikar: That is correct. The tests are deterministic.

lang/test/org/partiql/lang/eval/ExprValueFactoryTest.kt
abhikuhikar: I have removed the `ionValue` assertion from here as those assertions are covered in `EvaluatingCompilerDateTimeTests`.

lang/test/org/partiql/lang/eval/ExprValueFactoryTest.kt
abhikuhikar: I have removed the `ionValue` assertion from here as those assertions are covered in `EvaluatingCompilerDateTimeTests`.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
abhikuhikar: This is little tricky. Using `@JvmOverloads` is not allowed on Kotlin interfaces. The overridden methods in Kotlin cannot have default values for parameters. If we want to achieve this we probably will have to write the overloads manually. I am punting this for now until we find a better alternative. If not I will publish another PR to include the overloaded methods.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
abhikuhikar: I see. Added the check for the `precision`.

abhikuhikar:
Included a wrapper class to consolidate the Time/precision logic in one place. This addresses most of the concerns surrounding duplicative code and different modality of TIME values. Added documentation as requested and required.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
abhikuhikar: Done. Thanks.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -22,6 +22,7 @@ import org.partiql.lang.errors.ErrorCode.*
 import org.partiql.lang.errors.Property
 import org.partiql.lang.errors.Property.*
 import org.partiql.lang.errors.*
+import org.partiql.lang.eval.builtins.Time.Companion.MAX_PRECISION_FOR_TIME
```
almann, line 25: *Minor*: If the constant was named `MAX_PRECISION`, one only need to import the `Time` type and refer to it as `Time.MAX_PRECISION` which seems more concise.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2404,10 +2403,19 @@ class SqlParser(private val ion: IonSystem) : Parser {
                 }
             }
         }
+        // Extract the precision from the time if the precision is not specified.
+        // For e.g., TIME '23:12:12.123' should have precision of 3.
+        // Note that TIME '23:12:12.1230' will also have precision of 3 unless specified otherwise as the significant
+        // digits in the fractional part excluding trailing zeros is 3.
```
almann, line 2409: I would not expect `.1230` to have a precision of 3, but rather a precision of 4, what am I missing?

lang/src/org/partiql/lang/util/UtilityExtensions.kt
```diff
@@ -1,25 +1,34 @@
...
- * Licensed under the Apache License, Version 2.0 (the "License").
- *  You may not use this file except in compliance with the License.
- * A copy of the License is located at:
- *
- *      http://aws.amazon.com/apache2.0/
- *
- *  or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
- *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
- *  language governing permissions and limitations under the License.
- */
```
almann: We still need a copyright header, don't we? Or has that changed?

lang/src/org/partiql/lang/util/UtilityExtensions.kt
```diff
@@ -1,25 +1,34 @@
...
-
 package org.partiql.lang.util
```
almann, line 1: Should these be in a file called `TimeExtensions.kt` or some such?

lang/src/org/partiql/lang/util/UtilityExtensions.kt
```diff
@@ -1,25 +1,34 @@
...
 import java.time.ZoneOffset
 import kotlin.math.absoluteValue
 
 /**
  * Returns the string representation of the [ZoneOffset] in HH:mm format.
  */
-fun ZoneOffset.getOffsetHHmm(): String {
-    return "${if(totalSeconds >= 0) "+" else "-"}${(totalSeconds / 3600).absoluteValue.toString().padStart(2, '0')}:${((totalSeconds / 60) % 60).absoluteValue.toString().padStart(2, '0')}"
-}
+fun ZoneOffset.getOffsetHHmm(): String =
```
almann, line 13: Should these methods be internal?

lang/src/org/partiql/lang/util/UtilityExtensions.kt
```diff
@@ -1,25 +1,34 @@
...
+    else -> {
+        var trailingZeros = 0
+        var nanos = nano
+        while (nanos % 10 == 0) {
+            trailingZeros ++
+            nanos /= 10
+        }
+        MAX_PRECISION_FOR_TIME - trailingZeros
+    }
+}
```
almann, line 34: I don't think this is right.  Given some nanoseconds, you cannot infer precision.  The only way to infer precision is from the textual representation otherwise it must be given to us.

lang/test/org/partiql/lang/eval/ExprValueFactoryTest.kt
```diff
@@ -383,72 +385,54 @@ class ExprValueFactoryTest {
...
+
+    @Test
+    fun invalidPrecisionForTime() {
+        try {
+            Time.of(23, 12, 34, 344423, -1, 300)
+            Assert.fail("Expected evaluation error")
+        } catch (e: EvaluationException) {
+            Assert.assertEquals(ErrorCode.EVALUATOR_INVALID_PRECISION_FOR_TIME, e.errorCode)
+        }
```
almann, line 436: Do you have negative tests for a precision of  `>9`?  Do you have negative tests for various invalid times and offsets?  I know you're leaning on the `java.time` library, but we should make sure we have at least some tests here for those cases.

almann (changes requested):
Getting closer--thanks for incorporating feedback.  I have some real concerns about the inference of precision from nanos--you can't just throw away trailing zeros.

For example in Ion timestamp `...12:23:45.100` would be a time with a mantissa of milliseconds precision, those zeroes are significant.

Also I think you are eliding precision when the user gives a value that is not supported `> 9`.

lang/src/org/partiql/lang/util/UtilityExtensions.kt
dlurton: FWIW, Theo told me once that it wasn't necessary as long as there was a license file in the root of the repository.

lang/src/org/partiql/lang/eval/builtins/Time.kt
```diff
@@ -0,0 +1,116 @@
+package org.partiql.lang.eval.builtins
```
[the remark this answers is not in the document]
dlurton: this package is for  is for built-in functions, i.e. putting the `Time` class here fails the "which of these is not like the other" test.

I'd opt to put this in the `org.partiql.lang.eval` package myself, which is where all of the "built-in" `ExprValue` implementations live.

lang/src/org/partiql/lang/eval/builtins/Time.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    companion object {
+
+        const val PARTIQL_TIME_ANNOTATION = "\$partiql_time"
+        const val HOURS_PER_DAY = 24
+        const val MINUTES_PER_HOUR = 60
+        const val SECONDS_PER_MINUTE = 60
+        const val SECONDS_PER_HOUR = SECONDS_PER_MINUTE * MINUTES_PER_HOUR
+        const val NANOS_PER_SECOND = 1000000000.0
+        const val MAX_PRECISION_FOR_TIME = 9
```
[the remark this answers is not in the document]
dlurton: It may seem logical for organizational purposes to put the functions and constants we want to be "static" in a companion object, however, I would strongly discourage that and instead use top-level declarations, that are `private` (ideally), or `internal` (if needed), and only `public` if intended to be accessed by customer code.

Aside from being idiomatic to Kotlin (when in Kotlin-land, do as the Kotlonians do :) ), using top-level declarations for these is preferable because `companion object` are *not* "static" at all, they actually needlessly create an anonymously typed singleton `object` instance with these fields.  Companion objects also pollute the API as seen by Java and seem to provide much in the way of actual value.

In general, I'd avoid the use of `companion object`s unless truly needed.  I haven't yet seen a circumstance where I felt using the `companion object` was *truly* required.  The alternatives are almost always preferable.

Some of the code in this library was written before we came to this conclusion about `companion object`s, so unfortunately you'll see a number that don't actually need to be there.

lang/src/org/partiql/lang/ast/ast.kt
```diff
@@ -1008,7 +1009,7 @@ sealed class DateTimeType : ExprNode() {
         val minute: Int,
         val second: Int,
         val nano: Int,
-        val precision: Int = 9,
+        val precision: Int = MAX_PRECISION_FOR_TIME,
```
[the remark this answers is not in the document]
abhikuhikar: Removed the default parameter value.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
```diff
@@ -94,16 +92,28 @@ interface ExprValueFactory {
...
+     * For e.g., if [nano] is 126700000 and the [precision] is 3, the [nano] will be rounded to 127000000.
+     * Note that the [nano] will still store the value in nanoseconds (0's padded after the desired precision digits),
+     * however the [Time] instance will preserve the [precision] thereby preserving the entire original value.
+     * @param tz_minutes  the minutes of the UTC time-zone offset, from -1080 to 1080.
+     * If [tz_minutes] is null then the timezone offset is not defined.
+     * @return TimeExprValue
+     * @throws DateTimeException if the value of any field is out of range
      */
-    fun newTime(hour: Int, minute: Int, second: Int, nano: Int, precision: Int? = null, tz_minutes: Int? = null): ExprValue
+    fun newTime(hour: Int, minute: Int, second: Int, nano: Int, precision: Int = MAX_PRECISION_FOR_TIME, tz_minutes: Int? = null): ExprValue
```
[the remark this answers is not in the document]
abhikuhikar: I think we don't need this one. I have removed this.

lang/src/org/partiql/lang/eval/builtins/Time.kt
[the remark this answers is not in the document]
abhikuhikar: I see. I have moved it to `lang.eval.time`.

lang/src/org/partiql/lang/eval/builtins/Time.kt
```diff
@@ -0,0 +1,116 @@
...
+import java.time.format.DateTimeFormatter
+import kotlin.math.min
+
+/**
+ * Wrapper class representing the run time instance of TIME in PartiQL.
+ * - `TIME [(p)] HH:MM:ss[.ddd][+|-HH:MM]` PartiQL statement creates a run-time instance of this class with [zoneOffset] as [null].
+ * - `TIME WITH TIME ZONE [(p)] HH:MM:ss[.ddd][+|-HH:MM]` PartiQL statement creates a run-time instance of this class with both
+ * [localTime] and [zoneOffset] defined.
+ *
+ * @param localTime Represents the local time.
```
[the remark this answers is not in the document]
abhikuhikar: Thanks. Changed.

lang/src/org/partiql/lang/eval/builtins/Time.kt
```diff
@@ -0,0 +1,116 @@
...
+ * @param zoneOffset Represents the time zone of the TIME instance.
+ *  If the [zoneOffset] is null, the zone offset is not defined for the instance.
```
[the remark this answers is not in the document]
abhikuhikar: Thanks. Changed.

lang/src/org/partiql/lang/eval/builtins/Time.kt
```diff
@@ -0,0 +1,116 @@
...
+ * @param precision Represents the number of significant digits in fractional part of the second's value of the TIME instance.
+ *  The default precision is 9 meaning, the default precision is of nanoseconds.
+ *  If the precision is specified by the user, the fractional part of the second will be rounded to the specified precision.
+ *  For eg, `TIME (3) 23:46:58.1267` will be stored in this instance with [localTime] as `23:46:58.127000000` along with preserving the
+ *  [precision] value as 3.
+ *  Note that the [LocalTime] always stores the fractional part of the second in nanoseconds.
+ *  It is up to the application developers to make use of the preserved [precision] value as need be.
+ */
+data class Time(val localTime: LocalTime, val zoneOffset: ZoneOffset? = null, val precision: Int = MAX_PRECISION_FOR_TIME) {
```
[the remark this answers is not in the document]
abhikuhikar: Removed the default value.

lang/src/org/partiql/lang/eval/builtins/Time.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    init {
+        // Validate that the precision value is non-negative.
+        if (precision < 0) {
```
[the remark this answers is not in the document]
abhikuhikar: Added the check. Thanks.

lang/src/org/partiql/lang/eval/builtins/Time.kt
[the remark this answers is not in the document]
abhikuhikar: I see. Thanks @dlurton for sharing the great insight. I have moved these to top level.

lang/src/org/partiql/lang/eval/builtins/Time.kt
```diff
@@ -0,0 +1,116 @@
...
+         * For e.g., if [nano] is 126700000 and the [precision] is 3, the [nano] will be rounded to 127000000.
+         * Note that the [nano] will still store the value in nanoseconds (0's padded after the desired precision digits),
+         * however the [Time] instance will preserve the [precision] thereby preserving the entire original value.
+         */
+        @JvmOverloads
+        fun of(hour: Int, minute: Int, second: Int, nano: Int, precision: Int = MAX_PRECISION_FOR_TIME, tz_minutes: Int? = null) : Time {
+            // Round nanoseconds to the given precision.
+            val nanoWithPrecision = when  {
+                precision >= 9 -> nano
+                else -> (BigDecimal(nano / NANOS_PER_SECOND).setScale(precision, RoundingMode.HALF_UP) * NANOS_PER_SECOND).toInt()
```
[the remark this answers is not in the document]
abhikuhikar: Removed the `>` part as it is not allowed now.

lang/src/org/partiql/lang/eval/builtins/Time.kt
```diff
@@ -0,0 +1,116 @@
...
+                ZoneOffset.ofTotalSeconds( it * SECONDS_PER_MINUTE)
+            }
+            return Time(localTime, zoneOffset, precision)
+        }
+    }
+
+    val offsetTime
+        get() : OffsetTime? = zoneOffset?.let {
+            OffsetTime.of(localTime, it)
+        }
```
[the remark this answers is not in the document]
abhikuhikar: Added to the kdoc. Thanks.

lang/src/org/partiql/lang/eval/builtins/Time.kt
```diff
@@ -0,0 +1,116 @@
...
+            add("minute", ion.newInt(localTime.minute))
+            add("second", ion.newDecimal(BigDecimal(localTime.second + localTime.nano / NANOS_PER_SECOND).setScale(precision, RoundingMode.HALF_UP)))
+            add("timezone_hour", ion.newInt(zoneOffset?.totalSeconds?.div(SECONDS_PER_HOUR)))
+            add("timezone_minute", ion.newInt(zoneOffset?.totalSeconds?.div((SECONDS_PER_MINUTE))?.rem(60)))
+            addTypeAnnotation(PARTIQL_TIME_ANNOTATION)
+        }
+
+    private fun formatterPattern() : String {
+        return "HH:mm:ss" + if (precision > 0) "." + "S".repeat(min(9, precision)) else ""
+    }
```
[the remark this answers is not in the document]
abhikuhikar: Added a note. Thanks.

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: Changed the code to infer precision based on the length of the mantissa.

lang/src/org/partiql/lang/util/UtilityExtensions.kt
abhikuhikar: So I am opting to not include the license in these files.

lang/src/org/partiql/lang/util/UtilityExtensions.kt
abhikuhikar: Renamed the file. Thanks.

lang/src/org/partiql/lang/util/UtilityExtensions.kt
abhikuhikar: Changed to `internal`.

lang/src/org/partiql/lang/util/UtilityExtensions.kt
abhikuhikar: Removed this method.

lang/test/org/partiql/lang/eval/ExprValueFactoryTest.kt
abhikuhikar: That's a great catch. I have added the range checks for all the parameters of `Time.Of` method. I have also added the corresponding negative tests in `test.eval.Time`. I have wrapped around the `DateTimeException` in our own `EvaluationException`.

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: A wildcard import (`Time.*`) might make this ambiguous.

lang/src/org/partiql/lang/eval/time/Time.kt
```diff
@@ -32,11 +46,11 @@ import kotlin.math.min
  *  Note that the [LocalTime] always stores the fractional part of the second in nanoseconds.
  *  It is up to the application developers to make use of the preserved [precision] value as need be.
  */
-data class Time(val localTime: LocalTime, val zoneOffset: ZoneOffset? = null, val precision: Int = MAX_PRECISION_FOR_TIME) {
+data class Time(val localTime: LocalTime, val precision: Int = MAX_PRECISION_FOR_TIME, val zoneOffset: ZoneOffset? = null) {
```
almann, line 49: I don't think this `MAX_PRECISION_FOR_TIME` default is necessary or desirable.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -708,15 +708,15 @@ class SqlParser(private val ion: IonSystem) : Parser {
             }
             TIME -> {
                 val timeString = token!!.text!!
-                val precision = min(children[0].token?.value?.numberValue()?.toInt() ?: MAX_PRECISION_FOR_TIME, MAX_PRECISION_FOR_TIME)
+                val precision = children[0].token?.value?.numberValue()?.toInt()
```
almann, line 711: You probably want the `!!` here or better yet `?:` and throw in the null case something more meaningful.

almann: If precision is not specified, what is it?  I would assume you infer it from the literal, yes?

lang/test/org/partiql/lang/syntax/SqlParserDateTimeTests.kt
```diff
@@ -172,10 +163,32 @@ class SqlParserDateTimeTests : SqlParserTestBase() {
...
+        // TODO: These tests should pass. Check https://github.com/partiql/partiql-lang-kotlin/issues/395
+//        DateTimeTestCase("TIME '23:59:59.1234567890'") {
+//            litTime(timeValue(23, 59, 59, 123456789, 9, null))
+//        },
+//        DateTimeTestCase("TIME '23:59:59.1234567899'") {
+//            litTime(timeValue(23, 59, 59, 123456790, 9, null))
+//        },
+//        DateTimeTestCase("TIME '23:59:59.1234567890+18:00'") {
+//            litTime(timeValue(23, 59, 59, 123456789, 9, null))
+//        },
```
almann, line 176: Are we able to have these uncommented and somehow ignored (might be a limitation of test framework)?

lang/test/org/partiql/lang/eval/time/TimeTest.kt
```diff
@@ -0,0 +1,177 @@
...
+        ),
+        case(
+            hour = 23,
+            min = 23,
+            sec = 12,
+            nano = 123456789,
+            precision = 0,
+            tz_min = 18 * MINUTES_PER_HOUR,
+            expectedErrorCode = ErrorCode.EVALUATOR_DATETIME_EXCEPTION
+        ),
```
almann, line 112: It would be nice to have either a description field or comment indicating why this should fail (and other cases like it).  Not entirely obvious without looking at the case and even though I might infer an error, it might not be the same one you were thinking...

almann (commented):
I'm pretty close to approving this--some questions below.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
```diff
@@ -92,28 +91,23 @@ interface ExprValueFactory {
...
      * Note that the [nano] will still store the value in nanoseconds (0's padded after the desired precision digits),
      * however the [Time] instance will preserve the [precision] thereby preserving the entire original value.
+     * The valid values for precision are between 0 and 9 inclusive.
      * @param tz_minutes  the minutes of the UTC time-zone offset, from -1080 to 1080.
      * If [tz_minutes] is null then the timezone offset is not defined.
      * @return TimeExprValue
      * @throws DateTimeException if the value of any field is out of range
      */
-    fun newTime(hour: Int, minute: Int, second: Int, nano: Int, precision: Int = MAX_PRECISION_FOR_TIME, tz_minutes: Int? = null): ExprValue
+    fun newTime(hour: Int, minute: Int, second: Int, nano: Int, precision: Int, tz_minutes: Int? = null): ExprValue
```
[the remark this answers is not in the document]
abhikuhikar: I think it would be nice to have an API supported on `ExprValueFactory` to create `newTime` with `hour, min, sec....`.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
```diff
@@ -230,13 +224,9 @@ private class ExprValueFactoryImpl(override val ion: IonSystem) : ExprValueFacto
...
 
-    override fun newTime(value: Time): ExprValue {
-        return TimeExprValue(ion, value)
-    }
-
     override fun newTime(hour: Int, minute: Int, second: Int, nano: Int, precision: Int, tz_minutes: Int?): ExprValue {
         val timeValue = Time.of(hour, minute, second, nano, precision, tz_minutes)
-        return newTime(timeValue)
+        return TimeExprValue(ion, timeValue)
     }
```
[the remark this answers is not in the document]
abhikuhikar: Not sure if I understood this entirely.

lang/src/org/partiql/lang/eval/time/Time.kt
abhikuhikar: So here's the pickle. I think we should not expose this constructor (make `private`) at all because it does not make anything of `precision` and the caller might get confused. 
Static method `Time.of` takes `precision` and rounds the fractional part of the `second`. The constructor cannot embed this logic as `localTime` is immutable.  
Since it was still exposed, the `precision`  default was chosen to be `MAX_PRECISION_FOR_TIME`. 
We probably can introduce another overload of `Time.of(localTime: LocalTime, precision: Int, timeZone: TimeZone)`

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: 1. Good catch. Thanks. It will never be `null` so I will add `!!`.
2. That's true. The precision is inferred from the literal in `SqlParser` during parsing.

lang/test/org/partiql/lang/syntax/SqlParserDateTimeTests.kt
abhikuhikar: Not sure if I can `@ignore` these. I have kept those commented to keep track of the issue we want to resolve later.

lang/test/org/partiql/lang/eval/time/TimeTest.kt
abhikuhikar: Thanks a lot for pointing this out. This is indeed a happy case and should have failed saying "expected test to fail". I have modified the code.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -708,9 +708,9 @@ class SqlParser(private val ion: IonSystem) : Parser {
             }
             TIME -> {
                 val timeString = token!!.text!!
-                val precision = children[0].token?.value?.numberValue()?.toInt()
+                val precision = children[0].token!!.value!!.numberValue().toInt()
                 val time = LocalTime.parse(timeString, DateTimeFormatter.ISO_TIME)
```
almann, line 712: I wasn't being clear--I think the `?` is fine, but you probably want a `?: throw something` at the end of this to get a better exception unless you really want the NPE.

lang/src/org/partiql/lang/eval/time/Time.kt
almann: If that is the case then this constructor should be at least `internal`, but even still the default makes no sense because internal code will never use that default, no?

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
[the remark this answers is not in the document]
almann: Why do we need an extra constructor with all the fields here when I can just `factory.newTime(Time.of(...))`?

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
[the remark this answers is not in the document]
almann: I think it is just copy/paste from the other one--I mean by that logic should we have something similar for `Timestamp`?  `Time` should be responsible for how it gets constructed, this API need only expose binding to an instance of `Time` as the smallest API area.

lang/src/org/partiql/lang/eval/time/Time.kt
abhikuhikar: Sounds good. I will remove the default parameter and make the constructor `internal`.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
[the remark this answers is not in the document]
abhikuhikar: Sure. Will remove this API.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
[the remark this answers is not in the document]
abhikuhikar: Ohh that makes sense. It is redundant. I will remove this API.

lang/src/org/partiql/lang/eval/time/Time.kt
abhikuhikar: I have made the constructor `private` as it wasn't used anywhere and is not intended to be used (as is) anywhere.
I have instead added a static API `Time.of(localTime: LocalTime, precision: Int, zoneOffset: ZoneOffset? = null)` which  replaces the constructor and also uses the `precision` to round the `localTime`. I believe it is better this way.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2329,6 +2328,22 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
     }
 
+    /**
+     * Gets the precision from the time string of the format 'HH:MM:SS[.ddd....][+|-HH:MM]'.
+     */
+    private fun getPrecisionFromTimeString(timeString: String) : Int {
+        val matcher = genericTimeRegex.toPattern().matcher(timeString)
+        if (!matcher.find()) {
+            org.partiql.lang.eval.err("Time string does not match the format 'HH:MM:SS[.ddd....][+|-HH:MM]'",
+            propertyValueMapOf(),
```
dlurton, line 2338: Nit: indentation

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
```diff
@@ -34,4 +46,197 @@ class EvaluatingCompilerDateTimeTests : EvaluatorTestBase() {
...
+            case("TIME WITH TIME ZONE '12:25:12.123456' IS TIME", "true"),
+            case("TIME (2) WITH TIME ZONE '01:01:12' IS TIME", "true"),
+            case("'01:01:12' IS TIME", "false")
+        )
+    }
+
+    private val RANDOM_GENERATOR = generateRandomSeed()
+
+    private fun generateRandomSeed() : Random {
+        val seed = Random(1).nextInt()
```
dlurton, line 122: I don't think you intended to include `1` here since that will effectively be using the same seed with every run of the tests, thus, you won't get the best coverage.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
dlurton: I'd prefer to make them non-deterministic actually because better test coverage.  Reproducibility is the reason you `println` the seed that was used, above.  

Yes, non-determinism can lead to intermittent failures, but that is the point.  All the tests should always succeed even if the cases being fed are deterministic.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
```diff
@@ -1,13 +1,25 @@
...
+import java.math.RoundingMode
+import java.time.Instant
+import java.time.LocalTime
+import java.time.ZoneOffset
+import kotlin.math.absoluteValue
+import kotlin.random.Random
 
 class EvaluatingCompilerDateTimeTests : EvaluatorTestBase() {
 
+    private val RANDOM_TESTS_SIZE = 200
```
dlurton, line 21: This value is way too small, IMHO.  Make this large enough to cause the total time for all tests in this fixture to be at *least* several seconds long. I would even be OK with 20-30 seconds.

lang/test/org/partiql/lang/syntax/SqlParserDateTimeTests.kt
dlurton: The only way is to add a `skip` property to `DateTimeTestCase` and then filter out those cases where `testCase.skip == false`, which could easily be done.

dlurton (changes requested):
Seems pretty close to me, I think after the next revision I can give you my seal of approval.

almann (dismissed):
:ship: LGTM

lang/src/org/partiql/lang/syntax/SqlParser.kt
abhikuhikar: Done.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
abhikuhikar: I see. Good point. I thought the generated seed was different every time. I was wrong. Changed it. Thanks.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
abhikuhikar: Sounds good. I have changed this to `5000`

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
```diff
@@ -34,4 +45,197 @@ class EvaluatingCompilerDateTimeTests : EvaluatorTestBase() {
...
+            case("TIME (2) WITH TIME ZONE '12:59:59.134-05:30'", "12:59:59.13-05:30", TimeForTest(12, 59, 59, 130000000, 2, -330)),
+            case("TIME '12:25:12.123456' IS TIME", "true"),
+            case("TIME (2) '01:01:12' IS TIME", "true"),
+            case("TIME WITH TIME ZONE '12:25:12.123456' IS TIME", "true"),
+            case("TIME (2) WITH TIME ZONE '01:01:12' IS TIME", "true"),
+            case("'01:01:12' IS TIME", "false")
+        )
+    }
+
+    private val RANDOM_GENERATOR = generateRandomSeed()
```
dlurton, line 118: Nit:  also move to top level in file, make `const` if possible.  My IntelliJ wants me to change the naming convention of this field, does it not for you?

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
```diff
@@ -34,4 +45,197 @@ class EvaluatingCompilerDateTimeTests : EvaluatorTestBase() {
...
+        }
+        val timezoneMinutes = if (withTimezone) {
+            nextInt(-1080, 1081)
+        } else {
+            null
+        }
+        return TimeForTest(hour, minute, second, nano, precision, timezoneMinutes)
+    }
+
+    private val RANDOM_TIMES = List(RANDOM_TESTS_SIZE) {
```
dlurton, line 176: Nit:  also move to top level in file, make `const` if possible.  My IntelliJ wants me to change the naming convention of this field, does it not for you?

dlurton (commented):
Sorry it's taken me a while to get to this, and there's a few things I didn't notice previously.

I notice you have a lot of `UPPER_SNAKE_CASE` immutable instance properties in a various cases.  In general, Kotlin class properties should never use this naming convention.   

It's reserved for `const` and top-level immutable data only: https://kotlinlang.org/docs/coding-conventions.html#property-names  Unfortunately, the official coding conventions aren't really clear about the other naming conventions used for class instance properties, but in general we use `lowerSnakeCase`.

Most of those properties should be moved to the top-level in the file they reside in. However, that is not always the case.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
dlurton: I think something more like `50000` is appropriate, which takes about 9 seconds for the whole suite to run on my machine.

almann (approved):
:ship:

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
```diff
@@ -1,13 +1,24 @@
...
+import java.math.BigDecimal
+import java.math.RoundingMode
+import java.time.Instant
+import java.time.ZoneOffset
+import kotlin.math.absoluteValue
+import kotlin.random.Random
 
 class EvaluatingCompilerDateTimeTests : EvaluatorTestBase() {
 
+    private val RANDOM_TESTS_SIZE = 5000
```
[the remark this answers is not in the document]
abhikuhikar: Changed to `camelCase`. However, I am a little inclined on having it in the class rather than the top-level.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
```diff
@@ -34,4 +45,197 @@ class EvaluatingCompilerDateTimeTests : EvaluatorTestBase() {
...
+        )
+    }
+
+    private val RANDOM_GENERATOR = generateRandomSeed()
+
+    private fun generateRandomSeed() : Random {
+        val seed = Random.nextInt()
+        println("Randomly generated seed is ${seed}. Use this to reproduce failures in dev environment.")
+        return Random(seed)
+    }
```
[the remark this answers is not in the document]
abhikuhikar: Changed to `camelCase`.

lang/src/org/partiql/lang/eval/time/Time.kt
```diff
@@ -0,0 +1,163 @@
...
+         * Note that the [nano] will still store the value in nanoseconds (0's padded after the desired precision digits),
+         * however the [Time] instance will preserve the [precision] thereby preserving the entire original value.
+         * The valid values for precision are between 0 and 9 inclusive.
+         * @param tz_minutes  the minutes of the UTC time-zone offset, from -1080 to 1080.
+         * If [tz_minutes] is null then the timezone offset is not defined.
+         * @return TimeExprValue
+         * @throws EvaluationException if the value of any field is out of range
+         */
+        @JvmOverloads
+        fun of(hour: Int, minute: Int, second: Int, nano: Int, precision: Int, tz_minutes: Int? = null) : Time {
```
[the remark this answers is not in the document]
abhikuhikar: Added. Thanks.

lang/src/org/partiql/lang/errors/ErrorCode.kt
```diff
@@ -500,6 +500,21 @@ enum class ErrorCode(private val category: ErrorCategory,
...
+        "invalid precision used for TIME type"),
+
+    /**
+     * This is a generic error wrapper for the DateTimeException thrown by Java's [java.time] when attempting to create
+     * an instance of [java.time.LocalTime] or [java.time.OffsetTime].
+     * The exception is caught by [org.partiql.lang.eval.time.Time.Companion.of].
+     */
+    EVALUATOR_DATETIME_EXCEPTION(
+        ErrorCategory.EVALUATOR,
+        LOCATION,
```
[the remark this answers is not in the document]
abhikuhikar: Since the `DateTimeException` is wrapped around to throw this, it does throw an error with the meaningful message. 
For eg, `Time.of(12, 123, 12, 23, 3, 12)` throws the following error message. 
```
org.partiql.lang.eval.EvaluationException: Invalid value for MinuteOfHour (valid values 0 - 59): 123
```

lang/src/org/partiql/lang/eval/time/Time.kt
```diff
@@ -0,0 +1,163 @@
...
+                ChronoField.HOUR_OF_DAY.checkValidValue(hour.toLong())
+                ChronoField.MINUTE_OF_HOUR.checkValidValue(minute.toLong())
+                ChronoField.SECOND_OF_MINUTE.checkValidValue(second.toLong())
+                ChronoField.NANO_OF_SECOND.checkValidValue(nano.toLong())
+                tz_minutes?.let {
+                    ChronoField.OFFSET_SECONDS.checkValidIntValue((it * SECONDS_PER_MINUTE).toLong())
+                }
+            } catch (dte: DateTimeException) {
+                throw EvaluationException(dte, ErrorCode.EVALUATOR_DATETIME_EXCEPTION, propertyValueMapOf(), false)
+            }
```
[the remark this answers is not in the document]
abhikuhikar: I think this is a little redundant as mentioned in the above comment. Another thing we can do is not wrap around the `DateTimeException` and throw as is.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2398,17 +2413,24 @@ class SqlParser(private val ion: IonSystem) : Parser {
                     // Fall back on parsing a string without a time zone offset only if the offset is not specified.
                     // Add local system timezone offset in that case.
                     tryLocalTimeParsing(timeString)
-                    newTimeString = timeString + ZoneOffset.systemDefault().rules.getOffset(Instant.now()).toString()
+                    newTimeString = timeString + ZoneOffset.systemDefault().rules.getOffset(Instant.now()).getOffsetHHmm()
```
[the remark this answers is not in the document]
abhikuhikar: Sure. I will do this in the separate PR.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2398,17 +2413,24 @@ class SqlParser(private val ion: IonSystem) : Parser {
...
                     rem.head.err(e.localizedMessage, PARSE_INVALID_TIME_STRING)
                 }
             }
         }
+        // Extract the precision from the time string representation if the precision is not specified.
+        // For e.g., TIME '23:12:12.12300' should have precision of 5.
+        // The source span here is just the filler value and does not reflect the actual source location of the precision
+        // as it does not exists in case the precision is unspecified.
+        val precisionOfValue =  precision.token ?:
+            Token(LITERAL, ion.newInt(getPrecisionFromTimeString(newTimeString)), SourceSpan(-1, -1, -1))
```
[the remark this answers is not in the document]
abhikuhikar: I see. Changed it as suggested.

lang/test/org/partiql/lang/syntax/SqlParserDateTimeTests.kt
```diff
@@ -26,7 +26,12 @@ class SqlParserDateTimeTests : SqlParserTestBase() {
 
     @Test
     @Parameters
```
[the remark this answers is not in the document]
abhikuhikar: Not exactly sure what you mean here. Do you mean running an individual test in the file is difficult? I am able to run the individual test by clicking the "play icon" next to it.

lang/test/org/partiql/lang/syntax/SqlParserDateTimeTests.kt
```diff
@@ -17,7 +17,7 @@ class SqlParserDateTimeTests : SqlParserTestBase() {
 
     private val LOCAL_TIME_ZONE_OFFSET = (ZoneOffset.systemDefault().rules.getOffset(Instant.now()).totalSeconds / 60).toLong()
 
-    data class DateTimeTestCase(val source: String, val block: PartiqlAst.Builder.() -> PartiqlAst.PartiqlAstNode)
+    data class DateTimeTestCase(val source: String, val skipTest: Boolean = false, val block: PartiqlAst.Builder.() -> PartiqlAst.PartiqlAstNode)
     private data class Date(val year: Int, val month: Int, val day: Int)
 
     private val MONTHS_WITH_31_DAYS = listOf(1, 3, 5, 7, 8, 10, 12)
```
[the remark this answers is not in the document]
abhikuhikar: Changed to `camelCase`. I don't feel so strongly about moving this to top-level type. Because as per the logic above, all the immutable properties (`val`) in the class should go to top-level which is not always appropriate in my opinion. 

The error has come up previously as well. I have fixed it in the `CAST` PR. Nevertheless, I have changed it to `50000`.

lang/test/org/partiql/lang/eval/time/TimeTest.kt
```diff
@@ -0,0 +1,182 @@
...
+    )
+
+    fun case(hour: Int, min: Int, sec: Int, nano: Int, precision: Int, tz_min: Int?, expectedLocalTime: LocalTime, expectedZoneOffset: ZoneOffset?) =
+        TimeTestCase(hour, min, sec, nano, precision, tz_min, expectedLocalTime, expectedZoneOffset)
+
+    fun case(hour: Int, min: Int, sec: Int, nano: Int, precision: Int, tz_min: Int?, expectedErrorCode: ErrorCode) =
+        TimeTestCase(hour, min, sec, nano, precision, tz_min, expectedErrorCode = expectedErrorCode)
+
+    @Test
+    @Parameters
```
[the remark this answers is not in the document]
abhikuhikar: Same comment as above.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
abhikuhikar: Changed to `camelCase`.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
abhikuhikar: Changed to `camelCase`.

lang/test/org/partiql/lang/syntax/SqlParserDateTimeTests.kt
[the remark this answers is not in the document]
dlurton: 1. Run this test in IntelliJ.
2. In the test runner, it will expand to one test for every case
3. right-click one of the test cases in IntelliJ, notice that it says "nothing here" 
4. Run any test where the test cases are specified with either `@MethodSource` or `@ArgumentsSource` (i.e. a JUnit5 parameterized test)
5. Individual test cases will appear in the test runner as before, however... 
6. Right click on one test case - observe that you can run just that one test case.

dlurton (approved):

lang/test/org/partiql/lang/syntax/SqlParserDateTimeTests.kt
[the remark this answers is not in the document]
abhikuhikar: Ohh, I get it now. Yeah. It indeed is easier to run a single test. Thanks a lot. I will change this in one of the subsequent PRs. Closing this for now.

PR #397, opened 2021-04-30
Extends extract exprFunction to support TIME type.

abhikuhikar:
*Issue #, if available:*

*Description of changes:*
 - Extends `Extract` `ExprFunction` to support `TIME` type.
 - Enriches `Time` API to support returning values such as `timezone_hour`, `timezone_minute`, `secondsWithFractionalPart`.


lang/src/org/partiql/lang/eval/time/Time.kt
```diff
@@ -120,6 +120,24 @@ data class Time(val localTime: LocalTime, val precision: Int = MAX_PRECISION_FOR
...
+     * Returns the TIMEZONE_HOUR for the [zoneOffset] of this instance.
+     */
+    val timezoneMinute
+        get() : Int? = (zoneOffset?.totalSeconds?.div(SECONDS_PER_MINUTE))?.rem(SECONDS_PER_MINUTE)
+
+    /**
+     * Returns the seconds along with the fractional part of the second's value.
+     */
+    val secondsWithFractionalPart
+        get()  = BigDecimal(localTime.second + localTime.nano / NANOS_PER_SECOND)
```
almann, line 139: I think this is problematic (but maybe tests exist that prove me wrong). Aren't you doing integer division here so isn't that effectively zero?  I think what you want is is to scale the coefficient (`second + nano`) by the scale `9`.  But you probably should take into account precision if returning this as a `BigDecimal`.

lang/src/org/partiql/lang/eval/builtins/ExtractExprFunction.kt
```diff
@@ -66,11 +67,32 @@ internal class ExtractExprFunction(valueFactory: ExprValueFactory) : NullPropaga
         }.toDouble()
     }
 
+    private fun Time.extractedValue(datePart: DatePart) : Double {
+        return when (datePart) {
+            DatePart.HOUR -> localTime.hour.toDouble()
+            DatePart.MINUTE -> localTime.minute.toDouble()
+            DatePart.SECOND -> secondsWithFractionalPart.toDouble()
+            DatePart.TIMEZONE_HOUR -> timezoneHour?.toDouble() ?: errNoContext(
```
almann, line 75: Should this be called `DatePart` when the fields are _time_ related?

almann (changes requested):
I would like to make sure `secondsWithFractionalPart` works as you intend, but also probably should take into account precision and have its own tests (irrespective of the `ExprFunction`.

lang/src/org/partiql/lang/eval/builtins/ExtractExprFunction.kt
```diff
@@ -66,11 +67,32 @@ internal class ExtractExprFunction(valueFactory: ExprValueFactory) : NullPropaga
...
+                "Time unit ${datePart.name.toLowerCase()} not supported for TIME type without TIME ZONE",
+                internal = false
+            )
+            DatePart.TIMEZONE_MINUTE -> timezoneMinute?.toDouble() ?: errNoContext(
+                "Time unit ${datePart.name.toLowerCase()} not supported for TIME type without TIME ZONE",
+                internal = false
+            )
```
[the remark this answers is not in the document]
abhikuhikar: Postgresql throws an error in this case.

lang/src/org/partiql/lang/eval/time/Time.kt
abhikuhikar: `NANOS_PER_SECOND` is a `double` value so it's not an integer division. 
`secondsWithFractionalPart` is already rounded to the given `precision` as this is taken care of when the `Time` instance is created.

lang/src/org/partiql/lang/eval/builtins/ExtractExprFunction.kt
abhikuhikar: Will rename it to `DateTimePart`.

abhikuhikar: Would it be okay to punt this as renaming it to `DateTimePart` also encircles renaming a lot of places in the code including variable names, error messages, function names, etc.

lang/src/org/partiql/lang/eval/time/Time.kt
almann: Don't mix float and decimal math, you will be creating weird rounding out of your control.

lang/src/org/partiql/lang/eval/builtins/ExtractExprFunction.kt
almann: Add an issue.

lang/src/org/partiql/lang/eval/time/Time.kt
abhikuhikar: Not sure what you meant here. Do you mean being explicit here ? Something like this? 
```
    val secondsWithFractionalPart
        get()  = BigDecimal(localTime.second.toDouble() + localTime.nano / NANOS_PER_SECOND)
```

lang/src/org/partiql/lang/eval/builtins/ExtractExprFunction.kt
abhikuhikar: Added.

lang/src/org/partiql/lang/eval/time/Time.kt
almann: Meaning, I would not use `double` at all in any of this, keep to `BigDecimal` math.

lang/src/org/partiql/lang/eval/time/Time.kt
```diff
@@ -136,7 +136,7 @@ data class Time(val localTime: LocalTime, val precision: Int = MAX_PRECISION_FOR
      * Returns the seconds along with the fractional part of the second's value.
      */
     val secondsWithFractionalPart
-        get()  = BigDecimal(localTime.second + localTime.nano / NANOS_PER_SECOND)
+        get()  = localTime.second.toBigDecimal() + localTime.nano.toBigDecimal() / NANOS_PER_SECOND.toBigDecimal()
```
almann, line 139: Just to verify, none of these fields in `localTime` are `double`, right?  In other words, are they all `long`?

almann (approved):

lang/src/org/partiql/lang/eval/time/Time.kt
abhikuhikar: They are all `int`.

almann (approved):

Issue #202, opened 2020-02-13
Add ability to read the query from a file when using the CLI 


therapon:
Apart from specifying the query inline it'd be helpful to be able to read the query from a file as well

PR #398, opened 2021-04-30
Adds Compile-Time `Thread.interrupted()` checks

dlurton:
A few times over the past few years we've received reports from a PartiQL service where with a customer sending a very large  query (in terms of size of the SQL), where PartiQL would take a long time (>30s) to parse, rewrite, validate, and compile.  

Although we have optimized every case where this has happened, this PR applies a mitigation measure which can be used in case it happens again.  I have identified several "hot" loops where the state of `Thread.interrupted()` can be polled and if set an `InterruptedException` is thrown.  This is the hope that if such a query is encountered again, it can be aborted by the calling code rather than risk causing service outages.  This approach still might miss loops that we did not anticipate would become hot in the case of unexpected user input, but at least this should allow most such scenarios to be aborted.  If it later discovered that we missed such a hot loop in the future, it is a simple matter to add a call to `checkThreadInterrupted()`.

What this PR does *not* do is add an evaluation-time poll for `Thread.interrupted()`.  That will come in another PR if it is later determined to be needed.

### DIFF
--- a/lang/src/org/partiql/lang/CompilerPipeline.kt
+++ b/lang/src/org/partiql/lang/CompilerPipeline.kt
@@ -20,6 +20,7 @@ import org.partiql.lang.eval.*
 import org.partiql.lang.eval.builtins.*
 import org.partiql.lang.eval.builtins.storedprocedure.StoredProcedure
 import org.partiql.lang.syntax.*
+import org.partiql.lang.util.interruptibleFold
 
 /**
  * Contains all of the information needed for processing steps.
@@ -180,7 +181,7 @@ interface CompilerPipeline  {
     }
 }
 
-private class CompilerPipelineImpl(
+internal class CompilerPipelineImpl(
     override val valueFactory: ExprValueFactory,
     private val parser: Parser,
     override val compileOptions: CompileOptions,
@@ -198,10 +199,15 @@ private class CompilerPipelineImpl(
     override fun compile(query: ExprNode): Expression {
         val context = StepContext(valueFactory, compileOptions, functions, procedures)
 
-        val preProcessedQuery = preProcessingSteps.fold(query) { currentExprNode, step ->
-            step(currentExprNode, context)
-        }
+        val preProcessedQuery = executePreProcessingSteps(query, context)
 
         return compiler.compile(preProcessedQuery)
+    }
+
+    internal fun executePreProcessingSteps(
+        query: ExprNode,
+        context: StepContext
+    ) = preProcessingSteps.interruptibleFold(query) { currentExprNode, step ->
+        step(currentExprNode, context)
     }
 }
--- a/lang/src/org/partiql/lang/ast/AstDeserialization.kt
+++ b/lang/src/org/partiql/lang/ast/AstDeserialization.kt
@@ -248,7 +248,7 @@ class AstDeserializerBuilder(val ion: IonSystem) {
         }
 }
 
-private class AstDeserializerInternal(
+internal class AstDeserializerInternal(
     val astVersion: AstVersion,
     val ion: IonSystem,
     private val metaDeserializers: Map<String, MetaDeserializer>
@@ -262,7 +262,9 @@ private class AstDeserializerInternal(
         return deserializeExprNode(sexp)
     }
 
-    private fun validate(rootSexp: IonSexp) {
+    internal fun validate(rootSexp: IonSexp) {
+        checkThreadInterrupted()
+
         val nodeTag = rootSexp.nodeTag // Throws if nodeTag is invalid for the current AstVersion
         val nodeArgs = rootSexp.args
 
@@ -321,8 +323,9 @@ private class AstDeserializerInternal(
     /**
      * Given a serialized AST, return its [ExprNode] representation.
      */
-    private fun deserializeExprNode(metaOrTermOrExp: IonSexp): ExprNode =
-        deserializeSexpMetaOrTerm(metaOrTermOrExp) { target, metas ->
+    internal fun deserializeExprNode(metaOrTermOrExp: IonSexp): ExprNode {
+        checkThreadInterrupted()
+        return deserializeSexpMetaOrTerm(metaOrTermOrExp) { target, metas ->
             val nodeTag = target.nodeTag
             val targetArgs = target.args //args is an extension property--call it once for efficiency
             //.toList() forces immutability
@@ -417,6 +420,7 @@ private class AstDeserializerInternal(
                 NodeTag.TYPE -> errInvalidContext(nodeTag)
             }
         }
+    }
 
     private fun deserializeLit(targetArgs: List<IonValue>, metas: MetaContainer) = Literal(targetArgs.first(), metas)
     private fun deserializeMissing(metas: MetaContainer) = LiteralMissing(metas)
--- a/lang/src/org/partiql/lang/ast/ast.kt
+++ b/lang/src/org/partiql/lang/ast/ast.kt
@@ -24,19 +24,35 @@ import java.util.*
 sealed class AstNode : Iterable<AstNode> {
 
     /**
-     * returns all the children nodes.
+     * Returns all the children nodes.
+     *
+     * This property is [deprecated](see https://github.com/partiql/partiql-lang-kotlin/issues/396).  Use
+     * one of the following PIG-generated classes to analyze AST nodes instead:
+     *
+     * - [org.partiql.lang.domains.PartiqlAst.Visitor]
+     * - [org.partiql.lang.domains.PartiqlAst.VisitorFold]
      */
+    @Deprecated("DO NOT USE - see kdoc, see https://github.com/partiql/partiql-lang-kotlin/issues/396")
     abstract val children: List<AstNode>
 
     /**
      * Depth first iterator over all nodes.
+     *
+     * While collecting child nodes, throws [InterruptedException] if the [Thread.interrupted] flag has been set.
+     *
+     * This property is [deprecated](see https://github.com/partiql/partiql-lang-kotlin/issues/396).  Use
+     * one of the following PIG-generated classes to analyze AST nodes instead:
+     *
+     * - [org.partiql.lang.domains.PartiqlAst.Visitor]
+     * - [org.partiql.lang.domains.PartiqlAst.VisitorFold]
      */
+    @Deprecated("DO NOT USE - see kdoc for alternatives")
     override operator fun iterator(): Iterator<AstNode> {
         val allNodes = mutableListOf<AstNode>()
 
         fun depthFirstSequence(node: AstNode) {
             allNodes.add(node)
-            node.children.map { depthFirstSequence(it) }
+            node.children.interruptibleMap { depthFirstSequence(it) }
         }
 
         depthFirstSequence(this)
--- a/lang/src/org/partiql/lang/ast/passes/AstRewriterBase.kt
+++ b/lang/src/org/partiql/lang/ast/passes/AstRewriterBase.kt
@@ -15,6 +15,7 @@
 package org.partiql.lang.ast.passes
 
 import org.partiql.lang.ast.*
+import org.partiql.lang.util.checkThreadInterrupted
 
 /**
  * Provides a minimal interface for an AST rewriter implementation.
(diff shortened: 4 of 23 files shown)


lang/src/org/partiql/lang/eval/visitors/VisitorTransformBase.kt
```diff
@@ -1,11 +1,18 @@
 package org.partiql.lang.eval.visitors
 
 import org.partiql.lang.domains.PartiqlAst
+import org.partiql.lang.util.checkThreadInterrupted
 
 /**
  * Base-class for visitor transforms that provides additional functions outside of [PartiqlAst.VisitorTransform].
  */
 abstract class VisitorTransformBase : PartiqlAst.VisitorTransform() {
```
almann, line 9: This doc comment is incredibly oblique.  Maybe something like the following:

```suggestion
/**
 * Base-class for visitor transforms that should be used by implementors.
 * [PartiqlAst.VisitorTransform]. should **not** be used directly.
 */
abstract class VisitorTransformBase : PartiqlAst.VisitorTransform() {
```

Should implementors _ever_ implement `PartiqlAst.VisitorTransform`?  If not, we should add some comment to that effect there too.

Better yet, do we need the separation or is it because the AST visitor is generated?

almann (dismissed):
:shipit: LGTM, some minor questions below.

lang/src/org/partiql/lang/CompilerPipeline.kt
```diff
@@ -198,10 +199,15 @@ private class CompilerPipelineImpl(
...
-        val preProcessedQuery = preProcessingSteps.fold(query) { currentExprNode, step ->
-            step(currentExprNode, context)
-        }
+        val preProcessedQuery = executeProcessingSteps(query, context)
 
         return compiler.compile(preProcessedQuery)
     }
+
+    internal fun executeProcessingSteps(query: ExprNode, context: StepContext): ExprNode {
+        val preProcessedQuery = preProcessingSteps.interruptibleFold(query) { currentExprNode, step ->
```
alancai98, line 208: nit: could be inlined.

```
return preProcessingSteps.interruptibleFold(query) { currentExprNode, step ->
    step(currentExprNode, context)
}
```

lang/src/org/partiql/lang/ast/passes/AstRewriterBase.kt
```diff
@@ -31,8 +32,9 @@ interface AstRewriter {
 @Deprecated("New rewriters should implement PIG's PartiqlAst.VisitorTransform instead")
```
alancai98, line 32: Could update this comment to use `VisitorTransformBase` rather than `PartiqlAst.VisitorTransform`.

alancai98 (approved):
LGTM overall. Only have a couple nits/suggestions. Also, should these checks also be added to the v0.1.5 tag @dlurton ?

lang/src/org/partiql/lang/CompilerPipeline.kt
```diff
@@ -198,10 +199,15 @@ private class CompilerPipelineImpl(
...
 
-        val preProcessedQuery = preProcessingSteps.fold(query) { currentExprNode, step ->
-            step(currentExprNode, context)
-        }
+        val preProcessedQuery = executeProcessingSteps(query, context)
 
         return compiler.compile(preProcessedQuery)
     }
+
+    internal fun executeProcessingSteps(query: ExprNode, context: StepContext): ExprNode {
```
abhikuhikar, line 207: nit: Should this be renamed to `executePreProcessingSteps`?

lang/src/org/partiql/lang/ast/AstDeserialization.kt
```diff
@@ -291,8 +293,8 @@ private class AstDeserializerInternal(
             deserializeNode(target.asIonSexp(), metas)
         }
 
-    private fun <T> deserializeIonValueMetaOrTerm(targetValue: IonValue, deserializeNode: (IonValue, MetaContainer) -> T): T =
-        when (targetValue) {
+    private fun <T> deserializeIonValueMetaOrTerm(targetValue: IonValue, deserializeNode: (IonValue, MetaContainer) -> T): T {
+        return when (targetValue) {
```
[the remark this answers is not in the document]
abhikuhikar: +1

lang/src/org/partiql/lang/ast/ast.kt
```diff
@@ -26,17 +26,21 @@ sealed class AstNode : Iterable<AstNode> {
     /**
      * returns all the children nodes.
      */
+    @Deprecated("DO NOT USE - will be removed in future revision, see https://github.com/partiql/partiql-lang-kotlin/issues/396")
```
abhikuhikar, line 29: Should we specify the appropriate replacement for this?

lang/src/org/partiql/lang/ast/ast.kt
```diff
@@ -26,17 +26,21 @@ sealed class AstNode : Iterable<AstNode> {
...
     abstract val children: List<AstNode>
 
     /**
      * Depth first iterator over all nodes.
+     *
+     * While collecting child nodes, throws [InterruptedException] if the [Thread.interrupted] flag has been set.
      */
+    @Deprecated("DO NOT USE - will be removed in future revision, see https://github.com/partiql/partiql-lang-kotlin/issues/396")
```
abhikuhikar, line 37: Should we specify the appropriate replacement for this?

abhikuhikar (changes requested):
Some minor nits. 
Can you also add an `interrupt` check in the function `validateTopLevelNodes` in `SqlParser`?
Since it recurses over the entire tree, I remember it had very "high flames" in the CPU profiler.

lang/src/org/partiql/lang/CompilerPipeline.kt
dlurton: actually, the entire function should be inlined at line #202.

lang/src/org/partiql/lang/CompilerPipeline.kt
dlurton: actually, the entire function should be inlined at line #202.

lang/src/org/partiql/lang/ast/AstDeserialization.kt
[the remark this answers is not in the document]
dlurton: Was original location of a  `checkThreadInterrupted()`, but I neglected to revert this change.  

The unit tests I added did not hit this line tho, believe it or not, so I moved it elsewhere in the file where the unit test does hit it.

lang/src/org/partiql/lang/ast/ast.kt
dlurton: was trying to punt on that, since the discussion of what to use instead is longer than can fit in that comment.  will think of something else.

lang/src/org/partiql/lang/ast/ast.kt
dlurton: was trying to punt on that, since the discussion of what to use instead is longer than can fit in that comment. will think of something else.

lang/src/org/partiql/lang/ast/passes/AstRewriterBase.kt
```diff
@@ -31,8 +32,9 @@ interface AstRewriter {
...
 open class AstRewriterBase : AstRewriter {
 
-    override fun rewriteExprNode(node: ExprNode): ExprNode =
-        when (node) {
+    override fun rewriteExprNode(node: ExprNode): ExprNode {
+        checkThreadInterrupted()
```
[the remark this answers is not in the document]
dlurton: I think would have to see what that would look like in practice to offer any sort of strong opinion on the matter.  

In any case, I think it would require some modification of PIG, which is of course beyond the current scope.

lang/src/org/partiql/lang/ast/passes/AstVisitor.kt
```diff
@@ -27,7 +27,6 @@ import org.partiql.lang.ast.*
  *
  *  One `visit*` function is included for each base type in the AST.
  */
-@Deprecated("Use AstNode#iterator() or AstNode#children()")
```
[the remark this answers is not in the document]
dlurton: Well, I removed this one because it was providing information that contradicted the `@Deprecated` tags I added earlier in this PR. However, it should probably say: "Use org.lang.partiql.domains.PartiqlAst.Visitor instead"` instead, which will be in the next commit.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -257,7 +256,15 @@ internal class EvaluatingCompiler(
      */
     fun eval(ast: ExprNode, session: EvaluationSession): ExprValue = compile(ast).eval(session)
 
+    /**
+     * Compiles the specified [ExprNode] into a [ThunkEnv].
+     *
+     * This operation is deeply recursive.
```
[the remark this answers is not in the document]
dlurton: Properly doing justice to the "why" here is beyond the scope of what I'm trying to accomplish here (this function was entirely undocumented before, it's not like I'm making the situation worse by leaving it mostly undocumented), but some placeholder documentation seemed in order since leaving the thread interruption behavior undocumented seemed unwise.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -257,7 +256,15 @@ internal class EvaluatingCompiler(
...
+     *
+     * As a first step, throws [InterruptedException] if [Thread.interrupted] has been set.
```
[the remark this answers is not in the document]
dlurton: There's no I think if you `s/may/will` then that's accurate. In next commit.

lang/src/org/partiql/lang/eval/visitors/VisitorTransformBase.kt
dlurton: https://github.com/partiql/partiql-ir-generator/issues/72

dlurton: > Should implementors ever implement PartiqlAst.VisitorTransform? If not, we should add some comment to that effect there too.

Adding guidance to explain that customers should implement `VisitorTransformBase` instead of `PartiqlAst.VisitorTransform`.

> Better yet, do we need the separation or is it because the AST visitor is generated?

The separation is needed because the AST visitor is generated.  PIG would need to be updated to entirely avoid that.

lang/src/org/partiql/lang/ast/passes/AstRewriterBase.kt
dlurton: indeed. next rev.

dlurton:
> Can you also add an `interrupt` check in the function `validateTopLevelNodes` in `SqlParser`?

-- Added a check to `validateTopLevelNodes`

dlurton:
> LGTM overall. Only have a couple nits/suggestions. Also, should these checks also be added to the v0.1.5 tag @dlurton ?

That's next after this has been merged to `master`.

abhikuhikar (approved):
LGTM. Thanks.

almann (approved):
:shipit: For the future, I recommend that you try to avoid rebasing/squashing commits after you get feedback as it makes it impossible for the reviewer to do inter-diffs or only look at what you changed.

alancai98 (approved):
Noticed one small typo. Rest of changes LGTM

lang/src/org/partiql/lang/eval/visitors/VisitorTransformBase.kt
```diff
@@ -1,11 +1,23 @@
...
+import org.partiql.lang.util.checkThreadInterrupted
 
 /**
- * Base-class for visitor transforms that provides additional functions outside of [PartiqlAst.VisitorTransform].
+ * Base-class for visitor transforms that provides additional `transform*` functions that outside of
+ * the PIG-generated [PartiqlAst.VisitorTransform] class and adds a [Thread.interrupted] check
+ * to [transformExpr].
+ *
+ * All transforms should derive from this class instead of [PartiqlAst.VisitorTransform] so that they can
+ * be interrupted of they take a long time to process large ASTs.
```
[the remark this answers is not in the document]
dlurton: Fixed

alancai98 (approved):

dlurton:
> :shipit: For the future, I recommend that you try to avoid rebasing/squashing commits after you get feedback as it makes it impossible for the reviewer to do inter-diffs or only look at what you changed.

Yes, sorry, I realized that too late.

PR #402, opened 2021-05-05
Adds make_time ExprFunction

abhikuhikar:
*Issue #, if available:*

*Description of changes:*
 - Adds `make_time` `ExprFunction`


lang/src/org/partiql/lang/eval/builtins/MakeTimeExprFunction.kt
```diff
@@ -0,0 +1,82 @@
...
+            it.bigDecimalValue()
+        }
+
+        val tzMinutes: Int? = when (args.size) {
+            4 -> args[3].let {
+                it.validateType(ExprValueType.INT)
+                it.intValue()
+            }
+            else -> null
+        }
```
almann, line 61: Is it expected that I have to have exact types?  Should it be?

almann (commented):
Overall this seems good--though do we really need/want to have strict type requirements on numeric parameters?  I ask because I doubt the SQL implementation do, but we should verify this and be explicit about it.

lang/src/org/partiql/lang/eval/builtins/MakeTimeExprFunction.kt
abhikuhikar: According to Postgresql, yes. Although we might be able to allow `int` as well as `decimal` for `second`.
However, the others strictly need to be `int` as per Postgresql. 
I think our `make_time` is inspired from Postgresql solely and hence we should follow the semantics closely as we can.

docs/user/BuiltInFunctions.md
```diff
@@ -505,6 +505,39 @@ MAKE_DATE(missing, 02, 28)               -- null
 MAKE_DATE(2021, missing, 28)             -- null
 MAKE_DATE(2021, 02, missing)             -- null
 ```
+### MAKE_TIME
+
+Given the values for the hour (int), minute (int), second (BigDecimal) and optionally for timezone minutes (int) returns the associated time.
+
```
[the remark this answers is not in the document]
abhikuhikar: Postgresql throws an error in case the arguments do not match the required types.

almann (approved):
:ship:

almann (approved):

PR #403, opened 2021-05-07
Cherry-pick PR #398 to v0.1.6 branch and optimize test memory usage

dlurton:
Just like the title says, no new implementation, but there was changes to the unit tests:

For some reason I still don't fully understand, the tests added in PR #398 to the started consuming more than 4gb of RAM after being cherry-picked to the `v0.1.6` branch, which would cause the build fail since the JVM was limited to 4gb.  I tried increasing the RAM allocated to the JVM but wasn't able to get it to run correctly until 6gb and 7gb is the maximum amount of memory allowed by GitHub actions, which I know we will likely migrate to some day soon.  That was cutting it pretty close, so I was felt it was better to reduce memory consumption of the tests, so this PR includes changes that reduce the memory consumption of the tests.

### DIFF
--- a/lang/src/org/partiql/lang/ast/ast.kt
+++ b/lang/src/org/partiql/lang/ast/ast.kt
@@ -26,27 +26,14 @@ sealed class AstNode : Iterable<AstNode> {
     /**
      * Returns all the children nodes.
      *
-     * This property is [deprecated](see https://github.com/partiql/partiql-lang-kotlin/issues/396).  Use
-     * one of the following PIG-generated classes to analyze AST nodes instead:
-     *
-     * - [org.partiql.lang.domains.PartiqlAst.Visitor]
-     * - [org.partiql.lang.domains.PartiqlAst.VisitorFold]
      */
-    @Deprecated("DO NOT USE - see kdoc, see https://github.com/partiql/partiql-lang-kotlin/issues/396")
     abstract val children: List<AstNode>
 
     /**
      * Depth first iterator over all nodes.
      *
      * While collecting child nodes, throws [InterruptedException] if the [Thread.interrupted] flag has been set.
-     *
-     * This property is [deprecated](see https://github.com/partiql/partiql-lang-kotlin/issues/396).  Use
-     * one of the following PIG-generated classes to analyze AST nodes instead:
-     *
-     * - [org.partiql.lang.domains.PartiqlAst.Visitor]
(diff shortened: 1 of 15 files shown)


lang/src/org/partiql/lang/ast/ast.kt
```diff
@@ -24,19 +24,35 @@ import java.util.*
...
+     *
+     * While collecting child nodes, throws [InterruptedException] if the [Thread.interrupted] flag has been set.
+     *
+     * This property is [deprecated](see https://github.com/partiql/partiql-lang-kotlin/issues/396).  Use
+     * one of the following PIG-generated classes to analyze AST nodes instead:
+     *
+     * - [org.partiql.lang.domains.PartiqlAst.Visitor]
+     * - [org.partiql.lang.domains.PartiqlAst.VisitorFold]
      */
+    @Deprecated("DO NOT USE - see kdoc for alternatives")
```
alancai98, line 49: Should these be deprecated on the v0.1.* branches? I don't think PIG's `PartiqlAst` has been added on the v0.1.* branches yet.

lang/src/org/partiql/lang/ast/passes/AstVisitor.kt
```diff
@@ -27,7 +27,7 @@ import org.partiql.lang.ast.*
  *
  *  One `visit*` function is included for each base type in the AST.
  */
-@Deprecated("Use AstNode#iterator() or AstNode#children()")
+@Deprecated("Use org.lang.partiql.domains.PartiqlAst.Visitor instead")
```
alancai98, line 30: Similar to comment in `ast.kt`. Should this be included since `PartiqlAst` isn't part of the v0.1.* branches?

alancai98 (commented):
Just had a couple questions about deprecation messages mentioning PIG's `PartiqlAst`, which isn't currently part of the v0.1.* branches. Rest of the changes LGTM.

lang/src/org/partiql/lang/ast/ast.kt
dlurton: 1. Well, it won't break anyone.
2. These functions are likely to not be used by customers.
3. `PartiqlAst` does exist in this branch.

lang/src/org/partiql/lang/ast/passes/AstVisitor.kt
dlurton: same response as above.

alancai98 (approved):

PR #412, opened 2021-05-28
Adds comparison operators for DATE and TIME

### LINKED ISSUE #394: Compare equality of DATE and TIME for SQL 92 and java.time

abhikuhikar:
Resolves #394 
Adds comparison operations for `DATE` and `TIME` `ExprValues`.

### DIFF
--- a/lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
+++ b/lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
@@ -154,6 +154,7 @@ operator fun ExprValue.compareTo(other: ExprValue): Int {
 
 /**
  * Checks if the two ExprValues are directly comparable.
+ * Directly comparable is used in the context of the `<`/`<=`/`>`/`>=` operators.
  */
 internal fun ExprValue.isDirectlyComparableTo(other: ExprValue): Boolean =
     when {
--- a/lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
+++ b/lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
@@ -147,10 +147,21 @@ operator fun ExprValue.compareTo(other: ExprValue): Int {
     return when {
         type.isUnknown || other.type.isUnknown  ->
             throw EvaluationException("Null value cannot be compared: $this, $other", internal = false)
-        type.isDirectlyComparableTo(other.type) -> DEFAULT_COMPARATOR.compare(this, other)
-        else                                    -> errNoContext("Cannot compare values: $this, $other", internal = false)
+        isDirectlyComparableTo(other) -> DEFAULT_COMPARATOR.compare(this, other)
+        else                          -> errNoContext("Cannot compare values: $this, $other", internal = false)
     }
 }
+
+/**
+ * Checks if the two ExprValues are directly comparable.
+ */
+internal fun ExprValue.isDirectlyComparableTo(other: ExprValue): Boolean =
+    when {
+        // The ExprValue type for TIME and TIME WITH TIME ZONE is same
+        // and thus needs to be checked explicitly for the timezone values.
+        type == TIME && other.type == TIME -> timeValue().isDirectlyComparableTo(other.timeValue())
+        else -> type.isDirectlyComparableTo(other.type)
+    }
 
 /** Types that are cast to the [ExprValueType.isText] types by calling `IonValue.toString()`. */
 private val ION_TEXT_STRING_CAST_TYPES = setOf(BOOL, TIMESTAMP)
--- a/lang/src/org/partiql/lang/eval/NaturalExprValueComparators.kt
+++ b/lang/src/org/partiql/lang/eval/NaturalExprValueComparators.kt
@@ -32,6 +32,9 @@ import org.partiql.lang.util.*
  *    of precision or specific type.
  *      For `FLOAT` special values, `nan` comes before `-inf`, which comes before all normal
  *      numeric values, which is followed by `+inf`.
+ *  * `DATE` values follow and are compared by the date from earliest to latest.
+ *  * `TIME` values follow and are compared by the time of the day (point of time in a day of 24 hours)
+ *      from earliest to latest. Note that time without time zone is incomparable with time with time zone.
  *  * `TIMESTAMP` values follow and are compared by the point of time irrespective of precision and
  *    local UTC offset.
  *  * The [ExprValueType.isText] types come next ordered by their lexicographical ordering by
@@ -183,6 +186,26 @@ enum class NaturalExprValueComparators(private val nullOrder: NullOrder) : Compa
                     lVal.isZero() && rVal.isZero() -> return EQUAL // for negative zero
                     else -> return lVal.compareTo(rVal)
                 }
+            }
+        ) { return it }
+
+        // Date
+        ifCompared(
+            handle(lType == DATE, rType == DATE) {
+                val lVal = left.dateValue()
+                val rVal = right.dateValue()
+
+                return lVal.compareTo(rVal)
+            }
+        ) { return it }
+
+        // Time
+        ifCompared(
+            handle(lType == TIME, rType == TIME) {
+                val lVal = left.timeValue()
+                val rVal = right.timeValue()
+
+                return lVal.compareTo(rVal)
             }
         ) { return it }
 
--- a/lang/src/org/partiql/lang/eval/NaturalExprValueComparators.kt
+++ b/lang/src/org/partiql/lang/eval/NaturalExprValueComparators.kt
@@ -34,7 +34,8 @@ import org.partiql.lang.util.*
  *      numeric values, which is followed by `+inf`.
  *  * `DATE` values follow and are compared by the date from earliest to latest.
  *  * `TIME` values follow and are compared by the time of the day (point of time in a day of 24 hours)
- *      from earliest to latest. Note that time without time zone is incomparable with time with time zone.
+ *      from earliest to latest. Note that time without time zone is not directly comparable with time with time zone.
+ *      However, time without time zone value comes before time with time zone value when compared in the natural order.
  *  * `TIMESTAMP` values follow and are compared by the point of time irrespective of precision and
  *    local UTC offset.
  *  * The [ExprValueType.isText] types come next ordered by their lexicographical ordering by
@@ -205,7 +206,7 @@ enum class NaturalExprValueComparators(private val nullOrder: NullOrder) : Compa
                 val lVal = left.timeValue()
                 val rVal = right.timeValue()
 
-                return lVal.compareTo(rVal)
+                return lVal.naturalOrderCompareTo(rVal)
             }
         ) { return it }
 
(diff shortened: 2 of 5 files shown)


lang/src/org/partiql/lang/eval/time/Time.kt
```diff
@@ -180,4 +181,15 @@ data class Time private constructor(val localTime: LocalTime, val precision: Int
...
+    override fun compareTo(other: Time): Int {
+        return when {
+            // When the zone offsets are not null for both the operands, compare OffsetTime i.e. LocalTime with ZoneOffset
+            this.zoneOffset != null && other.zoneOffset != null -> this.offsetTime!!.compareTo(other.offsetTime)
+            // When the zone offsets are null for both the operands, compare just LocalTime
+            this.zoneOffset == null && other.zoneOffset == null -> this.localTime.compareTo(other.localTime)
+            // When one of the times is `time with time zone` and other is just `time` (without time zone), then they are incomparable.
+            else -> errNoContext("Cannot compare values: $this, $other", internal = false)
+        }
+    }
```
almann, line 194: So I believe that this is good for `<`/`>`/`<=`/`>=`/`=` operators, but not good for the natural comparator for `Time` values. 

Put another way, for the natural order it would be okay with all `Time` without time zone to compare before or after the `Time` with time zones, but I don't think it should be an error, otherwise `GROUP BY` and `ORDER BY` won't work.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
```diff
@@ -236,5 +236,52 @@ class EvaluatingCompilerDateTimeTests : EvaluatorTestBase() {
...
+                    // EvaluationException is thrown as expected, do nothing.
+                }
+            false -> {
+                val originalExprValue = eval(tc.query)
+                assertEquals(tc.expected, originalExprValue.toString())
+            }
+        }
+    }
+
+    data class ComparisonTestCase(val query: String, val expected: String?, val throws: Boolean)
```
almann, line 257: I would not make the `throws` a boolean, you should have it be an exception class or something more explicit.  Also, it is the case that one cannot have an `expected` non-null `String` **and** a `throws`, yes?  I would document that.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
```diff
@@ -236,5 +236,52 @@ class EvaluatingCompilerDateTimeTests : EvaluatorTestBase() {
...
+            case("TIME WITH TIME ZONE '12:12:13' > TIME WITH TIME ZONE '12:12:12.123'", "true"),
+            case("TIME WITH TIME ZONE '12:12:12.123+00:00' = TIME WITH TIME ZONE '12:12:12.123+00:00'", "true"),
+            case("TIME WITH TIME ZONE '12:12:12.123-08:00' > TIME WITH TIME ZONE '12:12:12.123+00:00'", "true"),
+            case("TIME WITH TIME ZONE '12:12:12.123-08:00' < TIME WITH TIME ZONE '12:12:12.123+00:00'", "false"),
+            case("CAST('12:12:12.123' AS TIME WITH TIME ZONE) = TIME WITH TIME ZONE '12:12:12.123'", "true"),
+            case("CAST(TIME WITH TIME ZONE '12:12:12.123' AS TIME) = TIME '12:12:12.123'", "true"),
+            case("TIME '12:12:13' < TIME WITH TIME ZONE '12:12:12.123'", throws = true),
+            case("TIME WITH TIME ZONE '12:12:13' < TIME '12:12:12.123'", throws = true),
+            case("TIME WITH TIME ZONE '12:12:13-08:00' < TIME '12:12:12.123-08:00'", throws = true),
+            case("TIME WITH TIME ZONE '12:12:13' > DATE '2012-02-29'", throws = true)
```
almann, line 283: I think you need test cases for `ORDER BY` and `GROUP BY` which leverage the comparator and have *different* behavior.

almann (changes requested):
I think we need to split out the comparison operator behavior from the natural order and have tests for `GROUP BY` and `ORDER BY` that explicitly test these differences.

lang/src/org/partiql/lang/eval/NaturalExprValueComparators.kt
```diff
@@ -32,6 +32,9 @@ import org.partiql.lang.util.*
  *    of precision or specific type.
  *      For `FLOAT` special values, `nan` comes before `-inf`, which comes before all normal
  *      numeric values, which is followed by `+inf`.
+ *  * `DATE` values follow and are compared by the date from earliest to latest.
+ *  * `TIME` values follow and are compared by the time of the day (point of time in a day of 24 hours)
+ *      from earliest to latest. Note that time without time zone is incomparable with time with time zone.
```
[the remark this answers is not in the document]
abhikuhikar: I have updated the comment. Also changed the logic to include the natural ordering.

lang/src/org/partiql/lang/eval/time/Time.kt
abhikuhikar: Got it. Thanks. I have removed the `Comparable` interface from which the `Time` class was derived and have included `naturalOrderCompareTo` function to include the natural ordering logic as well. The overridden `compareTo` method, in my opinion, should still throw an error when the `Time` instances are compared. However, since it is not used anywhere currently, I have decided to remove it for now.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
abhikuhikar: Makes sense. Generally we check for the `ErrorCode` when the `Exception` is thrown. However this error doesn't have one. I have removed the `throws` boolean and documented it well to indicate that the `null` `expected` string value throws an error.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
abhikuhikar: I have added the tests in `NaturalExprValueComparatorsTest.kt`.

lang/test/org/partiql/lang/eval/NaturalExprValueComparatorsTest.kt
```diff
@@ -364,7 +388,7 @@ class NaturalExprValueComparatorsTest : EvaluatorTestBase() {
     fun nonNullEqualityTests(equivalentPair: Pair<String, String>) {
         val (left, right) = equivalentPair
 
-        assertEval("$left = $right", "true")
+        assertExprEquals(valueFactory.newBoolean(true), eval("$left = $right"))
```
abhikuhikar, line 391: Since we have decided to not support `v0` AST serialization for `DATE` and `TIME` nodes, `assertEval` fails because it validates the serialization logic for `DATE` and `TIME`. I have used `assertExprEquals` in its place. What is the right thing to do here?

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
```diff
@@ -254,11 +254,16 @@ class EvaluatingCompilerDateTimeTests : EvaluatorTestBase() {
...
+     * [expected] is the expected value of the query.
+     * The [null] [expected] value indicates that the comparison test case throws an error.
+     */
+    data class ComparisonTestCase(val query: String, val expected: String?)
 
     private class ArgumentsForComparison : ArgumentsProviderBase() {
-        private fun case(query: String, expected: String) = ComparisonTestCase(query, expected, false)
-        private fun case(query: String, throws: Boolean) = ComparisonTestCase(query, null, throws)
+        private fun case(query: String, expected: String) = ComparisonTestCase(query, expected)
+        private fun case(query: String) = ComparisonTestCase(query, null)
```
almann, line 266: Minor, maybe spell this as `errorCase`

lang/test/org/partiql/lang/eval/NaturalExprValueComparatorsTest.kt
almann: One thing we might want to consider, though I might check with @dlurton, is somehow annotating these cases so we only do `assertExprEquals` in the cases where we don't test the serialization round trips.  That probably looks like changing how the tests are parameterized a bit, but probably the safest way to preserve legacy paths.

abhikuhikar: As discussed, we anyway want to get rid of the round trip tests.

abhikuhikar: To David's point, I think `assertExprEquals` should also suffice instead of asserting separately for value and type. Either way, I don't have any strong inclination towards either one.

lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
```diff
@@ -147,11 +147,22 @@ operator fun ExprValue.compareTo(other: ExprValue): Int {
...
             throw EvaluationException("Null value cannot be compared: $this, $other", internal = false)
-        type.isDirectlyComparableTo(other.type) -> DEFAULT_COMPARATOR.compare(this, other)
-        else                                    -> errNoContext("Cannot compare values: $this, $other", internal = false)
+        isDirectlyComparableTo(other) -> DEFAULT_COMPARATOR.compare(this, other)
+        else                          -> errNoContext("Cannot compare values: $this, $other", internal = false)
     }
 }
 
+/**
+ * Checks if the two ExprValues are directly comparable.
```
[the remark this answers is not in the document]
abhikuhikar: Changed. Thanks.

lang/src/org/partiql/lang/eval/time/Time.kt
```diff
@@ -46,7 +46,7 @@ internal const val MAX_PRECISION_FOR_TIME = 9
  *  Note that the [LocalTime] always stores the fractional part of the second in nanoseconds.
  *  It is up to the application developers to make use of the preserved [precision] value as need be.
  */
-data class Time private constructor(val localTime: LocalTime, val precision: Int, val zoneOffset: ZoneOffset? = null): Comparable<Time>{
+data class Time private constructor(val localTime: LocalTime, val precision: Int, val zoneOffset: ZoneOffset? = null) {
```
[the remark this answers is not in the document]
abhikuhikar: As discussed, we don't want the `compareTo` to behave this way.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
abhikuhikar: Changed.

almann (approved):
:ship:

dlurton (approved):

sunpeaksfivemile:
Is this sort what is going to ultimately be used to implement order by? 

If so does that mean that time, date, and timestamp don't sort according to how humans perceive time? 

If so that seems counter intuitive - perhaps we should coerce everything to a UTC timestamp and then sort off that.

I believe that Partiql spec section 12.2 is under specified since it only has a single type `timestamp` and ought to be clarified before order by is implemented

PR #420, opened 2021-06-15
Prevent the ORDER BY clause from being dropped in visitor transforms

### LINKED ISSUE #417: `ORDER BY` is dropped by three different VisitorTransforms
**If a user supplies an `ORDER BY` clause, it is simply ignored and they don't get an exception to tell them that the evaluator doesn't support `ORDER BY`.**

`ORDER BY` is dropped here:

- [SelectStarVisitorTransform](https://github.com/partiql/partiql-lang-kotlin/blob/7926b29535929fcd9db42993441e00a1bdb2f750/lang/src/org/partiql/lang/eval/visitors/SelectStarVisitorTransform.kt#L17)
- [VisitorTransformBase](https://github.com/partiql/partiql-lang-kotlin/blob/7926b29535929fcd9db42993441e00a1bdb2f750/lang/src/org/partiql/lang/eval/visitors/VisitorTransformBase.kt#L47)
- [GroupByPathExpressionVisitorTransform](https://github.com/partiql/partiql-lang-kotlin/blob/e0012b2a12099dd3251e1ab9aa5dab77ec593f03/lang/src/org/partiql/lang/eval/visitors/GroupByPathExpressionVisitorTransform.kt#L117)

In addition to fixing each of these, to verify that we're not dropping `ORDER BY` anywhere else, we should add an `EvaluatingCompilerTest` that ensures an  `ErrorCode.EVALUATOR_FEATURE_NOT_SUPPORTED_YET` is thrown.

dlurton:
Fixes #417

Dropping the `ORDER BY` clause prevents the `EvaluationException` from being thrown by `EvaluatingCompiler` due to not having evaluation support for `ORDER BY`, effectively ignoring the `ORDER BY` clause.

### DIFF
--- a/lang/src/org/partiql/lang/eval/visitors/GroupByPathExpressionVisitorTransform.kt
+++ b/lang/src/org/partiql/lang/eval/visitors/GroupByPathExpressionVisitorTransform.kt
@@ -98,19 +98,12 @@ class GroupByPathExpressionVisitorTransform(
 
         // The scope of the expressions in the FROM clause is the same as that of the parent scope.
         val from = this.transformExprSelect_from(node)
-
-        val fromLet = node.fromLet?.let { unshadowedTransformer.transformExprSelect_fromLet(node) }
-
-        val where = node.where?.let { unshadowedTransformer.transformExprSelect_where(node) }
-
-        val groupBy = node.group?.let { unshadowedTransformer.transformExprSelect_group(node) }
-
-        val having = node.having?.let { currentAndUnshadowedTransformer.transformExprSelect_having(node) }
-
-        val order = node.having?.let { currentAndUnshadowedTransformer.transformExprSelect_order(node) }
-
-        val limit = node.limit?.let { unshadowedTransformer.transformExprSelect_limit(node) }
-
+        val fromLet = unshadowedTransformer.transformExprSelect_fromLet(node)
+        val where = unshadowedTransformer.transformExprSelect_where(node)
+        val groupBy = unshadowedTransformer.transformExprSelect_group(node)
+        val having = currentAndUnshadowedTransformer.transformExprSelect_having(node)
+        val order = currentAndUnshadowedTransformer.transformExprSelect_order(node)
+        val limit = unshadowedTransformer.transformExprSelect_limit(node)
         val metas = unshadowedTransformer.transformExprSelect_metas(node)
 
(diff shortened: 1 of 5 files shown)


lang/src/org/partiql/lang/eval/visitors/GroupByPathExpressionVisitorTransform.kt
```diff
@@ -98,19 +98,12 @@ class GroupByPathExpressionVisitorTransform(
...
-
-        val fromLet = node.fromLet?.let { unshadowedTransformer.transformExprSelect_fromLet(node) }
-
-        val where = node.where?.let { unshadowedTransformer.transformExprSelect_where(node) }
-
-        val groupBy = node.group?.let { unshadowedTransformer.transformExprSelect_group(node) }
-
-        val having = node.having?.let { currentAndUnshadowedTransformer.transformExprSelect_having(node) }
-
-        val order = node.having?.let { currentAndUnshadowedTransformer.transformExprSelect_order(node) }
```
dlurton, line 110: Bug caused by:

```
`val order = node.having?...`
```

All the `?.let` calls were unnecessary here anyway too...

lang/src/org/partiql/lang/eval/visitors/SelectStarVisitorTransform.kt
```diff
@@ -22,6 +22,7 @@ class SelectStarVisitorTransform : VisitorTransformBase() {
                 where = node.where,
                 group = node.group,
                 having = node.having,
+                order = node.order,
```
dlurton, line 25: bug: this was omitted previously

lang/src/org/partiql/lang/eval/visitors/VisitorTransformBase.kt
```diff
@@ -40,7 +40,7 @@ abstract class VisitorTransformBase : PartiqlAst.VisitorTransform() {
         val having = transformExprSelect_having(node)
         val setq = transformExprSelect_setq(node)
         val project = transformExprSelect_project(node)
-        val order = node.having?.let { transformExprSelect_order(node) }
+        val order = transformExprSelect_order(node)
```
dlurton, line 43: similar to above.  bug:

```
val order = node.having?.let ...
```

almann (approved):
:ship:

PR #425, opened 2021-08-27
Draft PR for replacing ExprNode with PIG-generated PartiqlAst in EvaluatingCompiler

dlurton:
I did this over 2 days, mostly between meetings,  as an experiment to see how difficult it would be.

As it turns out, not that difficult, although this is not complete.  There are at least 3 distinct bugs causing a total of 26 `EvaluatingCompilerTests` to fail.  I also have not looked to see if at any of the other tests are failing.

Before this is completed, we have a few other things consider.

I mainly wanted to do this to show how much effort is actually involved.  I did this in ~4 hours in total.


dlurton:
Unfortunately, it looks like https://github.com/partiql/partiql-lang-kotlin/commit/e16f515926f459effc8dac0df2c1817a5d6994c3 introduced large amounts of formatting only changes to `EvaluatingCompiler` class and there are several hundreds of lines of conflicts in this PR for that class.  I will confer with PartiQL team folk on Monday about this, but unless those formatting changes can be reverted, it will be less work to start this over from scratch than it will be to try to resolve those conflicts.

PR #427, opened 2021-09-01
Setting up JMH for PartiQL

abhikuhikar:
*Description of changes:*
Sets up JMH for PartiQL


lang/src/jmh/kotlin/org/partiql/lang/PartiQLBenchmark.kt
```diff
@@ -0,0 +1,102 @@
...
+     */
+    @Benchmark
+    @BenchmarkMode(Mode.AverageTime)
+    @OutputTimeUnit(TimeUnit.MILLISECONDS)
+    fun testMethod2(state: MyState2) {
+        state.sum = state.a + state.b
+    }
+
+    /**
+     * Pass the final values to the [Blackhole] to fool JMH that it is indeed been used.
```
abhikuhikar, line 83: Should be `... to fool JVM...`

lang/src/jmh/kotlin/org/partiql/lang/PartiQLBenchmark.kt
```diff
@@ -0,0 +1,102 @@
...
+    @Benchmark
+    fun testMethod3(state: MyState, blackhole: Blackhole) {
+        val sum1 = state.a + state.b
+        val sum2 = state.a + state.a + state.b + state.b
+        blackhole.consume(sum1)
+        blackhole.consume(sum2)
+    }
+
+    /**
+     * Example PartiQL benchmark
```
abhikuhikar, line 94: Add one non-trivial end to end (parse and eval) benchmarks.

lang/build.gradle
```diff
@@ -54,6 +56,12 @@ dependencies {
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.7.0'
     testImplementation 'org.junit.vintage:junit-vintage-engine:5.7.0'
     testImplementation 'org.junit.jupiter:junit-jupiter-params:5.7.0'
+
+    jmh 'org.openjdk.jmh:jmh-core:0.9'
+    jmh 'org.openjdk.jmh:jmh-generator-annprocess:0.9'
+
+    implementation 'org.openjdk.jmh:jmh-core:0.9'
+    implementation 'org.openjdk.jmh:jmh-generator-annprocess:0.9'
```
dlurton, line 64: You're adding a dependency on JMH to the maven package, here.  We don't want that.  I think the reason this is required is because the benchmarks reside in under the `src` directory.

lang/src/jmh/kotlin/org/partiql/lang/PartiQLBenchmark.kt
```diff
@@ -0,0 +1,163 @@
+package org.partiql.lang
```
dlurton, line 1: This *must* not reside in this directory since that will be adding it to the `.jar` we publish to maven.  

This file should really reside in `lang/jmh/org/partiql/lang` instead of here.  To make this work, you need to modify the [`build.gradle`](https://github.com/partiql/partiql-lang-kotlin/blob/68e9f61e9a36090f2aeab80c560e786affa65845/build.gradle#L82-L92) in the root project by adding the lines noted below:

```
    plugins.withId('java', { _ ->
        sourceSets {
            main.java.srcDirs = ["src"]
            main.resources.srcDirs = ["resources"]
            test.java.srcDirs = ["test"]
            test.resources.srcDirs = ["test-resources"]

            // add these lines
            jmh.java.srcDirs = ["jmh"]
            jmh.resources.srcDirs = ["jmh-resources"]
        }        
    })
```

This may not be 100% exactly right but I'm certain something like this will be needed.

dlurton (changes requested):

lang/src/jmh/kotlin/org/partiql/lang/PartiQLBenchmark.kt
```diff
@@ -0,0 +1,163 @@
...
+ * Refer this [JMH tutorial](http://tutorials.jenkov.com/java-performance/jmh.html) for more information on [Benchmark]s,
+ * [BenchmarkMode]s, etc.
+ */
+open class PartiQLBenchmark {
+    @State(Scope.Thread)
+    open class MyState {
+        var a = 1
+        var b = 2
+        var sum = 0
+    }
```
dlurton, line 30: I'm sure this was helpful for getting it working on your machine but it doesn't belong in `main`.

lang/src/jmh/kotlin/org/partiql/lang/PartiQLBenchmark.kt
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * Use [State] objects to initialize some variables that your benchmark code needs,
+     * but which you do not want to be part of the code your benchmark measures.
+     */
+    @Benchmark
+    fun testMethod(state: MyState) {
+        // This is a demo/sample template for building your JMH benchmarks. Edit as needed.
+        // Put your benchmark code here.
+        state.sum = state.a + state.b
+    }
```
dlurton, line 41: I'm sure this was helpful for getting it working on your machine but it doesn't belong in `main`.

lang/src/jmh/kotlin/org/partiql/lang/PartiQLBenchmark.kt
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        @TearDown(Level.Trial)
+        fun doTearDown() {
+            println("Do TearDown")
+        }
+
+        var a = 1
+        var b = 2
+        var sum = 0
```
dlurton, line 68: I'm sure this was helpful for getting it working on your machine but it doesn't belong in `main`.

lang/src/jmh/kotlin/org/partiql/lang/PartiQLBenchmark.kt
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * You can use different [BenchmarkMode]. The benchmark mode tells JMH what you want to measure.
+     * [TimeUnit] specifies what time units you want the benchmark results printed in.
+     */
+    @Benchmark
+    @BenchmarkMode(Mode.AverageTime)
+    @OutputTimeUnit(TimeUnit.MILLISECONDS)
+    fun testMethod2(state: MyState2) {
+        state.sum = state.a + state.b
+    }
```
dlurton, line 121: I'm sure this was helpful for getting it working on your machine but it doesn't belong in `main`.

lang/src/jmh/kotlin/org/partiql/lang/PartiQLBenchmark.kt
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * Pass the final values to the [Blackhole] to fool JMH that it is indeed been used.
+     */
+    @Benchmark
+    fun testMethod3(state: MyState, blackhole: Blackhole) {
+        val sum1 = state.a + state.b
+        val sum2 = state.a + state.a + state.b + state.b
+        blackhole.consume(sum1)
+        blackhole.consume(sum2)
+    }
```
dlurton, line 132: I'm sure this was helpful for getting it working on your machine but it doesn't belong in `main`.

lang/src/jmh/kotlin/org/partiql/lang/PartiQLBenchmark.kt
abhikuhikar: Removed.

lang/src/jmh/kotlin/org/partiql/lang/PartiQLBenchmark.kt
abhikuhikar: I created the `jmh` sub-project but for some reason the benchmarks need to be under `./..src/jmh/kotlin/..` folder structure. I tried some other directory structures but the benchmark list wasn't populated for any other directory structure.

dlurton (dismissed):
See the commit I added.  No need for a separate project, just need to change the default location for the `jmh` source set that's added by the JMH plugin.

alancai98 (approved):

dlurton (approved):

Issue #367, opened 2021-02-24
Consider using LazyThreadSafetyMode.PUBLICATION instead of LazyThreadSafteyMode.NONE 


dlurton:
Our API exposes a number of functions that behind the scenes use `LazyThreadSafetyMode.NONE`.  Unfortunately, this "no thread safety" *literally* means that accessing it from multiple threads can cause an NPE.  I guess we assumed this would be written such that it could be initialized by multiple threads without  throwing an NPE, with race conditions only possibly occurring inside the initializer block. 

There are some parts of the API which might be thread-safe but for this simple problem.  The problem lies within the `if` block below.   Note that the implementation of `Lazy<T>` for `LazyThreadSafetyMode.PUBLICATION` has similar code factored to avoid the NPE.

This is Kotlin's implementation of `lazy` for `LazyThreadSafetyMode.NONE`:

```
internal class UnsafeLazyImpl<out T>(initializer: () -> T) : Lazy<T>, Serializable {
    private var initializer: (() -> T)? = initializer
    private var _value: Any? = UNINITIALIZED_VALUE

    override val value: T
        get() {
            if (_value === UNINITIALIZED_VALUE) {
                _value = initializer!!()  
                initializer = null
            }
            @Suppress("UNCHECKED_CAST")
            return _value as T
        }

    override fun isInitialized(): Boolean = _value !== UNINITIALIZED_VALUE

    override fun toString(): String = if (isInitialized()) value.toString() else "Lazy value not initialized yet."

    private fun writeReplace(): Any = InitializedLazyImpl(value)
}
```

alancai98:
Not sure if this should also be fixed for PIG since PartiqlAst's hashcode also uses `LazyThreadSafetyMode.NONE`. ([link to the free marker template](https://github.com/partiql/partiql-ir-generator/blob/7e9ae0f028ecc3dd50f6438350a4bcff1f004945/pig/resources/org/partiql/pig/templates/kotlin.ftl#L117))

PR #433, opened 2021-09-07
Use LazyThreadSafteyMode.PUBLICATION instead of NONE

dlurton:
Issue #367 

LazyThreadSafetyMode.NONE literally means that accessing the lazy value from multiple threads can cause an NPE.

Unfortunately, we assumed NONE would be implemented such that a lazy value could be initialized by multiple
threads without throwing an NPE, with race conditions only possibly occurring inside the initializer block.
However, that is actually the behavior is for LazyThreadSafeteyMode.PUBLICATION.


alancai98 (approved):

dlurton:
FWIW, without this fix I was able to reproduce the issue ~10% of the time with [this code](https://gist.github.com/dlurton/5bc62feba6eba5c1689f505ab5cf05f3), but was never able to reproduce it with the fix.

dlurton:
I should add; given the inconsistency of the fix, I don't think we should add a test for it.

Issue #410, opened 2021-05-18
Provide an option to configure default time zone


abhikuhikar:
Provide an option (either via `CompileOptions` or some such) to configure the default time zone.

PR #449, opened 2021-09-28
Configure default timezone

lziq:
*Issue #410*

### Solution Description:
We want to provide an option to configure default timezone offset through `EvaluationSession` class, since the default timezone offset should be considered as a kind of information provided when we evaluate the time. 

#### Source code: 
1. Added the new attribute `defaultTimezoneOffset` to `EvaluationSession` class. The default value is set to be UTC time. 
2. Removed the global field `DEFAULT_TIMEZONE_OFFSET`. 
3. Modified the source code for time evaluation and time cast, so that the default timezone is now provided by `EvaluationSession`. 
#### Test: 
1. Added test for `defaultTimezoneOffset` for `EvaluationSession` class. 
2. Added test for time evaluation. 
3. Added test for time cast. 
4. Created a local variable `defaultTimezoneOffset` to replace `DEFAULT_TIMEZONE_OFFSET` for test cases using it.


lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
```diff
@@ -227,7 +227,8 @@ private val genericTimeRegex = Regex("\\d\\d:\\d\\d:\\d\\d(\\.\\d*)?([+|-]\\d\\d
 fun ExprValue.cast(
     targetDataType: DataType,
     valueFactory: ExprValueFactory,
-    locationMeta: SourceLocationMeta?
+    locationMeta: SourceLocationMeta?,
+    env: Environment
```
dlurton, line 231: I think you should be passing `EvaluationSession` here, not `Environment`.  `Environment` has a lot of details related to the current state of the interpreter that aren't relevant to casting.  Generally, we want to avoid sharing those details as much as reasonably possible.

dlurton (changes requested):
Mostly looks good to me, just one minor change requested. Thanks.

lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
lziq: Made the change in the new commit.

lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
```diff
@@ -227,7 +227,8 @@ private val genericTimeRegex = Regex("\\d\\d:\\d\\d:\\d\\d(\\.\\d*)?([+|-]\\d\\d
 fun ExprValue.cast(
     targetDataType: DataType,
     valueFactory: ExprValueFactory,
-    locationMeta: SourceLocationMeta?
+    locationMeta: SourceLocationMeta?,
+    session: EvaluationSession
```
alancai98, line 231: Would be good to add some kdoc for `session`

lang/test/org/partiql/lang/eval/EvaluatingCompilerCastTest.kt
```diff
@@ -822,22 +837,45 @@ class EvaluatingCompilerCastTest : EvaluatorTestBase() {
...
+            case("""`'23:12:12.1267'`""", "TIME (2) WITH TIME ZONE", "23:12:12.13${defaultTimezoneOffset.getOffsetHHmm()}"),
             case("""`'23:12:12.1267'`""", "TIME", "23:12:12.1267"),
             case("""`'23:12:12.1267'`""", "TIME (2)", "23:12:12.13"),
-            case("""`'23:12:12.1267'`""", "TIME WITH TIME ZONE", "23:12:12.1267${DEFAULT_TIMEZONE_OFFSET.getOffsetHHmm()}"),
-            case("""`'23:12:12.1267'`""", "TIME (2) WITH TIME ZONE", "23:12:12.13${DEFAULT_TIMEZONE_OFFSET.getOffsetHHmm()}")
+            case("""`'23:12:12.1267'`""", "TIME WITH TIME ZONE", "23:12:12.1267${defaultTimezoneOffset.getOffsetHHmm()}"),
+            case("""`'23:12:12.1267'`""", "TIME (2) WITH TIME ZONE", "23:12:12.13${defaultTimezoneOffset.getOffsetHHmm()}")
+        ),
+        // Configuring default timezone offset through EvaluationSession
+        listOf(
```
alancai98, line 852: Nice collection of tests!

lang/test/org/partiql/lang/eval/EvaluatingCompilerCastTest.kt
```diff
@@ -921,7 +960,11 @@ class EvaluatingCompilerCastTest : EvaluatorTestBase() {
                 assertEquals(castCase.expectedErrorCode, p.errorCode)
             }
         }
-        else -> assertEquals(castCase.expected, eval(castCase.expression).toString())
+        else -> assertEquals(castCase.expected,
+                             eval(source = castCase.expression,
+                                  session = castCase.session
+                             ).toString()
+        )
```
alancai98, line 967: nit: I find this formatting slightly confusing. Consider:

```suggestion
        else -> assertEquals(
            castCase.expected,
            eval(source = castCase.expression, session = castCase.session).toString()
        )
```

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
```diff
@@ -56,12 +56,13 @@ class EvaluatingCompilerDateTimeTests : EvaluatorTestBase() {
...
 
-    data class TimeTestCase(val query: String, val expected: String, val expectedTime: TimeForValidation? = null)
+    data class TimeTestCase(val query: String, val expected: String, val expectedTime: TimeForValidation? = null, val session: EvaluationSession = EvaluationSession.standard())
 
     @ParameterizedTest
     @ArgumentsSource(ArgumentsForTimeLiterals::class)
     fun testTime(tc: TimeTestCase)  {
-        val originalExprValue = eval(tc.query)
+        val originalExprValue = eval(source = tc.query,
+                                     session = tc.session)
```
alancai98, line 65: nit: could put on one line:
```suggestion
        val originalExprValue = eval(source = tc.query, session = tc.session)
```

lang/test/org/partiql/lang/eval/EvaluationSessionTest.kt
```diff
@@ -52,8 +57,21 @@ class EvaluationSessionTest {
...
 
-        assertEquals(globals, block.invoke().globals)
+        assertEquals(globals, session.globals)
+        assertEquals(ZoneOffset.UTC, session.defaultTimezoneOffset)
+        assertNow(block)
+    }
+
+    @Test
+    fun settingDefaultTimezoneOffset() {
+        val defaultTimezoneOffset = ZoneOffset.UTC
```
alancai98, line 69: For this test, I would put a different `ZoneOffset` other than the default of UTC to check that it's actually getting set.

lang/test/org/partiql/lang/eval/EvaluationSessionTest.kt
```diff
@@ -18,11 +18,16 @@ import com.amazon.ion.*
 import org.partiql.lang.util.*
 import org.junit.*
 import org.junit.Test
+import java.time.ZoneOffset
+import java.util.*
```
alancai98, line 22: nit: unused import. Also we prefer to not use wildcard imports

alancai98 (changes requested):
Only had a few minor comments. Looks good overall!

abhikuhikar (commented):
`EvaluationSession` class also has a `now: Timestamp` member. It can have its own time zone offset defined. I was wondering if it would conflict with a separate default time zone. Should we consider using `now.localOffset` which is time zone offset in minutes? (It can be `null` though). Need to investigate more into this. 

For the PR, just one minor suggestion. Otherwise looks great to me.

lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
```diff
@@ -110,7 +112,11 @@ class EvaluatingCompilerDateTimeTests : EvaluatorTestBase() {
             case("TIME (2) '01:01:12' IS TIME", "true"),
             case("TIME WITH TIME ZONE '12:25:12.123456' IS TIME", "true"),
             case("TIME (2) WITH TIME ZONE '01:01:12' IS TIME", "true"),
-            case("'01:01:12' IS TIME", "false")
+            case("'01:01:12' IS TIME", "false"),
+            case("TIME WITH TIME ZONE '00:00:00'", "00:00:00-01:00", TimeForValidation(0,0,0,0,0, -60), EvaluationSession.build {defaultTimezoneOffset(ZoneOffset.ofHours(-1))}),
+            case("TIME WITH TIME ZONE '11:23:45.678'", "11:23:45.678+06:00", TimeForValidation(11,23,45,678000000,3, 360), EvaluationSession.build {defaultTimezoneOffset(ZoneOffset.ofHours(6))}),
+            case("TIME WITH TIME ZONE '11:23:45.678-05:30'", "11:23:45.678-05:30", TimeForValidation(11,23,45,678000000,3, -330), EvaluationSession.build {defaultTimezoneOffset(ZoneOffset.ofHours(6))}),
+            case("TIME (2) WITH TIME ZONE '12:59:59.13456'", "12:59:59.13-05:30", TimeForValidation(12, 59, 59, 130000000, 2, -330), EvaluationSession.build {defaultTimezoneOffset(ZoneOffset.ofHoursMinutes(-5, -30))})
```
abhikuhikar, line 119: You can use a similar helper function here as well.

lang/test/org/partiql/lang/eval/EvaluatingCompilerCastTest.kt
```diff
@@ -822,22 +839,45 @@ class EvaluatingCompilerCastTest : EvaluatorTestBase() {
...
+            // CAST(<text> AS <TIME WITH TIME ZONE>)
+            case("'23:12:12.1267'", "TIME WITH TIME ZONE", "23:12:12.1267+00:00", buildSession()),
+            case("'23:12:12.1267'", "TIME WITH TIME ZONE", "23:12:12.1267+11:00", buildSession(11)),
+            case("'23:12:12.1267'", "TIME (2) WITH TIME ZONE", "23:12:12.13-05:30", buildSession(-5, -30)),
+            case("""`'23:12:12.1267'`""", "TIME WITH TIME ZONE", "23:12:12.1267+00:00", buildSession()),
+            case("""`'23:12:12.1267'`""", "TIME WITH TIME ZONE", "23:12:12.1267+11:00", buildSession(11)),
+            case("""`'23:12:12.1267'`""", "TIME (2) WITH TIME ZONE", "23:12:12.13-05:30", buildSession(-5, -30)),
+            case("""`"23:12:12.1267"`""", "TIME WITH TIME ZONE", "23:12:12.1267+00:00", buildSession()),
+            case("""`"23:12:12.1267"`""", "TIME WITH TIME ZONE", "23:12:12.1267+11:00", buildSession(11)),
+            case("""`"23:12:12.1267"`""", "TIME (2) WITH TIME ZONE", "23:12:12.13-05:30", buildSession(-5, -30))
```
abhikuhikar, line 874: Looks much cleaner. Thanks.

abhikuhikar (changes requested):

abhikuhikar (approved):

alancai98 (approved):
Nice job :shipit:

dlurton (approved):

PR #450, opened 2021-09-29
Fixes structs to handle non-text struct fields.

### LINKED ISSUE #355: Prevent structs with non-text struct keys from being created
So, this is odd:
```
PartiQL> { 1: 42 }
   | 
===' 
{
  'null': 42
}
--- 
OK!
```
1. An integer as a field key?
1. Why does it show null there instead of 1?

Additionally...

```
val strangeStruct = eval("{1: 2}")
val strangeField = strangeStruct.first().name!!
println(strangeField.type)
println(strangeField.toString())
```

outputs:
```
INT
1
```
Lastly:

```
PartiQL> { 1: 42 }[1]
   | 
org.partiql.lang.eval.EvaluationException: Expected text: 1
	<UNKNOWN>: at line 1, column 11: <UNKNOWN>
```

So it would appear that we allow structs to be created with non-text field names but then once created the resulting value is more-or-less unusable.  

I think we need to add an evaluation-time check that will raise an error if an attempt is made to construct a struct with a non-text, null or missing key.

abhikuhikar:
* Resolves #355 

### DIFF
--- a/lang/src/org/partiql/lang/errors/ErrorCode.kt
+++ b/lang/src/org/partiql/lang/errors/ErrorCode.kt
@@ -649,6 +649,11 @@ enum class ErrorCode(private val category: ErrorCategory,
             "LIMIT value must be an integer but found ${errorContext.getProperty(Property.ACTUAL_TYPE)}}"
     },
 
+    EVALUATOR_NON_TEXT_STRUCT_FIELD (
+        ErrorCategory.EVALUATOR,
+        LOCATION,
+        "Struct field should be text."),
+
     EVALUATOR_NEGATIVE_LIMIT(
         ErrorCategory.EVALUATOR,
         LOCATION,
--- a/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
+++ b/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
@@ -868,11 +868,30 @@ internal class EvaluatingCompiler(
 
         val fieldThunks = fields.map {
             val (nameExpr, valueExpr) = it
+            if (nameExpr is Literal && !nameExpr.ionValue.isText) {
+                // Compile time error. Fail before the StructFieldThunks are evaluated.
+                err("Found struct field to be of type ${nameExpr.ionValue.type}",
+                    ErrorCode.EVALUATOR_NON_TEXT_STRUCT_FIELD,
+                    errorContextFrom(nameExpr.metas.sourceLocationMeta),
+                    internal = false
+                )
+            }
             StructFieldThunks(compileExprNode(nameExpr), compileExprNode(valueExpr))
         }
 
         return thunkFactory.thunkEnv(metas) { env ->
-            val seq = fieldThunks.map { it.valueThunk(env).namedValue(it.nameThunk(env)) }.asSequence()
+            val seq = fieldThunks.map {
+                val nameValue = it.nameThunk(env)
+                if (!nameValue.type.isText) {
+                    // Evaluation time error where variable reference might be evaluated to non-text struct field.
+                    err("Found struct field to be of type ${nameValue.type}",
+                        ErrorCode.EVALUATOR_NON_TEXT_STRUCT_FIELD,
+                        errorContextFrom(metas.sourceLocationMeta),
+                        internal = false
+                    )
+                }
+                it.valueThunk(env).namedValue(nameValue)
+            }.asSequence()
             createStructExprValue(seq, StructOrdering.ORDERED)
         }
     }
--- a/lang/test/org/partiql/lang/eval/EvaluatingCompilerExceptionsTest.kt
+++ b/lang/test/org/partiql/lang/eval/EvaluatingCompilerExceptionsTest.kt
@@ -242,4 +242,37 @@ class EvaluatingCompilerExceptionsTest : EvaluatorTestBase() {
                 Property.COLUMN_NUMBER to 11L,
                 Property.BINDING_NAME to "leading")
         )
+
+    @Test
+    fun intAsNonTextStructField() =
+        checkInputThrowingEvaluationException(
+            "{'apple' : 1, `banana` : 2, 1 : 3}",
+            ErrorCode.EVALUATOR_NON_TEXT_STRUCT_FIELD,
+            mapOf(
+                Property.LINE_NUMBER to 1L,
+                Property.COLUMN_NUMBER to 29L
+            )
+        )
+
+    @Test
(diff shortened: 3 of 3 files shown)


lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -868,11 +868,30 @@ internal class EvaluatingCompiler(
 
         val fieldThunks = fields.map {
             val (nameExpr, valueExpr) = it
+            if (nameExpr is Literal && !nameExpr.ionValue.isText) {
+                // Compile time error. Fail before the StructFieldThunks are evaluated.
+                err("Found struct field to be of type ${nameExpr.ionValue.type}",
+                    ErrorCode.EVALUATOR_NON_TEXT_STRUCT_FIELD,
```
abhikuhikar, line 874: Throw `SEMANTIC` exception here

lang/src/org/partiql/lang/errors/ErrorCode.kt
```diff
@@ -649,6 +649,11 @@ enum class ErrorCode(private val category: ErrorCategory,
             "LIMIT value must be an integer but found ${errorContext.getProperty(Property.ACTUAL_TYPE)}}"
     },
 
+    EVALUATOR_NON_TEXT_STRUCT_FIELD (
```
alancai98, line 652: nit: extra space before left paren

lang/src/org/partiql/lang/errors/ErrorCode.kt
```diff
@@ -649,6 +649,11 @@ enum class ErrorCode(private val category: ErrorCategory,
...
+        ErrorCategory.EVALUATOR,
+        LOCATION,
+        "Struct field should be text."),
```
alancai98, line 655: Is it more accurate to say struct field *names*/*keys* rather than struct field? A struct field has a name/key (restricted to text) and a field (could be any value).

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -868,11 +868,30 @@ internal class EvaluatingCompiler(
 
         val fieldThunks = fields.map {
             val (nameExpr, valueExpr) = it
+            if (nameExpr is Literal && !nameExpr.ionValue.isText) {
+                // Compile time error. Fail before the StructFieldThunks are evaluated.
+                err("Found struct field to be of type ${nameExpr.ionValue.type}",
```
alancai98: nit: similar to comment in `ErrorCode.kt`, error message should say struct field *name*/*key*. Same w/ on L887

lang/test/org/partiql/lang/eval/EvaluatingCompilerExceptionsTest.kt
```diff
@@ -242,4 +242,37 @@ class EvaluatingCompilerExceptionsTest : EvaluatorTestBase() {
...
+        )
+
+    @Test
+    fun variableReferenceToIntAsNonTextStructField() =
+        checkInputThrowingEvaluationException(
+            "SELECT {a : 2} FROM {'a' : 1}",
+            ErrorCode.EVALUATOR_NON_TEXT_STRUCT_FIELD,
+            mapOf(
+                Property.LINE_NUMBER to 1L,
+                Property.COLUMN_NUMBER to 8L
```
alancai98, line 253: There's some slight inconsistency with the error location. If the non-text struct field name is caught at compile time (as in the previous two tests), the error points to the field name. Whereas, in this case it points to the start of the struct constructor (L8). I think both cases should be consistent with the error pointing to the problematic struct field name.

lang/test/org/partiql/lang/eval/EvaluatingCompilerExceptionsTest.kt
```diff
@@ -242,4 +242,37 @@ class EvaluatingCompilerExceptionsTest : EvaluatorTestBase() {
...
+    fun timestampAsNonTextStructField() =
+        checkInputThrowingEvaluationException(
+            "{`2007-02-23T12:14Z` : 3}",
+            ErrorCode.EVALUATOR_NON_TEXT_STRUCT_FIELD,
+            mapOf(
+                Property.LINE_NUMBER to 1L,
+                Property.COLUMN_NUMBER to 2L
+            )
+        )
+
```
alancai98: May want to also add some tests with `NULL` and `MISSING` as the struct field name.

lang/src/org/partiql/lang/errors/ErrorCode.kt
alancai98: Also, I don't think the `messagePrefix` argument of `ErrorCode` should be used. It can lead to a misconfigured error message:

```
PartiQL> { 1 : 2 };
org.partiql.lang.eval.EvaluationException: Found struct field to be of type INT
	Evaluator Error: at line 1, column 3: Struct field should be text., <UNKNOWN> : <UNKNOWN>

	at org.partiql.lang.eval.ExceptionsKt.err(Exceptions.kt:51)
        ...
```

Overriding `getErrorMessage` and leaving `messagePrefix` as an empty string should get rid of the extra `, <UNKNOWN> : <UNKNOWN>` at the end.

alancai98 (changes requested):
I have some minor suggestions to adjust the error message and error location. Also may need a couple more tests with different struct field name types.

lang/test/org/partiql/lang/eval/EvaluatingCompilerExceptionsTest.kt
abhikuhikar: `metas` are not available with the `ExprValue` once it is evaluated.

lang/test/org/partiql/lang/eval/ScratchPad.kt
```diff
@@ -0,0 +1,8 @@
+package org.partiql.lang.eval
```
alancai98, line 1: File not needed

alancai98 (commented):
Looks good. Only have a couple small nits and an unused file.

alancai98:
For future tracking: when the evaluator is to support a permissive typing mode, these fields with a non-string struct key would be ignored. From the [spec](https://partiql.org/assets/PartiQL-Specification.pdf#page=25)
>**Treatment of mistyped attribute names** It is possible that an expression a_I that computes an attribute name
results into a non-string, i.e., a value that is not a legitimate attribute name. In such cases, under the permissive mode
the attribute-value pair will be dismissed. Under the type checking mode the query will fail.

Issue #422, opened 2021-06-17
Add support for `OFFSET` 


alancai98:
We currently have `LIMIT` supported and parser support for `ORDER BY`. Would be nice to also support [`OFFSET`](https://www.postgresql.org/docs/13/queries-limit.html). For real usability, should also have `ORDER BY` implemented in the evaluator (#47) first.

Implementing `OFFSET` could be a good introduction to the lexer, parser, evaluator, and PIG.

PR #453, opened 2021-10-05
Add offset parser

lziq:
*Issue #422, task 2: Implement `OFFSET` in lexer, parser, and `ExprNode` conversions*

*Description of changes:*
1. `ExprNode` representation of `OFFSET` 
2. Conversion between `PartiqlAst` and `ExprNode` for `OFFSET` 
3. V0 Ast Serliazation for `OFFSET` 
4. Parsing of `OFFSET` in Parser
4. Other minor changes to make code able to be compiled 
5. Test cases for `OFFSET` in Parser

### DIFF
--- a/lang/test/org/partiql/lang/syntax/SqlParserTest.kt
+++ b/lang/test/org/partiql/lang/syntax/SqlParserTest.kt
@@ -15,8 +15,10 @@
 package org.partiql.lang.syntax
 
 import com.amazon.ion.Decimal
-import com.amazon.ion.IonInt
-import com.amazon.ionelement.api.*
+import com.amazon.ionelement.api.ionDecimal
+import com.amazon.ionelement.api.ionInt
+import com.amazon.ionelement.api.ionString
+import com.amazon.ionelement.api.loadSingleElement
 import org.junit.Ignore
 import org.junit.Test
 import org.partiql.lang.ast.ExprNode
--- a/lang/test/org/partiql/lang/syntax/SqlParserTest.kt
+++ b/lang/test/org/partiql/lang/syntax/SqlParserTest.kt
@@ -3990,6 +3990,17 @@ class SqlParserTest : SqlParserTestBase() {
         )
     }
 
+    @Test
+    fun selectOrderbyLimitOffsetTest() = assertExpression("SELECT x FROM a ORDER BY y DESC LIMIT 10 OFFSET 5") {
+        select(
+            project = buildProject("x"),
+            from = scan(id("a")),
+            order = PartiqlAst.OrderBy(listOf(PartiqlAst.SortSpec(id("y"), PartiqlAst.OrderingSpec.Desc()))),
+            limit = buildLit("10"),
+            offset = buildLit("5")
+        )
+    }
+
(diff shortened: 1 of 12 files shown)


lang/test/org/partiql/lang/syntax/SqlParserTest.kt
```diff
@@ -3951,6 +3950,44 @@ class SqlParserTest : SqlParserTestBase() {
...
+    fun selectWhereLimitOffsetTest() = assertExpression("SELECT x FROM a WHERE y = 10 LIMIT 7 OFFSET 5") {
+        select(
+            project = buildProject("x"),
+            from = scan(id("a")),
+            where = PartiqlAst.Expr.Eq(listOf(id("y"), buildLit("10"))),
+            limit = buildLit("7"),
+            offset = buildLit("5")
+        )
+    }
+
```
alancai98, line 3990: Would be good to add a test using `ORDER BY`, which will be the most common use case when using `OFFSET`

lang/test/org/partiql/lang/syntax/SqlParserTest.kt
```diff
@@ -3951,6 +3950,44 @@ class SqlParserTest : SqlParserTestBase() {
         )
     }
 
+    //****************************************
+    // OFFSET clause parsing
+    //****************************************
```
alancai98, line 3955: These are some successful `OFFSET` parsing tests. May want to also add a few `OFFSET` parsing error tests. Consider:
- `OFFSET` with no subsequent argument
- `OFFSET` before `LIMIT`
- etc.

lang/test/org/partiql/lang/syntax/SqlParserTest.kt
```diff
@@ -15,9 +15,8 @@
 package org.partiql.lang.syntax
 
 import com.amazon.ion.Decimal
-import com.amazon.ionelement.api.ionDecimal
-import com.amazon.ionelement.api.ionInt
-import com.amazon.ionelement.api.ionString
+import com.amazon.ion.IonInt
+import com.amazon.ionelement.api.*
```
alancai98, line 19: nit: as per our [style guide](https://github.com/partiql/partiql-lang-kotlin/blob/main/docs/dev/CODE-STYLE.md#imports), try not to use star imports

alancai98 (commented):
Nice job, looks good overall! Would only suggest adding a few more tests and a minor import nit.

lang/test/org/partiql/lang/syntax/SqlParserTest.kt
```diff
@@ -3990,6 +3990,17 @@ class SqlParserTest : SqlParserTestBase() {
         )
     }
 
+    @Test
+    fun selectOrderbyLimitOffsetTest() = assertExpression("SELECT x FROM a ORDER BY y DESC LIMIT 10 OFFSET 5") {
```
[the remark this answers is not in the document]
lziq: Actually according to (section 12 https://partiql.org/assets/PartiQL-Specification.pdf), it should be upper-case.

PR #462, opened 2021-10-26
Better SqlParserTestBase 

lziq:
This PR aims to better the code in file 'SqlParserTestBase'. 

### Description of Changes: 
1. Changed variables name. e.g. originally, for actual values, some variables use `actual` or `parsed` as prefix, while some do not have prefix. Now change them into a uniform prefix `actual`. 
2. Removed unuseful code. 
3. Refactored the code in `assertExpression`.

PR #464, opened 2021-10-28
Deprecate `ExprNode` in `Parser`

lziq:
*This PR aims to deprecate `ExprNode` in `Parser`.*

*Subtasks:*
- [x] Implement `toAstStatement` method to directly transform `ParseNode` to PIG AST, and remove `toExprNode` method in `Parser`. 
&nbsp; PR: https://github.com/partiql/partiql-lang-kotlin/pull/460 
&nbsp; PR: https://github.com/partiql/partiql-lang-kotlin/pull/469 
- [x] Change the test framework so that it uses PIG AST generated directly from `toAstStatement` method. 
&nbsp; PR: https://github.com/partiql/partiql-lang-kotlin/pull/462
&nbsp; PR: https://github.com/partiql/partiql-lang-kotlin/pull/471
- [x] Remove dependencies on `parseExprNode`. 
&nbsp; PR: https://github.com/partiql/partiql-lang-kotlin/pull/470

### DIFF
--- a/lang/src/org/partiql/lang/syntax/SqlParser.kt
+++ b/lang/src/org/partiql/lang/syntax/SqlParser.kt
@@ -15,6 +15,7 @@
 package org.partiql.lang.syntax
 
 import com.amazon.ion.*
+import com.amazon.ionelement.api.*
 import org.partiql.lang.ast.*
 import org.partiql.lang.domains.PartiqlAst
 import org.partiql.lang.errors.ErrorCode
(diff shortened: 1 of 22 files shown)


lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -43,8 +45,6 @@ import java.time.temporal.Temporal
  */
 class SqlParser(private val ion: IonSystem) : Parser {
```
dlurton, line 46: Do we still need this?  Can we remove it?

dlurton (approved):

PR #469, opened 2021-10-31
Fixed bug for 3 failed test cases

lziq:
*This PR is aimed to fix the bug causing 3 test cases failed in FB-deprecate-ExprNode branch*

### Problem Description:
1. The `CreateIndex` in `ExprNode` AST has a field `tableName`, which is a `String` type, as shown [here](https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/src/org/partiql/lang/ast/ast.kt#L480). However, in PIG AST, it is an `identifier`, as shown [here](https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/resources/org/partiql/type-domains/partiql.ion#L275).
2. Similarly for `DropIndex`, refer to [here](https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/src/org/partiql/lang/ast/ast.kt#L507) (`ExprNode`) and [here](https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/resources/org/partiql/type-domains/partiql.ion#L280) (PIG AST).
3. Similarly for `DropTable`, refer to [here](https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/src/org/partiql/lang/ast/ast.kt#L490) (`ExprNode`) and [here](https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/resources/org/partiql/type-domains/partiql.ion#L270) (PIG AST).

### Changes Details: 
1. Changed the `String` type into `VariableReference` type in `ExprNode` version AST. 
2. Made corresponding changes in 'ExprNodeToStatement', 'StatmentToExprNode', 'AstSerialization' and 'AstDeserialization' files. 
3. Made corresponding changes in rewriter. 
4. Changed expected value in test cases.

### DIFF
--- a/lang/src/org/partiql/lang/ast/StatementToExprNode.kt
+++ b/lang/src/org/partiql/lang/ast/StatementToExprNode.kt
@@ -430,7 +430,16 @@ private class StatementTransformer(val ion: IonSystem) {
         return when(op) {
             is DdlOp.CreateTable -> CreateTable(op.tableName.text, metas)
             is DdlOp.DropTable -> DropTable(op.tableName.name.text, metas)
-            is DdlOp.CreateIndex -> CreateIndex(op.indexName.name.text, op.fields.map { it.toExprNode() }, metas)
+            is DdlOp.CreateIndex ->
+                CreateIndex(
+                    VariableReference(
+                        op.indexName.name.text,
+                        op.indexName.case.toCaseSensitivity(),
+                        org.partiql.lang.ast.ScopeQualifier.UNQUALIFIED,
+                        metas
+                    ),
+                op.fields.map { it.toExprNode() },
+                metas)
             is DdlOp.DropIndex ->
                 DropIndex(
                     tableName = op.table.name.text,
--- a/lang/src/org/partiql/lang/ast/StatementToExprNode.kt
+++ b/lang/src/org/partiql/lang/ast/StatementToExprNode.kt
@@ -442,7 +442,12 @@ private class StatementTransformer(val ion: IonSystem) {
                 metas)
             is DdlOp.DropIndex ->
                 DropIndex(
-                    tableName = op.table.name.text,
+                    tableId = VariableReference(
+                        id = op.table.name.text,
+                        case = op.table.case.toCaseSensitivity(),
+                        scopeQualifier = org.partiql.lang.ast.ScopeQualifier.UNQUALIFIED,
+                        metas = emptyMetaContainer
+                    ),
                     identifier = VariableReference(
                         id = op.keys.name.text,
                         case = op.keys.case.toCaseSensitivity(),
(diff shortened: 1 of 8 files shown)


lang/test/org/partiql/lang/syntax/SqlParserTestBase.kt
```diff
@@ -118,10 +118,13 @@ abstract class SqlParserTestBase : TestBase() {
...
 
+        val test1 = parsedExprNode.stripMetas()
+        val test2 = deserializedExprNodeFromSexp.stripMetas()
+
         assertEquals(
             "Parsed ExprNodes must match deserialized s-exp $astVersion AST",
-            parsedExprNode.stripMetas(),
-            deserializedExprNodeFromSexp.stripMetas())
+            test1,
+            test2)
```
lziq: Will remove the unnecessary code in the next commit.

lang/src/org/partiql/lang/ast/StatementToExprNode.kt
```diff
@@ -430,7 +430,16 @@ private class StatementTransformer(val ion: IonSystem) {
...
             is DdlOp.CreateTable -> CreateTable(op.tableName.text, metas)
             is DdlOp.DropTable -> DropTable(op.tableName.name.text, metas)
-            is DdlOp.CreateIndex -> CreateIndex(op.indexName.name.text, op.fields.map { it.toExprNode() }, metas)
+            is DdlOp.CreateIndex ->
+                CreateIndex(
+                    VariableReference(
+                        op.indexName.name.text,
+                        op.indexName.case.toCaseSensitivity(),
+                        org.partiql.lang.ast.ScopeQualifier.UNQUALIFIED,
+                        metas
```
lziq, line 439: Should be `emptyMetaContainer`
```suggestion
                        emptyMetaContainer
```

lziq:
Tips for reviewers: first review 'ast.kt' file, then 'ExprNodeToStatement', 'StatmentToExprNode', 'AstSerialization' and 'AstDeserialization' files, and finally rest of files.

lang/src/org/partiql/lang/ast/ast.kt
```diff
@@ -504,7 +504,7 @@ data class DropTable(
  * [AstNode] named `Identifier` which has `id` and `case` properties.
```
dlurton, line 504: See the comment above this line.

lang/src/org/partiql/lang/ast/ast.kt
```diff
@@ -477,7 +477,7 @@ data class CreateTable(
  * @param keys The expressions that extract the keys from the target to be applied to the index.
  */
 data class CreateIndex(
-    val tableName: String,
+    val tableId: VariableReference,
```
dlurton, line 480: Please follow the advice in the comment I've identified below instead of using `VariableReference` here.

The new `Identifier` class should *not* inherit from `ExprNode`, but it should implement `HasMetas`.

dlurton (approved):
The changes as you have them are acceptable, just be aware that normally we don't change the structure of the V0 AST at all.  The only reason it's fine in this case IMHO is the parts of the V0 AST you are changing are not listed in [README-AST-V0.md](https://github.com/partiql/partiql-lang-kotlin/blob/main/docs/dev/README-AST-V0.md).   I realize you need to in this case so I won't block this PR.

PR #472, opened 2021-11-04
WIP to fix missing DML ParseNode tokens

### LINKED ISSUE #465: Inconsistent exceptions and error messages for DML statements as arguments for EXEC
Tried passing DML statements as arguments to a system stored procedure, and got the following different errors for each:

#### DML INSERT
* Statement: `EXEC abc_proc INSERT INTO dogs <<{}>>`
* Error: `IllegalStateException: Can't transform class org.partiql.lang.ast.DataManipulation to a PartiqlAst.expr`

#### DML DELETE
* Statement: `EXEC abc_proc DELETE FROM dogs WHERE name = 'Jack'`
* Error: `ParserException: Keyword delete only expected at the top level in the query`

#### DML UPDATE
* Statement: `EXEC abc_proc UPDATE dogs SET dogs.name = '1'`
* Error: `ParserException: Keyword null only expected at the top level in the query`

#### DML REMOVE
* Statement: `EXEC abc_proc FROM dogs REMOVE dogs.name`
* Error: `ParserException: Keyword null only expected at the top level in the query`

To clarify, I was expecting all of them would return a parser exception with a clear message.

alancai98:
WIP until
- other DML edge cases are sorted out
  - `SET` without `UPDATE` (like following `FROM`)
  - `UPDATE` without `SET`
- source location for error is agreed upon
  - If there's a DML `FROM`, should the error point here?
  - If there's a DML `UPDATE`, should the error point here?

Initial attempt to fix #465 (Inconsistent exceptions and error messages for DML statements as arguments for EXEC). For some DML expressions following stored procedure calls, there would be a "null" appearing in the parser error. This is due to some DML `ParseNode`s not storing the associated `Token`.

At the moment, I think resolving the above WIP issues will involve refactoring `ParseNode` for DML in a non-trivial way.

### DIFF
--- a/lang/src/org/partiql/lang/syntax/SqlParser.kt
+++ b/lang/src/org/partiql/lang/syntax/SqlParser.kt
@@ -1247,7 +1247,7 @@ class SqlParser(private val ion: IonSystem) : Parser {
 
         KEYWORD -> when (head?.keywordText) {
             in BASE_DML_KEYWORDS -> parseBaseDml()
-            "update" -> tail.parseUpdate()
+            "update" -> tail.parseUpdate(head!!)
             "delete" -> tail.parseDelete(head!!)
             "case" -> when (tail.head?.keywordText) {
                 "when" -> tail.parseCase(isSimple = false)
@@ -1464,11 +1464,11 @@ class SqlParser(private val ion: IonSystem) : Parser {
         return ParseNode(FROM, null, listOf(operation, fromList) + children, rem)
     }
 
-    private fun List<Token>.parseBaseDmls() : ParseNode {
+    private fun List<Token>.parseBaseDmls(updateToken: Token? = null) : ParseNode {
         var rem = this;
         val nodes = ArrayList<ParseNode>()
         while (rem.head?.keywordText in BASE_DML_KEYWORDS) {
-            var node = rem.parseBaseDml()
+            var node = rem.parseBaseDml(updateToken)
             nodes.add(node)
             rem = node.remaining
         }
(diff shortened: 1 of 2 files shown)


lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -2857,7 +2858,10 @@ class SqlParser(private val ion: IonSystem) : Parser {
             else -> node.type.isTopLevelType
         }
         if (topLevelTokenSeen && isTopLevelType) {
-            node.throwTopLevelParserError()
+            when (node.type) {
+                DML_LIST -> node.children.first().throwTopLevelParserError()
+                else -> node.throwTopLevelParserError()
+            }
```
alancai98, line 2864: `ParseType` of `DML_LIST` has no associated token, so previously if there were multiple dml ops, the error would output a "null" token. This changes to use the first dml op in the list

lang/test/org/partiql/lang/errors/ParserErrorsTest.kt
```diff
@@ -2110,6 +2110,56 @@ class ParserErrorsTest : SqlParserTestBase() {
...
+
+    @Test
+    fun execWithDmlRemove() = checkInputThrowingParserException(
+        "EXEC abc_proc FROM dogs REMOVE dogs.name",
+        ErrorCode.PARSE_UNEXPECTED_TERM,
+        mapOf(
+            Property.LINE_NUMBER to 1L,
+            Property.COLUMN_NUMBER to 25L,
+            Property.TOKEN_TYPE to TokenType.KEYWORD,
+            Property.TOKEN_VALUE to ion.newSymbol("remove")))
```
alancai98, line 2151: Ideally, I think the error should point to the `FROM` on column 15. `FROM` is not currently denoted as a "topLevelType" ([source](https://github.com/partiql/partiql-lang-kotlin/blob/1462dd351fa7dbba0493b6b2d9e87af8a0b3ac3c/lang/src/org/partiql/lang/syntax/SqlParser.kt#L128)), so this may be where the issue comes from.

Issue #366, opened 2021-02-19
PartiQL should use idiomatic CSV parsing


jonathanfrenz:
PartiQL's CSV parser today doesn't take into account quotation marks around fields. Typical CSV writers will write a field with a comma in it to be surrounded by quotations. This is mentioned in the PartiQL code as a TODO [here](https://github.com/partiql/partiql-lang-kotlin/blob/7c6998283c20dd20e61316755c9ed3bfcbbade2c/lang/src/org/partiql/lang/eval/io/DelimitedValues.kt#L27-L29).

[CSVReader from Apache](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVParser.html) is a great example of how to implement this easily. A one-line fix could just be replacing [this line](https://github.com/partiql/partiql-lang-kotlin/blob/7c6998283c20dd20e61316755c9ed3bfcbbade2c/lang/src/org/partiql/lang/eval/io/DelimitedValues.kt#L89) with a CSVParser.parse call.

dlurton:
We should really utilizing an existing [RFC-4180](https://tools.ietf.org/html/rfc4180) CSV parser as a dependency in the CLI project and removing the hand-rolled one we created.

dlurton:
Example 1: the csv file has header row.  in this case, the columns should be bound to `_n` fields where n is the field number. i..e.

This selects the first three columns of `some_file.csv` which does not contain a header row:

```
select _1, _2, _3 from read_csv('some_file.csv', `{ has_header_row: true }`)
```

Example 2: the csv file has a header row.  in this case, the columns should be bound to their names:


```
select first_name, last_name, city from read_csv('some_file_with_header_row.csv', `{ has_header_row: false }`)
```

Example 3: the user does not know if there are column names.  We should try to autodetect.  


```
select * from read_csv('some_file_with_header_row.csv', `{ has_header_row: auto }`)
-- or -- 
select * from read_csv('some_file_with_header_row.csv')  -- has_header_row defaults to auto
```

Example 4: the user would like to specify a predefined csv format:

```
select * from read_csv('some_file_with_header_row.csv', `{ format: postgres }`)
```

[...]

PR #474, opened 2021-11-10
Replaced self-implemented CSV file reader with Apache CSVParser for CLI

lziq:
*This PR aims to resolve the issue on CLI reading CSV files reported in Issue #366*

### Problem Description
Originally, we were implementing the CSV file reader for CLI by ourselves. Our implementation does not follow the standard CSV format as defined in [rfc4180](https://datatracker.ietf.org/doc/html/rfc4180), and it casues the following problems: 
1. Cannot recognize the double quotes escape. As a result, the comma in between double quotes escape are considered as a delimiter in a line. 
2. Cannot ignore the the empty lines. 
3. Each row can have different numbers of fields. 

### Solution Description 
This PR migrates from the self-implemented CSV reader to a standard [Apache CSVParser library](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVParser.html). Thus, all the above problems except the last one are solved. Also, adopting this library can make it easier to realize more functionalities for CLI file reader, such as supporting reading files in other formats. 

### Changes Details
1. In 'DelimitedValues.kt', changed the file reader so now it uses the Apache `CSVParser` library to pares the CSV file. 
2. In 'DelimitedValues.kt', slightly changed the file writer so now it uses the Apache `CSVPrinter` library to help print the CSV file. 
3. Added test cases for all the problems solved. 
4. Changed the type of the variable `delimiter` from `String` into `Char`, since the `CSVParser` library only accepts the delimiter as `Char`. 

*For reviewers:*
First review 'DelimitedValues.kt', then 'DelimitedValuesTest.kt'. Other changes are minor. 
It might take up to 1.5 hours to review all the changes.

### DIFF
--- a/cli/test/org/partiql/cli/functions/ReadFileTest.kt
+++ b/cli/test/org/partiql/cli/functions/ReadFileTest.kt
@@ -81,6 +81,18 @@ class ReadFileTest {
     }
 
     @Test
+    fun readCsvWithIonSymbolAsInput() {
+        writeFile("data_with_ion_symbol_as_input.csv", "1,2")
+
+        val args = listOf("\"${dirPath("data_with_ion_symbol_as_input.csv")}\"", "{type:csv}").map { it.exprValue() }
+
+        val actual = function.call(env, args).ionValue
+        val expected = "[{_1:\"1\",_2:\"2\"}]"
+
+        assertEquals(ion.singleValue(expected), actual)
+    }
+
+    @Test
     fun readCsvWithDoubleQuotesEscape() {
         writeFile("data_with_double_quotes_escape.csv", "\"1,2\",2")
 
--- a/cli/test/org/partiql/cli/functions/ReadFileTest.kt
+++ b/cli/test/org/partiql/cli/functions/ReadFileTest.kt
@@ -54,7 +54,7 @@ class ReadFileTest {
         val actual = function.call(env, args).ionValue
         val expected = "[1, 2]"
 
-        assertEquals(actual, ion.singleValue(expected))
+        assertEquals(ion.singleValue(expected), actual)
     }
 
     @Test
@@ -65,7 +65,7 @@ class ReadFileTest {
         val actual = function.call(env, args).ionValue
         val expected = "[1, 2]"
 
-        assertEquals(actual, ion.singleValue(expected))
+        assertEquals(ion.singleValue(expected), actual)
     }
 
     @Test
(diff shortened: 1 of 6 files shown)


lang/src/org/partiql/lang/eval/io/DelimitedValues.kt
```diff
@@ -66,38 +68,29 @@ object DelimitedValues {
...
-                }
-            }
-        }
+        val seq = csvParser.asSequence().map { csvRecord ->
+            valueFactory.newStruct(
+                csvRecord.mapIndexed { i, value ->
+                    val name = when {
+                        i < columns.size -> columns[i]
+                        else -> syntheticColumnName(i)
+                    }
```
lziq, line 88: The code here allows the case that each row has different numbers of fields. I just follow the original code even though [rfc4180](https://datatracker.ietf.org/doc/html/rfc4180) does not allow it. Let me know if it is better to change it.

cli/test/org/partiql/cli/functions/ReadFileTest.kt
```diff
@@ -77,7 +77,55 @@ class ReadFileTest {
         val actual = function.call(env, args).ionValue
         val expected = "[{_1:\"1\",_2:\"2\"}]"
 
-        assertEquals(actual, ion.singleValue(expected))
+        assertEquals(ion.singleValue(expected), actual)
+    }
+
+    @Test
```
dlurton, line 83: Should `tsv` files also support the use of a header row?  If so, I don't see any tests for `tsv` files that include a header row.

cli/test/org/partiql/cli/functions/ReadFileTest.kt
```diff
@@ -54,7 +54,7 @@ class ReadFileTest {
         val actual = function.call(env, args).ionValue
         val expected = "[1, 2]"
 
-        assertEquals(actual, ion.singleValue(expected))
+        assertEquals(ion.singleValue(expected), actual)
```
dlurton, line 57: thanks, reversed expected/actual arguments is my pet peeve.  :)

lang/src/org/partiql/lang/eval/io/DelimitedValues.kt
```diff
@@ -66,38 +68,29 @@ object DelimitedValues {
...
-            hasHeader -> {
-                val line = reader.readLine()
-                    ?: throw IllegalArgumentException("Got EOF for header row")
-
-                line.split(delimiter)
-            }
-            else -> emptyList()
+        val csvFormat = when (hasHeader){
+            true -> CSVFormat.DEFAULT.withDelimiter(delimiter).withTrim().withFirstRecordAsHeader()
+            false -> CSVFormat.DEFAULT.withDelimiter(delimiter).withTrim()
```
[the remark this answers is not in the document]
dlurton: Is there a reason you've included the `.withTrim()` option?  I would avoid doing that.  If the user wants to remove those spaces it is possible to do so with the `TRIM()` function

lang/src/org/partiql/lang/eval/io/DelimitedValues.kt
```diff
@@ -131,34 +124,36 @@ object DelimitedValues {
...
-            }.joinTo(output, delimiter)
-            output.write(newline)
+            csvPrinter.printRecord(
+                names.map {
+                    val col = row.bindings[BindingName(it, BindingCase.SENSITIVE)]?.ionValue ?: ion.newNull()
+                    col.csvStringValue()
+                }
+            )
+
+            csvPrinter.flush()
```
dlurton, line 154: Why call `.flush()` here?  Seems like that could hurt performance since you're forcing more smaller writes to disk instead of allowing the OS to batch the writes together.

lang/src/org/partiql/lang/eval/io/DelimitedValues.kt
```diff
@@ -131,34 +124,36 @@ object DelimitedValues {
     fun writeTo(ion: IonSystem,
                 output: Writer,
                 value: ExprValue,
-                delimiter: String,
+                delimiter: Char,
                 newline: String,
-                writeHeader: Boolean): Unit {
-        val nullValue = ion.newNull()
+                writeHeader: Boolean) {
+        val csvPrinter = CSVPrinter(output, CSVFormat.DEFAULT.withDelimiter(delimiter).withRecordSeparator(newline))
```
dlurton, line 130: Probably aught to be using Kotlin's [use](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html) here since your manual invocation of `.close()` below will be skipped if an exception occurs.  (`.use` is Kotlin's equivalent of Java's [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html).)

dlurton (changes requested):

lang/src/org/partiql/lang/eval/io/DelimitedValues.kt
lziq: Will remove it in the next revision.

lang/src/org/partiql/lang/eval/io/DelimitedValues.kt
lziq: Will change it in the next revision.

cli/test/org/partiql/cli/functions/ReadFileTest.kt
lziq: Will add test for reading tsv files with a header in the next revision.

cli/test/org/partiql/cli/functions/ReadFileTest.kt
```diff
@@ -77,7 +77,43 @@ class ReadFileTest {
...
+    }
+
+    @Test
+    fun readCsvWithDoubleQuotesEscape() {
+        writeFile("data_with_double_quotes_escape.csv", "\"1,2\",2")
+
+        val args = listOf("\"${dirPath("data_with_double_quotes_escape.csv")}\"", "{type:\"csv\"}").map { it.exprValue() }
```
dlurton, line 87: Can the user specify Ion symbols instead of strings?  (I think they should.) Need at least one test for that too.

dlurton (approved):

PR #476, opened 2021-11-11
Inital definition of PartiQL's relational algebra and converstion from the AST

dlurton:
This is the absolute beginning of PartiQL's experimental relational algebra which will be used to optimize PartiQL queries.  It's very early in its design so far.  For instance, many more operators other than the 4 defined so far are needed.  These will be defined later, incrementally.

This PR will merge to the `query-planner` branch where we will be also be building an experimental planner. 

- `docs/dev/RELATIONAL-ALGEBRA.md`  is intended to be the long term home for the documentation of the relational algebra.  Start there for details.
- `partiql.ion` includes a permuted PIG domain for the algebra. 
- Also included is a `astToAlgebra` function which performs conversion from the AST to the relational algebra.

### DIFF
--- a/docs/dev/RELATIONAL-ALGEBRA.md
+++ b/docs/dev/RELATIONAL-ALGEBRA.md
@@ -1,0 +1,445 @@
+# PartiQL’s Relational Algebra
+
+This document is a work-in-progress document that describes the PartiQL’s equivalent of the relational algebra that will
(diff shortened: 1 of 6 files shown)


dlurton:
I'm just going to merge this since it was previously reviewed by @alancai98 and is going into a feature branch anyway.

Issue #475, opened 2021-11-10
Add ability to use `--query` command-line option without external data source


dlurton:
Whenever the `-q`/`--query` command-line option is specified to the CLI, it will *always* read Ion formatted data from stdin unless the `-i`/`--input` argument is included, in which case it reads Ion formatted data from the specified file.  We should provide a way for the user to disable this behavior because there is no way to turn it off.  For instance, we currently only support reading CSV files thru the `read_file` function, but this means we must have some sort of Ion input file to use the `--query` option.

I've worked around this by creating a file containing fake Ion data:

```sh
partiql --query "SELECT * FROM read_file('important-data.csv')" --input dummy-data.ion
```

Ideally, this query command can be executed without the need for `--input dummy-data.ion`, but as of today this will cause the `partiql` command to wait for Ion data to be piped in via `stdin`.

Ultimately, this significantly reduces the usefulness of the CLI in non-interactive scenarios, especially when querying non-Ion files.

PR #478, opened 2021-11-23
Fixed the bug on CLI command with input argument

lziq:
*Issue #475*

*Description of changes:*
1. Fixed the issue by assuming the input data is empty, if users do not explicit specify the input data argument while using CLI. 
2. Creates one test case to cover the case.

### DIFF
--- a/cli/src/org/partiql/cli/main.kt
+++ b/cli/src/org/partiql/cli/main.kt
@@ -18,7 +18,8 @@ package org.partiql.cli
 
 import com.amazon.ion.system.*
 import joptsimple.*
-import org.partiql.cli.functions.*
+import org.partiql.cli.functions.ReadFile
+import org.partiql.cli.functions.WriteFile
 import org.partiql.lang.*
 import org.partiql.lang.eval.*
 import org.partiql.lang.syntax.*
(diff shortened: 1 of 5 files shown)


cli/test/org/partiql/cli/CliTest.kt
```diff
@@ -128,4 +129,12 @@ class CliTest {
...
     }
+
+    @Test
+    fun withoutInput() {
+        val subject = makeCli("SELECT * FROM input_data")
+        val actual = subject.runAndOutput()
+
+        assertEquals("\n", actual)
+    }
 }
```
abhikuhikar, line 140: This behavior is incorrect. If the `--input` option is not provided this query should throw an error - `No such binding found for 'input_data'`. Right now `input_data` is bound to *`emptyInputStream`*. Consider removing this binding in case the cli option `--input` is not set.

abhikuhikar (changes requested):
Need to fix the behavior for binding named `input_data`. Also consider modifying `Cli.md` markdown to reflect the updated behavior.

cli/src/org/partiql/cli/main.kt
```diff
@@ -17,11 +17,11 @@
 package org.partiql.cli
 
 import com.amazon.ion.system.*
-import org.partiql.lang.eval.*
-import org.partiql.lang.syntax.*
-import org.partiql.cli.functions.*
 import joptsimple.*
+import org.partiql.cli.functions.*
```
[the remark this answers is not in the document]
lziq: Will apply this comment in the next commit.

cli/test/org/partiql/cli/CliTest.kt
lziq: Will apply this comment in the next commit.

abhikuhikar (approved):

Issue #473, opened 2021-11-10
cli bug: Stream closed when outputting ion text


dlurton:
When running the CLI with `-output <any file>` and `--output-format ION_TEXT` command line arguments, an unexpected exception is thrown.  For example:

```
$ partiql -q "1" --output-format ION_TEXT -o temp.ion                                                                                                                                                                                                             asdf
java.io.IOException: Stream Closed
	at java.base/java.io.FileOutputStream.writeBytes(Native Method)
	at java.base/java.io.FileOutputStream.write(FileOutputStream.java:341)
	at org.partiql.cli.Cli.run(Cli.kt:53)
	at org.partiql.cli.Main.runCli(main.kt:179)
	at org.partiql.cli.Main.main(main.kt:137)
```

The command causes the CLI to start, and since `-i` was specified, it will read start to read the contents of the `input_data` global variable from `stdin`.  To get evaluation going, I typed "1234" and hit enter in order to feed a single Ion value into the CLI.  Then the exception occurred.  

Note that this works fine when `-o ION_TEXT` is omitted:
```
$ partiql -q "1" --output-format ION_TEXT                                                                                                                                                                                                                       
1234
1
```

Or when an output format other than `ION_TEXT` is used.

The work-around to this issue is to omit the `-o` option and redirect output to a file:

```
$ partiql -q "1" --output-format ION_TEXT > temp.ion
```

PR #479, opened 2021-11-23
Fixed bug on CLI output IONTEXT to file

lziq:
*Issue #473*

### Problem Description
As show in https://github.com/partiql/partiql-lang-kotlin/blob/main/cli/src/org/partiql/cli/Cli.kt#L52-53, the `use` block in L52 invokes `close()` function on `output`. In case of `FileOutputStream`, it indeed closes the output stream. Thus, the `write()` function in L53 throws an exception. 

*Description of changes:*
1. Move L53 into the `use` block. 
2. Create a test case to cover file output.

### DIFF
--- a/cli/src/org/partiql/cli/Cli.kt
+++ b/cli/src/org/partiql/cli/Cli.kt
@@ -48,8 +48,8 @@ internal class Cli(private val valueFactory: ExprValueFactory,
             val result = compilerPipeline.compile(query).eval(EvaluationSession.build { globals(bindings) })
 
             when (format) {
-                ION_TEXT   -> {
-                    ionTextWriterBuilder.build(output).use { printIon(it, result) }
+                ION_TEXT   -> ionTextWriterBuilder.build(output).use {
+                    printIon(it, result)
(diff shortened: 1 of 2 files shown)


abhikuhikar (commented):
Change looks reasonable to me. However, after executing `partiql -q "1" --output-format ION_TEXT -o temp.ion`, it still waits to read from `stdin` but does not throw a `stream closed` error after providing input to `stdin` as it used to. However, I believe this should not be the case. The command should not wait for the input from `stdin`.

lziq:
> Change looks reasonable to me. However, after executing `partiql -q "1" --output-format ION_TEXT -o temp.ion`, it still waits to read from `stdin` but does not throw a `stream closed` error after providing input to `stdin` as it used to. However, I believe this should not be the case. The command should not wait for the input from `stdin`.

I think this issue is resolved in another PR, https://github.com/partiql/partiql-lang-kotlin/pull/478

abhikuhikar (dismissed):

abhikuhikar (approved):

PR #489, opened 2021-12-22
Removes sealed class DateTimeType as it was not needed.

### LINKED ISSUE #388: Reconsider or justify modeling of `DATE` and `TIME` literals in their own sealed class
Below we have `DateTimeType` as a `sealed` class inheriting from `ExprNode` and `DateTimeType.Date` (and eventually `DateTimeType.Time`) inheriting from `DateTimeType`.  

It's not clear in this case why `DateTimeType` needs to exist and why date and time literals do not inherit direclty from `ExprNode`.  
`DateTimeType` either needs to include documentation that justifies its existence or it should be removed.

https://github.com/partiql/partiql-lang-kotlin/blob/ce4f7a007967a947541de1ac36d3aa544a0e95bc/lang/src/org/partiql/lang/ast/ast.kt#L970-L989

abhikuhikar:
Resolves #388.

### DIFF
--- a/lang/src/org/partiql/lang/ast/ast.kt
+++ b/lang/src/org/partiql/lang/ast/ast.kt
@@ -997,6 +997,10 @@ enum class OrderingSpec {
 
 /**
  * AST Node corresponding to the DATE literal
+ *
+ * @param year represents the year of the DATE as Integer.
+ * @param month represents the month of the DATE as Integer.
+ * @param day represents the day of the DATE as Integer.
  */
 data class DateLiteral(
     val year: Int,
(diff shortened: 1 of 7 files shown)


lang/src/org/partiql/lang/ast/ast.kt
```diff
@@ -996,52 +996,43 @@ enum class OrderingSpec {
 }
 
 /**
- * The sealed class includes all the datetime types such as DATE, TIME, TIMESTAMP
- * Note that the ast nodes corresponding to the DATE, TIME and TIMESTAMP here are different from the [Literal] nodes.
- * You can create an Ion literal as [Timestamp] which will correspond to the [Literal] node and will have the type
- * [SqlDataType.TIMESTAMP]. However that will be different from the
- * `TIMESTAMP` here.
- * Note: TIME and TIMESTAMP are yet to be added.
+ * AST Node corresponding to the DATE literal
```
am357, line 999: Could we also `@param`s here?

abhikuhikar: Added.

am357 (approved):
ShipIt!

Issue #491, opened 2022-01-03
Add more test cases to read CSV files in other formats


lziq:
Currently, the CLI supports reading CSV file formats in `postgresql_csv`, `excel_csv`, `mysql_csv`, `mongodb_csv`, `mongodb_tsv`, and `postgresql_text`.  
However, as shown in https://github.com/partiql/partiql-lang-kotlin/blob/main/cli/test/org/partiql/cli/functions/ReadFileTest.kt#L160-L182, the test cases only cover `excel_csv` and `postgresql_csv` formats. We should add more test cases to cover other formats as well.

PR #493, opened 2022-01-06
More test cases to read CSV formats

lziq:
*Issue #491*

Added more test cases to read different CSV formats.

### DIFF
--- a/cli/test/org/partiql/cli/functions/ReadFileTest.kt
+++ b/cli/test/org/partiql/cli/functions/ReadFileTest.kt
@@ -206,30 +206,6 @@ class ReadFileTest {
     }
 
     @Test
-    fun readMongodbCsvFile() {
-        writeFile("simple_mongo.csv", "id,name,balance\n1,Bob,10000.00")
-
-        val args = listOf("\"${dirPath("simple_mongo.csv")}\"", "{type:\"mongodb_csv\", header:true}").map { it.exprValue() }
-
-        val actual = function.call(env, args).ionValue
-        val expected = "[{id:\"1\",name:\"Bob\",balance:\"10000.00\"}]"
-
-        assertEquals(ion.singleValue(expected), actual)
-    }
-
-    @Test
-    fun readMongodbTsvFile() {
-        writeFile("simple_mongo.tsv", "id\tname\tbalance\n1\tBob\t10000.00")
-
-        val args = listOf("\"${dirPath("simple_mongo.tsv")}\"", "{type:\"mongodb_tsv\", header:true}").map { it.exprValue() }
-
-        val actual = function.call(env, args).ionValue
-        val expected = "[{id:\"1\",name:\"Bob\",balance:\"10000.00\"}]"
-
-        assertEquals(ion.singleValue(expected), actual)
-    }
-
-    @Test
     fun readCustomizedCsvFile1() { // delimiter
         writeFile("customized.csv", "id name balance\n1 Bob 10000.00")
 
--- a/cli/test/org/partiql/cli/functions/ReadFileTest.kt
+++ b/cli/test/org/partiql/cli/functions/ReadFileTest.kt
@@ -171,12 +171,12 @@ class ReadFileTest {
 
     @Test
     fun readPostgreSQLCsvFile() {
-        writeFile("simple_postgresql.csv", "id,name,balance\n1,Bob,10000.00")
+        writeFile("simple_postgresql.csv", "id,name,balance\n1,B\"\"ob,10000.00")
 
         val args = listOf("\"${dirPath("simple_postgresql.csv")}\"", "{type:\"postgresql_csv\", header:true}").map { it.exprValue() }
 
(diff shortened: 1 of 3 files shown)


cli/test/org/partiql/cli/functions/ReadFileTest.kt
```diff
@@ -181,6 +181,54 @@ class ReadFileTest {
...
+    fun readMongodbTsvFile() {
+        writeFile("simple_mongo.tsv", "id\tname\tbalance\n1\tBob\t10000.00")
+
+        val args = listOf("\"${dirPath("simple_mongo.tsv")}\"", "{type:\"mongodb_tsv\", header:true}").map { it.exprValue() }
+
+        val actual = function.call(env, args).ionValue
+        val expected = "[{id:\"1\",name:\"Bob\",balance:\"10000.00\"}]"
+
+        assertEquals(ion.singleValue(expected), actual)
+    }
```
alancai98, line 230: I was browsing `CSVFormat`'s [documentation](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVFormat.html#MONGODB_CSV) to see more of the differences between the different CSV parsing formats. I noticed that for `mongodb_csv` and `mongodb_tsv` that it states:

> **Parsing is not supported yet.**

Could we verify if the parsing is actually supported by passing a CSV/TSV created by `mongoexport`? If it's actually not supported, I think we should remove those options from the CLI parsing.

cli/test/org/partiql/cli/functions/ReadFileTest.kt
```diff
@@ -181,6 +181,54 @@ class ReadFileTest {
         assertEquals(ion.singleValue(expected), actual)
     }
 
+    @Test
```
alancai98, line 184: These tests do test the different supported PartiQL CLI options, but they don't really show the difference between the formats (e.g. `readPostgreSQLTextFile` and `readMongodbTsvFile` have the same CSV string).

The documentation for CSVFormat's [POSTGRESQL_CSV](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVFormat.html#POSTGRESQL_CSV) states:

> This is a comma-delimited format with a LF character as the line separator. Values are double quoted and special characters are escaped with '"'. The default NULL string is "".

The documentation for CSVFormat's [MYSQL](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVFormat.html#MYSQL) states:

> This is a tab-delimited format with a LF character as the line separator. Values are not quoted and special characters are escaped with '\'. The default NULL string is "\\N".

I think using different CSV strings with those format nuances will provide better test cases for the different CSV formats.

cli/test/org/partiql/cli/functions/ReadFileTest.kt
lziq: I was using MongDB Commpass to create the CSV file. It just looks like a normal CSV file. 
[test.csv](https://github.com/partiql/partiql-lang-kotlin/files/7825251/test.csv)

If the `CSVForamt` docs says it is not supported yet. I think we should remove them.

cli/test/org/partiql/cli/functions/ReadFileTest.kt
```diff
@@ -171,12 +171,12 @@ class ReadFileTest {
 
     @Test
     fun readPostgreSQLCsvFile() {
-        writeFile("simple_postgresql.csv", "id,name,balance\n1,Bob,10000.00")
+        writeFile("simple_postgresql.csv", "id,name,balance\n1,B\"\"ob,10000.00")
```
lziq, line 174: Since double quotes is escape in postgre_sql format, we need 2 double quotes as the first one escape the second.

alancai98 (approved):
:shipit: Thanks for adding the tests!

lziq:
Hi, Almann. I found in 2 test cases the behavior of `StaticTypeVisitorTransform` is confusing. Can you spend some time looking at them and answering my question?

Case1:
                Schema:               “a” to <struct>, “foo“ to <int>
                Query:                  SELECT foo FROM a AS b

In this case, the actual behavior is to qualify`foo` as `b.foo`. However, should `foo` be considered as an ambiguous variable, as it can either refer to the global variable `foo` or get qualified as `b.foo`?

I found the comment here, https://code.amazon.com/packages/IonSQLSandbox/blobs/d9980a8a45ae65c6081f95f1dab35819ad89cf6c/--/lang/src/org/partiql/lang/eval/visitors/StaticTypeVisitorTransform.kt#L286, saying that in this case it should be resolved as an attribute. Do you know why we have to deal with it in this way?

Case2:
                Schema:               “a” to <struct>, “c” to <struct>, “foo” to <int>
                Query:                  SELECT foo FROM a AS b, c AS d

In this case, the actual behavior is to resolve `foo` as a global variable. Again, shouldn’t we throw an error as `foo` can be either `b.foo`, `d.foo` or the global variable `foo` here?


Best,
Ziqi

PR #523, opened 2022-02-09
CAST: Improve the CAST assertion assertEval

am357:
Improve the CAST assertion assertEval with removing a redundant
assertion for permissive mode since we already run tests for this
mode as part of the overall process.

Issue #, if available: 519

Addressed the comment from the original PR which was closed because it was having changes from `remove-unused-imports` change:
https://github.com/partiql/partiql-lang-kotlin/pull/520


alancai98 (approved):
(Reviewed previously in https://github.com/partiql/partiql-lang-kotlin/pull/520)

:shipit:

Issue #526, opened 2022-02-11
Unexpected/incorrect exception thrown by StaticTypeInferenceVisitorTransform when calling a function with a struct reference of the wrong type


lziq:
When I call StaticTypeInferenceVisitorTransform.transformStatement() on a query calling a function where I am passing in a field of a struct of the wrong type, I expect to get a SemanticException indicating "InvalidArgumentForFunction".  Instead I get a java.lang.IllegalStateException "No source location found on PartiqlAst: {$static_type=int}".

A minimal test case (using the test framework in StaticTypeInferenceVisitorTransformTest:

    @Test
    fun `test function of struct reference with wrong type`() {
        val tc = TestCase(
            name = "STRUCT REFERENCE WITH WRONG TYPE",
            originalSql = "SELECT upper(structCol.field) as foo from a",
            globals = mapOf(
                "a" to BagType(StructType(mapOf("structCol" to StructType(mapOf("field" to INT)))))
            ),
            // This is wrong, but doesn't interfere with reproducing the problem
            handler = expectQueryOutputType(BagType(StructType(mapOf("foo" to STRING))))
        )
        runTest(tc)
    }

This is the exception stack:

java.lang.IllegalStateException: No source location found on PartiqlAst: {$static_type=int}

	at org.partiql.lang.eval.visitors.StaticTypeInferenceVisitorTransform$VisitorTransform.getSourceLocation(StaticTypeInferenceVisitorTransform.kt:111)
	at org.partiql.lang.eval.visitors.StaticTypeInferenceVisitorTransform$VisitorTransform.functionHasValidArgTypes(StaticTypeInferenceVisitorTransform.kt:902)
	at org.partiql.lang.eval.visitors.StaticTypeInferenceVisitorTransform$VisitorTransform.returnTypeForPropagatingFunction(StaticTypeInferenceVisitorTransform.kt:933)
	at org.partiql.lang.eval.visitors.StaticTypeInferenceVisitorTransform$VisitorTransform.computeReturnTypeForFunctionCall(StaticTypeInferenceVisitorTransform.kt:862)
	at org.partiql.lang.eval.visitors.StaticTypeInferenceVisitorTransform$VisitorTransform.transformExprCall(StaticTypeInferenceVisitorTransform.kt:489)
	at org.partiql.lang.domains.PartiqlAst$VisitorTransform.transformExpr(partiql-domains.kt:15577)
[...]

PR #527, opened 2022-02-11
Changed Path AST node to use its root node source location

lziq:
*Issue #526*

*Description of changes:*
Changed Path AST node to use its root node source location

Now the error becomes what we expected: 
Screen Shot 2022-02-11 at 12 09 08 PM


lang/test/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransformTest.kt
```diff
@@ -5737,6 +5737,21 @@ class StaticTypeInferenceVisitorTransformTest : VisitorTransformTestBase() {
...
+                    expectedProblems = listOf(
+                        createInvalidArgumentTypeForFunctionError(
+                            sourceLocation = SourceLocationMeta(1L, 7L, 1L),
+                            functionName = "upper",
+                            expectedArgType = unionOf(STRING, SYMBOL),
+                            actualType = INT
+                        )
+                    )
+                )
+            ),
```
alancai98, line 5754: Could you also add the similar path test case brought up by the customer involving list references of the wrong type:

Something like `UPPER(x[0])` with a globals of `"x" to ListType(INT)`?

alancai98 (changes requested):
Fix looks good to me! Have one comment to add the other path test case brought up by the customer.

lang/test/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransformTest.kt
lziq: Will add the test case in the next commit.

alancai98 (approved):
:shipit:

PR #529, opened 2022-02-18
Clean up ExprFunction test

lziq:


lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -441,7 +441,7 @@ class SqlParser(
                             else -> errMalformedParseTree("Unsupported path component: ${it.type}")
                         }
                     }
-                    path(rootExpr, pathComponents, metas)
+                    path(rootExpr, pathComponents, rootExpr.metas) // Here we use its root source location, since itself has no source location (not a token).
```
alancai98, line 444: nit: for future PRs, it would be helpful to start with a clean commit history (i.e. not including commits from other PRs). GitHub shows this as a changed file even though your target branch, `main`, already has this change, which can be confusing for the reviewer.

You could start by creating the branch directly off of `main`:
```
git checkout main
git pull
git checkout feature-branch
... (make and commit changes) ...
... (create PR with feature-branch) ...
```

If the origin GitHub branch already has the commits, you can rebase and force push to the origin branch.

alancai98 (approved):
Changes already approved previously :shipit: . Can you also add a more detailed PR description?

PR #530, opened 2022-02-19
Deprecate ExprNode in test and related files.

lziq:
*Description of changes:*
1. Deprecated `compile(ExprNode)` and added `compile(PartiqlAst.Statement)` in CompilerPipeline.kt. Deprecated `compile(ExprNode)` and added `compile(PartiqlAst.Statement)` in EvaluatingCompiler.kt. 
2. Replaced the use of `ExprNode` with corresponding PIG AST node in EvaluatingCompilerNAryTests.kt. 
3. Removed dependencies on `ExprNode` in other files. 

Suggested reviewing order: 
1. CompilerPipline.kt and EvaluatingCompiler.kt. 
2. EvaluatingCompilerNAryTests.kt 
3. Rest of files.

### DIFF
--- a/lang/src/org/partiql/lang/ast/ast.kt
+++ b/lang/src/org/partiql/lang/ast/ast.kt
@@ -20,7 +20,7 @@ import org.partiql.lang.util.interruptibleMap
 import org.partiql.lang.util.stringValue
 import java.util.Arrays
 
-@Deprecated("Please use PIG-generated AST instead. ")
+@Deprecated("Please use PIG-generated AST instead.")
 /**
  * Base type for all AST nodes.
  */
@@ -63,7 +63,7 @@ sealed class AstNode : Iterable<AstNode> {
     }
 }
 
-@Deprecated("Please use PIG-generated AST instead. ")
+@Deprecated("Please use PIG-generated AST instead.")
 /**
  * The only nodes that inherit directly from ExprNode should just be generic expressions that can be used any
  * place a value is allowed.
--- a/lang/src/org/partiql/lang/ast/ast.kt
+++ b/lang/src/org/partiql/lang/ast/ast.kt
@@ -20,6 +20,7 @@ import org.partiql.lang.util.interruptibleMap
 import org.partiql.lang.util.stringValue
 import java.util.Arrays
 
+@Deprecated("Please use PIG-generated AST instead. ")
 /**
  * Base type for all AST nodes.
  */
@@ -62,6 +63,7 @@ sealed class AstNode : Iterable<AstNode> {
     }
 }
 
+@Deprecated("Please use PIG-generated AST instead. ")
 /**
  * The only nodes that inherit directly from ExprNode should just be generic expressions that can be used any
  * place a value is allowed.
--- a/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
+++ b/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
@@ -133,6 +133,7 @@ import java.math.BigDecimal
 import java.util.LinkedList
 import java.util.Stack
 import java.util.TreeSet
+import kotlin.math.exp
 
 /**
  * A basic compiler that converts an instance of [ExprNode] to an [Expression].
@@ -312,6 +313,9 @@ internal class EvaluatingCompiler(
         )
     }()
 
+    @Deprecated("ExprNode is deprecated. Please use PIG generated AST. ")
+    fun compile(exprNode: ExprNode): Expression = compile(exprNode.toAstStatement())
+
     /**
      * Compiles an [ExprNode] tree to an [Expression].
      *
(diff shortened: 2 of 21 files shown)


lang/src/org/partiql/lang/ast/ast.kt
```diff
@@ -20,6 +20,7 @@ import org.partiql.lang.util.interruptibleMap
 import org.partiql.lang.util.stringValue
 import java.util.Arrays
 
+@Deprecated("Please use PIG-generated AST instead. ")
```
lziq, line 23: Not pretty sure if this line is necessary.

lang/src/org/partiql/lang/ast/ast.kt
```diff
@@ -62,6 +63,7 @@ sealed class AstNode : Iterable<AstNode> {
     }
 }
 
+@Deprecated("Please use PIG-generated AST instead. ")
```
lziq, line 66: Same as above.

lang/src/org/partiql/lang/ast/ast.kt
am357: For this and the other one in `ast.kt`: just curious, what makes you unsure?

am357 (approved):
Overall looks legitimate. This a nitpick but can we remove the extra space in all the instances of the following?

```
"Please use PIG-generated AST instead. "
```

Could be in a separate PR.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -312,16 +313,19 @@ internal class EvaluatingCompiler(
         )
     }()
 
+    @Deprecated("ExprNode is deprecated. Please use PIG generated AST. ")
+    fun compile(exprNode: ExprNode): Expression = compile(exprNode.toAstStatement())
+
     /**
      * Compiles an [ExprNode] tree to an [Expression].
```
alancai98, line 320:
```suggestion
     * Compiles a [PartiqlAst] tree to an [Expression].
```

alancai98: (Could already be fixed in the other PR)

lang/test/org/partiql/lang/thread/ThreadInterruptedTests.kt
```diff
@@ -1,6 +1,7 @@
 package org.partiql.lang.thread
 
 import com.amazon.ion.system.IonSystemBuilder
+import com.amazon.ionelement.api.ionInt
```
alancai98, line 4: Unused imports of `Literal` and `.toIonElement`

lang/src/org/partiql/lang/ast/ast.kt
dlurton: Honestly, I'd consider adding this to every type in this file.

alancai98: They should be necessary. We are purposely deprecating `ExprNode` in favor of PIG-generated `PartiqlAst`.

Same applies to L66.

dlurton: Functions too.

Really, anything not `private`.

alancai98 (commented):
Looks good. Just noticed a couple minor things.

lang/test/org/partiql/lang/eval/EvaluatingCompilerNAryTests.kt
```diff
@@ -46,78 +43,88 @@ class EvaluatingCompilerNAryTests: EvaluatorTestBase() {
...
-            ), dummyMetas)
+        fun buildExprList(arg1: Long?, arg2: Long?, arg3: Long?) = listOf(
+            PartiqlAst.build { lit(arg1.toIonValue().toIonElement()) },
+            PartiqlAst.build { lit(arg2.toIonValue().toIonElement()) },
+            PartiqlAst.build { lit(arg3.toIonValue().toIonElement()) },
+        )
 
+        val query = PartiqlAst.build {
+            query(
+                when (tc.op) {
```
dlurton, line 107: Blah, I'd consider making `tc.op` an enum--don't use strings for this.

lang/test/org/partiql/lang/eval/EvaluatingCompilerNAryTests.kt
```diff
@@ -195,72 +201,86 @@ class EvaluatingCompilerNAryTests: EvaluatorTestBase() {
...
     @Parameters
     fun ternaryComparisonTest(tc: ComparisonTestCase) {
-        val exprNode = NAry(
-            tc.op,
-            tc.args.map { Literal(it.toIonValue(), dummyMetas) },
-            dummyMetas)
-
+        val query = PartiqlAst.build {
+            query(
+                when (tc.op) {
```
dlurton, line 206: Same--prefer `tc.op` to be an enum.

lang/src/org/partiql/lang/ast/ast.kt
dlurton: Of course, if you do this you'll have to evaluate all of the new compiler warnings and add an appropriate `@Supppress` attribute to avoid useless noise of compiler warnings.

lang/test/org/partiql/lang/thread/ThreadInterruptedTests.kt
lziq: Will remove them in the next commit.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
lziq: I think it is fixed in the other PR, https://github.com/partiql/partiql-lang-kotlin/pull/528/files#diff-141d6e1b17ce7a7bb13abcea1c75b9129b09c56c6ca86f7417487835a7d875deR278.

lang/test/org/partiql/lang/eval/EvaluatingCompilerNAryTests.kt
lziq: Will create enum for these operators in the next commit.

alancai98 (approved):
:shipit:

am357 (approved):

PR #533, opened 2022-02-23
Deprecate ExprNode metas

lziq:
*Motivation:*
The meta container PIG AST uses is `com.amazon.ionelement.api.MetaContainer`, while `ExprNode` uses `org.partiql.lang.ast.MetaContainer`. As we are deprecating `ExprNode`, we need to replace meta container as well. 

Note that the meta inside meta container is still the same. 

*Description of changes:*
1. Added `metaContainerOf(vararg Meta)` in util.kt file. 
2. Replaced `org.partiql.lang.ast.MetaContainer` with `com.amazon.ionelement.api.MetaContainer` in EvaluatingCompiler.kt, Thunk.kt and related files.

### DIFF
--- a/lang/src/org/partiql/lang/domains/util.kt
+++ b/lang/src/org/partiql/lang/domains/util.kt
@@ -3,6 +3,7 @@ package org.partiql.lang.domains
 import com.amazon.ionelement.api.MetaContainer
 import com.amazon.ionelement.api.emptyMetaContainer
 import com.amazon.ionelement.api.metaContainerOf
+import org.partiql.lang.ast.Meta
 import org.partiql.lang.ast.SourceLocationMeta
 import org.partiql.lang.ast.StaticTypeMeta
 import org.partiql.lang.errors.Property
@@ -15,6 +16,9 @@ fun PartiqlAst.Builder.id(name: String) =
 
 
 val MetaContainer.staticType: StaticTypeMeta? get() = this[StaticTypeMeta.TAG] as StaticTypeMeta?
+
+fun metaContainerOf(vararg metas: Meta): MetaContainer =
+    metaContainerOf(metas.map { Pair(it.tag, it) })
 
 /**
  * Returns a [MetaContainer] with *only* the source location of the receiver [MetaContainer], if present.
--- a/lang/src/org/partiql/lang/domains/util.kt
+++ b/lang/src/org/partiql/lang/domains/util.kt
@@ -17,6 +17,7 @@ fun PartiqlAst.Builder.id(name: String) =
 
 val MetaContainer.staticType: StaticTypeMeta? get() = this[StaticTypeMeta.TAG] as StaticTypeMeta?
 
+/** Constructs a container with the specified metas. */
 fun metaContainerOf(vararg metas: Meta): MetaContainer =
     metaContainerOf(metas.map { Pair(it.tag, it) })
 
--- a/lang/src/org/partiql/lang/eval/Exceptions.kt
+++ b/lang/src/org/partiql/lang/eval/Exceptions.kt
@@ -110,14 +110,14 @@ fun fillErrorContext(errorContext: PropertyValueMap, location: SourceLocationMet
     }
 }
 
-fun fillErrorContext(errorContext: PropertyValueMap, metaContainer: com.amazon.ionelement.api.MetaContainer) {
+fun fillErrorContext(errorContext: PropertyValueMap, metaContainer: MetaContainer) {
     val location = metaContainer[SourceLocationMeta.TAG] as? SourceLocationMeta
     if(location != null) {
         fillErrorContext(errorContext, location)
     }
 }
 
-fun errorContextFrom(metaContainer: com.amazon.ionelement.api.MetaContainer?): PropertyValueMap {
+fun errorContextFrom(metaContainer: MetaContainer?): PropertyValueMap {
     if(metaContainer == null) {
         return PropertyValueMap()
     }
--- a/lang/src/org/partiql/lang/eval/Exceptions.kt
+++ b/lang/src/org/partiql/lang/eval/Exceptions.kt
@@ -14,8 +14,8 @@
 
 package org.partiql.lang.eval
 
+import com.amazon.ionelement.api.MetaContainer
 import org.partiql.lang.SqlException
-import org.partiql.lang.ast.MetaContainer
 import org.partiql.lang.ast.SourceLocationMeta
 import org.partiql.lang.errors.ErrorCode
 import org.partiql.lang.errors.Property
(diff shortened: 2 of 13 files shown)


lang/src/org/partiql/lang/domains/util.kt
```diff
@@ -16,6 +17,9 @@ fun PartiqlAst.Builder.id(name: String) =
 
 val MetaContainer.staticType: StaticTypeMeta? get() = this[StaticTypeMeta.TAG] as StaticTypeMeta?
 
+fun metaContainerOf(vararg metas: Meta): MetaContainer =
```
alancai98, line 20: Missing some kdoc. I think you can just reuse what the original function uses https://github.com/partiql/partiql-lang-kotlin/blob/03b30653480d1d315dd321e5305183d93e2efd9e/lang/src/org/partiql/lang/ast/meta.kt#L174-L175

lang/src/org/partiql/lang/eval/Exceptions.kt
```diff
@@ -110,18 +110,18 @@ fun fillErrorContext(errorContext: PropertyValueMap, location: SourceLocationMet
     }
 }
 
-fun fillErrorContext(errorContext: PropertyValueMap, metaContainer: MetaContainer) {
-    val location = metaContainer.find(SourceLocationMeta.TAG) as? SourceLocationMeta
+fun fillErrorContext(errorContext: PropertyValueMap, metaContainer: com.amazon.ionelement.api.MetaContainer) {
```
alancai98, line 113: nit: don't need the `com.amazon.ionelement.api.` part before `MetaContainer` since it was imported

lang/src/org/partiql/lang/eval/Exceptions.kt
```diff
@@ -110,18 +110,18 @@ fun fillErrorContext(errorContext: PropertyValueMap, location: SourceLocationMet
...
+    val location = metaContainer[SourceLocationMeta.TAG] as? SourceLocationMeta
     if(location != null) {
         fillErrorContext(errorContext, location)
     }
 }
 
-fun errorContextFrom(metaContainer: MetaContainer?): PropertyValueMap {
+fun errorContextFrom(metaContainer: com.amazon.ionelement.api.MetaContainer?): PropertyValueMap {
```
alancai98, line 120: nit: don't need the `com.amazon.ionelement.api.` part before `MetaContainer` since it was imported

lang/src/org/partiql/lang/eval/Thunk.kt
```diff
@@ -133,7 +133,7 @@ internal abstract class ThunkFactory(
         // When this check is enabled we throw an exception the [MetaContainer] does not have a
         // [StaticTypeMeta].  This indicates a bug or unimplemented support for an AST node in
         // [StaticTypeInferenceVisitorTransform].
-        val staticType = metas.staticType?.type ?: error("Metas collection does not have a StaticTypeMeta")
+        val staticType = (metas["\$static_type"]as StaticTypeMeta?)?.type ?: error("Metas collection does not have a StaticTypeMeta")
```
alancai98, line 136: Can you modify the `.staticType` value for `StaticTypeMeta` rather than passing in the string: https://github.com/partiql/partiql-lang-kotlin/blob/03b30653480d1d315dd321e5305183d93e2efd9e/lang/src/org/partiql/lang/ast/StaticTypeMeta.kt#L82

alancai98:
Can we also make it more clear that `ExprNode`'s `MetaContainer` is deprecated in favor of `IonElement`'s`MetaContainer`? I would add `@Deprecated` annotations to the interfaces and public functions in [meta.kt](https://github.com/partiql/partiql-lang-kotlin/blob/03b30653480d1d315dd321e5305183d93e2efd9e/lang/src/org/partiql/lang/ast/meta.kt)
- `Meta` interface
- `HasMetas` interface
- `MetaContainer` interface
- `metaContainerOf` function
- `emptyMetaContainer` function
- `metaContainerOf` function

alancai98 (changes requested):

lziq:
> Can we also make it more clear that `ExprNode`'s `MetaContainer` is deprecated in favor of `IonElement`'s`MetaContainer`? I would add `@Deprecated` annotations to the interfaces and public functions in [meta.kt](https://github.com/partiql/partiql-lang-kotlin/blob/03b30653480d1d315dd321e5305183d93e2efd9e/lang/src/org/partiql/lang/ast/meta.kt)
> 
>     * `Meta` interface
> 
>     * `HasMetas` interface
> 
>     * `MetaContainer` interface
> 
>     * `metaContainerOf` function
> 
>     * `emptyMetaContainer` function
> 
>     * `metaContainerOf` function

I think we need to add `@Deprecated` annotation only to the following: 
>     * `MetaContainer` interface
> 
>     * `metaContainerOf` function
> 
>     * `emptyMetaContainer` function
> 
>     * `metaContainerOf` function

As we still use `Meta` and it is not being deprecated.

alancai98:
> > Can we also make it more clear that `ExprNode`'s `MetaContainer` is deprecated in favor of `IonElement`'s`MetaContainer`? I would add `@Deprecated` annotations to the interfaces and public functions in [meta.kt](https://github.com/partiql/partiql-lang-kotlin/blob/03b30653480d1d315dd321e5305183d93e2efd9e/lang/src/org/partiql/lang/ast/meta.kt)
> > ```
> > * `Meta` interface
> > 
> > * `HasMetas` interface
> > 
> > * `MetaContainer` interface
> > 
> > * `metaContainerOf` function
> > 
> > * `emptyMetaContainer` function
> > 
> > * `metaContainerOf` function
> > ```
> 
> I think we need to add `@Deprecated` annotation only to the following:
> 
> > ```
> > * `MetaContainer` interface
> > 
> > * `metaContainerOf` function
> > 
> > * `emptyMetaContainer` function
> > 
> > * `metaContainerOf` function
> > ```
> 
> As we still use `Meta` and it is not being deprecated.

Ok, my mistake. Only those four should have `@Deprecated` added.

lang/src/org/partiql/lang/domains/util.kt
lziq: Will add the kdoc in the next commit.

lang/src/org/partiql/lang/eval/Exceptions.kt
lziq: Will remove them in the next commit.

lang/src/org/partiql/lang/eval/Exceptions.kt
lziq: Will remove them in the next commit.

lang/src/org/partiql/lang/eval/Thunk.kt
lziq: Will make the change in the next commit.

alancai98 (approved):
:shipit:

Issue #546, opened 2022-03-10
Pretty Print a Parsed Tree


lziq:
Currently, there are some problems with CLI to show a parsed tree: 
<img width="602" alt="Screen Shot 2022-03-09 at 5 29 05 PM" src="https://user-images.githubusercontent.com/67429381/157569148-6781a920-5c77-4e18-a061-a6fd1f57573e.png">
1. It does not look very concise. 
2. some fields are are to understand, such as the 3 `null`s as shown above. 

Thus, we need a way to pretty print a parsed tree.

lziq:
Corresponding to PIG AST, we can create a recursive tree structure, `FeatureTree`, which has a list of child `FeatureTree`, and it has several attributes which reflect the main feature of a PIG AST node. 

To pretty print a PIG AST, we first transform it into `FeatureTree`, then we recursively pretty print the `FeatureTree`.

PR #547, opened 2022-03-10
Pretty Print a Parsed Tree

lziq:
*Description of changes:*
1. Added  `RecursionTree` 
2. Finished all the nodes transformation from PIG AST to `RecursionTree`.
3. Added test cases for pretty print all the PIG AST nodes.


cli/src/org/partiql/cli/prettyprintAST/PrettyPrinter.kt
```diff
@@ -0,0 +1,503 @@
...
+            }
+            is PartiqlAst.Statement.Dml -> TODO()
+            is PartiqlAst.Statement.Ddl -> TODO()
+            is PartiqlAst.Statement.Exec -> TODO()
+        }
+
+        return featureTree.convertToString()
+    }
+
+    private fun toFeatureTree(node: PartiqlAst.Expr, attrOfParent: String? = null): FeatureTree {
```
dlurton, line 35: Consider using the generated `org.partiql.lang.domains.PartiqlAst.Converter<T>`, which was designed for this kind of thing.

cli/src/org/partiql/cli/prettyprintAST/FeatureTree.kt
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @param astType is a string of the PIG AST node type
+ * @param value is the value in case node type is [PartiqlAst.Expr.Lit]
+ * @param attrOfParent is a string which represents which attribute it belongs to its parent
+ * @param children is a list of child FeatureTree
+ *
+ * The first 3 parameters are chosen as they represent the main feature of a PIG AST node.
+ *
+ */
+class FeatureTree(astType: String, value: String? = null, attrOfParent: String? = null, children: List<FeatureTree>? = null) {
```
dlurton, line 18: Why the name "feature"?  Doesn't really seem related to what this actually doing.

Why is `FeatureTree` better than s-expressions?

In general, there is going to be a long term maintenance cost to this, since it touches every node in the AST.  Is there a way to avoid having to touch every single node?

lziq: This is my COW task. 

1. What's your suggested name for this tree structure? (I'm not good at naming : ) )
2. As described in https://github.com/partiql/partiql-lang-kotlin/issues/546, I think the current s-expressions for a parsed tree is not readable enough. 
3. Since the PIG AST does not have something like a list of children, I think we have to touch each node, otherwise, there is no way to know the child tree nodes for each AST node.

cli/src/org/partiql/cli/prettyprintAST/PrettyPrinter.kt
lziq: I tried to use it, but I think it makes the code harder to understand. The reason is that only a part of PIG AST nodes have `Converter` interface. So if we convert some of the nodes using the interface, and some of nodes not using it, I think it does not look really good.

cli/src/org/partiql/cli/prettyprintAST/PrettyPrinter.kt
```diff
@@ -0,0 +1,508 @@
...
+
+import com.amazon.ion.system.IonSystemBuilder
+import org.partiql.lang.domains.PartiqlAst
+import org.partiql.lang.syntax.SqlParser
+import org.partiql.pig.runtime.SymbolPrimitive
+
+/**
+ * This class is used to pretty print PIG AST.
+ */
+class PrettyPrinter {
```
am357, line 11: Does it make sense to move this class and the `RecursiveTree` structure to the `lang` package? I can see some customers are asking for a `PrettyPrinter` that they can consume from within `lang`.

cli/src/org/partiql/cli/prettyprintAST/RecursionTree.kt
```diff
@@ -0,0 +1,51 @@
...
+import org.partiql.lang.domains.PartiqlAst
+
+/**
+ * PIG AST is not a recursive data structure, thus it is not easy to transform it directly to a pretty printed string.
+ * So we need to first transform it into RecursionTree, which is a recursive tree structure (it has a list of children which are also RecursionTree),
+ * then we can recursively pretty print the RecursionTree as we want.
+ *
+ * @param astType is a string of the PIG AST node type
+ * @param value is the value in case node type is [PartiqlAst.Expr.Lit]
+ * @param attrOfParent is a string which represents which attribute it belongs to its parent
```
am357, line 12: nit: an example would be helpful here.

cli/src/org/partiql/cli/prettyprintAST/PrettyPrinter.kt
am357: My thinking is in that case, users _must_ use CLI for this, even if they don't need to.

lziq: Seems our customer also asks for it here https://github.com/partiql/partiql-lang-kotlin/issues/552

am357 (approved):
LGTM

PR #558, opened 2022-03-24
Implemented a basic PartiQL query formmater

lziq:
Implemented PartiQL query formatter. For now, only the following scenarios for nested queries are supported
* SFW/CASE in SFW
* SFW/CASE in DML

The following scenarios for nested queries are not supported:
* SFW/CASE in CASE
* SFW/CASE in function call
* SFW/CASE in container type
* SFW/CASE in operators
    
Thus, in these cases, nested SFW/CASE queries are just printed as one line string, as the test cases below TODO comment in the test file shown.

### DIFF
--- a/lang/src/org/partiql/lang/prettyprint/QueryPrettyPrinter.kt
+++ b/lang/src/org/partiql/lang/prettyprint/QueryPrettyPrinter.kt
@@ -133,7 +133,18 @@ class QueryPrettyPrinter {
                 sb.append("INSERT INTO ")
                 writeAstNodeCheckSubQuery(dmlOp.target, sb, 0)
                 sb.append(" VALUES ")
-                writeAstNodeCheckSubQuery(dmlOp.values, sb, 0)
+                val bag = dmlOp.values as PartiqlAst.Expr.Bag
+                bag.values.forEach {
+                    val list = it as PartiqlAst.Expr.List
+                    sb.append('(')
+                    list.values.forEach { value ->
+                        writeAstNodeCheckSubQuery(value, sb, 0)
+                        sb.append(", ")
+                    }
+                    sb.removeLast(2)
+                    sb.append("), ")
+                }
+                sb.removeLast(2)
             }
             is PartiqlAst.DmlOp.InsertValue -> {
                 sb.append("INSERT INTO ")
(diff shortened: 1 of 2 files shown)


am357 (dismissed):

lang/src/org/partiql/lang/prettyprint/QueryPrettyPrinter.kt
```diff
@@ -0,0 +1,900 @@
...
+ *
+ * The idea is to use a StringBuilder and write a pretty-printed query to it according to the like
+ * of the parsed tree
+ */
+class QueryPrettyPrinter {
+    fun prettyPrintQuery(query: String): String {
+        val ion = IonSystemBuilder.standard().build()
+        val ast = SqlParser(ion).parseAstStatement(query)
+
+        return astToPrettyQuery(ast)
```
am357, line 25: nit: could also do the following:
```kotlin
return astToPrettyQuery(
 SqlParser(IonSystemBuilder
   .standard()
   .build()).parseAstStatement(query));
```

lang/src/org/partiql/lang/prettyprint/QueryPrettyPrinter.kt
```diff
@@ -0,0 +1,900 @@
...
+            is PartiqlAst.Type.SparkLong -> sb.append("SPARK_LONG")
+            is PartiqlAst.Type.SparkShort -> sb.append("SPARK_SHORT")
+            is PartiqlAst.Type.StringType -> sb.append("STRING")
+            is PartiqlAst.Type.StructType -> sb.append("STRUCT")
+            is PartiqlAst.Type.SymbolType -> sb.append("SYMBOL")
+            is PartiqlAst.Type.TimeType -> sb.append("TIME")
+            is PartiqlAst.Type.TimeWithTimeZoneType -> sb.append("TIME WITH TIME ZONE")
+            is PartiqlAst.Type.TimestampType -> sb.append("TIMESTAMP")
+            is PartiqlAst.Type.TupleType -> sb.append("TUPLE")
+            is PartiqlAst.Type.CustomType -> error("CustomType is not supported yet. ")
```
am357, line 801: For now, could we put the `CustomType.name` here with a TODO comment?

am357 (approved):

PR #559, opened 2022-03-28
For ExprFunction, replace Environment with EvaluationSession

dlurton:
I am working on a significant evaluator refactor to work with the new relational algebra.  There is a requirement that this be implemented as a refactored copy of the original `EvaluatingCompiler` to allow customers to switch between the old and new evaluators via runtime configuration.  One challenge associated with this is: the new evaluator does not use the same `Environment` class of the legacy evaluator and (until this PR) and the `ExprFunction.call*` functions all accepted `Environment` as a first argument.  This PR changes that argument to `session: EvaluationSession` instead.  Unfortunately, this is a breaking API change and customers who have their own `ExprFunction` implementations.  (Although this has gone through a breaking change in the last release anyway and most of our customers probably haven't migrated yet.)   There is, however, no other way that I can see to accomplish the goal of exposing the new physical plan evaluator as an alternative to the legacy AST evaluator.

Aside from the above:

- Exposing `Environment` as public API wasn't a great idea because it exposed implementation details that the customer should never depend on (i.e. all of the properties [here](https://github.com/partiql/partiql-lang-kotlin/blob/d6ad8d202797c0b1afbccb91dcbc11a507c7fdf7/lang/src/org/partiql/lang/eval/Environment.kt#L29-L33) except for `session`).
- No customer that I am aware of uses the `env: Environment` parameter.  Nor is there any reason to that I am aware of, outside of accessing its `session: EvaluationSession` property.

This PR also makes `Environment` and several other types `internal` as well.  These other types were all implementation details that customers should not be depending on and could not be made `internal` previously due to the requirement that `Environment` be public.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,4 @@
 org.gradle.parallel=true
 org.gradle.caching=true
+#kotlin.daemon.jvmargs="-Xmx8g"
+org.gradle.jvmargs="-Xmx8g"
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,2 @@
 org.gradle.parallel=true
 org.gradle.caching=true
-#kotlin.daemon.jvmargs="-Xmx8g"
-org.gradle.jvmargs="-Xmx8g"
--- a/lang/src/org/partiql/lang/eval/Environment.kt
+++ b/lang/src/org/partiql/lang/eval/Environment.kt
@@ -25,7 +25,7 @@ import java.util.TreeMap
  * @param session The evaluation session.
  * @param groups The map of [Group]s that is currently being built during query execution.
  */
-data class Environment(
+internal data class Environment(
     internal val locals: Bindings<ExprValue>,
     val current: Bindings<ExprValue> = locals,
     val session: EvaluationSession,
--- a/cli/src/org/partiql/cli/functions/ReadFile.kt
+++ b/cli/src/org/partiql/cli/functions/ReadFile.kt
@@ -16,7 +16,7 @@ package org.partiql.cli.functions
 
 import com.amazon.ion.IonStruct
 import org.apache.commons.csv.CSVFormat
-import org.partiql.lang.eval.Environment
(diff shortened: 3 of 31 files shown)


lang/src/org/partiql/lang/eval/Environment.kt
```diff
@@ -25,7 +25,7 @@ import java.util.TreeMap
  * @param session The evaluation session.
  * @param groups The map of [Group]s that is currently being built during query execution.
  */
-data class Environment(
+internal data class Environment(
```
lziq, line 28: Currently the kdoc is "the environment for execution". Do you think is makes more sense to change it to "variables environment used during evaluation"?

gradle.properties
```diff
@@ -1,2 +1,4 @@
 org.gradle.parallel=true
 org.gradle.caching=true
+#kotlin.daemon.jvmargs="-Xmx8g"
+org.gradle.jvmargs="-Xmx8g"
```
lziq, line 4: Why do we need to add these 2 lines?

lziq (commented):
So I think this PR mainly changes `Environment` from public to internal. I think we should do so, even if this could breaks users' code, as users should not access it by design.

gradle.properties
dlurton: I forgot to remove this--it might be needed a bit later.

lang/src/org/partiql/lang/eval/Environment.kt
dlurton: I think this version of the class is not *just* for variables--it includes `groups` too so the existing wording is more appropriate, IMHO.  Anyway, perhaps that is outside the scope of this CR.

lziq: Then can we change it as "environment used during evaluation"? I think "execution" in the original definition is not really clear.

lziq (approved):

Issue #541, opened 2022-03-03
Add warning suppressions to files using `ExprNode`


alancai98:
After deprecating `ExprNode` and adding deprecation warnings, the current gradle build emits a lot of warnings (see example [GitHub actions run](https://github.com/partiql/partiql-lang-kotlin/runs/5401106902?check_suite_focus=true#step:4:126)). For files that heavily use `ExprNode`, we should add suppress the deprecated warnings. Files with heavy usage include (but not limited to):
- AstNodeTest.kt
- SqlDataTypeTest.kt
- AstDeserialization.kt
- StatementToExprNode.kt
- ast.kt
- AstRewriterBase.kt
- ...

dlurton:
+1, came to create this issue, found that it already exists.

Issue #256, opened 2020-07-08
Enable "warnings as errors" for all gradle projects


dlurton:
We currently have a few compiler warnings during the build.  

We should fix them and enable the "warnings as errors" feature of the Kotlin compiler for all projects.

PR #562, opened 2022-04-04
Fix all compiler warnings

dlurton:
I've been doing some extensive work on this code base lately and the compiler warnings (which numbered ~900, IIRC) were driving me crazy since I couldn't tell if I was adding new warnings or not, so I decided to knock this out this morning.  It took only about 1.5 hours since most of them were related to `ExprNode` deprecation.  

In the future, it will be easy to tell if there are new warnings or not and we should consider a way of enforcing this.

This addresses #541 and all other warnings as well.

### DIFF
--- a/lang/test/org/partiql/lang/eval/EvaluatorTestBase.kt
+++ b/lang/test/org/partiql/lang/eval/EvaluatorTestBase.kt
@@ -12,6 +12,9 @@
  *  language governing permissions and limitations under the License.
  */
 
+// We don't need warnings about deprecated ExprNode.
+@file:Suppress("DEPRECATION")
+
 package org.partiql.lang.eval
 
 import com.amazon.ion.IonType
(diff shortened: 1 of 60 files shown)


lang/test/org/partiql/lang/eval/EvaluatorTestBase.kt
```diff
@@ -50,7 +53,6 @@ import kotlin.test.assertEquals
  *
  * As we parameterize PartiQL's other tests, we should migrate them away from using this base class as well.
  */
-@Deprecated("This class and everything in it should be considered deprecated.")
```
alancai98, line 53: Always wondered about why `EvaluatorTestBase` was deprecated. Any insight into its history?

alancai98 (approved):
:shipit: Thanks for fixing all the warnings! Left one non-blocking question out of curiosity

lang/test/org/partiql/lang/eval/EvaluatorTestBase.kt
dlurton: The intention was to migrate to the new test DSL that I made. That migration however never happened and the deprecation was just causing a ton of compiler warnings.

alancai98:
Since this PR resolves all the current warnings, may want to also enable future warnings as errors #256.

PR #563, opened 2022-04-04
Migrate to PIG 0.5.0

dlurton:
Migrates to the most recent version of PIG.  PIG now puts each domain into its own output files instead of one uber `partiql-domains.kt` file, which until this PR, took a ton of memory to compile and made IntelliJ's code completion choke with only 3-4 domains permuted from `partiql_ast`.

This modifies the gradle build such that:

- Uses PIG `0.5.0` instead of `0.4.0`.
- PIG output is new placed in `lang/build/generated/src/pig` (instead of `lang/domains`)
- The old `partiql-domains.kt` file is deleted automatically (to avoid having to remember to do this manually after working with an older revision of this repo).


lang/jmh/org/partiql/jmh/PartiQLBenchmark.kt
```diff
@@ -91,4 +90,4 @@ open class PartiQLBenchmark {
         val exprValue = state.expression.eval(state.session)
         blackhole.consume(exprValue)
     }
-}
```
dlurton, line 94: Not sure what's going on here but for some odd reason ktlint decided it no longer liked the lack of newline here in my local build.  Not sure why it was ok with it previously...

lang/build.gradle
```diff
@@ -58,32 +58,34 @@ dependencies {
...
 
-sourceSets {
-    jmh {
-        java.srcDirs = ['src/jmh/java']
-        resources.srcDirs = ['src/jmh/resources']
-        compileClasspath += sourceSets.main.runtimeClasspath
-    }
-}
-
-def partiqlDomainFile = 'partiql-domains.kt'
```
alancai98, line 69: Could the .gitignore in the `lang/domains` folder also be deleted in this PR?

https://github.com/partiql/partiql-lang-kotlin/blob/5a073ca0a2ea4308aad7acd041d66fdc589dcef5/lang/src/org/partiql/lang/domains/.gitignore#L1-L2

alancai98 (approved):

PR #286, opened 2020-09-21
Appropriately fast `LIKE` pattern compilation

dlurton:
Implements #284 

This is able to compile the like pattern `%<n>%` where `<n>` is 8000 `!` characters on my local machine in ~60ms.  The previous implementation took too long to measure on my local machine, even after #279, because it used an an unoptimized state machine.  I do not know if there is a name for the algorithm in this PR, but it does not use a state machine.

I haven't yet analysed the performance of the evaluating the compiled pattern--I will need to do that before this can be merged.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #575, opened 2022-04-13
Replace implementation of `LIKE` matching via compilation to `java.util.regex.Pattern`

jpschorr:
Implement `LIKE` matching via compilation to `java.util.regex.Pattern`

See also https://github.com/partiql/partiql-lang-kotlin/pull/286
See also https://github.com/partiql/partiql-lang-rust/pull/98


jpschorr:
JMH results from `./gradlew jmh`:

using `PatternParts`:
```
Benchmark                                   Mode  Cnt       Score      Error  Units
MultipleLikeBenchmark.testPartiQLCompiler   avgt   10     107.344 ±    3.372  us/op
MultipleLikeBenchmark.testPartiQLEvaluator  avgt   10  662893.671 ± 4098.512  us/op
MultipleLikeBenchmark.testPartiQLParser     avgt   10     190.963 ±    3.631  us/op
PartiQLBenchmark.testPartiQLCompiler        avgt   10       6.634 ±    0.083  us/op
PartiQLBenchmark.testPartiQLEvaluator       avgt   10       1.626 ±    0.031  us/op
PartiQLBenchmark.testPartiQLParser          avgt   10       8.085 ±    0.045  us/op
```


using `Pattern` (i.e. via `java.util.regex`):
```
Benchmark                                   Mode  Cnt       Score       Error  Units
MultipleLikeBenchmark.testPartiQLCompiler   avgt   10      97.674 ±    27.887  us/op
MultipleLikeBenchmark.testPartiQLEvaluator  avgt   10  297260.944 ± 19894.939  us/op
MultipleLikeBenchmark.testPartiQLParser     avgt   10     186.545 ±     2.887  us/op
PartiQLBenchmark.testPartiQLCompiler        avgt   10       6.625 ±     0.148  us/op
PartiQLBenchmark.testPartiQLEvaluator       avgt   10       1.809 ±     0.283  us/op
PartiQLBenchmark.testPartiQLParser          avgt   10       8.306 ±     0.321  us/op
```

lang/src/org/partiql/lang/eval/like/Pattern.kt
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Translates a SQL-style `LIKE` pattern to a regular expression.
+ *
+ * @param likePattern A `LIKE` match pattern (i.e. a string where '%' means '.*?' and '_' means '.').
+ * @param escapeChar The escape character for the `LIKE` pattern.
+ *
+ * @return a [Pattern] which is a regular expression corresponding to the specified `LIKE` pattern.
+ */
+internal fun parsePattern(likePattern: String, escapeChar: Int?): Pattern {
```
am357, line 16: Nit: For better ergonomics, reckon we could have the common escape chars as an Enum? (E.g. `'\\'.toInt()`)

am357 (dismissed):

lang/src/org/partiql/lang/eval/like/Pattern.kt
```diff
@@ -0,0 +1,89 @@
...
+private const val ANY_MANY = '%'.toInt()
+private const val ANY_ONE = '_'.toInt()
+
+/**
+ * Translates a SQL-style `LIKE` pattern to a regular expression.
+ *
+ * @param likePattern A `LIKE` match pattern (i.e. a string where '%' means '.*?' and '_' means '.').
+ * @param escapeChar The escape character for the `LIKE` pattern.
+ *
+ * @return a [Pattern] which is a regular expression corresponding to the specified `LIKE` pattern.
```
dlurton, line 14: Perhaps add an example of a `LIKE` pattern and the equivalent regex?

lang/src/org/partiql/lang/eval/like/Pattern.kt
dlurton: Also, does `escapeChar` need to be nullable?

lang/src/org/partiql/lang/eval/like/Pattern.kt
```diff
@@ -0,0 +1,89 @@
...
+    buf.append("^")
+
+    var isEscaped = false
+    var wildcardMin: Int? = null
+    var wildcardUnbounded = false
+    var literal = StringBuilder()
+
+    // If a wildcard (e.g. a sequence of '%' and '_') has been accumulated, write out the regex equivalent
+    val flushWildcard = {
+        if (wildcardMin != null) {
```
dlurton, line 27: consider replacing `if` with `when` here, especially where `else if` is used.

lang/src/org/partiql/lang/eval/like/Pattern.kt
```diff
@@ -0,0 +1,89 @@
...
+ * @param escapeChar The escape character for the `LIKE` pattern.
+ *
+ * @return a [Pattern] which is a regular expression corresponding to the specified `LIKE` pattern.
+ */
+internal fun parsePattern(likePattern: String, escapeChar: Int?): Pattern {
+    val buf = StringBuilder(likePattern.length + 8)
+    buf.append("^")
+
+    var isEscaped = false
+    var wildcardMin: Int? = null
```
dlurton, line 21: I would avoid use of nullable values here--(these become boxed `java.lang.Integer`s).  Can you use `-1` instead?

lang/src/org/partiql/lang/eval/like/Pattern.kt
```diff
@@ -0,0 +1,89 @@
...
+                        wildcardMin = (wildcardMin ?: 0) + 1
+                    }
+                    ANY_MANY -> {
+                        flushLiteral()
+                        wildcardMin = (wildcardMin ?: 0)
+                        wildcardUnbounded = true
+                    }
+                    else -> {
+                        flushWildcard()
+                        literal.append(Character.toChars(codepoint))
```
dlurton, line 73: This may work, but can you use `literal.appendCodepoint(codepoint)` instead of `literal.append(Character.toChars(codepoint))`?

lang/src/org/partiql/lang/eval/like/Pattern.kt
```diff
@@ -0,0 +1,89 @@
...
+                    }
+                }
+            } else {
+                flushWildcard()
+                literal.append(Character.toChars(codepoint))
```
dlurton, line 78: Same

dlurton (dismissed):
Overall LGTM, my comments are mostly nits.

lang/src/org/partiql/lang/eval/like/Pattern.kt
```diff
@@ -0,0 +1,89 @@
...
+ * @return a [Pattern] which is a regular expression corresponding to the specified `LIKE` pattern.
+ */
+internal fun parsePattern(likePattern: String, escapeChar: Int?): Pattern {
+    val buf = StringBuilder(likePattern.length + 8)
+    buf.append("^")
+
+    var isEscaped = false
+    var wildcardMin: Int? = null
+    var wildcardUnbounded = false
+    var literal = StringBuilder()
```
dlurton, line 23: this should be `val` not `var`.

lang/src/org/partiql/lang/eval/like/Pattern.kt
jpschorr: According to the spec:

> The optional third argument is a character string containing exactly one character, known as the "escape character", for use when a percent or underscore is required in the pattern without its special meaning.

The [PostGres docs say](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-LIKE):

> According to the SQL standard, omitting ESCAPE means there is no escape character (rather than defaulting to a backslash), and a zero-length ESCAPE value is disallowed. PostgreSQL's behavior in this regard is therefore slightly nonstandard.

There are a lot of existing test cases that pass `null` here: https://github.com/partiql/partiql-lang-kotlin/blob/12051302413f8d3c3c87aafa7551c0c93d8d0648/lang/test/org/partiql/lang/eval/like/PatternTests.kt#L15-L148

I didn't change the signature from the removed PatternParts: https://github.com/partiql/partiql-lang-kotlin/blob/9da23cfb07672580d4b56b671ad18f044a850e60/lang/src/org/partiql/lang/eval/like/PatternPart.kt#L27

lang/src/org/partiql/lang/eval/like/Pattern.kt
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Translates a SQL-style `LIKE` pattern to a regular expression.
+ *
+ * @param likePattern A `LIKE` match pattern (i.e. a string where '%' means '.*?' and '_' means '.').
+ * @param escapeChar The escape character for the `LIKE` pattern.
+ *
+ * @return a [Pattern] which is a regular expression corresponding to the specified `LIKE` pattern.
+ */
+internal fun parsePattern(likePattern: String, escapeChar: Int?): Pattern {
+    val buf = StringBuilder(likePattern.length + 8)
```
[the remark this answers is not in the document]
alancai98: What does the '8' here represent?

lang/test/org/partiql/lang/eval/like/PatternTests.kt
```diff
@@ -151,32 +151,60 @@ class PatternPartTests {
...
+            parsePattern("%foo%%%%bar*%baz%", '*'.toInt()).pattern(),
+        )
+        assertEquals("^.\\Qfoo\\E$", parsePattern("_foo", '\\'.toInt()).pattern())
+        assertEquals("^\\Qfoo\\E.$", parsePattern("foo_", '\\'.toInt()).pattern())
+        assertEquals("^\\Qfoo\\E.\\Qbar\\E$", parsePattern("foo_bar", '\\'.toInt()).pattern())
+        assertEquals("^\\Qfoo\\E.{2,2}\\Qbar\\E$", parsePattern("foo__bar", '\\'.toInt()).pattern())
+        assertEquals(
+            "^\\Qfoo\\E.\\Q.*?\\E.\\Qbar\\E$",
+            parsePattern("foo_.*?_bar", '\\'.toInt()).pattern(),
+        )
```
alancai98, line 201: Could also add a test for a pattern with at least two of a character (i.e. '_'). Example query: `foo_%_bar`, `foo_%_%bar`, or `foo%_%%_%bar`

lang/src/org/partiql/lang/eval/like/Pattern.kt
alancai98: There are some examples in PatternTests.kt but some equivalent regex would be nice here

alancai98 (approved):

alancai98:
Btw, I found https://regexr.com/ useful in explaining some regex patterns.

alancai98 (approved):

am357 (dismissed):

jpschorr:
I added some more benchmarks to demonstrate scaling:
- the "<x>15" tests are the same as above, with 15 `OR`ed `LIKE`s
- the "<x>30" tests add an additional 15 for a total of 30 `OR`ed `LIKE`s
- "<x>30WithData10" test runs the evaluation of the 30 `OR`ed `LIKE`s against 10x the data

Data demonstrates:
- that evaluation of both `PatternParts` and `java.util.regex` scale linearly in the number of `OR`ed `LIKE`s and the number of 'rows' evaluated
- the `java.util.regex` implementation runs in ~42% of the time (or a ~2.3x speedup)


JMH results from `./gradlew jmh`:

using `PatternParts:`

```
Benchmark                                                Mode  Cnt         Score        Error  Units
MultipleLikeBenchmark.testPartiQLCompiler15             avgt   10        95.412 ±     32.856  us/op
MultipleLikeBenchmark.testPartiQLCompiler30             avgt   10       142.728 ±      4.676  us/op
MultipleLikeBenchmark.testPartiQLEvaluator15            avgt   10    687755.031 ±  99042.520  us/op
MultipleLikeBenchmark.testPartiQLEvaluator30            avgt   10   1335957.886 ±  35773.013  us/op
MultipleLikeBenchmark.testPartiQLEvaluator30WithData10  avgt   10  13276802.415 ± 141398.143  us/op
MultipleLikeBenchmark.testPartiQLParser15               avgt   10       195.902 ±      5.871  us/op
MultipleLikeBenchmark.testPartiQLParser30               avgt   10       378.426 ±      9.067  us/op
```

using Pattern (i.e. via `java.util.regex`):

```
Benchmark                                               Mode  Cnt        Score        Error  Units
MultipleLikeBenchmark.testPartiQLCompiler15             avgt   10      108.562 ±     36.045  us/op
MultipleLikeBenchmark.testPartiQLCompiler30             avgt   10      159.411 ±     11.700  us/op
MultipleLikeBenchmark.testPartiQLEvaluator15            avgt   10   289924.568 ±   5236.064  us/op
MultipleLikeBenchmark.testPartiQLEvaluator30            avgt   10   561482.476 ±   5531.718  us/op
MultipleLikeBenchmark.testPartiQLEvaluator30WithData10  avgt   10  5713322.505 ± 246204.236  us/op
MultipleLikeBenchmark.testPartiQLParser15               avgt   10      198.732 ±      4.763  us/op
MultipleLikeBenchmark.testPartiQLParser30               avgt   10      381.148 ±     18.694  us/op
```

lang/jmh/org/partiql/jmh/MultipleLikeBenchmark.kt
```diff
@@ -0,0 +1,403 @@
...
+    @BenchmarkMode(Mode.AverageTime)
+    @OutputTimeUnit(TimeUnit.MICROSECONDS)
+    @Fork(value = 2)
+    fun testPartiQLParser15(state: MyState, blackhole: Blackhole) {
+        val expr = state.parser.parseAstStatement(state.query15)
+        blackhole.consume(expr)
+    }
+
+    /**
+     * Example PartiQL benchmark for compiling a query
```
am357, line 330: nit: sorry for missing this earlier, but would be great have the explanation of each benchmark in the kdocs.

alancai98 (dismissed):

am357 (approved):
LGTM w/ only one nit.

am357 (approved):

PR #578, opened 2022-04-15
Parse expectedResult instead of query fixing a bug in the evaluator test framework

dlurton:
Glad I caught this right away, could have been very confusing.  This makes me realize that my next PR should be some additional for the new evaluator test framework.


alancai98 (approved):
Ahh missed from the previous review. It's good we caught this now

PR #584, opened 2022-04-29
PIG domains for planner and plan evaluator

dlurton:
Originally part of #582 and based on #586  (which must be merged first) this PR contains all the PIG domain (tree data structure) changes needed for the planner and plan evaluator.  Documentation in `partiql.ion` should be relatively self-explanatory, please let me know if not.

### DIFF
--- a/lang/resources/org/partiql/type-domains/partiql.ion
+++ b/lang/resources/org/partiql/type-domains/partiql.ion
@@ -564,12 +564,12 @@ may then be further optimized by selecting better implementations of each operat
                 (filter predicate::expr source::bexpr)
 
                 // Basic join operator.  Covers cross, inner, left, right and full joins.
-                // For cross joins, set `join_type` to `(inner)` and the `predicate` to `(lit true)`.
+                // To represent a cross join, set `predicate` to `null`.
                 (join
                     join_type::join_type
                     left::bexpr
                     right::bexpr
-                    predicate::expr)
+                    predicate::(? expr))
 
                 // Skips `row_count` rows, then emits all remaining rows.
                 (offset row_count::expr source::bexpr)
--- a/lang/resources/org/partiql/type-domains/partiql.ion
+++ b/lang/resources/org/partiql/type-domains/partiql.ion
@@ -471,13 +471,13 @@ may then be further optimized by selecting better implementations of each operat
             // that is included in the final struct.
             (sum struct_part
                 // For `.*` in SELECT list
-                // If `<expr>` is a struct, the fields of that struct will be part of the merged struct.
-                // If `<expr>` is not a struct, The field `_n` will be included in the struct, where `n` is the
-                // ordinal of the field in the final merged struct. If `expr` returns a container that is not a struct,
-                // field names will be assigned in the format of `_n` where `n` is the ordinal position of the
-                // field within the merged struct.  If `expr` returns a scalar value, it will be coerced into a
-                // singleton bag expr and the previous logic will apply.
-                (struct_fields expr::expr)
+                // If `<part_expr>` is a struct, the fields of that struct will be part of the merged struct.
+                // If `<part_expr>` is not a struct, The field `_n` will be included in the struct, where `n` is the
+                // ordinal of the field in the final merged struct. If `<part_expr>` returns a container that is not a
+                // struct, field names will be assigned in the format of `_n` where `n` is the ordinal position of the
+                // field within the merged struct.  If `<part_expr>` returns a scalar value, it will be coerced into a
+                // singleton bag and the previous logic will apply.
+                (struct_fields part_expr::expr)
 
                 // For `<expr> [AS <id>]`.  If `field_name` returns a non-text value, in legacy mode an exception
                 // will be thrown.  In permissive mode, the field will be excluded from the final struct.
--- a/lang/resources/org/partiql/type-domains/partiql.ion
+++ b/lang/resources/org/partiql/type-domains/partiql.ion
@@ -528,7 +528,7 @@ may then be further optimized by selecting better implementations of each operat
             from_source
         )
 
-        // Change let_binding so that it has a var_decl instead of a only name to represent bindings.
+        // Change let_binding so that it has a var_decl instead of only a name to represent bindings.
         (exclude let_binding)
         (include (product let_binding value::expr decl::var_decl))
 
--- a/lang/resources/org/partiql/type-domains/partiql.ion
+++ b/lang/resources/org/partiql/type-domains/partiql.ion
@@ -475,8 +475,8 @@ may then be further optimized by selecting better implementations of each operat
                 // If `<expr>` is not a struct, The field `_n` will be included in the struct, where `n` is the
                 // ordinal of the field in the final merged struct. If `expr` returns a container that is not a struct,
                 // field names will be assigned in the format of `_n` where `n` is the ordinal position of the
-                // field within the struct.  If `expr` returns a scalar value, it will be coerced into a singleton bag
-                // expr and the previous logic will apply.
+                // field within the merged struct.  If `expr` returns a scalar value, it will be coerced into a
+                // singleton bag expr and the previous logic will apply.
                 (struct_fields expr::expr)
 
                 // For `<expr> [AS <id>]`.  If `field_name` returns a non-text value, in legacy mode an exception
@@ -499,15 +499,18 @@ may then be further optimized by selecting better implementations of each operat
                 // a struct-union expression.
                 //
                 // Example as struct constructor:
-                //    (struct (struct_field (lit a) (lit 42))
-                // Returns: { a: 42 }
+                //    (struct
+                //        (struct_field (lit a) (lit 42))
+                //        (struct_field (lit b) (lit 43)))
+                // Returns: { a: 42, b: 43 }
                 //
-                // Example as a struct-union. Given a global environment `{| foo: { a: 42 }, bar: { b: 43} |}`, then:
+                // Example as a struct-union. Given a global environment with `foo` bound to `{ a: 42 }` and `bar`
+                // bound to `{ b: 43}`, then:
                 //     (struct
                 //         (struct_fields (id foo))
                 //         (struct_fields (id bar)))
                 // Returns { a: 42, b: 43 }
-                // Note that `struct_field` and `struct_fields` may be used in combination also:
+                // Note that `struct_field` and `struct_fields` may be used in combination:
                 //     (struct
                 //         (struct_fields (id foo))
                 //         (struct_fields (id bar))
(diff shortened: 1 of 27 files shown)


alancai98 (commented):
I gave a cursory review of the PIG domains defined in `partiql.ion`. Looks reasonable based off what I remember from some offline discussions. I think @jpschorr and @am357 should also take a look to check the plans align with the reviewed query plan docs.

I also think it would be good to include some of the docs part of those offline discussions in the `partiql-docs` repo since the query plan and planner are to be open-sourced and public APIs.

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -426,4 +454,279 @@
...
+
+                // `struct` is the primary struct constructor and also encapsulates semantics needed for
+                // `SELECT <expr>.*`, and `SELECT <expr> AS y`.  It can be used as a regular struct constructor, or as
+                // a struct-union expression.
+                //
+                // Example as struct constructor:
+                //    (struct (struct_field (lit a) (lit 42))
+                // Returns: { a: 42 }
+                //
+                // Example as a struct-union. Given a global environment `{| foo: { a: 42 }, bar: { b: 43} |}`, then:
```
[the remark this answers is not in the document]
dlurton: It's supposed to denote a bindings environment which is similar to but different than a struct.  However, I really don't feel like providing a section describing notation here so I'll change this to be more clear without the notation.

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -426,4 +454,279 @@
...
+                // that, see bexpr.project in the partiql_physical domain.)  If evaluating `expr` results in a scalar
+                // value, it is converted into a singleton bag.
+                (scan expr::expr as_decl::var_decl at_decl::(? var_decl) by_decl::(? var_decl))
+
+                // Evaluates `predicate` within the scope of every row of `bexpr`, and only returns those
+                // rows for which `predicate` returns true.
+                (filter predicate::expr source::bexpr)
+
+                // Basic join operator.  Covers cross, inner, left, right and full joins.
+                // For cross joins, set `join_type` to `(inner)` and the `predicate` to `(lit true)`.
```
am357, line 565: For this and the other join definition in physical_plan: curious why going w/ this approach for representing cross join? Does going w/ the following make sense?

```bash
(sum join_type (cross) (inner) (left) (right) (full))

(join
    join_type::join_type
    left::bexpr
    right::bexpr
    predicate::(? expr))
```

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -426,4 +454,279 @@
...
+        // For `var_decl`, `name` with `index`.  The name of the variable can still be determined by looking at the
+        // `local_variable` with the same index.
+        (exclude var_decl scope_qualifier)
+        (include
+            (product var_decl index::int)
+        )
+
+        (with expr
+            // At this point, there should be no undefined variables in the plan since the area all rewritten to
+            // dynamic lookup function call sites (if enabled).
```
am357, line 650: Can you please elaborate, enabled by what? `CompilerPipeline`? Also perhaps there is an assume knowledge here that requires expanding.

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -426,4 +454,279 @@
...
+                // A resolved reference to a variable that was defined within a query.  Otherwise known as a local
+                // variable. "Resolved" means that the variable is guaranteed to exist and we know its register index.
+                // Elements:
+                // - `index`: the index of the `var_decl` that this variable refers to, i.e. this always corresponds to
+                // the `var_decl` with the same index.
+                (local_id index::int)
+
+                // Global variable reference--typically a table although it can actually be bound to any value.  Unlike
+                // local variables, global variables are not stored in registers.  Instead, they are typically stored
+                // in persistent storage.  Evaluating a `global_id` will return a value with an open iterator.  There
```
am357, line 662: Does it make sense to change `persistent storage` to `non-ephemeral meaning the value persists after query evaluation` (or something along those lines)? This is to cater for in-memory database systems or RDBMSs that can map a table/schema to memory.

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -426,4 +454,279 @@
...
+                (version int)
+            )
+
+            // Defines a field within a struct constructor or an expression which is expected to be a container
+            // that is included in the final struct.
+            (sum struct_part
+                // For `.*` in SELECT list
+                // If `<expr>` is a struct, the fields of that struct will be part of the merged struct.
+                // If `<expr>` is not a struct, The field `_n` will be included in the struct, where `n` is the
+                // ordinal of the field in the final merged struct. If `expr` returns a container that is not a struct,
```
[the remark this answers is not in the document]
dlurton: Sorry, I'm confused as to what you are asking here.  `lhs` typically refers to the "left hand side" of a binary expression but this isn't a binary expression?

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -426,4 +454,279 @@
...
+                // field names will be assigned in the format of `_n` where `n` is the ordinal position of the
+                // field within the struct.  If `expr` returns a scalar value, it will be coerced into a singleton bag
```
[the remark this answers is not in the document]
dlurton: sure.

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -426,4 +454,279 @@
...
+            // Note that `AS` aliases specified in a select list (i.e. `SELECT x AS y` are *not* variables, they are
+            // fields.)
+            // Modeling this with a separate node (as opposed to just a symbol) is beneficial because it is easy to
+            // identify all variable declarations within a logical plan during tree traversal, and because in later
+            // permuted domains we can add information to this type such as the variable's assigned index.
+            // Elements:
+            // - `name`: the name of the variable as specified by the query author or determined statically.
+            (product var_decl name::symbol)
+
+            // The operators of PartiQL's relational algebra. See `$projectDir/docs/dev/RELATIONAL-ALGEBRA.md` for
```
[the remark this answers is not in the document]
dlurton: That particular document is out of date.  I have a draft of a better version but it needs to be cleaned up and made ready  for public consumption, so I'll add this ticket to the current milestone: https://github.com/partiql/partiql-lang-kotlin/issues/599

lang/resources/org/partiql/type-domains/partiql.ion
dlurton: I bristled when I first saw this particular modeling of cross joins as well.  This is fundamentally similar to how they are modeled with the AST today.  ...

To demonstrate why it doesn't make sense to add `(cross)` to `join_type`:  how should the evaluator behave when the `predicate` is not null and the `join_type` is `(cross)`?  

- Throw an exception?  (That would be an edge case that would have to be handled many places, which would increase the noise level in the implementation.)
- Honor the predicate anyway?  But that would be identical to an inner join, making `(cross)` redundant.

lang/resources/org/partiql/type-domains/partiql.ion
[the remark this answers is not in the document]
am357: Ah sorry, indeed, it's confusing. I meant in `expr::expr`, changing  the property name `expr`; or making the reference in the documentation more clear b/c IMO for fresh eyes, it's confusing.

lang/resources/org/partiql/type-domains/partiql.ion
dlurton: I would keep references to implementation specific details such as `CompilerPipeline` (actually `PlannerPipeline`, introduced in #590) out of this document since it can/should be used to generate the ASTs of multiple partiql implementations.  However, based on that argument, dynamic lookup function call sites are also an implementation detail.  I'll remove that.

lang/resources/org/partiql/type-domains/partiql.ion
dlurton: But the term "non-ephemeral" wouldn't apply to an in-memory only database, would it?  I'll rephrase this to be more inclusive of different types of storage systems.

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -426,4 +454,279 @@
...
+(define partiql_physical
+    (permute_domain partiql_logical_resolved
+        (include
+            // Identifies an implementation that has been selected for an instance of a physical operator and
+            // identifies any static arguments required. This will initially have the `(impl default)` value, with
+            // different implementations being selected as needed.
+            // Elements:
+            // - `name`: the unique name of the implementation.  Each operator has a different namespace containing its
+            // default and PartiQL-specific
+            (product impl name::symbol static_args::(* ion 0))
```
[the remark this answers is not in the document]
dlurton: added

lang/resources/org/partiql/type-domains/partiql.ion
am357: We could ignore the predicate when the `join_type` is `cross`. Although still not ideal, having it defined explicitly is better than implicit IMO. Thinking about this further and also [looking at the grammar](https://jakewheat.github.io/sql-overview/sql-92-grammar.html#joined-table), perhaps we an also go with the following model (or something similar)?

```
(sum join
  (product left left::bexpr predicate::expr)
  (product right right::bexpr predicate::expr)
  (product inner left::bexpr right::bexpr predicate::expr)
  (product cross  left::bexpr right::bexpr))
```

dlurton: 1.  `left`, `right` should have the same elements as `inner`--they each requires a left and right `bexpr` and a `predicate`.
2.  With this proposal, the code that consumes the generated classes is actually more complex.  I.e. in the `VisitorTransform`s, I now have to override 4 different methods instead of 1.  Pattern matching too has similar concerns.
3. With `inner`, I can still have still set the predicate to `(lit true)` and get something that's semantically equivalent to `cross`).  This makes `cross` kind of redundant.  Such redundancies generally increase complexity.

It's funny--I distinctly remember having this exact discussion (but related to `ExprNode`) with @almann and @therapon  that I had your position back then.  They argued strongly for modeling `cross` as an `inner` with a `(lit true)` predicate.  It's taken time but I've come around.

lang/resources/org/partiql/type-domains/partiql.ion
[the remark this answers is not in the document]
dlurton: How's `part_expr`?

lang/resources/org/partiql/type-domains/partiql.ion
am357: For `1`, yes that was a miss from my side. For `2` we could have a wrapper type with having a member property as `kind` that references the join sum. I see the `3` orthogonal to choice of having `cross` or not as this is still the case without it. 

I hope I come around one day, but at this stage, I just can't see having an implicit logic to model rather than a precise model is a better choice. Let me know if you want to discuss offline and we can take it from there.

dlurton: From offline discussion we will make `predicate` nullable so that if the user did not specify a predicate (as with `cross join`), the AST will not lie by saying there was a `(lit true)`.

am357 (approved):

lang/resources/org/partiql/type-domains/partiql.ion
am357: I approved, but putting this non-blocking comment here: Although we disregard a join w/ predicate as cross join, to be more defensive, I think we could still go w/ having the `(cross_join)` as a join type to ensure the absence of predicate is not accidental to infer the cross join.

PR #585, opened 2022-04-29
Exclude planner & plan evaluator from tests for features not yet implemented

dlurton:
Originally part of #582  and based on #584 (which must be merged first), this PR primarily sets the target evaluator for all tests related to the following features to be the AST evaluator, thereby excluding the planner and plan evaluator from them.  (These features are not yet on the backlog for the planner & plan evaluator.)  Note that the new planner and evaluator are not in this branch yet and will be introduced in a separate pull request.  There are also a few new tests that I added to provide simpler test cases for a few scenarios that were a little difficult to get working with the new planner and evaluator.

- `ORDER BY`
- `GROUP BY`
- `HAVING`
- Aggregate functions (`SUM`, `MAX`, `MIN`, etc)
- `PIVOT`
- `UNPIVOT`

### DIFF
--- a/lang/test/org/partiql/lang/eval/EvaluatorTestSuite.kt
+++ b/lang/test/org/partiql/lang/eval/EvaluatorTestSuite.kt
@@ -368,7 +368,7 @@ internal val EVALUATOR_TEST_SUITE: IonResultTestSuite = defineTestSuite {
             """
         )
     }
-    group("select-where") {
+    group("select_where") {
         test(
             "selectWhereStringEqualsSameCase",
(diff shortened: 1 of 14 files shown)


lang/test/org/partiql/lang/eval/EvaluatorTestSuite.kt
```diff
@@ -439,6 +439,52 @@ internal val EVALUATOR_TEST_SUITE: IonResultTestSuite = defineTestSuite {
             """
         )
 
+        test(
```
dlurton, line 442: I guess I did add a few tests below--for some reason I cannot recall, however, I see no reason to remove them.

PR #587, opened 2022-04-29
Make ThunkFactory generic 

dlurton:
Originally part of #582, this PR is based on #584 (which must be merged first).

The legacy AST evaluator's `ThunkFactory` and related types were tightly coupled to the `org.partiql.lang.eval.Environment` class.  However, the new plan evaluator replaces this class with a different one.  This PR makes `ThunkFactory` and related types generic so that a different environment class may be used with the physical plan evaluator without having to duplicate `ThunkFactory` et al.

### DIFF
--- a/lang/src/org/partiql/lang/SqlException.kt
+++ b/lang/src/org/partiql/lang/SqlException.kt
@@ -18,6 +18,7 @@ import org.partiql.lang.errors.ErrorCode
 import org.partiql.lang.errors.Property
 import org.partiql.lang.errors.PropertyValueMap
 import org.partiql.lang.errors.UNKNOWN
+import org.partiql.lang.util.propertyValueMapOf
 
 /**
  * General exception class for the interpreter.
@@ -43,10 +44,12 @@ import org.partiql.lang.errors.UNKNOWN
 open class SqlException(
     override var message: String,
     val errorCode: ErrorCode,
-    val errorContext: PropertyValueMap? = null,
+    errorContextArg: PropertyValueMap? = null,
     cause: Throwable? = null
-) :
-    RuntimeException(message, cause) {
+) : RuntimeException(message, cause) {
+
+    val errorContext: PropertyValueMap = errorContextArg ?: propertyValueMapOf()
+
 
     /**
      * Indicates if this exception is due to an internal error or not.
--- a/lang/src/org/partiql/lang/SqlException.kt
+++ b/lang/src/org/partiql/lang/SqlException.kt
@@ -50,7 +50,6 @@ open class SqlException(
 
     val errorContext: PropertyValueMap = errorContextArg ?: propertyValueMapOf()
 
-
     /**
      * Indicates if this exception is due to an internal error or not.
      *
--- a/lang/src/org/partiql/lang/SqlException.kt
+++ b/lang/src/org/partiql/lang/SqlException.kt
@@ -34,7 +34,7 @@ import org.partiql.lang.util.propertyValueMapOf
  *
  * @param message the message for this exception
  * @param errorCode the error code for this exception
- * @param propertyValueMap context for this error
+ * @param errorContextArg context for this error, contains details like line & column number among other attributes.
  * @param cause for this exception
  *
  * @constructor a custom error [message], the [errorCode], error context as a [propertyValueMap] and optional [cause] creates an
--- a/lang/src/org/partiql/lang/eval/Thunk.kt
+++ b/lang/src/org/partiql/lang/eval/Thunk.kt
@@ -21,7 +21,6 @@ import org.partiql.lang.domains.staticType
 import org.partiql.lang.errors.ErrorBehaviorInPermissiveMode
 import org.partiql.lang.errors.ErrorCode
 import org.partiql.lang.errors.Property
-import org.partiql.lang.util.propertyValueMapOf
 
 /**
  * A thunk with no parameters other than the current environment.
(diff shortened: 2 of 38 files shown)


lang/src/org/partiql/lang/eval/Thunk.kt
```diff
@@ -28,8 +28,10 @@ import org.partiql.lang.errors.Property
  * See https://en.wikipedia.org/wiki/Thunk
  *
  * This name was chosen because it is a thunk that accepts an instance of `Environment`.
```
alancai98, line 30: Comment may need updated since `Environment` was made generic. Similarly on L41.

lang/src/org/partiql/lang/eval/Thunk.kt
```diff
@@ -38,8 +40,12 @@ internal typealias ThunkEnv = (Environment) -> ExprValue
  *
  * This name was chosen because it is a thunk that accepts an instance of `Environment` and an [ExprValue] as
  * its arguments.
+ *
+ * @param TEnv The type of the environment.  Generic so that the legacy AST compiler and the new compiler may use
+ * different types here.
+ * @param TArg The type of the additional argument.
```
alancai98, line 46: `TArg` doesn't exist. Do you mean `T` here?

lang/src/org/partiql/lang/eval/Thunk.kt
```diff
@@ -60,8 +66,10 @@ internal typealias ThunkExceptionHandlerForPermissiveMode = (Throwable, SourceLo
  */
 data class ThunkOptions private constructor(
     val handleExceptionForLegacyMode: ThunkExceptionHandlerForLegacyMode = DEFAULT_EXCEPTION_HANDLER_FOR_LEGACY_MODE,
-    val handleExceptionForPermissiveMode: ThunkExceptionHandlerForPermissiveMode = DEFAULT_EXCEPTION_HANDLER_FOR_PERMISSIVE_MODE
+    val handleExceptionForPermissiveMode: ThunkExceptionHandlerForPermissiveMode = DEFAULT_EXCEPTION_HANDLER_FOR_PERMISSIVE_MODE,
+    val thunkReturnTypeAssertions: ThunkReturnTypeAssertions = ThunkReturnTypeAssertions.DISABLED,
```
alancai98, line 70: Can you add some documentation for this parameter?

lang/src/org/partiql/lang/SqlException.kt
```diff
@@ -43,10 +44,11 @@ import org.partiql.lang.errors.UNKNOWN
 open class SqlException(
     override var message: String,
     val errorCode: ErrorCode,
-    val errorContext: PropertyValueMap? = null,
+    errorContextArg: PropertyValueMap? = null,
```
alancai98, line 47: kdoc for this parameter needs to be updated.

alancai98 (commented):
Changes to `ThunkFactory` and its uses seem reasonable to me. Left some minor comments related to documentation that needs updated due to this change.

lang/src/org/partiql/lang/eval/Thunk.kt
dlurton: I think I can probably just delete these lines--I don't think they added much.

lang/src/org/partiql/lang/eval/Thunk.kt
dlurton: Meant to rename `T` to `TArg` actually.

lang/src/org/partiql/lang/eval/Thunk.kt
dlurton: done.

lang/test/org/partiql/lang/eval/EvaluatingCompilerOffsetTests.kt
```diff
@@ -110,7 +110,10 @@ class EvaluatingCompilerOffsetTests : EvaluatorTestBase() {
     @ParameterizedTest
     @ArgumentsSource(ArgsProviderValid::class)
     fun validTests(tc: EvaluatorTestCase) = runEvaluatorTestCase(
-        tc.copy(excludeLegacySerializerAssertions = true),
+        tc.copy(
+            excludeLegacySerializerAssertions = true,
+            targetPipeline = EvaluatorTestTarget.COMPILER_PIPELINE, // planner & phys. alg. have no support for OFFSET (yet)
```
alancai98, line 115: From your comment in the PR that set the test exclusions, #586, you left a [comment](https://github.com/partiql/partiql-lang-kotlin/pull/586#discussion_r867138242) saying that this comment was wrong. Perhaps there was some missed commit in the merge?

(The original version was just `tc.copy(excludeLegacySerializerAssertions = true)` and the exclusions were set by the individual `EvaluatorTestCase`s.)

alancai98 (commented):
Changes since last commit look good other than the one comment in [EvaluatingCompilerOffsetTests.kt](https://github.com/partiql/partiql-lang-kotlin/pull/587/files#diff-160849dd4f8a7a040522130f82d28a7f11199b8b1669d0475e7f0da5e1614e8a).

lang/test/org/partiql/lang/eval/EvaluatingCompilerOffsetTests.kt
dlurton: Oh, yeah, it must have returned after a merge.  Will fix.  Thanks.

alancai98 (approved):
:shipit:

PR #588, opened 2022-04-29
Query planner passes

dlurton:
Originally part of #582, this commit is based on #587, which must be merged first.  However, it may be helpful to review the PR introducing the `PlannerPipeline` first (#590), if more context for these changes is desired.

This PR contains 3 query planner passes:

- Convert AST to logical plan 
- Logical plan -> Resolved logical plan.  This pass assigns indexes to all local variables, and resolves references to global variables. (undefined variables are detected and reported here).
- Resolved logical plan -> physical plan

These are not yet integrated with anything--that's in #590.

### DIFF
--- a/lang/src/org/partiql/lang/domains/util.kt
+++ b/lang/src/org/partiql/lang/domains/util.kt
@@ -1,5 +1,6 @@
 package org.partiql.lang.domains
 
+import com.amazon.ionelement.api.IonElement
 import com.amazon.ionelement.api.MetaContainer
 import com.amazon.ionelement.api.emptyMetaContainer
 import com.amazon.ionelement.api.metaContainerOf
@@ -13,6 +14,19 @@ import org.partiql.lang.eval.BindingCase
 // TODO:  once https://github.com/partiql/partiql-ir-generator/issues/6 has been completed, we can delete this.
 fun PartiqlAst.Builder.id(name: String) =
     id(name, caseInsensitive(), unqualified())
+
+// TODO:  once https://github.com/partiql/partiql-ir-generator/issues/6 has been completed, we can delete this.
+fun PartiqlLogical.Builder.id(name: String) =
+    id(name, caseInsensitive(), unqualified())
+
+// TODO:  once https://github.com/partiql/partiql-ir-generator/issues/6 has been completed, we can delete this.
+fun PartiqlLogical.Builder.pathExpr(exp: PartiqlLogical.Expr) =
+    pathExpr(exp, caseInsensitive())
+
+// Workaround for a bug in PIG that is fixed in its next release:
+// https://github.com/partiql/partiql-ir-generator/issues/41
+fun List<IonElement>.asAnyElement() =
+    this.map { it.asAnyElement() }
 
 val MetaContainer.staticType: StaticTypeMeta? get() = this[StaticTypeMeta.TAG] as StaticTypeMeta?
 
@@ -60,17 +74,17 @@ fun PartiqlAst.CaseSensitivity.toBindingCase(): BindingCase = when (this) {
 }
 
 /**
- * Returns the [SourceLocationMeta] as an error context if the [SourceLocationMeta.TAG] exists in the passed
- * [metaContainer]. Otherwise, returns an empty map.
+ * Converts a [PartiqlLogical.CaseSensitivity] to a [BindingCase].
  */
-fun errorContextFrom(metaContainer: MetaContainer?): PropertyValueMap {
-    if (metaContainer == null) {
-        return PropertyValueMap()
-    }
-    val location = metaContainer[SourceLocationMeta.TAG] as? SourceLocationMeta
-    return if (location != null) {
-        org.partiql.lang.eval.errorContextFrom(location)
-    } else {
-        PropertyValueMap()
-    }
+fun PartiqlLogical.CaseSensitivity.toBindingCase(): BindingCase = when (this) {
+    is PartiqlLogical.CaseSensitivity.CaseInsensitive -> BindingCase.INSENSITIVE
+    is PartiqlLogical.CaseSensitivity.CaseSensitive -> BindingCase.SENSITIVE
+}
+
+/**
+ * Converts a [PartiqlLogical.CaseSensitivity] to a [BindingCase].
+ */
+fun PartiqlPhysical.CaseSensitivity.toBindingCase(): BindingCase = when (this) {
+    is PartiqlPhysical.CaseSensitivity.CaseInsensitive -> BindingCase.INSENSITIVE
+    is PartiqlPhysical.CaseSensitivity.CaseSensitive -> BindingCase.SENSITIVE
 }
--- a/lang/src/org/partiql/lang/planner/GlobalBindings.kt
+++ b/lang/src/org/partiql/lang/planner/GlobalBindings.kt
@@ -1,0 +1,50 @@
+package org.partiql.lang.planner
+
+import org.partiql.lang.eval.BindingCase
+import org.partiql.lang.eval.BindingName
+
+/** Indicates the result of an attempt to resolve a global binding. */
+sealed class ResolutionResult {
+    /**
+     * A success case, indicates the [uniqueId] of the match to the [BindingName] in the global scope.
+     * Typically, this is defined by the storage layer.
+     */
+    data class GlobalVariable(val uniqueId: String) : ResolutionResult()
+
+    /**
+     * A success case, indicates the [index] of the only possible match to the [BindingName] in a local lexical scope.
+     * This is `internal` because [index] is an implementation detail that shouldn't be accessible outside of this
+     * library.
+     */
+    internal data class LocalVariable(val index: Int) : ResolutionResult()
+
+    /** A failure case, indicates that resolution did not match any variable. */
+    object Undefined : ResolutionResult()
+}
+
+fun interface GlobalBindings {
+    /**
+     * Implementations try to resolve a global variable which is typically a database table, as identified by a
+     * [bindingName].  The [bindingName] includes both the name as specified by the query author and a [BindingCase]
+     * which indicates if query author included double quotes (") which mean the lookup should be case-sensitive.
+     *
+     * Implementations of this function must return:
+     *
+     * - [ResolutionResult.GlobalVariable] if [bindingName] matches a global variable (typically a database table).
+     * - [ResolutionResult.Undefined] if no identifier matches [bindingName].
+     *
+     * When determining if a variable name matches a global variable, it is important to consider if the comparison
+     * should be case-sensitive or case-insensitive.  @see [BindingName.bindingCase].  In the event that more than one
+     * variable matches a case-insensitive [BindingName], the implementation must still select one of them
+     * without providing an error. (This is consistent with Postres's behavior in this scenario.)
+     *
+     * Note that while [ResolutionResult.LocalVariable] exists, it is intentionally marked `internal` and cannot
+     * be used by outside of this project..
+     */
+    fun resolve(bindingName: BindingName): ResolutionResult
+}
+
+private val EMPTY = GlobalBindings { ResolutionResult.Undefined }
+
+/** Convenience function for obtaining an instance of [GlobalBindings] with no defined variables. */
+fun emptyGlobalBindings(): GlobalBindings = EMPTY
--- a/lang/src/org/partiql/lang/planner/PassResult.kt
+++ b/lang/src/org/partiql/lang/planner/PassResult.kt
@@ -9,7 +9,8 @@ sealed class PassResult<TResult> {
 
     /**
      * Indicates query planning was not successful and includes a list of errors and warnings that were encountered
-     * along the way.
+     * along the way.  Encountering both errors and warnings, as well as multiple errors is possible since we are not
+     * required to stop when encountering the first error.
      */
     data class Error<TResult>(val errors: List<Problem>) : PassResult<TResult>()
 }
--- a/lang/src/org/partiql/lang/planner/PassResult.kt
+++ b/lang/src/org/partiql/lang/planner/PassResult.kt
@@ -1,0 +1,15 @@
+package org.partiql.lang.planner
+import org.partiql.lang.errors.Problem
+
+sealed class PassResult<TResult> {
+    /**
+     * Indicates query planning was successful and includes a list of any warnings that were encountered along the way.
+     */
+    data class Success<TResult>(val result: TResult, val warnings: List<Problem>) : PassResult<TResult>()
+
+    /**
+     * Indicates query planning was not successful and includes a list of errors and warnings that were encountered
+     * along the way.
+     */
+    data class Error<TResult>(val errors: List<Problem>) : PassResult<TResult>()
+}
--- a/lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
+++ b/lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
@@ -16,7 +16,7 @@ internal fun PartiqlAst.Statement.toLogicalPlan(): PartiqlLogical.Plan =
     PartiqlLogical.build {
         plan(
             AstToLogicalVisitorTransform.transformStatement(this@toLogicalPlan),
-            version = PLAN_VERSION_NUMBER.toLong()
+            version = PLAN_VERSION_NUMBER
         )
     }
 
--- a/lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
+++ b/lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
@@ -1,6 +1,5 @@
 package org.partiql.lang.planner.transforms
 
-import com.amazon.ionelement.api.ionBool
 import org.partiql.lang.domains.PartiqlAst
 import org.partiql.lang.domains.PartiqlAstToPartiqlLogicalVisitorTransform
 import org.partiql.lang.domains.PartiqlLogical
@@ -161,7 +160,7 @@ private object FromSourceToBexpr : PartiqlAst.FromSource.Converter<PartiqlLogica
                 joinType = AstToLogicalVisitorTransform.transformJoinType(node.type),
                 left = convert(node.left),
                 right = convert(node.right),
-                predicate = node.predicate?.let { AstToLogicalVisitorTransform.transformExpr(it) } ?: lit(ionBool(true)),
+                predicate = node.predicate?.let { AstToLogicalVisitorTransform.transformExpr(it) },
                 node.metas
             )
         }
--- a/lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
+++ b/lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
@@ -1,0 +1,168 @@
+package org.partiql.lang.planner.transforms
+
+import com.amazon.ionelement.api.ionBool
+import org.partiql.lang.domains.PartiqlAst
+import org.partiql.lang.domains.PartiqlAstToPartiqlLogicalVisitorTransform
+import org.partiql.lang.domains.PartiqlLogical
+
+/**
+ * Transforms an instance of [PartiqlAst.Statement] to [PartiqlLogical.Statement].
+ *
+ * Performs no semantic checks.
+ *
+ * This conversion (and the logical algebra) are early in their lifecycle and so only a very limited subset of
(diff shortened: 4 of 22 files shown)


lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -0,0 +1,168 @@
...
+        TODO("Support for UNPIVOT")
+    }
+
+    override fun convertJoin(node: PartiqlAst.FromSource.Join): PartiqlLogical.Bexpr =
+        PartiqlLogical.build {
+            join(
+                joinType = AstToLogicalVisitorTransform.transformJoinType(node.type),
+                left = convert(node.left),
+                right = convert(node.right),
+                predicate = node.predicate?.let { AstToLogicalVisitorTransform.transformExpr(it) } ?: lit(ionBool(true)),
```
dlurton, line 164: Don't forget: feedback from #584 made `predicate` nullable to represent cross joins so you'll have to remove this the `.let { ... } ?: lit(...))` here....

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -0,0 +1,167 @@
...
+            )
+        }
+
+    override fun transformStatementDml(node: PartiqlAst.Statement.Dml): PartiqlLogical.Statement {
+        TODO("Support for DML")
+    }
+
+    override fun transformStatementDdl(node: PartiqlAst.Statement.Ddl): PartiqlLogical.Statement {
+        TODO("Support for DDL")
+    }
```
lziq, line 122: I'm not really sure about this, but I think DML & DDL are not parts of logical plans. So is it better to throw error for these methods, instead of `TODO`?

lang/test/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransformTests.kt
```diff
@@ -0,0 +1,152 @@
...
+        println(SexpAstPrettyPrinter.format(algebra.toIonElement().asAnyElement().toIonValue(ion)))
+        Assertions.assertEquals(tc.expectedAlgebra, algebra)
+    }
+
+    @ParameterizedTest
+    @ArgumentsSource(ArgumentsForToLogicalTests::class)
+    fun `to logical`(tc: TestCase) = runTestCase(tc)
+
+    class ArgumentsForToLogicalTests : ArgumentsProviderBase() {
+        override fun getParameters() = listOf(
```
lziq, line 49: Maybe it's good to also add test cases for `LET`, `OFFSET` & `LIMIT` clause?

lziq (commented):
Just finished review for [AstToLogicalVisitorTransform.kt](https://github.com/partiql/partiql-lang-kotlin/pull/588/files#diff-7d67a5a9e5b054b8588f5a1326cdf8342bccec8eb49da7843aaeb4e4d4882f6b) and related test files and left a few comments. Generally, it looks good to me. Will continue to review rest of the changes.

lang/src/org/partiql/lang/domains/util.kt
```diff
@@ -14,6 +15,19 @@ import org.partiql.lang.eval.BindingCase
...
+    id(name, caseInsensitive(), unqualified())
+
+// TODO:  once https://github.com/partiql/partiql-ir-generator/issues/6 has been completed, we can delete this.
+fun PartiqlLogical.Builder.pathExpr(exp: PartiqlLogical.Expr) =
+    pathExpr(exp, caseInsensitive())
+
+// Workaround for a bug in PIG that is fixed in its next release:
+// https://github.com/partiql/partiql-ir-generator/issues/41
+fun List<IonElement>.asAnyElement() =
+    this.map { it.asAnyElement() }
```
am357, line 29: While at it, shouldn't we just release the PIG and remove this?

lang/src/org/partiql/lang/planner/GlobalBindings.kt
```diff
@@ -0,0 +1,50 @@
+package org.partiql.lang.planner
+
+import org.partiql.lang.eval.BindingCase
+import org.partiql.lang.eval.BindingName
+
+/** Indicates the result of an attempt to resolve a global binding. */
```
am357, line 6: I understand we have kdoc for `GlobalBindings.resolve()` but can we also add a definition for _Global Binding_ here?

BTW. 
I think it's helpful to have the link to the upcoming documentation in applicable constructs under `planner`.

lang/src/org/partiql/lang/planner/GlobalBindings.kt
```diff
@@ -0,0 +1,50 @@
...
+     * Typically, this is defined by the storage layer.
+     */
+    data class GlobalVariable(val uniqueId: String) : ResolutionResult()
+
+    /**
+     * A success case, indicates the [index] of the only possible match to the [BindingName] in a local lexical scope.
+     * This is `internal` because [index] is an implementation detail that shouldn't be accessible outside of this
+     * library.
+     */
+    internal data class LocalVariable(val index: Int) : ResolutionResult()
```
am357, line 19: Considering that `GlobalBindings.resolve()` is meant to return `GlobalVariable` or `undefined`, should this be here?

lang/src/org/partiql/lang/planner/PassResult.kt
```diff
@@ -0,0 +1,15 @@
...
+import org.partiql.lang.errors.Problem
+
+sealed class PassResult<TResult> {
+    /**
+     * Indicates query planning was successful and includes a list of any warnings that were encountered along the way.
+     */
+    data class Success<TResult>(val result: TResult, val warnings: List<Problem>) : PassResult<TResult>()
+
+    /**
+     * Indicates query planning was not successful and includes a list of errors and warnings that were encountered
```
am357, line 11: `errors and warnings` -> `errors`?

lang/src/org/partiql/lang/planner/transforms/LogicalResolvedToPhysicalVisitorTransform.kt
```diff
@@ -0,0 +1,94 @@
...
+ */
+internal fun PartiqlLogicalResolved.Plan.toPhysicalPlan() =
+    LogicalResolvedToPhysicalVisitorTransform().transformPlan(this)
+
+internal val DEFAULT_IMPL = PartiqlPhysical.build { impl("default") }
+
+internal class LogicalResolvedToPhysicalVisitorTransform : PartiqlLogicalResolvedToPartiqlPhysicalVisitorTransform() {
+
+    /** Copies [PartiqlLogicalResolved.Bexpr.Scan] to [PartiqlPhysical.Bexpr.Scan], adding the default impl. */
+    override fun transformBexprScan(node: PartiqlLogicalResolved.Bexpr.Scan): PartiqlPhysical.Bexpr {
```
am357, line 19: For this and others that apply: can we add `withDefaultImpl` suffix (or something similar) to the names. Also where in the APIs are we're receiving `impl` as a parameter. (sorry if it's already somewhere and I've missed it).

lang/src/org/partiql/lang/planner/transforms/LogicalToLogicalResolvedVisitorTransform.kt
```diff
@@ -0,0 +1,393 @@
...
+ * `(id <name> <unique-index>)`) or `(global_id <name> <unique-id>)`.  The latter is usually a reference to
+ * a database table.  `<unique-id>` is supplied by the integrating PartiQL service by means of the [globals] specified
+ * by callers of this function.  Note that in general, all `(scan (global_id ...) ...)` operators will later be
+ * rewritten to an optimized physical read operator.
+ *
+ * The [problemHandler] is notified of any undefined variables.  Resolution does not stop on the first error, rather
+ * we keep going to provide the end user any additional error messaging, unless [ProblemHandler.handleProblem] throws
+ * an exception when an error is logged.  **If any undefined variables are detected, in order to allow traversal to
+ * continue, a fake index value is used in place of a real one and the resolved logical plan returned by this function
+ * is guaranteed to be invalid.** **Therefore, it is the responsibility therefore of callers to check if any problems
```
am357, line 28: A redundant _therefore`?

lang/src/org/partiql/lang/planner/transforms/Util.kt
```diff
@@ -0,0 +1,13 @@
+
+package org.partiql.lang.planner.transforms
+
+/**
+ * This is the version number of the logical and physical plans supported by this version of PartiQL.
+ *
+ * It would be nice to embed this in the PIG domain somehow, but this isn't supported, so we have to include it
```
am357, line 7: Didn't we introduce the version number in the new domains. I'm sure you feel the same way, but this is not great. Even if going with this hard-coding, the version doesn't tell anything about backward compatibility; shouldn't we adopt a format like `SemVer`?

lang/src/org/partiql/lang/planner/transforms/VariableIdAllocator.kt
```diff
@@ -0,0 +1,45 @@
...
+ * declared within the plan, which becomes the `locals` sub-node of the `plan` node.
+ */
+internal fun PartiqlLogical.Plan.allocateVariableIds(): Pair<PartiqlLogical.Plan, List<PartiqlLogicalResolved.LocalVariable>> {
+
+    var allLocals = mutableListOf<PartiqlLogicalResolved.LocalVariable>()
+    val planWithAllocatedVariables = VariableIdAllocator(allLocals).transformPlan(this)
+    return planWithAllocatedVariables to allLocals.toList()
+}
+
+private const val VARIABLE_ID_META_TAG = "\$variable_id"
```
am357, line 20: Nit.: I'm thinking if we need to get all the constants in one place under `planner`.

am357 (changes requested):
Had a review and left some comments w/ requesting code changes if they make sense.

lang/src/org/partiql/lang/domains/util.kt
dlurton: Yes, but https://github.com/partiql/partiql-ir-generator/issues/41 hasn't been fixed yet.

lang/src/org/partiql/lang/planner/GlobalBindings.kt
dlurton: It's `internal`, so it's not exposed to customers.  Customer code is never intended to resolve local variables, so this fits the situation.  Having this here does simplify a few things in relation in the `LogicalToLogicalResolvedVisitorTransform` class.  Namely, I found this to be preferable to having two variations of this `sealed` class: one that is public and only has `GlobalVariable` and `Undefined`, and another that is private and has `GlobalVariable`, `Undefined` as well as  `LocalVariable`. 

Making one member of a `public` sealed class `internal` is perhaps a little unusual, but unusual doesn't equate to undesirable, necessarily.  Was there a specific concern you had with this approach?

lang/src/org/partiql/lang/planner/PassResult.kt
dlurton: This is designed to allow us to continue at least the current pass in the event of a semantic error instead of throwing an exception.  That means its entirely possible to encounter both warnings and errors.

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
dlurton: - DML is definitely needed since is really just another type of query--the results of which are applied to the data store instead of returned to the caller.  That said, it is out of scope for now.
- DDL isn't needed in the short term but there's no reason why it can't be part of the logical and physical plans at some point.  It's also entirely possible that DDL never has a representation in the logical or physical plans.  There may not be a benefit.  I am undecided because I don't have any customer requirements to help guide the decision.

lang/src/org/partiql/lang/planner/GlobalBindings.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    /** A failure case, indicates that resolution did not match any variable. */
+    object Undefined : ResolutionResult()
+}
+
+fun interface GlobalBindings {
```
lziq, line 25: Maybe rename this as `GlobalVariableToIdBindings`? The current naming might be confused with the global bindings for evaluation.

lang/src/org/partiql/lang/planner/transforms/LogicalResolvedToPhysicalVisitorTransform.kt
dlurton: I'm not sure I understand the question exactly... but... 1) it is expected that a customer-supplied pass will change these from `(impl default)` to something that is customer specific.  2) right now, the `impl` is ignored by `PhysicalBexprToThunkConverterImpl` (introduced in #592), but I expect to fix that in a future PR that has as yet not been produced.  This will probably be the same PR that introduces the API that allows customers add passes that rewrite physical plans.

lang/src/org/partiql/lang/planner/transforms/LogicalToLogicalResolvedVisitorTransform.kt
dlurton: Out of order, actually.  Fixing...

dlurton: also added a couple dozen lines about `$__dynamic_lookup__`

lang/src/org/partiql/lang/planner/transforms/Util.kt
dlurton: I really think this aught to be something solved by PIG actually so I created https://github.com/partiql/partiql-ir-generator/issues/121

lang/src/org/partiql/lang/planner/transforms/VariableIdAllocator.kt
dlurton: In general, I'd rather not have one file with a bunch of unrelated constants.  Note that this is `private`, putting it somewhere else means it can't be `private` anymore and will pollute the package namespace.  Putting the constant here also has the advantage that the definition is close to its uses, which improves readability IMHO.

lang/test/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransformTests.kt
dlurton: From the kdoc above:

> Test cases in this class might seem a little light--that's because [AstToLogicalVisitorTransform] is getting heavily exercised during many other integration tests.  These should be considered "smoke tests".

Detailed test cases were not intended here.  We have tests which run `LET`, `OFFSET` and `LIMIT` e2e.  I lean against exhaustive testing of every unit in this type of project because the integration tests and the e2e behavior they test is what has the most value.  Also, detailed tests like this increase maintenance overhead since any changes to those nodes means that this is yet another place that needs to be updated.

lang/test/org/partiql/lang/planner/transforms/LogicalToLogicalResolvedVisitorTransformTests.kt
```diff
@@ -0,0 +1,685 @@
...
+    @ArgumentsSource(DynamicIdSearchCases::class)
+    fun `dynamic_lookup search order cases`(tc: TestCase) = runTestCase(tc)
+    class DynamicIdSearchCases : ArgumentsProviderBase() {
+        // The important thing being asserted here is the contents of the dynamicId.search, which
+        // defines the places we'll look for variables that are unresolved at compile time.
+        override fun getParameters() = listOf(
+            // Not in an SFW query (empty search path)
+            TestCase(
+                "undefined1 + undefined2",
+                Expectation.Success(
```
dlurton, line 606: Need to add test cases for `globalsFirst = false`, which is missing here.

lang/src/org/partiql/lang/planner/GlobalBindings.kt
dlurton: Going with `UniqueIdResolver`

lang/src/org/partiql/lang/planner/GlobalBindings.kt
dlurton: I have rephrased this to be more clear.

lang/src/org/partiql/lang/planner/transforms/Util.kt
dlurton: I'll add a link to the ticket.

lziq: Do you also mind adding some comments to 
```
            (record plan
                (stmt statement)
                (version int)
            )
```
in the `partiql.ion` file, to state that the version number here will be removed and assigned to PIG domains once https://github.com/partiql/partiql-ir-generator/issues/121 is resolved?

lang/src/org/partiql/lang/planner/transforms/LogicalResolvedToPhysicalVisitorTransform.kt
lziq: Can we change the class name to `LogicalResolvedToPhysicalDefaultVisitorTransform` to indicate that users should have their own `impl` and implement `PartiqlLogicalResolvedToPartiqlPhysicalVisitorTransform` by themselves?

lziq (commented):
Just finished the review. Overall it looks good to me.

lang/src/org/partiql/lang/planner/GlobalBindings.kt
am357: I wasn't worried about exposing it, I just thought this data class doesn't belong here. I think your explanation makes sense, so let's leave it as is.

lang/src/org/partiql/lang/planner/transforms/LogicalResolvedToPhysicalVisitorTransform.kt
am357: > I'm not sure I understand the question exactly... but... 1) it is expected that a customer-supplied pass will change these from (impl default) to something that is customer specific. 2) right now, the impl is ignored by PhysicalBexprToThunkConverterImpl (introduced in #592), but I expect to fix that in a future PR that has as yet not been produced. This will probably be the same PR that introduces the API that allows customers add passes that rewrite physical plans.

So will customers be able to use `default` for some of the operators and pass their own for others in a pass?

lang/src/org/partiql/lang/planner/transforms/Util.kt
am357: > I would like to leave this as-is for now if you don't mind.

I think this is something that we need to fix before we expose the API; once we have the version as is and the API is exposed, then it's too late. Happy to discuss further offline.

lang/src/org/partiql/lang/planner/transforms/LogicalResolvedToPhysicalVisitorTransform.kt
dlurton: > So will customers be able to use default for some of the operators and pass their own for others in a pass?

That is the intent, yes, however, the `default` implementation will not always be the most performant or scalable.  In such circumstances, it is expected that the customer will supply their own implementation.  For instance, the user may supply an implementation of `distinct` which stores "seen" values on disk while the `default` implementation would do this using heap.

lang/src/org/partiql/lang/planner/transforms/Util.kt
dlurton: Offline discussion indicates this doesn't need to be addressed in this PR, but should be addressed before this hits `main`.  Creating an issue in the related milestone to track this.

lang/src/org/partiql/lang/planner/transforms/LogicalResolvedToPhysicalVisitorTransform.kt
dlurton: > Can we change the class name to LogicalResolvedToPhysicalDefaultVisitorTransform to indicate that users should have their own impl and implement PartiqlLogicalResolvedToPartiqlPhysicalVisitorTransform by themselves?

Going with `LogicalResolvedToDefaultPhysicalVisitorTransform` and `toDefaultPhysicalPlan()`

lziq:
The new commit makes sense to me and there is no more file change requested for this PR from me.

lang/src/org/partiql/lang/planner/transforms/Util.kt
am357: Just an update on this; as a comprise, requested having the version as `SemVer` hard-coded for now. @dlurton had a point on using `patch`—which is for bug fixes—might not be applicable to our use-case. We ultimately agreed on having `major` and `minor` notions in the version for now.

am357 (approved):

PR #592, opened 2022-04-30
Add Plan Evaluator

dlurton:
~~Yes, there's a failed unit test here.  I will need to fix that before this can be merged--that does not however, mean can't 
get started reviewing this before then.~~ EDIT: No longer, this is now fixed.

This was originally part of #582 and is based on #590, which must be merged first.

Compilation of a query plan is now handled by two classes:

- `PhysicalBexprToThunkConverter`, which compiles the physical algebra itself.
- `PhysicalExprToThunkConverterImpl`, which is a copy and paste of `EvaluatingCompiler` with:
    - All references to "PartiqlAst" replaced with "PartiqlPhysical"
    - The following changes:
        - Removal of support for the `(select ...)` node (including the features out of scope for now: `ORDER BY`,  `GROUP BY`, `HAVING`, aggregate functions (`SUM`, `MAX`, `MIN`, etc),  `PIVOT` and `UNPIVOT`).
        - Support for the enhanced `(struct ...)` constructor (that supports struct unions)
        - The AST's `id` node removed.  (This is handled now by either `local_id` or `global_id`, where variables can be resolved at compile time and by the `$__dynamic_lookup__(...)` `ExprFunction` if they cannot (and this is enabled).
        - Support for `bindings_to_values` expression added (see #584)
    - (I'll try go to thru this class and add PR comments to illuminate where these changes are.)

The two classes are mutually recursive.  Each invokes the other to compile value or relational expressions as needed.

### Why Is the Duplication Needed

The duplication is *unfortunate* but necessary and can hopefully be removed in the short term.  The goal is, once we're confident in the new plan compiler and it has reached feature parity, to delete `EvaluatingCompiler`.  All of the physical plan's node types (value and relational expressions) exist in the `PartiqlPhysical` domain while the AST's node types exist in the `PartiqlAst` domain.  If this duplication is to persist long term, we might need to look at how to eliminate it (possibly thru a new PIG feature), since maintaining both `PhysicalExprToThunkConverterImpl` and `EvaluatingCompiler` long term could get onerous.

## Other Notables That Need Reviewing

- `EvaluatorState` and its uses.  This is the equivalent of `Environment` for the plan compiler.
- `sequence { }`, `RelationIterator` etc and their uses.  Kotlin coroutines are used during relational operator evaluation.


lang/src/org/partiql/lang/eval/visitors/PartiqlPhysicalSanityValidator.kt
```diff
@@ -0,0 +1,121 @@
...
+import org.partiql.lang.eval.TypedOpBehavior
+import org.partiql.lang.eval.err
+import org.partiql.lang.eval.errorContextFrom
+import org.partiql.lang.planner.EvaluatorOptions
+import org.partiql.lang.util.propertyValueMapOf
+import org.partiql.pig.runtime.LongPrimitive
+
+/**
+ * Provides rules for basic AST sanity checks that should be performed before any attempt at further phsycial
+ * plan processing. This is provided as a distinct [PartiqlPhysical.Visitor] so that [PlanCompiler] may assume that
```
dlurton, line 24: `PlanCompiler` is the old name.

am357 (approved):

lang/src/org/partiql/lang/planner/PlannerPipeline.kt
```diff
@@ -231,18 +276,21 @@ interface PlannerPipeline {
...
-            val builtinFunctions = createBuiltinFunctions(valueFactory)
-            // TODO: uncomment when DynamicLookupExprFunction exists
-//            val builtinFunctions = createBuiltinFunctions(valueFactory) + DynamicLookupExprFunction()
+            val builtinFunctions = createBuiltinFunctions(valueFactory) + DynamicLookupExprFunction()
             val builtinFunctionsMap = builtinFunctions.associateBy {
                 it.signature.name
             }
 
+            // customFunctions must be on the right side of + here to ensure that they overwrite any
+            // built-in functions with the same name.
```
am357, line 285: I  know this for later but I'm noting—overwrite any built-int—as a design decision that we need to make on the open type system.

lang/test/org/partiql/lang/eval/evaluatortestframework/PlannerPipelineFactory.kt
```diff
@@ -0,0 +1,113 @@
...
+import org.partiql.lang.eval.UndefinedVariableBehavior
+import org.partiql.lang.planner.EvaluatorOptions
+import org.partiql.lang.planner.MetadataResolver
+import org.partiql.lang.planner.PassResult
+import org.partiql.lang.planner.PlannerPipeline
+import org.partiql.lang.planner.ResolutionResult
+import kotlin.test.assertNotEquals
+import kotlin.test.assertNull
+
+internal class PlannerPipelineFactory : PipelineFactory {
```
am357, line 18: Could use a `kdoc` here too.

am357 (approved):

PR #597, opened 2022-05-04
Added type, function & operator interfaces

lziq:
*Issue #593*

*Description of changes:*
Added type, function & operator interfaces


lang/src/org/partiql/lang/typesystem/interfaces/function/SqlFunction.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.partiql.lang.typesystem.interfaces.type.TypeWithParameters
+
+/**
+ * Used to define a sql function.
+ */
+interface SqlFunction {
+    /**
+     * Function name
+     */
+    fun getFuncName(): String
```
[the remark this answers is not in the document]
lziq: I think `getFuncName` is not a getter for any field of this Kotlin class. Similar to `inferReturnType` & `invoke`, etc. they are just functions which Open Type System will call to collect necessary information related to the sql function.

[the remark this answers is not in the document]
lziq: Probably the naming of it is misleading, and probably we need to remove the word "get", just leaving it as `funcName`.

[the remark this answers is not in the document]
lziq: Let me know if you think it is better to change the naming.

lang/src/org/partiql/lang/typesystem/interfaces/function/SqlFunction.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Variadic argument of this function. Null value means no variadic argument.
+     */
+    fun getVariadicArgType(): Type?
+
+    /**
+     * Function return type inference
+     */
+    fun inferReturnType(): List<TypeWithParameters>
```
dlurton, line 34: Just looking at this function... if "infer return type" means what I think it means, this function also needs an argument `List<Type>` here, since one of the inputs to inference will be the arguments of a call site.

lziq: The argument types are already known by `requiredArgTypes`, `optionalArgTypes` and `variadicArgType`, right?

lziq: Can you be more clear about why we need the argument of `List<Type>` here?

lang/src/org/partiql/lang/typesystem/interfaces/function/SqlFunction.kt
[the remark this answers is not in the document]
lziq: As we discussed offline, we decided to change these functions into fields, since it looks more concise for use.

lang/src/org/partiql/lang/typesystem/interfaces/function/SqlFunction.kt
lziq: I spent some time thinking about this, and I don't think we really need it. We can just add a property `returnType` for sql functions.

lang/src/org/partiql/lang/typesystem/interfaces/operator/OpAlias.kt
```diff
@@ -0,0 +1,9 @@
+package org.partiql.lang.typesystem.interfaces.operator
+
+enum class OpAlias {
```
am357, line 3: A `kdoc` would be helpful here.

lang/src/org/partiql/lang/typesystem/interfaces/operator/PqlOperator.kt
```diff
@@ -0,0 +1,11 @@
+package org.partiql.lang.typesystem.interfaces.operator
+
+/**
+ * Used to define a sql operator
+ */
+interface PqlOperator {
```
am357, line 6: `PartiqlOperator`?

lang/src/org/partiql/lang/typesystem/interfaces/operator/operators/binary/IntersectOp
```diff
@@ -0,0 +1,12 @@
+package org.partiql.lang.typesystem.interfaces.operator.operators.binary
```
am357, line 1: Looks like the file misses `.kt`

lang/src/org/partiql/lang/typesystem/interfaces/type/SqlType.kt
```diff
@@ -0,0 +1,33 @@
+package org.partiql.lang.typesystem.interfaces.type
+
+import org.partiql.lang.eval.ExprValueType
+
+/**
+ * This interface is used to define a sql type
+ */
+interface SqlType {
+    /**
+     * A sql type has a unique type name
```
am357, line 10: Within namespace or across namespaces?

lang/src/org/partiql/lang/typesystem/interfaces/type/TypeExtension.kt
```diff
@@ -0,0 +1,24 @@
+package org.partiql.lang.typesystem.interfaces.type
+
+import org.partiql.lang.eval.ExprValue
+
+/**
+ * A sql type with actual parameters. If it is a non-parametric type, [parameters] should be an empty list.
+ */
+data class TypeWithParameters(
```
am357, line 8: `TypeWithParameters` -> `SqlTypeWithParameters`?

am357 (commented):
I still need to think further about the interfaces in`TypeExtentions` but added some comments in the meantime. I also think it'd be good to move the `typesystem` under `experimental` for now.

lang/src/org/partiql/lang/typesystem/interfaces/operator/PqlOperator.kt
lziq: Does `SqlOperator` sound good?

am357 (approved):

Issue #600, opened 2022-05-09
MIN and MAX support for Timestamp data type


alancai98:
The MIN and MAX aggregation functions currently don't support the Timestamp data type. The ordering should be consistent with the ordering provided by `ORDER BY`.

dlurton:
Duplicates https://github.com/partiql/partiql-lang-kotlin/issues/203

dlurton:
`MIN` and `MAX` should actually work with *any* data type, not just numbers and timestamps.

alancai98:
Ah, thanks for finding the duplicate issue.

Adding a reproducible test case for the timestamp case:

```
PartiQL> SELECT MAX(ts) FROM [`2022-05-09T`, `2022-01-01T`] AS ts;
org.partiql.lang.eval.EvaluationException: Expected number: 2022-05-09
	Evaluator Error: at line 1, column 1: Unexpected value type, <UNKNOWN> : <UNKNOWN>

	at org.partiql.lang.eval.ThunkFactory.populateErrorContext(Thunk.kt:387)
	at org.partiql.lang.eval.LegacyThunkFactory.handleException(Thunk.kt:552)
	at org.partiql.lang.eval.EvaluatingCompiler$compileSelect$1$2$getQueryThunk$$inlined$thunkEnv$lang$2.invoke(Thunk.kt:210)
	at org.partiql.lang.eval.EvaluatingCompiler$compileSelect$1$2$getQueryThunk$$inlined$thunkEnv$lang$2.invoke(Thunk.kt:129)
	at org.partiql.lang.eval.EvaluatingCompiler$compileSelect$1$2$getQueryThunk$$inlined$thunkEnv$lang$4$1.invoke(Thunk.kt:711)
	at org.partiql.lang.eval.EvaluatingCompiler$compileSelect$1$2$getQueryThunk$$inlined$thunkEnv$lang$4$1.invoke(Thunk.kt:129)
	at org.partiql.lang.eval.LegacyThunkFactory.handleException(Thunk.kt:550)
	at org.partiql.lang.eval.EvaluatingCompiler$compileSelect$1$2$getQueryThunk$$inlined$thunkEnv$lang$4.invoke(Thunk.kt:210)
	at org.partiql.lang.eval.EvaluatingCompiler$compileSelect$1$2$getQueryThunk$$inlined$thunkEnv$lang$4.invoke(Thunk.kt:129)
[...]

PR #604, opened 2022-05-11
Enable `min` & `max` to work on all the data types and across data ty…

lziq:
*Issue #600*

*Description of changes:*
1. Enable `MIN` & `MAX` to work on all the data types. 
2. Enable `MIN` & `MAX` to work across data types. 
3. Refactor test framework for aggregate functions. 
4. Add more tests to aggregate fnctions.


lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -192,30 +191,26 @@ internal class EvaluatingCompiler(
...
         }
 
     /** Dispatch table for built-in aggregate functions. */
     private val builtinAggregates: Map<Pair<String, PartiqlAst.SetQuantifier>, ExprAggregatorFactory> =
         run {
-            val countAccFunc: (Number?, ExprValue) -> Number = { curr, _ -> curr!! + 1L }
-            val sumAccFunc: (Number?, ExprValue) -> Number = { curr, next ->
-                curr?.let { it + next.numberValue() } ?: next.numberValue()
+            val countAccFunc: (ExprValue?, ExprValue) -> ExprValue = { accumulated, _ -> (accumulated!!.longValue() + 1L).exprValue() }
+            val sumAccFunc: (ExprValue?, ExprValue) -> ExprValue = { accumulated, nextItem ->
```
alancai98, line 209: nit: after playing around in the REPL/CLI a bit, I think the error messages could be improved for the `SUM` and `AVG` aggregation functions involving non-numeric types. Currently when giving a non-numeric type, we get something like the following error:

```
org.partiql.lang.eval.EvaluationException: Expected number: "abc"
```

Could consider providing a more descriptive error message (e.g. "SUM expects numeric types but was given a STRING 'abc'") in this PR or cutting a separate issue.

lang/test/org/partiql/lang/eval/SimpleEvaluatingCompilerTests.kt
```diff
@@ -125,11 +126,115 @@ class SimpleEvaluatingCompilerTests : EvaluatorTestBase() {
...
         runEvaluatorTestCase("max(`[1, 2e0, 3e0]`)", expectedResult = "3e0")
         runEvaluatorTestCase("max(`[1, 2d0, 3d0]`)", expectedResult = "3d0")
         runEvaluatorTestCase("max(`[1, 2e0, 3d0]`)", expectedResult = "3d0")
         runEvaluatorTestCase("max(`[1, 2d0, 3e0]`)", expectedResult = "3e0")
+
+        // String
+        runEvaluatorTestCase(
+            query = "max(['a', 'abc', '3'])",
+            expectedResult = "'abc'",
+            expectedResultFormat = ExpectedResultFormat.STRING
```
alancai98, line 141: For all the tests added here, can we use the [ExpectedResultFormat.ION](https://github.com/partiql/partiql-lang-kotlin/blob/b7e3422d79bfc33ae23ce1a946b9dd6a7156b800/lang/test/org/partiql/lang/eval/evaluatortestframework/ExpectedResultFormat.kt#L10-L16)? It's the strictest (and preferred) option in general for testing the evaluator behavior.

lang/test/org/partiql/lang/eval/SimpleEvaluatingCompilerTests.kt
```diff
@@ -125,11 +126,115 @@ class SimpleEvaluatingCompilerTests : EvaluatorTestBase() {
 
     @Test
     fun max() {
+        // Number
```
alancai98, line 129: Also need some tests for `MIN`.

alancai98 (commented):
Implementation looks good. Left a nit about improving the error message for aggregation functions that only take in numeric values. Other comments related to testing.

lang/test/org/partiql/lang/eval/SimpleEvaluatingCompilerTests.kt
lziq: I am okay with this suggestion. However, it seems a bit wired to me to write `expectedResult` as `"\"abc\""`. In case that someone does not know the default format is ION instead of STRING, he may think `"abc"` is a double quoted variable.

lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
```diff
@@ -110,25 +110,25 @@ val ExprValue.address: ExprValue?
     get() = asFacet(Addressed::class.java)?.address
 
 fun ExprValue.booleanValue(): Boolean =
-    scalar.booleanValue() ?: errNoContext("Expected boolean: $ionValue", errorCode = ErrorCode.EVALUATOR_UNEXPECTED_VALUE_TYPE, internal = false)
+    scalar.booleanValue() ?: errNoContext("A value of BOOLEAN type is expected, however, what's passed is: $ionValue, with type of ${ionValue.type}", errorCode = ErrorCode.EVALUATOR_UNEXPECTED_VALUE_TYPE, internal = false)
```
alancai98, line 113: The error message is more descriptive, but it's not quite what I had in mind for the previous [comment](https://github.com/partiql/partiql-lang-kotlin/pull/604/files#r870764266). For the aggregation functions that only accept numeric types (e.g. `SUM` and `AVG`), I felt the error could make it more explicit that a non-numeric value was provided to the aggregation function. I felt the current error on `ExprValue` doesn't make it clear about the numeric requirement for those aggregation functions. Let me know if this distinction is unclear.

A better error message imo would be something like:
```
org.partiql.lang.eval.EvaluationException: SUM expects number values but was provided "abc" with type of STRING.
```

lang/test/org/partiql/lang/eval/SimpleEvaluatingCompilerTests.kt
alancai98: All the tests in this file should use the [ExpectedResultFormat.ION](https://github.com/partiql/partiql-lang-kotlin/blob/b7e3422d79bfc33ae23ce1a946b9dd6a7156b800/lang/test/org/partiql/lang/eval/evaluatortestframework/ExpectedResultFormat.kt#L10-L16). I guess the string escaping could be weird w/ `ExpectedResultFormat.ION`, but we should definitely not use `ExpectedResultFormat.STRING`. From the description of [ExpectedResultFormat.STRING](https://github.com/partiql/partiql-lang-kotlin/blob/b7e3422d79bfc33ae23ce1a946b9dd6a7156b800/lang/test/org/partiql/lang/eval/evaluatortestframework/ExpectedResultFormat.kt#L33-L40):

```
    /**
     * The expected value is an arbitrary string.  [org.partiql.lang.eval.ExprValue.toString]` is called on the result
     * of the query and the expected/actual assertion is performed with regular string equality.
     *
     * This is suboptimal (really, don't use this in new tests) but it is easier to support this here than it is to
     * refactor hundreds of expected values.
     */
    STRING
```

If you're worried about the quote escaping being confusing, you could use the triple-single quotes to not have to escape the strings: ```` ```"foo"```  ````, but I don't think the escaping is confusing.

lang/test/org/partiql/lang/eval/SimpleEvaluatingCompilerTests.kt
```diff
@@ -124,12 +125,197 @@ class SimpleEvaluatingCompilerTests : EvaluatorTestBase() {
...
+        runEvaluatorTestCase(
+            "min([`2020-01-01T00:00:00Z`, '1'])",
+            expectedResult = "`2020-01-01T00:00:00Z`",
+            expectedResultFormat = ExpectedResultFormat.STRING
+        )
+        runEvaluatorTestCase(
+            "min([`{{\"abcd\"}}`, '1'])",
+            expectedResult = "'1'",
+            expectedResultFormat = ExpectedResultFormat.STRING
+        )
```
alancai98, line 319: Should replicate these "Across data types" tests for `MAX`

lang/test/org/partiql/lang/eval/SimpleEvaluatingCompilerTests.kt
```diff
@@ -124,12 +125,197 @@ class SimpleEvaluatingCompilerTests : EvaluatorTestBase() {
     }
 
     @Test
-    fun max() {
+    fun maxAndMin() {
```
alancai98, line 128: Sorry missed this from the previous review, but it can be annoying to test an individual query since all of these tests are in one test function. I'd suggest either using `ParameterizedTest` or breaking individual or closely-related tests into separate test functions.

lziq: If we want to use `ParameterizedTest`, I think we should refactor the whole file to use that. However, I think this is outside the scope of this PR.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -180,8 +179,8 @@ internal class EvaluatingCompiler(
      * Base class for [ExprAggregator] instances which accumulate values and perform a final computation.
      */
     private inner class Accumulator(
-        var current: Number? = 0L,
-        val nextFunc: (Number?, ExprValue) -> Number,
+        var current: ExprValue?,
```
dlurton, line 182: The only thing that makes me a little nervous about this change is that we don't really have a lot of tests surrounding aggregates in general.... Nothing like what you're adding below for the other functions.  Please consider adding additional tests.

dlurton: In particular, with `sum` we need to ensure there are overflow (and underflow) conditions being tested in legacy and permissive modes.

lziq: Also, can you give some examples related to overflow and underflow?

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -192,37 +191,45 @@ internal class EvaluatingCompiler(
...
     private val builtinAggregates: Map<Pair<String, PartiqlAst.SetQuantifier>, ExprAggregatorFactory> =
         run {
-            val countAccFunc: (Number?, ExprValue) -> Number = { curr, _ -> curr!! + 1L }
-            val sumAccFunc: (Number?, ExprValue) -> Number = { curr, next ->
-                curr?.let { it + next.numberValue() } ?: next.numberValue()
+            fun checkIsNumberType(funcName: String, value: ExprValue) {
+                if (!value.type.isNumber) {
+                    throw errNoContext(
+                        message = "Aggregate function $funcName expects arguments of NUMBER type but the following value was provided: ${value.ionValue}, with type of ${value.type}",
+                        errorCode = ErrorCode.EVALUATOR_UNEXPECTED_VALUE_TYPE,
```
dlurton, line 212: I wonder if this is the appropriate error code?  Remember that error codes also have their own messages associated with them, with the intent of making them localizable.  Some customers use the exception message but there's at least one that I know of that uses the error messages associated with the error code.  Anyway, the error message associated with this code is: "Unexpected value type".  Consider adding a new error code with a similar / friendly error message.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -192,37 +191,45 @@ internal class EvaluatingCompiler(
...
 
     /** Dispatch table for built-in aggregate functions. */
     private val builtinAggregates: Map<Pair<String, PartiqlAst.SetQuantifier>, ExprAggregatorFactory> =
         run {
-            val countAccFunc: (Number?, ExprValue) -> Number = { curr, _ -> curr!! + 1L }
-            val sumAccFunc: (Number?, ExprValue) -> Number = { curr, next ->
-                curr?.let { it + next.numberValue() } ?: next.numberValue()
+            fun checkIsNumberType(funcName: String, value: ExprValue) {
+                if (!value.type.isNumber) {
+                    throw errNoContext(
```
dlurton, line 210: How is `throw` even allowed here?  I am not sure at all why this even builds, considering that `errNoContext` returns `Nothing`.  I'm surprised there's no compile error here.... actually, having tested this, it does generate a compile warning.  Please remove `throw`.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
dlurton: > Can we create a new issue to add more test cases?

I would try to put them in this PR if I could since you can't be certain you're not introducing some bug.

> Also, can you give some examples related to overflow and underflow?

This would be when the result of summation exceeds the maximum value that can be represented with an 8 byte value.  i.e. `SUM([1, 9223372036854775807])`.  What should happen?  Should it throw an exception, or should it coerce into a decimal which can grow to larger values?  (I'm not entirely certain myself since I haven't given this much thought.)  Underflow is the opposite: i.e. `SUM([-1, -9223372036854775808])`  In any case, I looked and I don't think we have any tests for this behavior today, and you should make sure this change doesn't alter it.

lziq: I tried `SUM([1, 9223372036854775807])` for both the original code and this revision. Both of them gave: 
```
org.partiql.lang.eval.EvaluationException: Int overflow or underflow
	Evaluator Error: at line 1, column 1: INT-8 overflow or underflow
```

alancai98 (approved):

Issue #606, opened 2022-05-18
Fix the `write_file` function to write Ion bags


johnedquinn:
## Overview
Following the [GettingStarted guide (regarding the section for reading/writing)](https://github.com/partiql/partiql-lang-kotlin/blob/main/docs/user/CLI.md#:~:text=The%20REPL%20also%20has%20the%20capability%20to%20write%20files%20with%20the%20write_file%20function%3A), there seems to be an issue with writing BAGs.

Similarly, following the [CLI guide](https://github.com/partiql/partiql-lang-kotlin/blob/main/docs/user/CLI.md#:~:text=write_file(%27out.tsv%27%2C%20SELECT%20name%2C%20type%20FROM%20animals%2C%20%7B%27type%27%3A%27tsv%27%2C%20%27header%27%3Atrue%7D)) causes issues when writing TSV.

## Error
```
org.partiql.lang.eval.EvaluationException: Invalid type for argument 2 of write_file.
	Evaluator Error: at line 1, column 1: Invalid argument type for write_file argument number 2, expected: [STRING] got: BAG
```

## Other Notes
I also ran this in permissive mode, and no files were created.

## To Reproduce
Create file `data.ion` with contents:
```
{ 'city': 'Seattle', 'state': 'WA' }
{ 'city': 'Bellevue', 'state': 'WA' }
{ 'city': 'Honolulu', 'state': 'HI' }
{ 'city': 'Rochester', 'state': 'NY' }
```

Command No. 1:
```
PartiQL> SELECT city FROM read_file('data.ion') AS c, `["HI", "NY"]` AS s WHERE c.state = s;
```

Output No. 1:
```
===' 
<<
  {
    'city': `Honolulu`
  },
  {
    'city': `Rochester`
  }
>>
--- 
OK!
```

Command No. 2:
```
PartiQL> write_file('out.ion', SELECT * FROM _, {'type': 'ion'})
   | 
```

Output No. 2:
```
org.partiql.lang.eval.EvaluationException: Invalid type for argument 2 of write_file.
	Evaluator Error: at line 1, column 1: Invalid argument type for write_file argument number 2, expected: [STRING] got: BAG

	at org.partiql.lang.eval.ExceptionsKt.err(Exceptions.kt:58)
	at org.partiql.lang.eval.ExceptionsKt.errInvalidArgumentType(Exceptions.kt:87)
[...]

PR #608, opened 2022-05-21
Fix write file

johnedquinn:
## Related Issue
Related to Issue #606

## Description
- Fixes the `write_file()` function
  - The old function required the input to be a string, but it must be a generic type.
  - For example:
    - a SELECT statement returns a bag
    - a SELECT .. ORDER BY statement returns a list
- Updates the README as well

### DIFF
--- a/cli/src/org/partiql/cli/functions/WriteFile.kt
+++ b/cli/src/org/partiql/cli/functions/WriteFile.kt
@@ -15,9 +15,7 @@
 package org.partiql.cli.functions
 
 import com.amazon.ion.IonStruct
-import com.amazon.ion.IonValue
 import com.amazon.ion.system.IonTextWriterBuilder
-import org.partiql.lang.domains.PartiqlAst
 import org.partiql.lang.eval.EvaluationSession
 import org.partiql.lang.eval.ExprValue
 import org.partiql.lang.eval.ExprValueFactory
@@ -27,7 +25,6 @@ import org.partiql.lang.types.FunctionSignature
 import org.partiql.lang.types.StaticType
 import org.partiql.lang.util.asIonStruct
 import org.partiql.lang.util.booleanValue
-import org.partiql.lang.util.isBag
 import org.partiql.lang.util.stringValue
 import java.io.FileOutputStream
 import java.io.OutputStream
--- a/cli/src/org/partiql/cli/functions/WriteFile.kt
+++ b/cli/src/org/partiql/cli/functions/WriteFile.kt
@@ -15,7 +15,9 @@
 package org.partiql.cli.functions
 
 import com.amazon.ion.IonStruct
+import com.amazon.ion.IonValue
 import com.amazon.ion.system.IonTextWriterBuilder
+import org.partiql.lang.domains.PartiqlAst
 import org.partiql.lang.eval.EvaluationSession
 import org.partiql.lang.eval.ExprValue
 import org.partiql.lang.eval.ExprValueFactory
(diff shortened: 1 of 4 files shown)


lziq:
Do you mind also adding some test for it? Currently, in https://github.com/partiql/partiql-lang-kotlin/blob/main/cli/test/org/partiql/cli/functions/WriteFileTest.kt, there is only unit test which tests `callWithRequired` & `callWithOptional` methods for the `write_file` function. Do you think it makes sense to refactor it and perform integration test, where the input is a query with `write_file` function?

johnedquinn:
> Do you mind also adding some test for it? Currently, in https://github.com/partiql/partiql-lang-kotlin/blob/main/cli/test/org/partiql/cli/functions/WriteFileTest.kt, there is only unit test which tests `callWithRequired` & `callWithOptional` methods for the `write_file` function. Do you think it makes sense to refactor it and perform integration test, where the input is a query calling `write_file` function?

@lziq -- Addressed. Let me know if all good! Thanks.

lziq (approved):
File changes make sense to me.

cli/src/org/partiql/cli/functions/WriteFile.kt
```diff
@@ -33,7 +33,7 @@ import java.io.OutputStreamWriter
 internal class WriteFile(valueFactory: ExprValueFactory) : BaseFunction(valueFactory) {
     override val signature = FunctionSignature(
         name = "write_file",
-        requiredParameters = listOf(StaticType.STRING, StaticType.STRING),
+        requiredParameters = listOf(StaticType.STRING, StaticType.ANY),
```
dlurton, line 36: `StaticType.STRING` is redundant when `StaticType.ANY` is present, since `StaticType.ANY` by definition includes the string type.

Consider making this just:
```
requiredParameters = listOf(StaticType.ANY),
```

cli/test/org/partiql/cli/CliTestUtility.kt
```diff
@@ -0,0 +1,61 @@
...
+import java.io.ByteArrayOutputStream
+import java.io.OutputStream
+
+/**
+ * A helper class to initialize the internal-scoped [Cli] and run queries for the purpose of unit and integration
+ * testing.
+ */
+class CliTestUtility() {
+
+    companion object {
```
dlurton, line 19: So, in Kotlin if you want a statically scoped method (like Java's `static`), it is generally preferable to use top-level functions instead.  These look and work like Java's `static` functions, but they really aren't.  I would remove the companion object and move the function to line 13 or so.  Same with other companion objects.  In general, companion objects should be avoided if all you want is something like Java's `static` functions.

cli/src/org/partiql/cli/functions/WriteFile.kt
lziq: I think we need both of `StaticType.STRING` and `StaticType.ANY`, because they refer to how many arguments this function has.

cli/test/org/partiql/cli/CliTestUtility.kt
johnedquinn: Thanks! Will address.

cli/src/org/partiql/cli/functions/WriteFile.kt
johnedquinn: Unsure what you mean -- but lmk if below is correct:

`listOf()` returns the 2 required parameters (1: file path, 2: content to write). In our case, the file path needs to be a string, and the second argument (content), can be anything that our query returns (bags, lists, etc).

dlurton: Oh.  Wow I misread that.  Never mind.  Resolving.

lziq (approved):

Issue #602, opened 2022-05-09
Implement `CEIL` and `FLOOR` functions


alancai98:
Similar to #601, would be good to support `CEIL` and `FLOOR` functions for decimal types. [PostgreSQL](https://www.postgresql.org/docs/current/functions-math.html)'s syntax and semantics seem a reasonable starting point for these functions.

`CEIL`/`CEILING`:
> ceil ( numeric ) → numeric ceil ( double precision ) → double precisionNearest integer greater than or equal to argumentceil(42.2) → 43ceil(-42.8) → -42
--
ceiling ( numeric ) → numeric ceiling ( double precision ) → double precisionNearest integer greater than or equal to argument (same as ceil)ceiling(95.3) → 96

`FLOOR`:
> floor ( numeric ) → numeric
floor ( double precision ) → double precision
    Nearest integer less than or equal to argument
    floor(42.8) → 42
    floor(-42.8) → -43

RCHowell:
I will take this as a first issue to familiarize myself with the PR process.

RCHowell:
https://github.com/partiql/partiql-lang-kotlin/pull/613

Issue #551, opened 2022-03-13
Add convenience StaticType for "TEXT"


mcliedtke:
There are a number of types defined in [`StaticType`][1] including some convenience types like `NULL_OR_MISSING` which is the union of the `NULL` and `MISSING` types. I believe a common enough use case would be to accept "text" without needing to distinguish between `String` or `Symbol`, the Ion domain objects have both `IonString` and `IonSymbol` extending `IonText` as a common type. Would it be appropriate to add an additional convenience enum type to `StaticType`  called `TEXT` which is the union of `STRING` and `SYMBOL`.

I came across this when writing a custom function and defining the function signature. I would like to accept either Strings or Symbols. This would a minor QoL improvement. If it would be acceptable, I could work on the PR.

[1]: https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/src/org/partiql/lang/types/StaticType.kt

alancai98:
Hi @mcliedtke,

We think it would be a good idea to add the convenience enum type which is the union of `STRING` and `SYMBOL`. We already have the concept of this type in some of our extension functions (e.g. https://github.com/partiql/partiql-lang-kotlin/blob/d6ad8d202797c0b1afbccb91dcbc11a507c7fdf7/lang/src/org/partiql/lang/ast/passes/inference/StaticTypeExtensions.kt#L24).

Regarding the name, `TEXT` seems like an appropriate name and is what Ion's types use along with [PostgreSQL's docs](https://www.postgresql.org/docs/14/functions-string.html). Another name suggested by @am357 could be `STRING_OR_SYMBOL` which aligns with `NULL_OR_MISSING` and provides some specificity.

Feel free to create a PR adding this new type. Let us know if you have any questions.

mcliedtke:
Thanks Alan. I'm fine either way with the naming, let me know if you have a preference - `STRING_OR_SYMBOL` aligns with `NULL_OR_MISSING`, but `TEXT` matches those other references and also leaves the door open potentially for a `NUMERIC` convenience type as well (and `INT_OR_FLOAT_OR_DECIMAL` seems clunky). So I would personally lean slightly towards `TEXT`.

For the line you pointed out in `StaticTypeExtensions`, would that need to be updated as well
[...]

Issue #494, opened 2022-01-11
Support for math/numerical functions


KieranHunt:
Hello!

It seems like this implementation of PartiQL doesn't support some common math/numerical functions. For example:

```
org.partiql.lang.eval.EvaluationException: No such function: pi
...
org.partiql.lang.eval.EvaluationException: No such function: pow
...
org.partiql.lang.eval.EvaluationException: No such function: power
```

A good list of these functions is defined in the MariaDB docs: https://mariadb.com/kb/en/numeric-functions/

PR #613, opened 2022-05-26
partiql-lang-kotlin-602 Implement `CEIL` and `FLOOR` functions

RCHowell:
Issue #602 

*Description of changes:*
I've created a class to hold math function implementations. This issue is only for CEIL and FLOOR.


lang/src/org/partiql/lang/eval/builtins/MathFunctions.kt
```diff
@@ -0,0 +1,46 @@
...
+        UnaryNumericToInt("ceil", valueFactory) { ceil(it.toDouble()) },
+        UnaryNumericToInt("ceiling", valueFactory) { ceil(it.toDouble()) },
+        UnaryNumericToInt("floor", valueFactory) { floor(it.toDouble()) },
+    )
+}
+
+private val numericTypes =
+    setOf(StaticType.INT2, StaticType.INT4, StaticType.INT8, StaticType.INT, StaticType.FLOAT, StaticType.DECIMAL)
+
+class UnaryNumericToInt(
```
alancai98, line 30: Can some documentation (i.e. [kdoc](https://kotlinlang.org/docs/kotlin-doc.html)) be added for this function? 

Also consider making this `private` or `internal`

lang/src/org/partiql/lang/eval/builtins/MathFunctions.kt
```diff
@@ -0,0 +1,46 @@
...
+    identifier: String,
+    private val valueFactory: ExprValueFactory,
+    private val func: (Number) -> Number,
+) : ExprFunction {
+
+    override val signature = FunctionSignature(
+        identifier,
+        listOf(AnyOfType(numericTypes)),
+        returnType = StaticType.FLOAT,
```
alancai98, line 39: Why is the `returnType` `FLOAT`? Shouldn't it be a `INT`?

lang/src/org/partiql/lang/eval/builtins/MathFunctions.kt
```diff
@@ -0,0 +1,46 @@
...
+    )
+}
+
+private val numericTypes =
+    setOf(StaticType.INT2, StaticType.INT4, StaticType.INT8, StaticType.INT, StaticType.FLOAT, StaticType.DECIMAL)
+
+class UnaryNumericToInt(
+    identifier: String,
+    private val valueFactory: ExprValueFactory,
+    private val func: (Number) -> Number,
```
alancai98, line 33: Hmm I think it's slightly confusing to have this class's constructor have a function returning a `Number` when it's named `UnaryNumericToInt`. IMO, the implicit conversion to int on L44 can be easily misused. I think it would make more sense for `func` to return an `Int` rather than number and have the callers to the constructor (e.g. L20-23) explicitly perform the conversion in the `func` argument.

lang/test/org/partiql/lang/eval/builtins/functions/MathFunctionsTest.kt
```diff
@@ -0,0 +1,35 @@
...
+            ExprFunctionTestCase("ceiling(1)", "1"),
+            ExprFunctionTestCase("ceiling(1.0)", "1"),
+            ExprFunctionTestCase("ceiling(1.1)", "2"),
+            ExprFunctionTestCase("ceiling(-42.8)", "-42"),
+            ExprFunctionTestCase("floor(1)", "1"),
+            ExprFunctionTestCase("floor(1.0)", "1"),
+            ExprFunctionTestCase("floor(1.1)", "1"),
+            ExprFunctionTestCase("floor(-42.8)", "-43"),
+        )
+    }
```
alancai98, line 34: Can also add some error/negative test cases. Consider using the `checkInvalidArgType` and `checkInvalidArity` APIs that Ziqi had created. Example using those APIs for the `size` ExprFunction: https://github.com/partiql/partiql-lang-kotlin/blob/ecbaebc70b66d437d5ebed53e31b4c4e8315aeca/lang/test/org/partiql/lang/eval/builtins/functions/SizeEvaluationTest.kt#L42-L57

lang/test/org/partiql/lang/eval/builtins/functions/MathFunctionsTest.kt
```diff
@@ -0,0 +1,35 @@
...
+    @ParameterizedTest
+    @ArgumentsSource(MathFunctionsPassCases::class)
+    fun runPassTests(tc: ExprFunctionTestCase) = runEvaluatorTestCase(
+        tc.source,
+        expectedResult = tc.expectedLegacyModeResult
+    )
+
+    class MathFunctionsPassCases : ArgumentsProviderBase() {
+
+        override fun getParameters(): List<Any> = listOf(
```
alancai98, line 20: Nice set of tests! Can also add some tests
- using the Ion literal constructor (i.e. using backticks). For example: ``ceil(`1`)``  returns `1`
- testing with floats. For example: `ceil(1.1e0)` returns `2`
  - similarly, we may also need to define the semantics for special float values (i.e. `+inf`, `-inf`, and `NaN`)

lang/src/org/partiql/lang/eval/builtins/BuiltinFunctions.kt
```diff
@@ -52,7 +52,7 @@ internal fun createBuiltinFunctions(valueFactory: ExprValueFactory) =
         SizeExprFunction(valueFactory),
         FromUnixTimeFunction(valueFactory),
         UnixTimestampFunction(valueFactory)
-    )
+    ) + MathFunctions.create(valueFactory)
```
alancai98, line 55: Could we also add some documentation similar to what we have in [BuiltInFunctions.md](https://github.com/partiql/partiql-lang-kotlin/blob/ecbaebc70b66d437d5ebed53e31b4c4e8315aeca/docs/user/BuiltInFunctions.md)? Feel free to create another markdown file for just these math functions.

alancai98 (changes requested):
Nice first PR! I left some comments related to some API questions related to `UnaryNumericToInt` and some more tests + docs that can be added. Looks good otherwise.

alancai98:
PR is also the start to https://github.com/partiql/partiql-lang-kotlin/issues/494.

RCHowell:
Thank you for the feedback. I will approach this with #494 in mind.

lang/test/org/partiql/lang/eval/builtins/functions/MathFunctionsTest.kt
RCHowell: Yeah I was curious about the negative tests, because they don't necessarily test the function -- more like testing the logic for how function signatures are checked/enforced. I can add some negative tests as well.

lang/src/org/partiql/lang/eval/builtins/BuiltinFunctions.kt
RCHowell: I will create a document for the math functions since this will grow soon.

lang/test/org/partiql/lang/eval/builtins/functions/MathFunctionsTest.kt
RCHowell: CEIL / FLOOR of +inf / -inf / NaN should be +inf / -inf / NaN respectively

RCHowell: I will need to address this later after discussion

lang/src/org/partiql/lang/eval/builtins/MathFunctions.kt
```diff
@@ -0,0 +1,54 @@
...
+private val numericTypes =
+    setOf(StaticType.INT2, StaticType.INT4, StaticType.INT8, StaticType.INT, StaticType.FLOAT, StaticType.DECIMAL)
+
+/**
+ * A convenience class to wrap `(Double) -> Int` as a PartiQL ExprFunction
+ *
+ * @property identifier Symbol for the given function
+ * @property valueFactory
+ * @property function Function to invoke for the given signature
+ * @constructor
```
alancai98, line 36: Could add a brief description for the `valueFactory`. Also don't think we typically add the `@constructor` kdoc tag.

lang/test/org/partiql/lang/eval/builtins/functions/MathFunctionsTest.kt
alancai98: I think some of the float special value tests still need to be added

alancai98 (changes requested):
I think the PR is nearly ready. Need to still add some special float size tests and make some changes following #551 being merged.

docs/user/BuiltInFunctions.md
```diff
@@ -930,3 +930,29 @@ FROM_UNIXTIME(0.1)          -- `1970-01-01T00:00:00.1-00:00`
...
+```sql
+CEIL(1.1) = 2
+CEIL(-42.8) = -42
+```
+
+### FLOOR
+
+Returns the nearest integer less than or equal to the input.
+
+Signature : `FLOOR: Integer|Float|Decimal -> Integer`
```
alancai98, line 952: Since the function signature was changed for CEIL/FLOOR to return a numeric, can update the function signature and/or the function descriptions.

lang/src/org/partiql/lang/eval/builtins/MathFunctions.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    fun create(valueFactory: ExprValueFactory): List<ExprFunction> = listOf(
+        UnaryNumeric("ceil", valueFactory) { ceil(it) },
+        UnaryNumeric("ceiling", valueFactory) { ceil(it) },
+        UnaryNumeric("floor", valueFactory) { floor(it) },
+    )
+}
+
+/**
+ * A convenience class to wrap `(Double) -> Int` as a PartiQL ExprFunction
```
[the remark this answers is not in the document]
alancai98: Also may be worth calling out that special FLOAT values will result in FLOATs and INTs otherwise.

alancai98 (dismissed):
Looks good other than some minor documentation updates following the function signature change. Feel free to merge after those comments are addressed.

Issue #524, opened 2022-02-10
SQL IN Query is not working as expected


mustafaakin:
Hello, we have the following query with IN clause that is returning empty results:

```
SELECT * FROM users WHERE name IN (SELECT name FROM roles)
```

However, the same query works in Postgres.  Alternatively following works, but of course, it's not same query semantically:

```
SELECT * FROM users u WHERE EXISTS (SELECT r.name FROM roles r WHERE r.name = u.name)
```

What are we doing/assuming that is wrong?

Sample code to make trying easier:

```kotlin
  val ion = IonSystemBuilder.standard().build()
  val pipeline = CompilerPipeline.standard(ion)

  val globalVariables = Bindings.buildLazyBindings<ExprValue> {
    addBinding("users") {
      val values = ion.loader.load(
        """
        {
        "name": "mustafa",
        "age": 30,
        }
        {
        "name": "ilteris",
        "age": 26,
        }
        {
        "name": "onur",
        "age": 24,
        }""".trimMargin()
      )
      pipeline.valueFactory.newFromIonValue(values)
    }
    addBinding("roles") {
      val values = ion.loader.load(
        """
        {
        "username": "mustafa",
        "admin": true,
        }
        {
        "name": "ilteris",
        "admin": false,
        }""".trimMargin()
      )
      pipeline.valueFactory.newFromIonValue(values)
    }
  }

  val query = "SELECT * FROM users WHERE name IN (SELECT name FROM roles)"
  val e = pipeline.compile(query)
  val session = EvaluationSession.build {
    globals(globalVariables)
  }

  val result = e.eval(session)
  for (exprValue in result) {
    println(exprValue)
  }
```

lziq:
This query works `"SELECT * FROM users WHERE name IN (SELECT VALUE name FROM roles)"`. 

The sub-query you wrote `SELECT name FROM roles` outputs `<<{'name': 'mustafa'}, {'name': 'ilteris'}>>`, which is a bag of structs. The `name` in `users` is a string. That's why it returns you empty result.

mustafaakin:
Yes, it works. Thanks a lot!. By the way, is this expected per SQL spec and what Postgres and other databases are doing a workaround?

dlurton:
`SELECT VALUE` is a work-around, however the original query should work according to SQL-92.

lziq:
I think it is probably not possible to make the original query work. 

[...]

PR #614, opened 2022-05-27
Enable the IN clause to also compare the value of single-pair structs

johnedquinn:
## Related Issue
Related to #524

## Description
Allows the IN clause to compare the value of single-pair structs, to maintain compatibility with SQL-92. To do this, it is necessary to make two changes:
- If the rightOp is already computed as a sequence of single-pair structs (where each pair's value is a literal), we can use the optimized approach and immediately compare with that value.
- If the rightOp is not computed (for example: another SELECT returning a bag/list of structs, or a bag of structs with pairs that contain un-evaluated values), then we need to evaluate.

### Examples

```sql
SELECT age FROM << { 'name': 'John', 'age': 22 } >> WHERE name IN (SELECT name FROM <<{ 'name': 'John' }>>)
```
should result in:
```ion
<<
  { age: 22 }
>>
```
## Notice

### DIFF
--- a/cli/test/org/partiql/cli/CliTest.kt
+++ b/cli/test/org/partiql/cli/CliTest.kt
@@ -156,6 +156,30 @@ class CliTest {
     }
 
     @Test
+    fun withSelectIn() {
+        makeCli(
+            query = "SELECT age FROM << { 'name': 'John', 'age': 22 } >> WHERE name IN (SELECT name FROM <<{ 'name': 'John' }>>)",
+            output = FileOutputStream(testFile),
+            outputFormat = OutputFormat.ION_TEXT
+        ).run()
+        val actual = testFile.bufferedReader().use { it.readText() }
+
+        assertEquals("{age:22}\n", actual)
+    }
+
+    @Test
+    fun withSelectInButEmpty() {
+        makeCli(
+            query = "SELECT age FROM << { 'name': 'John', 'age': 22 } >> WHERE name IN (SELECT name FROM <<{ 'name': 'john' }>>)",
+            output = FileOutputStream(testFile),
+            outputFormat = OutputFormat.ION_TEXT
+        ).run()
+        val actual = testFile.bufferedReader().use { it.readText() }
+
+        assertEquals("\n", actual)
+    }
+
+    @Test
     fun withoutInput() {
         val subject = makeCli("1")
         val actual = subject.runAndOutput()
--- a/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
+++ b/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
@@ -754,7 +754,9 @@ internal class EvaluatingCompiler(
         val rightOp = args[1]
         fun PartiqlAst.Expr.Struct.isSingleValueOptimizedStruct() = this.fields.size == 1 && this.fields[0].second is PartiqlAst.Expr.Lit
 
-        fun isOptimizedCase(values: List<PartiqlAst.Expr>): Boolean = values.all { it is PartiqlAst.Expr.Lit && !it.value.isNull }
+        fun isOptimizedCase(values: List<PartiqlAst.Expr>): Boolean = values.all {
+            (it is PartiqlAst.Expr.Lit && !it.value.isNull) || (it is PartiqlAst.Expr.Struct && it.isSingleValueOptimizedStruct())
+        }
 
         fun optimizedCase(values: List<PartiqlAst.Expr>): ThunkEnv {
             // Put all the literals in the sequence into a pre-computed map to be checked later by the thunk.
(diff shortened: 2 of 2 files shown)


dlurton:
This change only applies to `EvaluatingCompiler`.  Since we're intending to remove the `EvaluatingCompiler` at some point and its replacement isn't complete yet, there are two options:

1. Make this change in `EvaluatingCompiler` **and** its replacement (`PhysicalBexprToThunkConverter`, introduced in #592 and will be merged to `main` today if all goes well). 
2. Make this change in only `PhysicalBexprToThunkConverter`, which means this fix will have limited applicability until the `PhysicalBexprToThunkConverter` is complete and `EvaluatingCompiler` is removed.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -743,6 +747,13 @@ internal class EvaluatingCompiler(
             // NOTE: we cannot use a [HashSet<>] here because [ExprValue] does not implement [Object.hashCode] or
             // [Object.equals].
             val precomputedLiteralsMap = values
+                .asSequence()
+                .map { value ->
+                    when (value) {
+                        is PartiqlAst.Expr.Struct -> if (value.isSingleValueOptimizedStruct()) value.fields[0].second else value
+                        else -> value
+                    }
+                }
```
dlurton, line 756: What's the use-case for this optimization?  Has this been reported by a customer as being slow?  If not, I would avoid optimizing this as it may be premature.  The other optimizations here were added as a direct result of customer feedback.

cli/test/org/partiql/cli/CliTest.kt
```diff
@@ -155,6 +155,30 @@ class CliTest {
...
+        makeCli(
+            query = "SELECT age FROM << { 'name': 'John', 'age': 22 } >> WHERE name IN (SELECT name FROM <<{ 'name': 'john' }>>)",
+            output = FileOutputStream(testFile),
+            outputFormat = OutputFormat.ION_TEXT
+        ).run()
+        val actual = testFile.bufferedReader().use { it.readText() }
+
+        assertEquals("\n", actual)
+    }
+
```
dlurton, line 181: Please add these tests to `EvaluatorTestSuite`--instead they are not needed here if the exist there.  

Also, please add a simpler versions of the test case which will be easier to understand the actual feature being tested:

These all must return `true`:

```
`foo` in << { 'foo': } >>
`foo` in [ { 'foo': } ]
`foo` in SEXP({ 'foo': })
```
And these must all return false:

```
`bar` in << { 'foo': } >>
`bar` in [ { 'foo': } ]
`bar` in SEXP({ 'foo': })
```

cli/test/org/partiql/cli/CliTest.kt
```diff
@@ -155,6 +155,30 @@ class CliTest {
...
+    fun withSelectInButEmpty() {
+        makeCli(
+            query = "SELECT age FROM << { 'name': 'John', 'age': 22 } >> WHERE name IN (SELECT name FROM <<{ 'name': 'john' }>>)",
+            output = FileOutputStream(testFile),
+            outputFormat = OutputFormat.ION_TEXT
+        ).run()
+        val actual = testFile.bufferedReader().use { it.readText() }
+
+        assertEquals("\n", actual)
+    }
```
dlurton, line 180: Anyway, looking again at your other tests, I don't think there's a need to test the CLI separately from the `EvaluatingCompilerInTests`.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
johnedquinn: ACK. I can modify. Thanks!

cli/test/org/partiql/cli/CliTest.kt
johnedquinn: ACK. I can remove. Thanks!

johnedquinn:
> This change only applies to `EvaluatingCompiler`. Since we're intending to remove the `EvaluatingCompiler` at some point and its replacement isn't complete yet, there are two options:
> 
> 1. Make this change in `EvaluatingCompiler` **and** its replacement (`PhysicalBexprToThunkConverter`, introduced in [Add Plan Evaluator #592](https://github.com/partiql/partiql-lang-kotlin/pull/592) and will be merged to `main` today if all goes well).
> 2. Make this change in only `PhysicalBexprToThunkConverter`, which means this fix will have limited applicability until the `PhysicalBexprToThunkConverter` is complete and `EvaluatingCompiler` is removed.

@dlurton -- Since this is working on a feature branch off of main, do you think it would make sense to create a separate PR to add to `PhysicalBexprToThunkConverter`? Then, this one can be merged immediately on approval, and I can work on merging to the other branch? Unless, of course, the other branch is expected to merge by EOD. Then, I'll just park this, pull once other is merged, then add to `PhysicalBexprToThunkConverter ` and merge in the same PR later. Whatever works.

dlurton:
@johnedquinn I am hoping it can be merged by EOD today but that will depend on getting the approval of @alancai98 

I would not make that separate PR due to the possibility of it being "inadvertently deprioritized".

johnedquinn:
@dlurton -- ACK. I'll park this and wait for the merge to main.

**Action items**:
- once #612 is merged to main, I'll pull and add my changes for the new evaluator compiler

johnedquinn:
Accidentally merged all of the new changes in main into this feature branch. Starting a new one to make it easier for reviewers. Following up with PR #621 .

PR #617, opened 2022-05-31
Cli add partiql input

johnedquinn:
## Description of Changes
- Adds the `--input-format` flag to the CLI
- Adds the ability to handle PartiQL data format for input (before was only Ion)
- Modifies the `--input` flag for CLI to not loop through the sequence of Ion values. Expect a single Ion value. (For the Ion input format)
- Modifies the `--output-format` option ION_TEXT to output the real value of the results (not loop through the bag)
- Updates the `read_file` function to read in a single Ion value
   - Before, the `read_file` function used to loop through the values of an Ion file and place all values within a bag. This limits our ability to differentiate between different Ion values. See the next bullet-point for more information.
- Updates the `write_file` function to write the actual result.
  - Before, the `write_file` function used to loop through the values within a bag/list and print those values. This would not allow for printing expressions like literals.
- Updates the documentation

### DIFF
--- a/cli/src/org/partiql/cli/Cli.kt
+++ b/cli/src/org/partiql/cli/Cli.kt
@@ -35,8 +35,9 @@ import java.io.OutputStreamWriter
 internal class Cli(
     private val valueFactory: ExprValueFactory,
     private val input: InputStream,
+    private val inputFormat: InputFormat,
     private val output: OutputStream,
-    private val format: OutputFormat,
+    private val outputFormat: OutputFormat,
     private val compilerPipeline: CompilerPipeline,
     private val globals: Bindings<ExprValue>,
     private val query: String
@@ -48,26 +49,50 @@ internal class Cli(
     }
 
     override fun run() {
+        when (inputFormat) {
+            InputFormat.ION -> runWithIonInput()
+            InputFormat.PARTIQL -> runWithPartiQLInput()
+        }
+    }
+
+    private fun runWithIonInput() {
         IonReaderBuilder.standard().build(input).use { reader ->
-            val inputIonValue = valueFactory.ion.iterate(reader).asSequence().map { valueFactory.newFromIonValue(it) }
-            val inputExprValue = valueFactory.newBag(inputIonValue)
-            val bindings = Bindings.buildLazyBindings<ExprValue> {
-                // If `input` is a class of `EmptyInputStream`, it means there is no input data provided by user.
-                if (input !is EmptyInputStream) { addBinding("input_data") { inputExprValue } }
-            }.delegate(globals)
-
+            val bindings = when (reader.next()) {
+                null -> Bindings.buildLazyBindings<ExprValue> {}.delegate(globals)
+                else -> getBindingsFromIonValue(valueFactory.newFromIonReader(reader))
+            }
             val result = compilerPipeline.compile(query).eval(EvaluationSession.build { globals(bindings) })
+            outputResult(result)
+        }
+    }
 
-            when (format) {
-                OutputFormat.ION_TEXT -> ionTextWriterBuilder.build(output).use {
-                    printIon(it, result)
-                    output.write(System.lineSeparator().toByteArray(Charsets.UTF_8))
-                }
-                OutputFormat.ION_BINARY -> valueFactory.ion.newBinaryWriter(output).use { printIon(it, result) }
-                OutputFormat.PARTIQL -> OutputStreamWriter(output).use { it.write(result.toString()) }
-                OutputFormat.PARTIQL_PRETTY -> OutputStreamWriter(output).use {
-                    ConfigurableExprValueFormatter.pretty.formatTo(result, it)
-                }
+    private fun runWithPartiQLInput() {
+        val partiql =
+            compilerPipeline.compile(input.readAllBytes().toString(Charsets.UTF_8)).eval(EvaluationSession.standard())
+        val bindings = getBindingsFromIonValue(partiql)
+        val result = compilerPipeline.compile(query).eval(EvaluationSession.build { globals(bindings) })
+        outputResult(result)
+    }
+
+    private fun getBindingsFromIonValue(value: ExprValue): Bindings<ExprValue> {
+        return Bindings.buildLazyBindings<ExprValue> {
+            // If `input` is a class of `EmptyInputStream`, it means there is no input data provided by user.
+            if (input !is EmptyInputStream) {
+                addBinding("input_data") { value }
+            }
+        }.delegate(globals)
+    }
+
+    private fun outputResult(result: ExprValue) {
+        when (outputFormat) {
+            OutputFormat.ION_TEXT -> ionTextWriterBuilder.build(output).use {
+                printIon(it, result)
+                output.write(System.lineSeparator().toByteArray(Charsets.UTF_8))
+            }
+            OutputFormat.ION_BINARY -> valueFactory.ion.newBinaryWriter(output).use { printIon(it, result) }
+            OutputFormat.PARTIQL -> OutputStreamWriter(output).use { it.write(result.toString()) }
+            OutputFormat.PARTIQL_PRETTY -> OutputStreamWriter(output).use {
+                ConfigurableExprValueFormatter.pretty.formatTo(result, it)
             }
         }
     }
--- a/cli/src/org/partiql/cli/Cli.kt
+++ b/cli/src/org/partiql/cli/Cli.kt
@@ -61,6 +61,11 @@ internal class Cli(
                 null -> Bindings.buildLazyBindings<ExprValue> {}.delegate(globals)
                 else -> getBindingsFromIonValue(valueFactory.newFromIonReader(reader))
             }
+            if (reader.next() != null) {
+                val message = "As of June 2022, PartiQL requires that Ion files contain only a single Ion value for " +
+                    "processing. Please consider wrapping multiple values in a list."
+                throw IllegalStateException(message)
+            }
             val result = compilerPipeline.compile(query).eval(EvaluationSession.build { globals(bindings) })
             outputResult(result)
         }
--- a/cli/src/org/partiql/cli/Cli.kt
+++ b/cli/src/org/partiql/cli/Cli.kt
@@ -60,7 +60,7 @@ internal class Cli(
                 else -> getBindingsFromIonValue(valueFactory.newFromIonReader(reader))
             }
             if (reader.next() != null) {
-                val message = "As of June 2022, PartiQL requires that Ion files contain only a single Ion value for " +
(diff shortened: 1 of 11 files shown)


docs/user/CLI.md
```diff
@@ -23,18 +23,18 @@ Option                                Description
 -e, --environment <File>              initial global environment (optional)
 -h, --help                            prints this help
 -i, --input <File>                    input file, requires the query option (optional)
+-if, --input-format <InputFormat>     input format, requires the query option -- [PARTIQL (default), ION]
```
alancai98, line 26: Is the default PARTIQL or ION? The default in [main.kt](https://github.com/partiql/partiql-lang-kotlin/blob/f33050e6a7f09ad7b9318af19f9ea1c5084bae4d/cli/src/org/partiql/cli/main.kt#L105) is ION. If the defaults are PARTIQL, may need to update the above text (and the [--help text](https://github.com/partiql/partiql-lang-kotlin/blob/f33050e6a7f09ad7b9318af19f9ea1c5084bae4d/cli/src/org/partiql/cli/main.kt#L50-L68)) to include the --input-format option.

cli/src/org/partiql/cli/Cli.kt
```diff
@@ -48,35 +47,53 @@ internal class Cli(
...
+                    "processing. Please consider wrapping multiple values in a list."
+                throw IllegalStateException(message)
+            }
             val result = compilerPipeline.compile(query).eval(EvaluationSession.build { globals(bindings) })
+            outputResult(result)
+        }
+    }
+
+    private fun runWithPartiQLInput() {
+        val partiql =
```
alancai98, line 73: nit: non-descriptive variable. Should reference that we're reading in the input environment.

cli/test/org/partiql/cli/CliTest.kt
```diff
@@ -58,6 +60,7 @@ class CliTest {
         Cli(
             valueFactory,
             input?.byteInputStream(Charsets.UTF_8) ?: EmptyInputStream(),
+            InputFormat.ION,
```
alancai98, line 63: Could have some tests with `InputFormat.PARTIQL`. Also I don't think you should hard-code the this argument to the `Cli` but rather pass it in as another argument to `makeCli`. You could set the default `inputFormat` for `makeCli` to `ION` depending on what the actual default is.

cli/test/org/partiql/cli/CliTestUtility.kt
```diff
@@ -26,6 +26,7 @@ fun makeCliAndGetResult(
     val cli = Cli(
         valueFactory,
         input?.byteInputStream(Charsets.UTF_8) ?: EmptyInputStream(),
+        InputFormat.ION,
```
alancai98, line 29: For consistency with the other `Cli` constructor args, I'd not hardcode the `inputFormat` argument but would rather pass it as an argument to `makeCliAndGetResult`.

alancai98 (changes requested):
Thanks for diving deep into some of the inconsistencies with the existing PartiQL CLI + REPL! I think it's headed in the right direction and will improve the usability of the CLI + REPL for newcomers.

Left some mostly minor comments related to the documentation. Also think there needs to be some tests with the `input-file` CLI option set to `PARTIQL`.

johnedquinn:
ACK. I'll check the comments out and address. Appreciate it!

cli/src/org/partiql/cli/functions/ReadFile.kt
```diff
@@ -65,13 +66,25 @@ internal class ReadFile(valueFactory: ExprValueFactory) : BaseFunction(valueFact
...
     private fun ionReadHandler(): (InputStream, IonStruct) -> ExprValue = { input, _ ->
-        @Suppress("DEPRECATION")
-        valueFactory.newBag(valueFactory.ion.iterate(input).asSequence().map { valueFactory.newFromIonValue(it) })
+        IonReaderBuilder.standard().build(input).use { reader ->
+            val value = when (reader.next()) {
+                null -> valueFactory.missingValue
+                else -> valueFactory.newFromIonReader(reader)
+            }
+            if (reader.next() != null) {
+                val message = "As of June 2022, PartiQL requires that Ion files contain only a single Ion value for " +
```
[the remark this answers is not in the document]
johnedquinn: Thanks! Will do.

docs/user/CLI.md
```diff
@@ -439,65 +441,59 @@ PartiQL> SELECT city FROM read_file('data.ion') AS c, `["HI", "NY"]` AS s WHERE
...
-{
-  city:Honolulu
-}
-{
-  city:Rochester
-}
+
+A file called `out.ion` will be created in the `cli` directory with the following contents:
+```ion
+[
```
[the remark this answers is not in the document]
johnedquinn: ACK. Nice catch.

lziq:
I like the idea, but I think we still need to discuss it with everyone to see if others agree. Also, there are some backward compatibility problems we need to take care of, such as throwing error when reading Ion file with multiple Ion values.

docs/user/CLI.md
```diff
@@ -23,18 +23,18 @@ Option                                Description
 -e, --environment <File>              initial global environment (optional)
 -h, --help                            prints this help
 -i, --input <File>                    input file, requires the query option (optional)
+-if, --input-format <InputFormat>     input format, requires the query option -- [PARTIQL (default), ION]
 -o, --output <File>                   output file, requires the query option (default: stdout)
---of, --output-format <OutputFormat:  output format, requires the query option (default: PARTIQL)
-  (ION_TEXT|ION_BINARY|PARTIQL|PARTIQL_PRETTY)>
+-of, --output-format <OutputFormat>   output format, requires the query option [PARTIQL (default), PARTIQL_PRETTY, ION_TEXT, ION_BINARY]
```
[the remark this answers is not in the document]
alancai98: Don't think this comment was addressed. Might also need addressed for [main.kt](https://github.com/partiql/partiql-lang-kotlin/blob/860f78023f0a2a1042d9797668c5da241b24163a/cli/src/org/partiql/cli/main.kt#L133-L135)

docs/user/CLI.md
```diff
@@ -43,72 +43,84 @@ project root).
...
-OK!
 ```
 
 Press control-D to exit the REPL.
 
-## Advanced REPL Features
+### Advanced REPL Features
 
-To view the AST of an SQL statement, type one and press enter only *once*, then type `!!` and press enter:
+To view the AST of a PartiQL statement, type one and press enter only *once*, then type `!!` and press enter:
```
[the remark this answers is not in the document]
alancai98: bump^

docs/user/CLI.md
```diff
@@ -409,28 +411,28 @@ PartiQL> SELECT * FROM stores AS s
...
----
-OK!
 ```
 
-# Reading/Writing Files
-The REPL provides the `read_file` function to stream data from a file. The files needs to be placed in the folder `cli`. 
-For example:
+## Reading/Writing Files
+The REPL provides the `read_file` function to stream data from a file. The files need to be placed in the folder `cli`, 
+and they must contain only a single Ion value (typically a list).
```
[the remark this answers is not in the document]
alancai98: bump^

cli/src/org/partiql/cli/main.kt
```diff
@@ -93,6 +97,13 @@ private val inputFileOpt = optParser.acceptsAll(listOf("input", "i"), "input fil
     .withRequiredArg()
     .ofType(File::class.java)
 
+private val inputFormatOpt = optParser.acceptsAll(listOf("input-format", "if"), "output format, requires the query option")
```
alancai98, line 100: Might have missed this from the first review. Hint is incorrect

```suggestion
private val inputFormatOpt = optParser.acceptsAll(listOf("input-format", "if"), "input format, requires the query option")
```

cli/test/org/partiql/cli/CliTest.kt
```diff
@@ -47,125 +43,142 @@ class CliTest {
...
     fun runQueryOnSingleValue() {
-        val subject = makeCli("SELECT * FROM input_data", "{a: 1}")
-        val actual = subject.runAndOutput()
+        val query = "SELECT * FROM input_data"
+        val input = "[{'a': 1}]"
+        val expected = "$partiqlBagAnnotation[{a: 1}]"
 
-        assertAsIon("{a: 1}", actual)
+        val ionInputResult = makeCliAndGetResult(query, input, inputFormat = InputFormat.ION)
+        val partiqlInputResult = makeCliAndGetResult(query, input, inputFormat = InputFormat.PARTIQL)
```
alancai98, line 53: Thanks for adding some tests with the `InputFormat.PARTIQL`. I think we may need a couple additional tests:
- use `InputFormat.PARTIQL` to load an input that uses PartiQL syntax (e.g. a bag with the angle bracket syntax: `<<{'a': 1}>>`)
- use `InputFormat.ION` to load an input with PartiQL syntax (e.g. bag w/ angle brackets), which should error

AFAICT, the inputs for these tests just use Ion syntax and nothing specific to PartiQL.

alancai98 (changes requested):
Some other minor comments related to documentation + tests. Otherwise looks good. Also there were some other comments that I don't believe were addressed (GitHub sometimes hides comments if there are too many of them and you have to hit "Load more").

image

Agree w/ @lziq that we need to decide as a team on the breaking semantic changes to the CLI + REPL. I think these changes make the CLI more usable.

alancai98:
Btw to keep the commit history slightly cleaner for PRs as `main` is updated, I generally prefer rebasing `main` on to the working branch rather than merging. This can help prevent unrelated merge commits from appearing in the PR.

johnedquinn:
Yeah, I'm still getting used to the approval process and didn't realize there were hidden comments. Addressed. Thanks for pointing that out @alancai98

johnedquinn:
As discussed with team, going to add a flag to use the old implementation -- along with modifying the exception message specifying the flag use.

cli/src/org/partiql/cli/Cli.kt
```diff
@@ -35,48 +33,92 @@ import java.io.OutputStreamWriter
...
     private val compilerPipeline: CompilerPipeline,
     private val globals: Bindings<ExprValue>,
-    private val query: String
+    private val query: String,
+    private val wrapIon: Boolean
 ) : PartiQLCommand {
 
+    init {
+        if (wrapIon && inputFormat != InputFormat.ION) {
+            throw IllegalArgumentException("Specifying --wrap-ion requires that the input format be ${InputFormat.ION}.")
```
alancai98, line 47: nit: "Specifying -w/--wrap-ion requires..."

alancai98 (approved):
Thanks for addressing the feedback and adding in the `-w`/`--wrap-ion` flag to the CLI. Left some minor comments and a comment related to `read_file`. Feel free to decide if they're relevant.

PR #619, opened 2022-05-31
Rewrite built-in types

lziq:
*Issue #594*

*Description of changes:*
1. Modified `SqlType` interfaces: 
    a. Changed `typeName` into `typeAliases`, since one type can have multiple aliases. 
    b. Removed `parentType` as there is no need for type inheritence in PartiQL for now. 
2. Modified `ParametricType` interfaces. 
    a. Extended `validateParameters` with 2 more methods, which are `requiredParameters` & `optionalParameters`. They are used to define required parameters and optional parameters for parametric types. And now `validateParameters` is only used to perform other checks, such as value of each parameter and constraints across parameters. Now `validateParameters` throws exceptions if any check is not passed. 
3. Rewrote all the built-in types using the new type interfaces. 

Note that `any` type https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/resources/org/partiql/type-domains/partiql.ion#L424 will be rewritten when inferencer is refactored.

### DIFF
--- a/lang/src/org/partiql/lang/typesystem/builtin/types/nonprimitive/CharType.kt
+++ b/lang/src/org/partiql/lang/typesystem/builtin/types/nonprimitive/CharType.kt
@@ -1,0 +1,41 @@
+package org.partiql.lang.typesystem.builtin.types.nonprimitive
+
+import org.partiql.lang.eval.ExprValue
+import org.partiql.lang.eval.ExprValueType
+import org.partiql.lang.typesystem.builtin.types.primitive.IntType
+import org.partiql.lang.typesystem.builtin.types.primitive.StringType
+import org.partiql.lang.typesystem.builtin.types.valueFactory
+import org.partiql.lang.typesystem.interfaces.type.BuiltInType
+import org.partiql.lang.typesystem.interfaces.type.ParametricType
+import org.partiql.lang.typesystem.interfaces.type.SqlType
+import org.partiql.lang.typesystem.interfaces.type.ValueWithType
+import org.partiql.lang.util.asIonInt
+
+object CharType : BuiltInType(), ParametricType {
+    override val typeAliases: List<String>
+        get() = listOf("char", "character")
+
+    override val exprValueType: ExprValueType
+        get() = ExprValueType.STRING
+
+    override val parentType: SqlType
+        get() = StringType
+
+    override val isPrimitiveType: Boolean
+        get() = false
+
+    override val requiredParameters: List<SqlType> = emptyList()
+
+    override val optionalParameters: List<Pair<SqlType, ExprValue>> = listOf(
+        IntType to valueFactory.newInt(Integer.MAX_VALUE)
+    )
+
+    override fun validateParameters(parameters: List<ValueWithType>) {
+        // Check if length is larger than or equal to 0
+        val length = parameters[0]
+        val value = length.value.ionValue.asIonInt().longValue()
+        if (value < 0) {
+            throw IllegalArgumentException("Compile Error: The parameter of $this type, length, should be larger than or equal to 0")
+        }
+    }
+}
--- a/lang/src/org/partiql/lang/typesystem/builtin/types/nonprimitive/CharType.kt
+++ b/lang/src/org/partiql/lang/typesystem/builtin/types/nonprimitive/CharType.kt
@@ -14,8 +14,10 @@ import org.partiql.lang.util.asIonInt
  * The standard sql type CHARACTER(n), where "n" is the fixed number of characters of the string literal
  */
 object CharType : BuiltInType(), ParametricType {
-    private const val unboundedLength = Integer.MAX_VALUE
-    private const val maxLength = Integer.MAX_VALUE - 1
+    /**
+     * If the parameter's value is -1, it means the user did not explicitly specify the parameter of the VARCHAR type.
+     */
+    private const val unboundedLength = -1
 
     override val typeAliases: List<String>
         get() = listOf("char", "character")
@@ -38,9 +40,6 @@ object CharType : BuiltInType(), ParametricType {
         val value = length.value.ionValue.asIonInt().longValue()
         if (value < 0) {
             throw IllegalArgumentException("Compile Error: The parameter of $this type, length, should be larger than or equal to 0")
-        }
-        if (value > maxLength) {
-            throw IllegalArgumentException("Compile Error: The parameter of $this type, length, should be less than or equal to $maxLength")
         }
     }
 }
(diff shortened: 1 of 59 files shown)


lang/src/org/partiql/lang/typesystem/builtin/types/nonprimitive/CharType.kt
```diff
@@ -0,0 +1,37 @@
...
+    override val exprValueType: ExprValueType
+        get() = ExprValueType.STRING
+
+    override val isPrimitiveType: Boolean
+        get() = false
+
+    override val requiredParameters: List<SqlType> = emptyList()
+
+    override val optionalParameters: List<Pair<SqlType, ExprValue>> = listOf(
+        IntType to valueFactory.newInt(Integer.MAX_VALUE)
```
am357, line 26: Curious, why `MAX_VALUE`?

lang/src/org/partiql/lang/typesystem/builtin/types/nonprimitive/CharType.kt
```diff
@@ -0,0 +1,37 @@
...
+import org.partiql.lang.eval.ExprValueType
+import org.partiql.lang.typesystem.builtin.types.primitive.IntType
+import org.partiql.lang.typesystem.builtin.types.valueFactory
+import org.partiql.lang.typesystem.interfaces.type.BuiltInType
+import org.partiql.lang.typesystem.interfaces.type.ParametricType
+import org.partiql.lang.typesystem.interfaces.type.SqlType
+import org.partiql.lang.typesystem.interfaces.type.TypeParameters
+import org.partiql.lang.util.asIonInt
+
+object CharType : BuiltInType(), ParametricType {
```
am357, line 13: For this and others that apply: let's have a decent `kdoc` for the types with examples.

lang/src/org/partiql/lang/typesystem/builtin/types/nonprimitive/DoublePrecisionType.kt
```diff
@@ -0,0 +1,18 @@
+package org.partiql.lang.typesystem.builtin.types.nonprimitive
+
+import org.partiql.lang.eval.ExprValueType
+import org.partiql.lang.typesystem.interfaces.type.BuiltInType
+
+/**
+ * DOUBLE_PRECISION type is actually FLOAT8
+ */
+object DoublePrecisionType : BuiltInType() {
```
am357, line 9: Why aren't we considering this and other `int` and `float` types as primitive?

lziq: There is no `DoublePrecision` as one of `ExprValueType`.

lziq: So basically, I want to classify types into primitive types and non-primitive types. If a sql type corresponds to one `ExprValueType`, it is a primitive type. As far as I can see, this classification is not used anywhere, but I think it helps us to understand differences between types.

lang/src/org/partiql/lang/typesystem/builtin/types/nonprimitive/CharType.kt
lziq: I think trinodb does the same, https://github.com/trinodb/trino/blob/master/core/trino-spi/src/main/java/io/trino/spi/type/VarcharType.java#L47

am357: But do we know why?

lang/src/org/partiql/lang/typesystem/builtin/types/nonprimitive/CharType.kt
lziq: Do we want to write documentations to explain what each type does?

am357: > Do we want to write documentations (as separate files) to explain what each type does?
That can also be done, but at its very least we need some sort of documentation in the code.

lang/src/org/partiql/lang/typesystem/builtin/types/nonprimitive/CharType.kt
lziq: This is just my speculation. I think normally, the parameter of `VARCHAR` is not going to be too large, so `Integer.MAX_VALUE` is large enough to deal with all the use cases.

lziq: Probably, we can also add a constraint in `validateParameters` method and requires the parameter to be less than `Integer.MAX_VALUE`.

lziq: Sorry for the confusion. Actually, TrinoDB uses `Integer.MAX_VALUE` to represent unbounded length, and requires the parameter of `VARCHAR` to be less than `Integer.MAX_VALUE`.

lang/src/org/partiql/lang/typesystem/builtin/types/nonprimitive/CharType.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    override val optionalParameters: List<Pair<SqlType, ExprValue>> = listOf(
+        IntType to valueFactory.newInt(unboundedLength)
+    )
+
+    override fun validateParameters(parameters: TypeParameters) {
+        // Check if length is larger than 0
+        val length = parameters[0]
+        val value = length.value.ionValue.asIonInt().longValue()
+        if (value <= 0) {
```
am357, line 41: Don’t we still need a `maxLength` for the when the user provides the `length`?

lang/src/org/partiql/lang/typesystem/builtin/types/nonprimitive/DoublePrecisionType.kt
am357: I still need to think about this more, but for now, if we're going with this categorization we should make it clear in in `kdoc`.

lang/src/org/partiql/lang/typesystem/builtin/types/nonprimitive/CharType.kt
am357: I was thinking we could have a value less that `INTEGER.MAX_VALUE` to control the amount of text (the actual size in bytes) just like what postgres does with [1GB](https://www.postgresql.org/docs/current/datatype-character.html) value but that’s a non-blocker for now.

am357 (approved):

PR #621, opened 2022-05-31
Enable the IN operator to also compare the value of single-pair structs (Revised)

### LINKED ISSUE #524: SQL IN Query is not working as expected

johnedquinn:
### Related Issue
Resolves #524 

### Note
I actually made a PR #614, but I merged all of @dlurton's changes in, and it made it impossible to read. This is a new PR, and I will be closing/cancelling the other.

## Description
Allows the IN clause to compare the value of single-pair structs, to maintain compatibility with SQL-92. To do this, it is necessary to make two changes:
- If the rightOp is already computed as a sequence of single-pair structs (where each pair's value is a literal), we can use the optimized approach and immediately compare with that value.
- If the rightOp is not computed (for example: another SELECT returning a bag/list of structs, or a bag of structs with pairs that contain un-evaluated values), then we need to evaluate.

### Examples

```sql
SELECT age FROM << { 'name': 'John', 'age': 22 } >> WHERE name IN (SELECT name FROM <<{ 'name': 'John' }>>)
```
should result in:
```ion
<<
  { age: 22 }
>>
```

### DIFF
--- a/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
+++ b/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
@@ -754,9 +754,7 @@ internal class EvaluatingCompiler(
         val rightOp = args[1]
         fun PartiqlAst.Expr.Struct.isSingleValueOptimizedStruct() = this.fields.size == 1 && this.fields[0].second is PartiqlAst.Expr.Lit
 
-        fun isOptimizedCase(values: List<PartiqlAst.Expr>): Boolean = values.all {
-            (it is PartiqlAst.Expr.Lit && !it.value.isNull) || (it is PartiqlAst.Expr.Struct && it.isSingleValueOptimizedStruct())
-        }
+        fun isOptimizedCase(values: List<PartiqlAst.Expr>): Boolean = values.all { it is PartiqlAst.Expr.Lit && !it.value.isNull }
 
         fun optimizedCase(values: List<PartiqlAst.Expr>): ThunkEnv {
             // Put all the literals in the sequence into a pre-computed map to be checked later by the thunk.
@@ -766,13 +764,6 @@ internal class EvaluatingCompiler(
             // NOTE: we cannot use a [HashSet<>] here because [ExprValue] does not implement [Object.hashCode] or
             // [Object.equals].
             val precomputedLiteralsMap = values
-                .asSequence()
-                .map { value ->
-                    when (value) {
-                        is PartiqlAst.Expr.Struct -> if (value.isSingleValueOptimizedStruct()) value.fields[0].second else value
-                        else -> value
-                    }
-                }
                 .filterIsInstance<PartiqlAst.Expr.Lit>()
                 .mapTo(TreeSet<ExprValue>(DEFAULT_COMPARATOR)) {
                     valueFactory.newFromIonValue(
--- a/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
+++ b/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
@@ -752,7 +752,6 @@ internal class EvaluatingCompiler(
         val args = expr.operands
         val leftThunk = compileAstExpr(args[0])
         val rightOp = args[1]
-        fun PartiqlAst.Expr.Struct.isSingleValueOptimizedStruct() = this.fields.size == 1 && this.fields[0].second is PartiqlAst.Expr.Lit
 
         fun isOptimizedCase(values: List<PartiqlAst.Expr>): Boolean = values.all { it is PartiqlAst.Expr.Lit && !it.value.isNull }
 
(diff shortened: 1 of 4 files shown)


lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -817,6 +819,12 @@ internal class EvaluatingCompiler(
                                 when (it.type) {
                                     ExprValueType.NULL -> nullSeen = true
                                     ExprValueType.MISSING -> missingSeen = true
+                                    // Allow comparison with 1-pair structs to remain compatible SQL-92
+                                    ExprValueType.STRUCT -> {
+                                        if (it.ionValue.asIonStruct().size() != 1) return@forEach
+                                        if (it.iterator().next().exprEquals(leftValue))
+                                            return@thunkEnvOperands valueFactory.newBoolean(true)
+                                    }
```
lziq, line 827:
```suggestion
                                    ExprValueType.STRUCT -> if (it.ionValue.asIonStruct().size() == 1 && it.iterator().next().exprEquals(leftValue)) {
                                        return@thunkEnvOperands valueFactory.newBoolean(true)
                                    }
```

lziq:
And it seems there are some merge conflicts. I think it will be good to go after you resolve them.

lziq (approved):

johnedquinn:
I'm currently holding off on merging this PR for one reason. If you take a look at the linked issue, their failing query is:

```sql
SELECT * FROM users WHERE name IN (SELECT name FROM roles)
```

Their sub-query is returning a single-column table (bag) with the attribute `name`. They are able to use the main query's variable `name` (a literal -- likely string) to compare the single-column table with the IN clause.

The same query in PartiQL does NOT work. In PartiQL, the sub-query returns a bag of structs (tuples). Something like:
```partiql
<< { 'name': 'John' }, { 'name': 'Ralph' } >>
```

So, when the main query variable `name` attempts to compare to the sub-query result (a bag of structs, not a bag of strings), there is a type mismatch. While we can modify the IN clause to compare the left operand (in this case, a String) with the value of the key-value pair found in the right operand, it might do us harm.

Essentially, what if a user wanted to perform the following:

```sql
SELECT * FROM << { 'name': 'John', 'age': 22 } >> WHERE { 'name': name } IN (SELECT name FROM << { 'name': 'John' } >>)
```

This query explicity wants a comparison between a struct `{ 'name': name }` and a bag of structs `<< { 'name': 'John' } >>`. If we add a work-around, to recognize when the right operand is a bag of single-value structs, then we might be limiting the end-user when they DO in fact want to compare structs.

Need to work with the team to see what we want to do.

johnedquinn:
See another example:

```partiql
users = << { 'name': 'Ziqi Li', 'age': '22' }, { 'name': 'Ziqi Li', 'age': '23'} >>
info = << { 'name': 'Ziqi Li', 'age': '22', 'height': '5.10', 'addr': '1234 Terry Ave' } >>
```

```sql
SELECT * FROM info WHERE { 'name': name, 'age': age } IN (SELECT name, age FROM users)
```

am357:
> I'm currently holding off on merging this PR for one reason. If you take a look at the linked issue, their failing query is...

Glad you've brought it up. My reading from [spec.'s section 9.1](https://partiql.org/assets/PartiQL-Specification.pdf#subsection.9.1) is that we should not coerce the subquery expression if its the `rhs` of `IN`:

---
In each of the following cases a SFW subquery coerces into a scalar:
- if it appears as the rhs of a comparison operator (=, >, etc) where the lhs is not an array literal. And, vice
versa, if it appears as the lhs of a comparison operator where the rhs is not an array literal. (If it is the lhs of a
comparison operator where the lhs is an array literal, it coerces into array, per Section 9.2.)
- if it is an SFW subquery expression that (a) is not the collection expression of a FROM clause item and **(b) is not
the rhs of an IN. (If it is the rhs of an IN then it should not be coerced;** see note on semantics of IN, Section ??.)
---

Considering above, I also share your concern on loosing the ability of comparing structs. Let's have a further discussion w/ other maintainers.

johnedquinn:
Thanks for referencing the respective section in the spec ! @am357

Issue #255, opened 2020-07-07
Command history for the REPL


dlurton:
Add the functionality of [`rlwrap`](https://linux.die.net/man/1/rlwrap) to the REPL.

hglr24:
Please, this drives me insane :((  
Currently, I have to copy-paste queries into a text editor just to make small changes

dlurton:
If you're on windows, you can try using WSL and install `rlwrap`...  Right now that's my best suggestion for windows users.

Issue #112, opened 2019-08-02
Arrow keys dont work in the REPL


swaranga:
Using Arrow keys (up, down, left, right) in the REPL prints control characters instead of navigating in the current buffer.

therapon:
My current workaround on Unix based OSes is to use [`rlwrap`](https://linux.die.net/man/1/rlwrap). nice utility that wraps a command and adds readline features including arrows, search backwards etc. 

My Windows knowledge is not that great--I do not have workaround for Windows.

Issue #63, opened 2019-05-29
Implement a parser for REPL commands


raganhan:
Repl commands are being naively parsed and need an actual parser to have more control

PR #625, opened 2022-06-01
Replaces REPL with JLine based shell

RCHowell:
This commit replaces the existing REPL with a JLine shell. The new shell provides many out-of-the-box features, as well as reduces the amount of code and the complexity of the existing REPL. I validated this by successfully running all of the commands in the CLI.md doc as well as the tutorial PDF.

This shell gives us
- History
- Arrow key functionality
- Searchable history via ctrl+r
- Highlighted output

As well as _many_ shell widgets available via JLine.

This resolves
- https://github.com/partiql/partiql-lang-kotlin/issues/112
- https://github.com/partiql/partiql-lang-kotlin/issues/255

and sets up for
- https://github.com/partiql/partiql-lang-kotlin/issues/63


johnedquinn:
This is awesome. BTW, when your start the shell, do you get the same warning I do?
```text
WARNING: Unable to create a system terminal, creating a dumb terminal (enable debug logging for more information)
```
after running:
```shell
./gradlew :cli:run --console=plain -q
```

Other notes:
- After entering a statement and pressing enter, it doesn't output a `|`
- Pressing enter (without a statement) throws a ParserException. Old implementation used to just enter.

johnedquinn:
This is awesome. BTW, when your start the shell, do you get the same warning I do?
```text
WARNING: Unable to create a system terminal, creating a dumb terminal (enable debug logging for more information)
```
after running:
```shell
./gradlew :cli:run --console=plain -q
```

Other notes:
- After entering a statement and pressing enter, it doesn't output a `|`
- Pressing enter (without a statement) throws a ParserException. Old implementation used to just enter.

cli/src/org/partiql/shell/Shell.kt
```diff
@@ -0,0 +1,285 @@
...
+import java.nio.file.Paths
+import java.util.Locale
+import java.util.Properties
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicBoolean
+import javax.annotation.concurrent.GuardedBy
+
+private const val PROMPT_1 = "PartiQL> "
+private const val PROMPT_2 = "   | "
```
johnedquinn, line 54: `PROMPT_2` is never used.

johnedquinn, line 54: `PROMPT_2` is never used.

cli/src/org/partiql/shell/Shell.kt
```diff
@@ -0,0 +1,285 @@
...
+/**
+ * Initial work to replace the REPL with JLine3. I have attempted to keep this similar to Repl.kt, but have some
+ * opinions on ways to clean this up in later PRs.
+ */
+class Shell(
+    private val valueFactory: ExprValueFactory,
+    private val output: OutputStream,
+    private val parser: Parser,
+    private val compiler: CompilerPipeline,
+    private val initialGlobal: Bindings<ExprValue>,
```
johnedquinn, line 80: Since this is used to only initialize `globals`, it might not need to be a class member

johnedquinn, line 80: Since this is used to only initialize `globals`, it might not need to be a class member

cli/src/org/partiql/shell/Shell.kt
RCHowell: Let me update so PROMPT_2 is used as well as multiple newlines. I do not get that warning. Have you tried running the executable script rather than via Gradle? Does that change anything?

johnedquinn: Yeah, just checked via the executable, and it works perfectly. History works smoothly, and the prompt `>` works.

Running via Gradle gets rid of all the ability to use the history and the prompt.

johnedquinn: Yeah, just checked via the executable, and it works perfectly. History works smoothly, and the prompt `>` works.

Running via Gradle gets rid of all the ability to use the history and the prompt.

RCHowell: Yep. I left a comment about this in `docs/user/CLI.md`

RCHowell:
Updated for whitespace and secondary prompt


https://user-images.githubusercontent.com/5731503/171750751-4c96295f-92b1-4345-a455-8071eab7eadf.mov

docs/user/CLI.md
```diff
@@ -58,14 +58,20 @@ The CLI can be run in two manners, non-interactive and interactive (REPL).
...
 To start an interactive read, eval, print loop (REPL) execute:
 
-```shell
+```bash
+# With Gradle
 rlwrap ./gradlew :cli:run --console=plain
+
+# Via Executable
+./gradlew :cli:build
+unzip ./cli/build/distributions/partiql-cli-0.6.1-SNAPSHOT.zip .
```
johnedquinn, line 67: This unzip command doesn't work. Seems like it should be:
```shell
unzip ./cli/build/distributions/partiql-cli-0.6.1-SNAPSHOT.zip -d ./cli
```

docs/user/CLI.md
```diff
@@ -58,14 +58,20 @@ The CLI can be run in two manners, non-interactive and interactive (REPL).
 
 To start an interactive read, eval, print loop (REPL) execute:
 
-```shell
+```bash
+# With Gradle
 rlwrap ./gradlew :cli:run --console=plain
```
johnedquinn, line 63: Might want to remove this. See my general comment.

johnedquinn:
Unsure what changed between last review and this one, but using rlwrap seems to be broken. When I enter a query, it prints the query again and expects another enter. Unsure why it's happening, but it seems like we might just want to start using jline instead of rlwrap. Consider removing rlwrap from the documentation, and asking the user to either run:

```shell
./gradlew :cli:run --console=plain -q
```
 or the JLine command:
```
# Already in the docs.
```
JLine seems to work perfectly for interactive queries. For non-interactive, small bug where there is a trailing `%`. To reproduce:
```
./cli/partiql-cli-0.6.1-SNAPSHOT/bin/partiql -q "TIME WITH TIME ZONE '23:15:15.1277+05:30'"
```

output: `23:15:15.1277+05:30%`

If we do want to move to only using JLine, it could be nice to create a gradle task within `cli` to automatically unzip and place the executable at either project or subproject root. Perhaps a future PR, but something to consider.

RCHowell:
I've created another issue for better packaging of the REPL. We could do a ShadowJar or native image via Graal. I'm surprised there is an issue with the trailing percent sign because that should be unrelated to Jline changes. If you specify the `-q` flag, no shell code is executed

RCHowell:
I'll debug and include the "no-trailing-%" in this

RCHowell:
The trailing percent is unrelated to the shell code, and it shows with the current build of main branch. I'll send an updated PR with the doc changes

johnedquinn (approved):
LGTM. This is a great shell experience.

Issue #560, opened 2022-03-30
Incorrect Definition of Equality for Some Unit Test Assertions


dlurton:
There are a couple of places in `EvaluatorTestBase` ([here](https://github.com/partiql/partiql-lang-kotlin/blob/18c06a89b739bcb497795fe66cf6739c6efc270b/lang/test/org/partiql/lang/eval/EvaluatorTestBase.kt#L182) and [here](https://github.com/partiql/partiql-lang-kotlin/blob/18c06a89b739bcb497795fe66cf6739c6efc270b/lang/test/org/partiql/lang/eval/EvaluatorTestBase.kt#L585)) that utilize ParitQL's own definition of equality for comparing expected vs actual values for test results.  The problem there is that PartiQL's equality can automatically coerce dissimilar types, (i.e. ints and decimals, strings and symbols, among others).   This means that the expected result of a test might have the incorrect type, but be of an equivalent value to the actual result.  

Changing the aforementioned places to use Ion's equivalence (i.e. converting both actual and expected to `IonValue` before comparing) which is far more strict, causes ~650 unit tests to fail.  A cursory review of the failed tests shows that mostly this is due to PartiQL's coercion that no longer works here, but there could be some subtle bugs lurking here, i.e. expressions that return the wrong type of value.  

Full disclosure: I wrote the lines of code above.  :(

dlurton:
Actually, this isn't limited to those two locations--there is at least one other test class using the incorrect equality, there might be more.  Need to look at all uses of [`DEFAULT_COMPARATOR` and `.exprEquals`](https://github.com/partiql/partiql-lang-kotlin/blob/3febaeec46a0b703781494699c445f2034e9c6c4/lang/src/org/partiql/lang/eval/ExprValueExtensions.kt#L167-L170) in unit tests.  The only tests that should be using these are those that specifically test those functions.  All other tests should be using Ion's equality instead.

PR #630, opened 2022-06-04
Modifies DATE/TIME tests to assert by Ion and modifies TIME output

### LINKED ISSUE #567: CAST tests for DATE values assert expected results using string equality instead of Ion equality
Note the call to `.toString`, at the end of the line here:

https://github.com/partiql/partiql-lang-kotlin/blob/fe306c6cf47c61e0fd31693241c2c987651a7753/lang/test/org/partiql/lang/eval/CastTestBase.kt#L217

This is a sub-optimal method of performing assertions on expected results--actually, the result should be specified in Ion, and the assertion should be performed on the resulting Ion values.

Something like:

```
assertEquals(ion.singleValue(castCase.expected), eval(castCase.expression, compileOptions = CompileOptions.build(compileOptionBlock)).ionValue)
```

For this to work, there's a couple hundred test cases here that need their expected values converted to Ion.

johnedquinn:
## Relevant Issue
- Closes #567 

## Description
- Modifies the tests to not compare by string but by ExprValue
- Modifies the output string of TIME
  - In order to compare by ExprValue, we need to make sure that we retain offset via TIME WITH TIME ZONE


lang/test/org/partiql/lang/eval/EvaluatingCompilerCastTest.kt
```diff
@@ -35,6 +35,6 @@ class EvaluatingCompilerCastTest : CastTestBase() {
     @ArgumentsSource(DateTimeConfiguredCastArguments::class)
     fun dateTimeConfiguredCast(dateTimeConfiguredCastCase: ConfiguredCastCase) =
         dateTimeConfiguredCastCase.assertCase(
-            ExpectedResultFormat.STRING
+            ExpectedResultFormat.PARTIQL
```
dlurton, line 38: While this is an improvement over string equality and is perfectly acceptable for `TIME` types, I might have considered `ExpectedResultFormat.ION` instead. See https://github.com/partiql/partiql-lang-kotlin/issues/560 for an explanation as to why none of the expected value formats in use today are truly ideal.  However, `ExpectedResultFormat.ION` is usually the most ideal one available today.

lang/test/org/partiql/lang/eval/CastTestBase.kt
```diff
@@ -1359,6 +1344,21 @@ abstract class CastTestBase : EvaluatorTestBase() {
...
+                case("TIME '23:12:12.1267+05:30'", "\"23:12:12.1267\"", CastQuality.LOSSLESS),
+                case("TIME (3) '23:12:12.1267'", "\"23:12:12.127\"", CastQuality.LOSSLESS),
+                case("TIME (3) '23:12:12.1267-05:30'", "\"23:12:12.127\"", CastQuality.LOSSLESS),
+                case("TIME (3) '23:12:12.1267+05:30'", "\"23:12:12.127\"", CastQuality.LOSSLESS),
+                case("TIME WITH TIME ZONE '23:12:12.1267'", "\"23:12:12.1267${defaultTimezoneOffset.getOffsetHHmm()}\"", CastQuality.LOSSLESS),
+                case("TIME WITH TIME ZONE '23:12:12.1267-05:30'", "\"23:12:12.1267-05:30\"", CastQuality.LOSSLESS),
+                case("TIME WITH TIME ZONE '23:12:12.1267+05:30'", "\"23:12:12.1267+05:30\"", CastQuality.LOSSLESS),
+                case("TIME (3) WITH TIME ZONE '23:12:12.1267'", "\"23:12:12.127${defaultTimezoneOffset.getOffsetHHmm()}\"", CastQuality.LOSSLESS),
+                case("TIME (3) WITH TIME ZONE '23:12:12.1267-05:30'", "\"23:12:12.127-05:30\"", CastQuality.LOSSLESS),
+                case("TIME (3) WITH TIME ZONE '23:12:12.1267+05:30'", "\"23:12:12.127+05:30\"", CastQuality.LOSSLESS)
```
lziq, line 1361: Why do we want to make this change? L1334-L1346 are not testing symbols as commented in L1333.

johnedquinn: It is testing symbols. See the Symbol section of [the Ion Spec](https://amzn.github.io/ion-docs/docs/spec.html#string). TLDR: Symbols can be specified with backticks.

lziq: My bad. I thought it was in PartiQL format.

lziq (approved):

PR #632, opened 2022-06-07
[DRAFT/EXPERIMENT] Add ability to specify custom operator "mini-compilers"

dlurton:
## Superceded by #636

Publishing this to get it before PartiQL devs eyes and collect initial thoughts on the approach.

Still TODO before this can be merged:

- [ ] Sign off of PartiQL developers on basic approach.
- [ ] Add additional tests with some custom operators (currently all tests use the default operators which prove this works but still should try a custom operator implementation in a test or three)
- [ ] Add `PlannerPipeline` function to add additional operator implementations.

----

### DIFF
--- a/lang/src/org/partiql/lang/eval/physical/operators/DefaultOperatorCompilers.kt
+++ b/lang/src/org/partiql/lang/eval/physical/operators/DefaultOperatorCompilers.kt
@@ -1,0 +1,298 @@
+package org.partiql.lang.eval.physical.operators
+
+import coercePredicateResult
+import com.amazon.ionelement.api.BoolElement
+import com.amazon.ionelement.api.MetaContainer
+import createFilterRelItr
+import org.partiql.lang.domains.PartiqlPhysical
+import org.partiql.lang.eval.ExprValue
+import org.partiql.lang.eval.ExprValueFactory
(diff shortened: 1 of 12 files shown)


lang/src/org/partiql/lang/eval/physical/operators/DefaultOperatorCompilers.kt
```diff
@@ -0,0 +1,298 @@
...
+        createPhysicalOperatorCompiler<PartiqlPhysical.Bexpr.Filter>(DEFAULT_IMPL_NAME) { node, compiler, _ ->
+            val predicateThunk = compiler.compile(node.predicate)
+            val sourceThunk = compiler.compile(node.source)
+
+            relationThunk(node.metas) { env ->
+                val sourceToFilter = sourceThunk(env)
+                createFilterRelItr(sourceToFilter, predicateThunk, env)
+            }
+        },
+        createPhysicalOperatorCompiler<PartiqlPhysical.Bexpr.Join>(DEFAULT_IMPL_NAME) { node, compiler, valueFactory ->
```
dlurton, line 80: Consider adding a `valueFactory` to the `EvaluatorState` class so it doesn't need to be passed separately here.

Issue #610, opened 2022-05-24
Document ways in which the PartiQL specification differs from the Kotlin reference implementation


alancai98:
It will be important to document how the [PartiQL specification](https://partiql.org/assets/PartiQL-Specification.pdf) differs from the Kotlin reference implementation. This can include differences in supported features, syntax, and semantics.

PR #634, opened 2022-06-08
Add doc with initial comparison of Kotlin implementation with PartiQL spec

alancai98:
Starting point for #610. 

Adds a document comparing the Kotlin implementation with the [PartiQL spec](https://partiql.org/assets/PartiQL-Specification.pdf). Currently lists out:
- Features defined in spec and whether it's implemented in `partiql-lang-kotlin`
- Features implemented in `partiql-lang-kotlin` and whether it's formally defined in the PartiQL spec

In the future
- Can consider adding a similar doc for the Rust implementation. The first table can be applied to other implementations.
- Integrate the conformance test suite (i.e. `partiql-tests`) with the document to make the comparison more automated

Note: this is just the starting point. About half of the spec has yet to be compared.


alancai98:
[Rendered markdown](https://github.com/partiql/partiql-lang-kotlin/blob/0e0aef61c1b8b8b099944d42404314a7925fa045/docs/user/kotlin-impl-spec-comparison.md)

docs/user/kotlin-impl-spec-comparison.md
```diff
@@ -0,0 +1,62 @@
...
+| 4	                         |Tuple navigation with array notation	|Yes	| `a[s]`; shorthand for `a.s` where `a` is a tuple, `s` is either a string literal or expression CAST to a string	                                                                                                   |
+| 4	                         |Composition of naviagations	|Yes	| `r.no[1]`	                                                                                                                                                                                                         |
+| 4	                         |Tuple naviagation in tuples w/ duplicate attributes	|Yes	| 	                                                                                                                                                                                                                  |
+| 4.1	                       |Tuple path evaluation on wrongly typed data	|Partially^1	| Permissive mode behavior returns `MISSING` as defined in the spec. "STRICT" type-checking mode doesn't exist in Kotlin implementation	                                                                             |
```
am357, line 21: I understand we have a reference to "strict mode" in the Spec. but we also have references to "Type checking mode" which has been put against "Permissive mode". Which one are we denoting here by "STRICT"?

docs/user/kotlin-impl-spec-comparison.md
```diff
@@ -0,0 +1,62 @@
...
+PartiQL spec.
+
+Note: the list is only partially completed. Sections 6.2 onwards from the spec have yet to be compared with the Kotlin
+implementation.
+
+## Features defined in spec
+
+| Spec section	              |Feature	|Implemented in `partiql-lang-kotlin`	| Notes	                                                                                                                                                                                                             |
+|----------------------------|---	|---	|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| 3.3	                       |Syntax improvement `SELECT` optionally written as last clause	|No	| 	                                                                                                                                                                                                                  |
```
am357, line 15: Improvement and not a blocker: we could have direct links to the Spec. like:

```suggestion
| [3.3](https://partiql.org/assets/PartiQL-Specification.pdf#subsection.3.3)	                       |Syntax improvement `SELECT` optionally written as last clause	|No	| 	                                                                                                                                                                                                                  |
```

docs/user/kotlin-impl-spec-comparison.md
alancai98: I've been under the impression that "type checking mode" was the `STRICT` mode in the spec, so `STRICT` in this document refers to "type checking mode".

docs/user/kotlin-impl-spec-comparison.md
alancai98: Will do. TIL you could link to specific sections of a PDF

docs/user/kotlin-impl-spec-comparison.md
alancai98: Yeah, I think using "type checking mode" rather than `STRICT` is more clear and representative of the spec. I'll change in the next commit.

am357 (approved):

Issue #631, opened 2022-06-07
How to interact with AWS resources with REPL


quiver:
Some AWS services(DynamoDB, etc.) support PartiQL natively.
Now how can I interact with these services from this REPL?

[Tutorial](https://partiql.org/tutorial.html) does not mention how to specify external data sources at all.

Is this repository intended for general purpose Particle implementation and do I have to write an AWS integration wrapper on top of this?

Best regards.

am357:
@quiver currently PartiQL does not expose any API to enable evaluation on external sources with data at rest. Considering this, a common approach as of today is to use our current APIs to retrieve an AST from the parsed query and evaluate further by walking through and transforming the AST through rewrites. We have some [examples](https://github.com/partiql/partiql-lang-kotlin/tree/main/examples/src/kotlin/org/partiql/examples) that can help clarifying this further. For example, see the following:
- [ParserExample.kt](https://github.com/partiql/partiql-lang-kotlin/blob/main/examples/src/kotlin/org/partiql/examples/ParserExample.kt)

Please let us know if you need further assistance and we can take it from there.

am357:
No worries @quiver. Thanks for raising this because this issue has struck @johnedquinn by an idea to provide a helper CLI helper function for DynamoDB queries; he'll update this issue with more details.

johnedquinn:
Hi @quiver, I really liked your idea about interacting with AWS resources from the CLI. It would be greatly beneficial to be able to send PartiQL queries to services and perhaps perform additional local computation on the result.

We are thinking of creating an `extensions` library that will hold functions used by the CLI (among other things).  If you'd like to do something similar with an external service for your own work, it's as simple as extending the `BaseFunction` abstract class. `QueryDDB` is such an example (in the PR). Hope this helps.

Here's my draft PR: #635

quiver:
@johnedquinn 

> to provide a helper CLI helper function for DynamoDB queries

Actually, I was using [stevearc/dql](https://github.com/stevearc/dql) to access DynamoDB with SQL. It should be a lot easier with PartiQL, but no one seemed to develop apps that way.

I'll try your feature this weekend.

Cheers

quiver:
Hi, @johnedquinn

This is exactly what I wanted.

I wrote about it on my company's tech blog(in Japanese).

https://dev.classmethod.jp/articles/query-dynamodb-from-partiql-repl/

Thank you for your work.

johnedquinn:
[...]

PR #635, opened 2022-06-09
Initializes extensions module and adds the query_ddb function to the CLI

johnedquinn:
## Relevant Issue
Related to #631

## Description
- Initializes the `extensions` module
- Sets the `extensions` as a dependency to `cli`
- Moves the `BaseFunction` class to the `extensions` module
- Creates a new `query_ddb` function, along with test cases
- Links the function to the CLI
- Updates documentation

### DIFF
--- a/cli/src/org/partiql/cli/functions/QueryDDB.kt
+++ b/cli/src/org/partiql/cli/functions/QueryDDB.kt
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ *  You may not use this file except in compliance with the License.
+ * A copy of the License is located at:
+ *
+ *      http://aws.amazon.com/apache2.0/
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+ *  language governing permissions and limitations under the License.
+ */
+
+package org.partiql.cli.functions
+
+import com.amazon.ion.system.IonReaderBuilder
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
+import org.partiql.lang.eval.EvaluationSession
+import org.partiql.lang.eval.ExprValue
+import org.partiql.lang.eval.ExprValueFactory
+import org.partiql.lang.eval.stringValue
+import org.partiql.lang.types.FunctionSignature
+import org.partiql.lang.types.StaticType
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClientBuilder
+import com.amazonaws.services.dynamodbv2.document.ItemUtils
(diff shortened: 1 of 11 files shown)


docs/user/CLI.md
```diff
@@ -661,6 +661,61 @@ All the available options for customized CSV files are as follows:
 6. Set quote sign (single character only): `'quote': '"'`
 7. Set delimiter sign (single character only): `'delimiter': ','`
 
+## Querying DDB
```
am357, line 664: `DDB` -> `Amazon DynamoDB`?

extensions/src/org/partiql/extensions/cli/functions/QueryDDB.kt
```diff
@@ -0,0 +1,81 @@
...
+import org.partiql.lang.eval.ExprValue
+import org.partiql.lang.eval.ExprValueFactory
+import org.partiql.lang.eval.stringValue
+import org.partiql.lang.types.FunctionSignature
+import org.partiql.lang.types.StaticType
+
+/**
+ * [QueryDDB] is a function to perform a single query on DDB using PartiQL.
+ * It uses the default configuration provider to establish a connection with AWS. Please reference the official
+ * AWS documentation for specifying which account/profile to use via credentials overrides. Reference the CLI.md
```
am357, line 31: Perhaps it's worthwhile to move this comment to the default `constructor` as we might end up extending the class to initialize with other cred. providers as well. If going w/ this change a slight change to `README.md` to reflect this would be great as well.

docs/user/CLI.md
johnedquinn: ACK.

extensions/src/org/partiql/extensions/cli/functions/QueryDDB.kt
am357: Looking at this again, I think we can just remove this from here b/c the creds. are supplied via the client.

am357 (approved):
ShipIt!

cli/src/org/partiql/cli/functions/QueryDDB.kt
```diff
@@ -0,0 +1,67 @@
...
+ * It uses the default configuration provider to establish a connection with AWS. Please reference the official
+ * AWS documentation for specifying which account/profile to use via credentials overrides. Reference the CLI.md
+ * file within this repository for more information.
+ */
+internal class QueryDDB(valueFactory: ExprValueFactory) : BaseFunction(valueFactory) {
+
+    private val client: AmazonDynamoDB = AmazonDynamoDBClientBuilder.defaultClient()
+
+    override val signature = FunctionSignature(
+        name = "query_ddb",
```
RCHowell: At some point, we will want to namespace things. Also, right now the functions must be manually added to the compiler pipeline. We should consider java/jvm service loader rather than modifying CLI directly. This would enable customers to provide their own functions

RCHowell: That being said, this is a good start. I want to overhaul function resolution after my bag op work

johnedquinn: Agreed. Service loader will be the best path forward.

PR #636, opened 2022-06-09
Support non-default physical operator implementations

dlurton:
This PR makes the new plan compiler support custom physical operator implementations.  It does this by adding factories for each type of physical relational operator.  This PR also uses these factories for the built-in default operator implementations.

This PR supercedes #632 which would have been very powerful and flexible, but also more difficult for customers to use while also simultaneously being easier to abuse.  It also would have meant more boilerplate code that customers would have needed to implement on their own--i.e. tree traversal, and it would have exposed implementation details I think we'd rather not have publicly exposed. (i.e. thunks and `EvaluatorState.registers`).  This alternative approach solves all of these concerns at the expense of being more complex to implement up front (in this PR), but will have cost savings for PartiQL and its customers in terms of reduced support cost.


lang/src/org/partiql/lang/eval/physical/SetValueFunc.kt
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Sets the value of a variable which is stored in an [EvaluatorState].
+ *
+ * To obtain an instance of [SetVariableFunc] that sets a specific variable, see [toSetVariableFunc], which is
+ * intentionlly `internal` to avoid exposing it to customers.  This is because [EvaluatorState.registers] contains the
+ * current value of all variables in an SFW query.  As such, allowing direct access to the registers is very risky--for
+ * example, if the wrong register is set the query will have the incorrect result or trigger an evaluation-time
+ * exception.
+ *
+ * To mitigate this risk, [EvaluatorState.registers] is marked to `internal` to avoid exposing it publicly and
```
dlurton, line 16: Some redudant verbiage here.

lang/src/org/partiql/lang/eval/physical/SetValueFunc.kt
```diff
@@ -0,0 +1,34 @@
...
+typealias SetVariableFunc = (EvaluatorState, ExprValue) -> Unit
+
+/**
+ * Creates a [SetVariableFunc] tied to the variable identified in the [PartiqlPhysical.VarDecl] receiver object that
+ * can be used to set the value of the variable at evaluation time.  This method is considerably safer than direct
+ * access to the [EvaluatorState.registers] array, which is marked as `internal` to reduce the likelihood of being
+ * clobbered by a physical operator implementation which was not supplied by this library.
+ */
+internal fun PartiqlPhysical.VarDecl.toSetVariableFunc(): SetVariableFunc =
+    { state, value -> state.registers[this.index.value.toIntExact()] = value }
```
dlurton, line 34: Can/should call `this.index.value.toIntExact()` outside the closure.

lang/src/org/partiql/lang/eval/physical/operators/DefaultOperatorFactories.kt
```diff
@@ -0,0 +1,169 @@
...
+import org.partiql.lang.eval.name
+import org.partiql.lang.eval.physical.EvaluatorState
+import org.partiql.lang.eval.physical.SetVariableFunc
+import org.partiql.lang.eval.physical.evalLimitRowCount
+import org.partiql.lang.eval.physical.evalOffsetRowCount
+import org.partiql.lang.eval.relation.RelationType
+import org.partiql.lang.eval.relation.relation
+import org.partiql.lang.eval.unnamedValue
+import org.partiql.lang.planner.transforms.DEFAULT_IMPL_NAME
+
```
dlurton, line 16: Could probably add some kdoc to this.

lang/src/org/partiql/lang/eval/physical/operators/DefaultOperatorFactories.kt
```diff
@@ -0,0 +1,169 @@
...
+                    ExprValueType.LIST, ExprValueType.BAG -> valueToScan.asSequence()
+                    else -> sequenceOf(valueToScan)
+                }
+
+                relation(RelationType.BAG) {
+                    val rowsIter: Iterator<ExprValue> = rows.iterator()
+                    while (rowsIter.hasNext()) {
+                        val item = rowsIter.next()
+
+                        // Note that although we could access [EvaluatorState.registers] directly within relationThunk,
```
dlurton, line 40: This isn't strictly true anymore.

lang/src/org/partiql/lang/eval/physical/operators/JoinPhysicalOperatorFactory.kt
```diff
@@ -0,0 +1,30 @@
...
+        /** Invokes the left-side bindings expression of the join.*/
+        leftBexpr: BindingsExpr,
+        /** Invokes the right-side bindings expression of the join.*/
+        rightBexpr: BindingsExpr,
+        /** Invokes the join's predicate expression. */
+        predicateExpr: ValueExpr?,
+        /** Sets the left-side variables to `null`. Needed for right joins when there is rows matching the predicate. */
+        setLeftSideVariablesToNull: (EvaluatorState) -> Unit,
+        /** Sets the right-side variables to `null`. Needed for left joins when there is rows matching the predicate. */
+        setRightSideVariablesToNull: (EvaluatorState) -> Unit
```
dlurton, line 28: > joins when there is rows
->
joins when there *are no* rows

lang/src/org/partiql/lang/eval/physical/operators/JoinRelationIterators.kt
```diff
@@ -0,0 +1,72 @@
+package org.partiql.lang.eval.physical.operators
+
+import org.partiql.lang.eval.physical.EvaluatorState
+import org.partiql.lang.eval.relation.RelationIterator
+import org.partiql.lang.eval.relation.RelationType
+import org.partiql.lang.eval.relation.relation
+
```
dlurton, line 7: Could add some kdoc

lang/src/org/partiql/lang/eval/physical/operators/PhysicalOperatorKind.kt
```diff
@@ -0,0 +1,19 @@
+package org.partiql.lang.eval.physical.operators
+
+/**
+ * Indicates which physical relational operator a [PhysicalOperatorFactory] can create instances of.
+ *
+ * This is part of [PhysicalOperatorFactoryKey] and also implies which subclass of [PhysicalOperatorFactory] the
+ * implementation must derive from; e.g. An [PhysicalOperatorFactory]s with [PROJECT] in its key must implement
```
dlurton, line 7: `A` not `An`.

lang/src/org/partiql/lang/eval/physical/operators/ValueExpr.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.partiql.lang.eval.physical.EvaluatorState
+
+/**
+ * Evaluates a PartiQL expression returning an [ExprValue].
+ *
+ * [BindingsExpr] implementations need a mechanism to evaluate such expressions, and said mechanism should
+ * avoid exposing implementation details (i.e. [org.partiql.lang.eval.physical.PhysicalPlanThunk]) of the evaluator.
+ * This implementation accomplishes that and is intended as a publicly usable API that is supported long term.
+ */
+interface ValueExpr {
```
dlurton, line 14: Any suggestions from the other reviewers for the name of this interface?  I just relased that `ValueExpr` is `ExprValue` reversed, which might be slightly confusing.

lang/src/org/partiql/lang/eval/physical/SetValueFunc.kt
```diff
@@ -0,0 +1,34 @@
+package org.partiql.lang.eval.physical
```
dlurton, line 1: Wrong filename

lang/src/org/partiql/lang/eval/physical/SetValueFunc.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import org.partiql.lang.domains.PartiqlPhysical
+import org.partiql.lang.eval.ExprValue
+import org.partiql.lang.util.toIntExact
+
+/**
+ * Sets the value of a variable which is stored in an [EvaluatorState].
+ *
+ * To obtain an instance of [SetVariableFunc] that sets a specific variable, see [toSetVariableFunc], which is
+ * intentionlly `internal` to avoid exposing it to customers.  This is because [EvaluatorState.registers] contains the
```
dlurton, line 11: *intentionally

hpolloni (approved):
LGTM

lang/src/org/partiql/lang/eval/relation/RelationIterator.kt
```diff
@@ -24,7 +24,7 @@ enum class RelationType { BAG, LIST }
  * including filters and joins that requires advancing through possibly all remaining rows to see if any remaining row
  * matches the predicate.  This is awkward to implement and would force eager evaluation of the [Iterator].
  */
-internal interface RelationIterator {
+interface RelationIterator {
```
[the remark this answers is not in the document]
lziq: Same for `relation` method and `RelationThunk` class.

lziq:
Generally looks good to me. Left a few comments above. Also, do we really want to call `PhyscialExprToThunkConverter` a converter instead of a compiler?

lang/src/org/partiql/lang/eval/physical/operators/PhysicalOperatorFactory.kt
```diff
@@ -0,0 +1,9 @@
+package org.partiql.lang.eval.physical.operators
+
+/**
+ * Marker interface with unique [key], allowing all [PhysicalOperatorFactory] implementations to exist in a
+ * `Map<OperatorFactoryKey, OperatorFactory>`.
+ */
+interface PhysicalOperatorFactory {
```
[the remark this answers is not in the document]
dlurton: > Can we rename this class to PhysicalOpeator and rename create method to eval? 

But it doesn't evaluate anything, hence `eval` would *not* be the right name.  It creates a `BindingsExpr` that *can* do the evaluation.  Hence the terms `create` and `Factory`.  I'm open to changing these terms, but not to `eval`.

lang/src/org/partiql/lang/eval/physical/operators/PhysicalOperatorKind.kt
```diff
@@ -0,0 +1,19 @@
...
+ * [ProjectPhysicalOperatorFactory], with [FILTER] the factory must implement [FilterPhysicalOperatorFactory], and so
+ * on.
+ */
+enum class PhysicalOperatorKind {
+    PROJECT,
```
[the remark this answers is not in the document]
dlurton: Unfortunately the term "project" is used in multiple ways, which can be confusing.  We do sometimes say "select projection", referring to the expressions between the `select` and `from` keyword, but that is actually not very accurate. 

In this case, the `PROJECT` operator is (perhaps questionably named) relational operator that simply means "project some values from some data source".  It's up to the partiql hosting application to define what that data source actually is, and to utilize it in a query plan.

dlurton:
> do we really want to call `PhyscialExprToThunkConverter` a converter instead of a compiler? Probably something like `PhysicalPlanCompiler`

I would not be opposed to this, but I would consider it outside the scope of this PR.

lang/src/org/partiql/lang/eval/physical/operators/PhysicalOperatorFactory.kt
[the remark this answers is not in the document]
lziq: What about `PhysicalOperator` with `invoke`?

lang/src/org/partiql/lang/eval/physical/operators/PhysicalOperatorKind.kt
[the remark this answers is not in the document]
lziq: Agree

lziq:
> > do we really want to call `PhyscialExprToThunkConverter` a converter instead of a compiler? Probably something like `PhysicalPlanCompiler`
> 
> I would not be opposed to this, but I would consider it outside the scope of this PR.

An issue created: https://github.com/partiql/partiql-lang-kotlin/issues/638

lang/src/org/partiql/lang/eval/physical/operators/BindingsExpr.kt
```diff
@@ -0,0 +1,14 @@
+package org.partiql.lang.eval.physical.operators
+
+import org.partiql.lang.eval.physical.EvaluatorState
+import org.partiql.lang.eval.relation.RelationIterator
+
+/**
+ * An implementation of a relational operator (a.k.a bindings expression).
```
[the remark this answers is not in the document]
dlurton: I don't understand this question.  Are you suggesting it has the wrong name?

lang/src/org/partiql/lang/eval/physical/operators/PhysicalOperatorFactory.kt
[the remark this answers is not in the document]
dlurton: `PhysicalOperatorFactory` doesn't actually evaluate anything.  What it does is create the thing that actually does the evaluation.  Therefore, it would not be appropriate to rename it to `PhysicalOperator`, nor to rename `create` to `eval` or `invoke`.    I have some new names for these items in the next commit I'm about to push--hopefully this will become clearer then.

lang/src/org/partiql/lang/eval/physical/operators/ValueExpr.kt
dlurton: `ValueExpr` will be renamed to `ValueExpression`

lang/src/org/partiql/lang/eval/relation/RelationIterator.kt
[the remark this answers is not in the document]
dlurton: `RelationIterator` I believe is the best name--even if we define the term "relation" a little differently from Codd's relational algebra.  I've added some kdoc to explain this.

lang/src/org/partiql/lang/eval/physical/operators/BindingsExpr.kt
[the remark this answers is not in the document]
lziq: Sorry for not being clear. I was trying to say this line of comment does not make sense to me. Relation should be something different from relational operator, however, this comment is saying that a relation is an implementation of a relational operator.

lang/src/org/partiql/lang/eval/physical/operators/ValueExpr.kt
lziq: I think you are right

lziq (approved):
Thanks for addressing comments. Also, some of my concerns were resolved in offline discussion.

PR #639, opened 2022-06-16
Adds syntax highlighting to the CLI

johnedquinn:
## Relevant Issue
TBD

## Description
- Adds syntax highlighting file (PartiQL.nanorc)
- Adds the ability to use your own nanorc file in your `~/.nano` directory
- Adds the ability to use `--monochrome` or `-m` on the REPL to avoid using syntax highlighting


cli/src/org/partiql/cli/main.kt
```diff
@@ -109,6 +109,8 @@ private val inputFormatOpt = optParser.acceptsAll(listOf("input-format", "if"),
 private val wrapIonOpt = optParser.acceptsAll(listOf("wrap-ion", "w"), "wraps Ion input file values in a bag, requires the input format to be ION, requires the query option")
     .availableIf(queryOpt)
 
+private val monochromeOpt = optParser.acceptsAll(listOf("monochrome", "m"), "removes syntax highlighting for the REPL")
+
```
RCHowell, line 113: What is the motivation behind this? I'm wondering if there's even a need. Maybe if someone has a terminal color scheme with poor contrast? I'm not sure if we need it, but that's just my opinion

cli/resources/org/partiql/cli/syntax/PartiQL.nanorc
```diff
@@ -0,0 +1,34 @@
+syntax "PartiQL" "\.pql$"
+
+# Numbers
+color magenta "([0-9]+(\.)?[0-9]*|[0-9]*(\.)?[0-9]+)([eE][+-]?[0-9]+)*"
+
+# PartiQL Data Types
+icolor green "\<(smallint|integer2|int2|integer4|int4|integer8|int8|char|character|time)\>"
+icolor green "\<(bool|boolean|string|symbol|clob|blob|struct|list|sexp|bag)\>"
+icolor green "\<(varchar|date|smallint|int|integer|bigint|float|double|decimal)\>"
```
RCHowell, line 9: nit, `smallint` is here twice, all else looks good

cli/src/org/partiql/shell/Shell.kt
```diff
@@ -79,6 +79,7 @@ class Shell(
     private val parser: Parser,
     private val compiler: CompilerPipeline,
     private val initialGlobal: Bindings<ExprValue>,
+    private val isMonochrome: Boolean = false
```
RCHowell, line 82: Probably fine for now, I am keen on parsing all options to a configuration object and passing that rather than individual options

cli/src/org/partiql/shell/Shell.kt
```diff
@@ -122,11 +123,17 @@ class Shell(
     }
 
     private fun run(exiting: AtomicBoolean) = TerminalBuilder.builder().build().use { terminal ->
-
+        val userSyntaxFile = homeDir.resolve(".nano/PartiQL.nanorc")
```
RCHowell, line 126: We need to document this path if a customer wishes to use their own .nanorc file

cli/src/org/partiql/cli/main.kt
johnedquinn: Yeah, just if a user experiences poor contrast with their theme.

cli/src/org/partiql/shell/Shell.kt
```diff
@@ -261,6 +269,8 @@ class Shell(
             }
         }
     }
+
+    class ShellConfiguration(val isMonochrome: Boolean = false)
```
alancai98, line 273: Could also document what this `ShellConfiguration` class is for and document why certain attributes are included here rather than in the `Shell` constructor.

cli/src/org/partiql/shell/ShellHighlighter.kt
```diff
@@ -20,17 +20,17 @@ import org.jline.reader.LineReader
...
+import java.net.URI
 import java.util.regex.Pattern
 
 private val SUCCESS: AttributedStyle = AttributedStyle.DEFAULT.foreground(AttributedStyle.GREEN)
 private val ERROR: AttributedStyle = AttributedStyle.DEFAULT.foreground(AttributedStyle.RED)
 private val INFO: AttributedStyle = AttributedStyle.DEFAULT
 private val WARN: AttributedStyle = AttributedStyle.DEFAULT.foreground(AttributedStyle.YELLOW)
 
-class ShellHighlighter(private val syntaxFile: Path) : Highlighter {
+class ShellHighlighter(syntaxUri: URI) : Highlighter {
```
alancai98, line 31: Somewhat related to this PR's changes but does the `ShellHighlighter` need to be public?

cli/resources/org/partiql/cli/syntax/PartiQL.nanorc
```diff
@@ -0,0 +1,34 @@
+syntax "PartiQL" "\.pql$"
```
alancai98, line 1: I'm not too familiar w/ JLine's syntax highlighter requirements but is it restricted to just `nanorc` files? I know @jpschorr had worked on some syntax highlighters for Ace, Python Pygments, Ruby Rouge, and Sublime Text (example integration in this recent Rust PR: https://github.com/partiql/partiql-lang-rust/pull/136). It would be nice to use one of those syntax highlighters or to use his generator to create a `nanorc` syntax highlighter to keep a consistent + complete syntax highlighting experience.

alancai98: An added benefit would be Ion syntax highlighting within PartiQL queries.

cli/resources/org/partiql/cli/syntax/PartiQL.nanorc
```diff
@@ -0,0 +1,34 @@
...
+
+# Numbers
+color magenta "([0-9]+(\.)?[0-9]*|[0-9]*(\.)?[0-9]+)([eE][+-]?[0-9]+)*"
```
alancai98, line 4: I don't think this handles decimals specified with the `d`/`D`

image

alancai98 (commented):
Some really nice work here! Syntax highlighting would definitely be a great feature to add for the PartiQL REPL experience.

Left some questions about the syntax highlighting support from JLine (and if we're restricted to just `nanorc` files) and additional documentation.

cli/src/org/partiql/shell/Shell.kt
RCHowell: The general thinking is, if the option/configuration comes in as user input, then it belongs in the ShellConfiguration. The ShellConfiguration could optionally be dumped into `$HOME/.partiql/shell.ion` or something of the likes. Then we deserialize that config to the object.

This allows use to have re-usable configurations as the shell is expanded

cli/resources/org/partiql/cli/syntax/PartiQL.nanorc
RCHowell: ^^ agreed. We can actually implement the highlighter however we want. Nanorc was chosen because this is built into JLine3, but you can actually implement your own highlighting. What I've typically done (which is what Trino does) is run the input through the lexer AND parser. The lexer will get you token types, and then you can even use the parser to highlight everything red if there's a parse error. It requires having a catch-all "unrecognized" rule

cli/src/org/partiql/shell/Shell.kt
alancai98: So should some of the other REPL/Shell options be included in this configuration? The `-e`/`environment` option specifying the initial global environment would seem to belong in the `ShellConfiguration`. Similarly for `-p`/`--permissive` specifying if the evaluator should be run in permissive typing mode.

cli/src/org/partiql/cli/main.kt
```diff
@@ -139,6 +142,7 @@ private val outputFormatOpt = optParser.acceptsAll(listOf("output-format", "of")
  *      * -w --wrap-ion: wraps Ion input file values in a bag, requires the input format to be ION, requires the query option
  *      * -o --output: output file (default: STDOUT)
  *      * -of --output-format: (default: PARTIQL) [ION_TEXT, ION_BINARY, PARTIQL, PARTIQL_PRETTY]
+ *      * -m --monochrome: removes syntax highlighting for the REPL
```
alancai98, line 145: This option is listed under the "Non interactive only" heading. I think it should be moved up above that heading

cli/resources/org/partiql/cli/syntax/PartiQL.nanorc
johnedquinn: This next revision is better because it resolves this. I'll leave more info below. Already updated the PR description. Ion still isn't added for syntax, but now it's a much more straightforward path.

cli/resources/org/partiql/cli/syntax/PartiQL.nanorc
johnedquinn: New revision uses the SqlLexer and SqlParser instead of nanorc. Better experience overrall. See the new video in the description.

johnedquinn:
Decided to go ahead and use the SqlLexer and SqlParser instead of the nanorc files because nanorc doesn't allow advanced features such as multiple grammars (hence our inability to differentiate between PartiQL quoted identifiers and Ion strings). Makes the most sense to use our grammar to identify tokens.

I've updated the description with a new video. Later, we can add highlighting for Ion, and this sets us up for that.

cli/resources/org/partiql/cli/syntax/PartiQL.nanorc
RCHowell: ^^ ask and you shall receive haha. This is awesome, nice work!

cli/src/org/partiql/shell/ShellHighlighter.kt
```diff
@@ -11,32 +11,134 @@
...
+        val tokens: List<Token>
+        try {
+            tokens = lexer.tokenize(input)
+        } catch (e: Exception) {
+            return AttributedString(input, AttributedStyle().foreground(AttributedStyle.RED))
+        }
 
-    override fun highlight(reader: LineReader, line: String): AttributedString = syntax.highlight(line)
+        // Replace Token Colors
+        tokens.forEach { builder = createAttributeStringBuilder(it, lineIndexesMap, builder, false) }
```
RCHowell, line 70: This confuses me. Why are we continually updating the value of builder almost like a fold?

Why not do
```
val builder = AttributedStringBuilder()
tokens.forEach { addTokenToBuilder(input, lineIndexesMap, builder, false)
```

RCHowell: I guess I'm confused why the builder is returned

cli/src/org/partiql/shell/ShellHighlighter.kt
```diff
@@ -11,32 +11,134 @@
...
+            TokenType.NULL -> AttributedStyle.BLUE
+            else -> AttributedStyle.WHITE
+        }
+        style = style.foreground(attrCode)
+
+        return when (token.type) {
+            TokenType.IDENTIFIER, TokenType.OPERATOR -> style.bold()
+            else -> style
+        }
+    }
```
RCHowell, line 141: Consider putting these in a map. Kotlin `when` is pretty clean, but it just leads to a lot of branches for what is just a lookup

cli/src/org/partiql/shell/ShellHighlighter.kt
```diff
@@ -11,32 +11,134 @@
...
 import java.io.PrintStream
-import java.nio.file.Path
 import java.util.regex.Pattern
 
 private val SUCCESS: AttributedStyle = AttributedStyle.DEFAULT.foreground(AttributedStyle.GREEN)
 private val ERROR: AttributedStyle = AttributedStyle.DEFAULT.foreground(AttributedStyle.RED)
 private val INFO: AttributedStyle = AttributedStyle.DEFAULT
 private val WARN: AttributedStyle = AttributedStyle.DEFAULT.foreground(AttributedStyle.YELLOW)
 
-class ShellHighlighter(private val syntaxFile: Path) : Highlighter {
```
alancai98, line 31: nit: parens aren't necessary following `ShellHighlighter`

cli/src/org/partiql/shell/ShellHighlighter.kt
```diff
@@ -11,32 +11,134 @@
...
+        var builder = AttributedStringBuilder()
+        builder.append(AttributedString(input))
+
+        // Map Between Line Number and Index
+        val lineIndexesMap = mutableMapOf<Int, Int>()
+        var currentLine = 0
+        lineIndexesMap[currentLine] = -1
+        input.forEachIndexed { index, char ->
+            if (char == '\n') lineIndexesMap[++currentLine] = index
+        }
```
alancai98, line 59: What is meant by "Index" here? Does it refer to the index of the PartiQL statement's token's position in the sequence?

cli/src/org/partiql/shell/ShellHighlighter.kt
```diff
@@ -11,32 +11,134 @@
...
+        a.append(stringBuilder.subSequence(0, index))
+        a.append(stringBuilder.subSequence(index, index + length), style)
+        a.append(stringBuilder.subSequence(index + length, stringBuilder.length))
+        return a
+    }
+
+    /**
+     * Sets the color and thickness of the string based on the [token] type
+     */
+    private fun getStyle(token: Token): AttributedStyle {
```
alancai98, line 120: May be useful to abstract the Token styling to a separate configuration a user can modify and provide through the `ShellConfiguration`. For accessibility (e.g. color-blindness), it may be helpful for a user to modify the Token styling to something other than the defaults. Feel free to create an issue to track since that configurability may be out of scope for this PR.

RCHowell: I would agree this is out of scope, but should definitely be a separate issue because this is a great idea

cli/src/org/partiql/shell/ShellHighlighter.kt
johnedquinn: Modifying to build a string. The only caveat is for parsing errors which needs to highlight a specific word red. You'll like the next revision posting soon. Thanks for the comment. Makes sense.

cli/src/org/partiql/shell/ShellHighlighter.kt
johnedquinn: ACK. I'll create an issue for highlighting based on user input. Then, we can store the default input in our resources. I'm not going to move the current syntax rules away from the when() since I'd want to clarify exactly what the input should be like, but that can come later. I'll make that a desired outcome of the issue. Will link.

But yeah, I totally agree. That's why I also created the `--monochrome` flag to turn off syntax highlighting if people have issues seeing it.

cli/src/org/partiql/shell/ShellHighlighter.kt
johnedquinn: See my comment about syntax files. If okay, I'll make this a requirement for the next issue to define a syntax file for users to pass in. Then, we can move this to a standardized file.

RCHowell (approved):

cli/src/org/partiql/shell/ShellHighlighter.kt
alancai98: Not sure if this issue has been created yet. Did you want me to create it?

PR #643, opened 2022-06-20
Adds other compile options to the CLI

### LINKED ISSUE #544: [cli] Add other `CompileOption` modes to CLI/REPL
Currently the CLI supports a few compile options (see [CLI documentation](https://github.com/partiql/partiql-lang-kotlin/blob/3febaeec46a0b703781494699c445f2034e9c6c4/docs/user/CLI.md#partiql-cli)). Would be helpful to test and experiment with other [CompileOptions](https://github.com/partiql/partiql-lang-kotlin/blob/3febaeec46a0b703781494699c445f2034e9c6c4/lang/src/org/partiql/lang/eval/CompileOptions.kt) using the CLI/REPL. E.g.:
- `TypedOpBehavior` -- LEGACY vs HONOR_PARAMETERS
- `ProjectionIterationBehavior` -- FILTER_MISSING vs UNFILTERED
- `UndefinedVariableBehavior` -- ERROR vs MISSING

### LINKED ISSUE #566: PartiQL fails to detect failed integer casts - in the REPL
I would expect the following to fail with a runtime exception or return `MISSING` in permissive mode but it does neither:

```
PartiQL> cast(3276882344 as smallint)
   | 
===' 
3276882344
--- 
OK!
```

The result is the same regardless of if the CLI is started with the new `-p` option which enables permissive mode.   The reason is because the CLI doesn't set `TypedOpBehavior.HONOR_PARAMS` in the `CompileOptions` of the pipeline.  So this can be fixed by adding a command-line option to change this setting and/or set it by default.

johnedquinn:
## Relevant Issues
- Closes #566
- Closes #544

## Description
- Adds 3 useful CLI flags to the CLI options

### DIFF
--- a/cli/src/org/partiql/cli/main.kt
+++ b/cli/src/org/partiql/cli/main.kt
@@ -151,7 +151,7 @@ private val outputFormatOpt = optParser.acceptsAll(listOf("output-format", "of")
  * Options:
  * * -e --environment: takes an environment file to load as the initial global environment
  * * -p --permissive: run the query in permissive typing mode (returns MISSING rather than error for data type
- * * -t --typed-op-behavior: indicates how CAST should behave: (default: LEGACY) [LEGACY, HONOR_PARAMETERS]
+ * * -t --typed-op-behavior: indicates how CAST should behave: (default: HONOR_PARAMETERS) [LEGACY, HONOR_PARAMETERS]
  * * -r --projection-iter-behavior: Controls the behavior of ExprValue.iterator in the projection result: (default: FILTER_MISSING) [FILTER_MISSING, UNFILTERED]
  * * -v --undefined-variable-behavior: Defines the behavior when a non-existent variable is referenced: (default: ERROR) [ERROR, MISSING]
  * mismatches)
--- a/cli/src/org/partiql/cli/main.kt
+++ b/cli/src/org/partiql/cli/main.kt
@@ -96,7 +96,7 @@ private val typedOpBehaviorOpt = optParser.acceptsAll(listOf("typed-op-behavior"
     .withRequiredArg()
     .ofType(TypedOpBehavior::class.java)
     .describedAs("(${TypedOpBehavior.values().joinToString("|")})")
-    .defaultsTo(TypedOpBehavior.LEGACY)
+    .defaultsTo(TypedOpBehavior.HONOR_PARAMETERS)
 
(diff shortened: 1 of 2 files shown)


cli/src/org/partiql/cli/main.kt
```diff
@@ -131,6 +151,9 @@ private val outputFormatOpt = optParser.acceptsAll(listOf("output-format", "of")
  * Options:
  * * -e --environment: takes an environment file to load as the initial global environment
  * * -p --permissive: run the query in permissive typing mode (returns MISSING rather than error for data type
+ * * -t --typed-op-behavior: indicates how CAST should behave: (default: LEGACY) [LEGACY, HONOR_PARAMETERS]
+ * * -r --projection-iter-behavior: Controls the behavior of ExprValue.iterator in the projection result: (default: FILTER_MISSING) [FILTER_MISSING, UNFILTERED]
+ * * -v --undefined-variable-behavior: Defines the behavior when a non-existent variable is referenced: (default: ERROR) [ERROR, MISSING]
```
alancai98, line 156: Could some tests also be added for these other compile options? Something similar to what was done in [CliTest.kt](https://github.com/partiql/partiql-lang-kotlin/blob/b3e62fc676740fd18ffcb93099e55a1af14f9136/cli/test/org/partiql/cli/CliTest.kt#L213-L224) for the permissive CLI mode.

cli/src/org/partiql/cli/main.kt
```diff
@@ -90,6 +92,24 @@ private val queryOpt = optParser.acceptsAll(listOf("query", "q"), "PartiQL query
 
 private val permissiveModeOpt = optParser.acceptsAll(listOf("permissive", "p"), "runs the query in permissive mode")
 
+private val typedOpBehaviorOpt = optParser.acceptsAll(listOf("typed-op-behavior", "t"), "indicates how CAST should behave")
+    .withRequiredArg()
+    .ofType(TypedOpBehavior::class.java)
+    .describedAs("(${TypedOpBehavior.values().joinToString("|")})")
+    .defaultsTo(TypedOpBehavior.LEGACY)
```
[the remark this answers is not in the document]
johnedquinn: Yeah, it's definitely a bit weird. The only reason I kept it as LEGACY is because the CompilerPipeline also has the default impl as LEGACY.

cli/src/org/partiql/cli/main.kt
johnedquinn: Sure. Will do.

alancai98 (approved):
:shipit:

Issue #651, opened 2022-06-24
!add_to_global_env is broken in the Shell


johnedquinn:
Seems like the !add_to_global_env function is broken in the shell.

## Reproduce

Type out:
```
!add_to_global_env { 'a': 3 }
```

It works if you copy/paste:
```
!add_to_global_env {
  'a': 3
}
```

Also, if you copy/paste:
```
!add_to_global_env { 'a': 3 }
```

The syntax highlighting doesn't work on the first line.

And it doesn't work if you type it out (not copy/paste or history)

## To Look At

Looks like it might have to do with #646 -- but need to look more into it.

PR #657, opened 2022-06-29
Fixes add_to_global_env highlighting

RCHowell:
[*Issue #, if available:*](https://github.com/partiql/partiql-lang-kotlin/issues/651)

*Description of changes:*

Fixed the `!add_to_global_env` highlighting. This will highlight the PartiQL value in the `!add_to_global_env` command using the lexer/parser. 

No, small bug fix

No

```
```

No

### DIFF
--- a/cli/src/org/partiql/shell/ShellHighlighter.kt
+++ b/cli/src/org/partiql/shell/ShellHighlighter.kt
@@ -62,7 +62,7 @@ internal class ShellHighlighter() : Highlighter {
             return if (hasAddToGlobalEnv) {
                 AttributedString(line, AttributedStyle.DEFAULT.foreground(AttributedStyle.CYAN))
             } else {
-                AttributedString("")
+                AttributedString(line)
             }
         }
 
--- a/cli/src/org/partiql/shell/ShellHighlighter.kt
+++ b/cli/src/org/partiql/shell/ShellHighlighter.kt
@@ -60,7 +60,7 @@ internal class ShellHighlighter() : Highlighter {
 
         if (input.isBlank()) {
             return if (hasAddToGlobalEnv) {
-                ADD_TO_GLOBAL_ENV_COMMAND
+                AttributedString(line, AttributedStyle.DEFAULT.foreground(AttributedStyle.CYAN))
             } else {
                 AttributedString("")
             }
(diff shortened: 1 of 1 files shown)


RCHowell:
https://user-images.githubusercontent.com/5731503/176507457-e1b38404-c3de-4a38-9552-a3600e3015b0.mov

cli/src/org/partiql/shell/ShellHighlighter.kt
```diff
@@ -47,8 +49,22 @@ internal class ShellHighlighter() : Highlighter {
...
+            line.substring(ADD_TO_GLOBAL_ENV.length, line.length)
+        } else {
+            line
+        }
+
+        if (input.isBlank()) {
+            return if (hasAddToGlobalEnv) {
+                ADD_TO_GLOBAL_ENV_COMMAND
+            } else {
+                AttributedString("")
```
johnedquinn, line 65:
```suggestion
                AttributedString(line)
```

cli/src/org/partiql/shell/ShellHighlighter.kt
```diff
@@ -47,8 +49,22 @@ internal class ShellHighlighter() : Highlighter {
...
+        val hasAddToGlobalEnv = line.toLowerCase().startsWith(ADD_TO_GLOBAL_ENV)
+        val input = if (hasAddToGlobalEnv) {
+            line.substring(ADD_TO_GLOBAL_ENV.length, line.length)
+        } else {
+            line
+        }
+
+        if (input.isBlank()) {
+            return if (hasAddToGlobalEnv) {
+                ADD_TO_GLOBAL_ENV_COMMAND
```
johnedquinn, line 63: After pulling this locally, if you type in `!add_to_global_env` and press Return immediately, there's an out of bounds exception. Seems like it's from this line. We want to color the entire line and return. If we go back to when we first added highlighting (or before highlighting), we were allowed to do this. See the suggestion that looks like it fixes this:
```suggestion
                AttributedString(line, AttributedStyle.DEFAULT.foreground(AttributedStyle.CYAN))
```

johnedquinn (changes requested):
Looks great. Decided to pull the changes for manual testing, and one thing is broken. Essentially pressing enter after `!add_to_global_env`. See my suggestions.

johnedquinn (approved):
LGTM.

PR #666, opened 2022-07-12
[superceded] Partial support for INSERT and DELETE DML statements

dlurton:
#### Description of Changes

This PR adds support to PartiQL for DML statements taking the following forms:

- `INSERT INTO foo << { 'id': 1, 'name': 'bob' }, { 'id': 2, 'name' : 'sue' } >>` (multi record insert)
- `INSERT INTO foo VALUE 1` (single record insert)
- `INSERT INTO foo SELECT c.id, c.name FROM customer AS c` (insert the results of a query into another table)
- `DELETE FROM foo` (delete all records in a table)
- `DELETE FROM foo AS f WHERE f.zipCode = '90210'` (delete all records matching a predicate)

This is accomplished by rewriting DML statements into standard queries and enhancing the API of `PlannerPipeline` to better support execution of these queries.  For details on the rewrites, see changes to `LogicalResolvedToDefaultPhysicalVisitorTransform` and its tests to see how this works.

---

This PR is still in WIP!  This PR has been completed enough to verify that all of the above DML statements work, however, the following must be completed before this may be merged:

- [ ] Address those `TODO` comments and function calls (those that are within the scope described above, that is).
- [ ] Throw/handle proper exceptions & error messaging.
- [ ] Additional unit testing (particularly surrounding error cases)
- [x] Update `CHANGELOG.md`

#### Backward Incompatible Changes

The API of `PlannerPipeline` (an experimental API that few, if anybody, is using) changes.  Namely, the result of compilation is now a `QueryPlan` which is analogous to the previous `Expression` interface but returns `QueryResult` instead of `ExprValue`.  `QueryResult` is a sealed class that may be one of `QueryResult.Value` (the result of an SFW query or other expression and contains an `ExprValue`) or `QueryResult.DmlCommand`, which is the result of a DML operation and contains a series of "commands" which must be applied by the application embedding PartiQL to effect the DML.

#### Dependencies 

> *Does your PR introduce a new external dependency? (y/n), if yes, please explain the reason. In addition, please

No new dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - path quantifiers  (section 4.4 of the GPML paper)
   - restrictors and selector  (section 5.1 of the GPML paper)
   - pre-filters and post-filters (section 5.2 of the GPML paper)
+- The `PlannerPipeline` API now has experimental and partial support for `INSERT` and `DELETE` DML statements.
+  Examples of supported statements include:
+  - `INSERT INTO foo << { 'id': 1, 'name': 'bob' }, { 'id': 2, 'name' : 'sue' } >>` (multi record insert)
+  - `INSERT INTO foo VALUE 1` (single record insert)
+  - `INSERT INTO foo SELECT c.id, c.name FROM customer AS c` (insert the results of a query into another table)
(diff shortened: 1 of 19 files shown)


dlurton:
Superceded by https://github.com/partiql/partiql-lang-kotlin/pull/675

Issue #601, opened 2022-05-09
Implement the `ROUND` function


alancai98:
A common use case would be to round decimals and get the specified precision. This doesn't seem to be a SQL standard function, but it's a common enough use case that we should implement it. 

PostgreSQL provides two versions that could both be included in the builtin functions.

> round ( numeric ) → numericround ( double precision ) → double precisionRounds to nearest integer. For numeric, ties are broken by rounding away from zero. For double precision, the tie-breaking behavior is platform dependent, but “round to nearest even” is the most common rule.round(42.4) → 42
--
round ( v numeric, s integer ) → numericRounds v to s decimal places. Ties are broken by rounding away from zero.round(42.4382, 2) → 42.44

PR #667, opened 2022-07-14
Round function

yliuuuu:
*Issue #, if available:* 

#601 Round Function Implementation

*Description of changes:* 

Implement the round function. Based on the following: 

  | source  | targetScale | output      | description |
 |---------|------------|-----------|-------------|
  | Integer | omitted/0   | Integer     | round to the nearest Integer |
  | Integer | > 0         | Decimal     | transform to Decimal |
  | Integer | < 0         | Integer     | round to [targetScale] decimal places |
  | Inexact | omitted/0   | Integer     | round to the nearest Integer |
  | Inexact | > 0         | Not allowed | allowing this potentially introduce accuracy loss |
  | Inexact | < 0         | Integer     | round to [targetScale] decimal places |
  | Exact   | omitted/0   | Integer     | round to the nearest Integer |
  | Exact   | > 0         | Decimal    | round to [targetScale] decimal places|
  | Exact   | < 0         | Integer     | round to [targetScale] decimal places |

No.

No.

```
```

No.

- [ ] The round function Documentation 
- [ ] Confirm the rounding behavior
- [ ] Confirm the function input/output
- [ ] CHANGELOG.md
- [ ] BuildinFunction.md

### DIFF
--- a/lang/src/org/partiql/lang/eval/builtins/MathFunctions.kt
+++ b/lang/src/org/partiql/lang/eval/builtins/MathFunctions.kt
@@ -1,5 +1,6 @@
 package org.partiql.lang.eval.builtins
 
+import com.amazon.ion.Decimal
 import org.partiql.lang.errors.ErrorCode
 import org.partiql.lang.eval.EvaluationSession
 import org.partiql.lang.eval.ExprFunction
@@ -7,14 +8,13 @@ import org.partiql.lang.eval.ExprValue
 import org.partiql.lang.eval.ExprValueFactory
 import org.partiql.lang.eval.errIntOverflow
 import org.partiql.lang.eval.errNoContext
+import org.partiql.lang.eval.intValue
 import org.partiql.lang.eval.numberValue
 import org.partiql.lang.types.FunctionSignature
 import org.partiql.lang.types.StaticType
 import org.partiql.lang.util.bigDecimalOf
 import org.partiql.lang.util.compareTo
-import org.partiql.lang.util.div
-import org.partiql.lang.util.plus
-import org.partiql.lang.util.times
+import org.partiql.lang.util.isZero
 import org.partiql.lang.util.unaryMinus
 import java.lang.Double.NEGATIVE_INFINITY
 import java.lang.Double.NaN
@@ -33,6 +33,7 @@ object MathFunctions {
         UnaryNumeric("ceil", valueFactory) { ceil(it) },
         UnaryNumeric("ceiling", valueFactory) { ceil(it) },
         UnaryNumeric("floor", valueFactory) { floor(it) },
+        RoundFunction(valueFactory)
     )
 }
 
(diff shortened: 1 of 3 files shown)


lziq:
One thing to note that in most of other sql languages, `NUMERIC` is the same as `DECIMAL`. https://learnsql.com/blog/numeric-decimal-data-types-sql/#:~:text=There%20is%20one%20notable%20difference,numbers%20than%20the%20stated%20precision. In PartiQL, `NUMERIC` is a wrapper of all the number types, and I am not sure whether this is correct.

lziq:
As Postgresql specifies [here](https://www.postgresql.org/docs/8.1/functions-math.html), I think we should only implement `round(v numeric, s int)` in this PR and skip `round(dp or numeric)` for now. As the latter's return type is dependent on the source argument type, which requires us to have functions overloading.

alancai98 (changes requested):
I haven't looked too closely at the whole implementation or test cases, but I think we need to decide on the float behavior first (i.e. round away from zero vs round to nearest even).

Sorry if this wasn't documented anywhere, but we'll also need to update the [BuiltinFunction.md](https://github.com/partiql/partiql-lang-kotlin/blob/388b049fca455745925683665be65784954967df/docs/user/BuiltInFunctions.md) file w/ `ROUND`. You can look at what was done for `CEIL`/`FLOOR` in this PR: https://github.com/partiql/partiql-lang-kotlin/pull/613.

alancai98:
Other general comments for PRs:
1. if possible, try to ensure the commit history doesn't contain other unrelated changes (e.g. this PR contains prior commits from #662; merging that PR into `main` should clean up the PR diff)
2. is there a reason this is not included in the changelog? I think adding `ROUND` should be included
3. it's generally good practice to link an issue to the pull request: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue. Doing so can prevent issues from staying open when they've been resolved/implemented. It can be done by
- using a "closing keyword" before the issue number
- manually linking the issue (go to the "Development" heading -> choose issue to link)

lang/src/org/partiql/lang/eval/builtins/MathFunctions.kt
```diff
@@ -43,22 +56,153 @@ private class UnaryNumeric(
...
+}
+
+/**
+ * | source  | targetScale | output      | description|
+ * | Integer | omitted/0   | Integer     | round to the nearest Integer
+ * | Integer | > 0         | Decimal     | transform to Decimal
+ * | Integer | < 0         | Integer     | round to [targetScale] decimal places
+ * | Inexact | omitted/0   | Integer     | round to the nearest Integer
+ * | Inexact | > 0         | Not allowed | allowing this potentially introduce accuracy loss
+ * | Inexact | < 0         | Integer     | round to [targetScale] decimal places
```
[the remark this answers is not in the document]
alancai98: Don't personally have a preference on which route to take. We generally follow postgresql, which says the behavior is "platform dependent" but 

> ... “round to nearest even” is the most common rule

Also raises the question of if we should allow both options (either through some compiler option or separate function or separate function argument.

lang/src/org/partiql/lang/eval/builtins/MathFunctions.kt
```diff
@@ -92,3 +92,116 @@ private fun Number.exprValue(valueFactory: ExprValueFactory): ExprValue = when (
...
+        optionalParameter = StaticType.unionOf(StaticType.INT2, StaticType.INT4, StaticType.INT8, StaticType.INT)
+    )
+
+    override fun callWithOptional(session: EvaluationSession, required: List<ExprValue>, opt: ExprValue): ExprValue {
+        return round(required.first().numberValue(), opt.intValue()).exprValue(valueFactory)
+    }
+
+    override fun callWithRequired(session: EvaluationSession, required: List<ExprValue>): ExprValue {
+        return round(required.first().numberValue(), 0).exprValue(valueFactory)
+    }
```
lziq, line 122: If I were you, I would prefer to write something as follow: 
```
    override fun callWithOptional(session: EvaluationSession, required: List<ExprValue>, opt: ExprValue): ExprValue {
        return round(required.first(), opt.intValue())
    }

    override fun callWithRequired(session: EvaluationSession, required: List<ExprValue>): ExprValue {
        val source = required.first()
        if (source.type == ExprValueType.Float) {
            // Either throws an error or deals with it
        } else {
            return round(source, 0)
        }
    }
    
    private fun round(source: ExprValue, targetScale: Int): ExprValue =
        when (source.type) {
            is ExprValueType.INT -> source.value.longValue().toBigDecimal()
            is ExprValueType.DECIMAL -> source.value.bigDecimalValue()
            else -> error("Unsupported type for round() function")
        }
        
   private fun roundDecimal(value: BigDecimal): BigDecimal {
        // Implementation
   }
```

lziq: I might be wrong here.

Issue #184, opened 2020-02-13
Add support for set operators `UNION`, `INTERSECT` `EXCEPT`


therapon:
## Overview

Add support for set operators `UNION`, `INTERSECT` `EXCEPT`

## Tasks
- [x] https://github.com/partiql/partiql-docs/pull/7
- [x] Define OUTER variants in the lexer/parser/ast
- [x] Implement the simpler OUTER variants
- [ ] Implement SQL conformant UNION, INTERSECT, EXCEPT -- depends on partiql-spec [#45](https://github.com/partiql/partiql-lang/issues/22).

A follow-up to this ticket can be #1015, for the  `CORRESPONDING BY` clause in these operators.

RCHowell:
The RFC for set operator behavior has been merged. Implementation is next.
 
https://github.com/partiql/partiql-docs/blob/main/drafts/0007-RFC-BAG-OPERATORS.md

RCHowell:
Added a sub task list to issue description

alancai98:
> The RFC for set operator behavior has been merged. Implementation is next.
> 
> https://github.com/partiql/partiql-docs/blob/main/drafts/0007-RFC-BAG-OPERATORS.md

FYI the above RFC was moved from `drafts/` to the `RFC`s folder: https://github.com/partiql/partiql-docs/blob/main/RFCs/0007-rfc-bag-operators.md

Issue #241, opened 2020-05-08
Model n-ary nodes better


dlurton:
We currently model "n-ary" nodes as direct variants of the `expr` sum type:
 
```
(sum expr
    // Arithmetic operators
    (plus (* expr 2))
    (minus (* expr 2))
    (times (* expr 2))
    (divide (* expr 2))
    (modulo (* expr 2))
    (string_concat (* expr 2))

    // Logical operators
    (and (* expr 2))
    (or (* expr 2))

    // Comparison operators
    (eq (* expr 2))
    (ne (* expr 2))
    (gt (* expr 2))
    (gte (* expr 2))
    (lt (* expr 2))
    (lte (* expr 2))
    (in_container (* expr 2))
    ...)
```

However a better method of modeling them that allows for greater re-use of Kotlin code is to model the function as another sum variant:

```
(sum expr
    ...
    (nary nary_op (* expr 2))
    ...)

(sum nary_op)
    // Arithmetic operators
    (plus)
    (minus)
    (times)
    (divide)
    (modulo)
    (string_concat)

    // Logical operators
    (and)
    (or)

    // Comparison operators
    (eq)
    (ne)
    (gt)
    (gte)
    (lt)
    (lte)
    (in_container)
)
```

alancai98:
Aged out. Modeling of nodes in AST and plan changed significantly.

Issue #239, opened 2020-05-08
Model `bag`, `list`, `sexp` better


dlurton:
To remain consistent with the V0 AST and reduce the impact on QLDB's rewriter rules, we are at modeling the bag, and list constructor expression as independent nodes:

```
(sum expr 
    ... other expressions ...
    (bag (* expr 0))
    (list (* expr 0)))
```
However, this may lead to some code duplication because they are modeled as separate types.  To reduce the potential for code duplication, we should consider modeling the type of sequence as an attribute instead.  Something like:

```text
(sum expr 
    ... other expressions ...
    (seq seq_type (* expr 0)))

(sum seq_type
    (sexp_seq)
    (list_seq)
    (bag_seq))
```

PR #678, opened 2022-07-20
Adds OUTER set operators to parser and type domain

RCHowell:
*Issue #, if available:*

Part 2 of — https://github.com/partiql/partiql-lang-kotlin/issues/184

*Description of changes:*

These changes add the OUTER set operators defined in [RFC-0007 ](https://github.com/partiql/partiql-docs/blob/main/RFCs/0007-rfc-bag-operators.md) to the type universe, lexer, and parser.

*Have you updated the `Unreleased` section of `CHANGELOG.md` with your changes? (y/n), If not, please explain why:* No, N/A

 also mention any other alternatives you've considered and the reason they've been discarded?:* N

```
```

also mention any other alternatives you've considered and the reason they've been discarded?:* N

### DIFF
--- a/lang/resources/org/partiql/type-domains/partiql.ion
+++ b/lang/resources/org/partiql/type-domains/partiql.ion
@@ -130,6 +130,9 @@ may then be further optimized by selecting better implementations of each operat
             (union setq::set_quantifier operands::(* expr 2))
             (except setq::set_quantifier operands::(* expr 2))
             (intersect setq::set_quantifier operands::(* expr 2))
+            (outer_union setq::set_quantifier operands::(* expr 2))
+            (outer_except setq::set_quantifier operands::(* expr 2))
+            (outer_intersect setq::set_quantifier operands::(* expr 2))
 
             // Other expression types
             (path root::expr steps::(* path_step 1))
--- a/lang/src/org/partiql/lang/syntax/LexerConstants.kt
+++ b/lang/src/org/partiql/lang/syntax/LexerConstants.kt
@@ -445,14 +445,19 @@ internal val DATE_TIME_PART_KEYWORDS: Set<String> = DateTimePart.values()
     listOf("union", "all") to ("union_all" to OPERATOR),
     listOf("intersect", "all") to ("intersect_all" to OPERATOR),
     listOf("except", "all") to ("except_all" to OPERATOR),
+    listOf("union", "distinct") to ("union_distinct" to OPERATOR),
+    listOf("intersect", "distinct") to ("intersect_distinct" to OPERATOR),
+    listOf("except", "distinct") to ("except_distinct" to OPERATOR),
+
     listOf("outer", "union") to ("outer_union" to OPERATOR),
     listOf("outer", "intersect") to ("outer_intersect" to OPERATOR),
     listOf("outer", "except") to ("outer_except" to OPERATOR),
-
-    // tomfoolery with multi lexeme token folding
-    listOf("outer_intersect", "all") to ("outer_intersect_all" to OPERATOR),
     listOf("outer_union", "all") to ("outer_union_all" to OPERATOR),
+    listOf("outer_intersect", "all") to ("outer_intersect_all" to OPERATOR),
     listOf("outer_except", "all") to ("outer_except_all" to OPERATOR),
+    listOf("outer_union", "distinct") to ("outer_union_distinct" to OPERATOR),
+    listOf("outer_intersect", "distinct") to ("outer_intersect_distinct" to OPERATOR),
+    listOf("outer_except", "distinct") to ("outer_except_distinct" to OPERATOR),
 
     listOf("character", "varying") to ("character_varying" to KEYWORD),
     listOf("double", "precision") to ("double_precision" to KEYWORD),
--- a/lang/src/org/partiql/lang/syntax/LexerConstants.kt
+++ b/lang/src/org/partiql/lang/syntax/LexerConstants.kt
@@ -441,9 +441,19 @@ internal val DATE_TIME_PART_KEYWORDS: Set<String> = DateTimePart.values()
     listOf("not", "in") to ("not_in" to OPERATOR),
     listOf("is", "not") to ("is_not" to OPERATOR),
     listOf("not", "between") to ("not_between" to OPERATOR),
+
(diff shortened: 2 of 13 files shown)


lang/src/org/partiql/lang/syntax/LexerConstants.kt
```diff
@@ -441,9 +441,19 @@ internal val DATE_TIME_PART_KEYWORDS: Set<String> = DateTimePart.values()
     listOf("not", "in") to ("not_in" to OPERATOR),
     listOf("is", "not") to ("is_not" to OPERATOR),
     listOf("not", "between") to ("not_between" to OPERATOR),
+
+    listOf("union", "all") to ("union_all" to OPERATOR),
```
alancai98, line 445: (not completely relevant to this PR's changes but may be relevant to future set operator PRs): I know that there's an implicit `DISTINCT` when no set quantifier is provided for the set operators, but I think the lexer isn't properly handling the case when the `DISTINCT` set quantifier is defined.

E.g:
```
PartiQL> A UNION DISTINCT B;
org.partiql.lang.syntax.ParserException: Unexpected keyword
        Parser Error: at line 1, column 9: unexpected keyword found, KEYWORD : distinct
```

Since it's part of the [RFC](https://github.com/partiql/partiql-docs/blob/main/RFCs/0007-rfc-bag-operators.md#definition), we may need to handle the explicit `DISTINCT` cases as well.

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -130,6 +130,9 @@ may then be further optimized by selecting better implementations of each operat
             (union setq::set_quantifier operands::(* expr 2))
             (except setq::set_quantifier operands::(* expr 2))
             (intersect setq::set_quantifier operands::(* expr 2))
+            (outer_union setq::set_quantifier operands::(* expr 2))
+            (outer_except setq::set_quantifier operands::(* expr 2))
+            (outer_intersect setq::set_quantifier operands::(* expr 2))
```
alancai98, line 135: Could there be a modeling of set operators in `partiql.ion` where the `OUTER` is included in the existing `union`, `except`, and `intersect` nodes rather than having to create other nodes? Not sure if that modeling would be cleaner and/or easier to work with than the current modeling of set operators.

lang/src/org/partiql/lang/syntax/SqlParser.kt
```diff
@@ -888,12 +888,18 @@ class SqlParser(
                     // TODO:  we are losing case-sensitivity of the function name here.  Do we care?
                     call(idArg.name.text, args.drop(1), metas)
                 }
+                "union" -> union(distinct(), args, metas)
```
alancai98, line 891: We also need some tests (probably in [SqlParser.kt](https://github.com/partiql/partiql-lang-kotlin/blob/0b271115b44b74b63aa1e7207d6f23d2da89b630/lang/test/org/partiql/lang/syntax/SqlParserTest.kt)) to make sure the `PartiqlAst` output by the parser is correct.

alancai98 (changes requested):
Left some comments about the modeling of `OUTER` in set operators and not needing to support `OUTER` in the older AST versions. Also noticed that specifying `DISTINCT` following the set operator results in a parser error.

lang/src/org/partiql/lang/syntax/LexerConstants.kt
RCHowell: I'll fix the DISTINCT in the next revision

lang/resources/org/partiql/type-domains/partiql.ion
RCHowell: These are actually notably different operators. I would much prefer to make an abstraction here rather than how it is done today. These should all just be `(bagop op::bagop_type setq::set_quantifier operands::(* expr 2))`. It can get frustrating how little abstraction we have .. lots of duplication which continues to pile up

RCHowell: As pointed out by Alan, there are existing issues to improve the IR
- https://github.com/partiql/partiql-lang-kotlin/issues/241 — NAry ops
- https://github.com/partiql/partiql-lang-kotlin/issues/239 — collection types

lang/src/org/partiql/lang/syntax/LexerConstants.kt
```diff
@@ -535,12 +550,18 @@ enum class OperatorPrecedenceGroups(val precedence: Int) {
  */
 @JvmField internal val OPERATOR_PRECEDENCE = mapOf(
     // set operator group
+    "union" to OperatorPrecedenceGroups.SET.precedence,
```
alancai98, line 553: I think this map is missing the `*_distinct` variants:
- union_distinct
- intersect_distinct
- except_distinct
- outer_union_distinct
- outer_intersect_distinct
- outer_except_distinct

lang/test/org/partiql/lang/syntax/SqlParserTest.kt
```diff
@@ -3987,6 +3987,13 @@ class SqlParserTest : SqlParserTestBase() {
         """
     )
 
+    @Test
```
alancai98, line 3990: Sorry if this wasn't documented, but for new tests in this file, we should be using the [assertExpression](https://github.com/partiql/partiql-lang-kotlin/blob/0dee7a6ce31c5482b5769f6507aa32006c923ae9/lang/test/org/partiql/lang/syntax/SqlParserTestBase.kt#L110) that asserts on the `PartiqlAst` rather than the s-expression representation of the AST. The s-expression version of the AST can be harder to read and directly interpret compared to the PIG-builder.

Could the tests from L3990-4040 use the `assertExpression` that asserts on the PIG-builder? Example test using the PIG-builder `assertExpression` https://github.com/partiql/partiql-lang-kotlin/blob/0dee7a6ce31c5482b5769f6507aa32006c923ae9/lang/test/org/partiql/lang/syntax/SqlParserTest.kt#L1793-L1804

alancai98 (commented):
Thanks for adding more tests and addressing the feedback. Left a comment about some operators missing from the precedence map and some rewritten `SqlParserTest` test cases.

alancai98 (approved):
:shipit:

Issue #677, opened 2022-07-20
Reduce the domain of acceptable values for DML targets.  


dlurton:
The domain of values allowed in the `partiql_logical.dml.target` element is too broad.  It allows for any expression, but in reality this is only a subset of expressions that is allowed by the grammar of PartiQL.

Of course, we it would be possible to create a `sum` type like `expr` but containing only the valid DML targets, but that care would have to be taken doing this without duplicating some logic related to variable resolution which is already implemented for standard expressions.  This task is to do this without copy-pasta.

PR #680, opened 2022-07-21
Narrow domain of DML targets (#677)

dlurton:
*Issue #, if available:*

#677 

*Description of changes:*

The domain of values allowed in the `partiql_logical.dml.target` element was too broad because it allowed for any expression.

This commit changes this to `partiql_logical.dml.target` to be simply `identifier`, which is the only currently supported form of DML targets in the planner.

No, this PR only slightly changes stuff already in the change log.

"Technically", the only change is to  `partiql_logical.dml.target`--an API that is previously unreleased.

No.

### DIFF
--- a/lang/resources/org/partiql/type-domains/partiql.ion
+++ b/lang/resources/org/partiql/type-domains/partiql.ion
@@ -697,7 +697,7 @@ may then be further optimized by selecting better implementations of each operat
                 // A DML operation, such as `INSERT`, `UPDATE` or `DELETE`
                 (dml
                     // The target is an expression that is indicates the table whose data is to be manipulated.
-                    // According to the PartiQL's grammar this can be an identifier or a simplified path expression
+                    // With current PartiQL Parser `SqlParser`, this can be an identifier or a simplified path expression
                     // consisting of only literal path steps (and with no wildcard or unpivot operators).
                     // Note: partiql_ast uses the `expr` sum type for this, which is too broad.  We're not
                     // changing that at this time because `partiql_ast` is established public API.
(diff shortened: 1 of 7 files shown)


lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -697,14 +697,11 @@ may then be further optimized by selecting better implementations of each operat
...
                     // The target is an expression that is indicates the table whose data is to be manipulated.
-                    // In practice this can be an identifier or a simplified path expression consisting of only
-                    // literal path steps (and with no wildcard or unpivot operators).  This restriction must
-                    // be enforced by the PartiQL implementation.
-                    // TODO: consider ways to reduce the domain of acceptable values here.  Of course, we it would
-                    // be possible to create a `sum` type like `expr` but containing only the valid DML targets,
-                    // but that would might necessitate duplicating some logic related to variable resolution which is
-                    // already implemented for standard expressions.  This is out of scope for current efforts.
-                    target::expr
+                    // According to the PartiQL's grammar this can be an identifier or a simplified path expression
```
am357, line 700: By PartiQL grammar do you mean what we have in [type-domain](https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/resources/org/partiql/type-domains/partiql.ion#L376)? If so, can you please clarify? Also maybe it's worthwhile to refer to the RFC PR here as a temporary reference.

am357 (commented):
LGTM left a clarification comment before approval.

lang/resources/org/partiql/type-domains/partiql.ion
dlurton: No, I mean what's supported by `SqlParser` in this project.  I guess, given your DML rfc, we need to start thinking about how your INSERT RFC differs from the reference implementation.

am357: That's a fair call, in that case, can we be more explicit in the text b/c that can cause confusion:

```suggestion
                    // With current PartiQL Parser `SqlParser`, this can be an identifier or a simplified path expression
```

am357 (approved):

PR #685, opened 2022-07-26
Added OTS CAST nodes & refactored CAST compiling

lziq:
*Purpose of PR:*
- Refactor scalar type AST nodes. Originally, each scalar type has a corresponding type AST node defined in the `partiql.ion` file. Now, we use one common AST node to model all of them. 

*Note:*
1. This PR still couples scalar types with compiler. 
2. This PR introduces backward incompatible changes. 
3. As of now, we assume a type parameter is an Ion literal.


lang/src/org/partiql/lang/util/CastToCastOts.kt
```diff
@@ -0,0 +1,185 @@
...
+
+import com.amazon.ionelement.api.ionNull
+import org.partiql.lang.domains.PartiqlAst
+import org.partiql.lang.domains.PartiqlPhysical
+import org.partiql.pig.runtime.LongPrimitive
+
+/**
+ * Conventional type IDs
+ */
+class TypeId {
```
lziq, line 11: The types defined will be gradually pulled out later as we pursue the pluggability goal of OTS.

vgapeyev (approved):

PR #676, opened 2022-07-19
Add `FilterScanToKeyLookup` pass and enhance integration tests to show how rewrites and custom operators can be used to optimize queries

dlurton:
## Superceded by #684 and #686 

*Description of changes:*

This PR enhances the planner integration tests introduced in #675  to show how a real-life database system would utilize physical planner passes and custom operators to optimize queries that include all of a table's primary key fields in the `WHERE` predicate.

This PR introduces the rewrite `FilterScanToKeyLookup`, which has been made generic enough to be included in the PartiQL library by default.  Note however that it is not enabled by default.  

For examle, this filter applies a predicate to to a full scan of the table `foo`:

```
(filter 
    (impl ...)
    <expr>
    (scan (impl ...) (global_id uuid_for_foo) varDecl(0)
)
```

If `<expr>` encompasses all of the `foo` table's primary key fields (in the form of `f.keyField1 = <expr-1> [and f.keyFieldN = <expr-n>]`), then the above is rewritten to an optimized form, which looks a record up from the underlying data store using its key:

```
(project
    (impl custom_get_by_key uuid_for_foo)
    (list <pk-field-value-expr_1> [<pk-field-value-expr_n>]...)
```

---

*Have you updated the `Unreleased` section of `CHANGELOG.md` with your changes? (y/n), If not, please explain why:*

TODO

*Does your PR include any backward-incompatible changes? (y/n), if yes, please explain the reason. In addition, please
 also mention any other alternatives you've considered and the reason they've been discarded?:*

No backward incompatible changes.

```
For this purpose, we define backward-incompatible changes as changes that—when consumed—can potentially result in 
errors for users that are using our public APIs or the entities that have `public` visibility in our code-base.
```

*Does your PR introduce a new external dependency? (y/n), if yes, please explain the reason. In addition, please
also mention any other alternatives you've considered and the reason they've been discarded?:*

No new dependencies.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #686, opened 2022-07-26
Refactor/improve planner integration tests

dlurton:
Originally part of #676  which was too big.   Also this PR builds on the other PR that was split from that larger PR, #684, which must be merged before this PR.

*Description of changes:*

Refactors and improves the query planner integration tests.  Namely, this includes the new `FilterScanToKeyLookup` optimization, but also renames a few things to make them more generic--the integration tests are not only for DML anymore.

No, because this only modifies stuff previously discussed in the file.

No backward incompatible API changes.

No new dependencies.


lang/test/org/partiql/lang/planner/memorydb/MemoryDatabase.kt
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * Locates a table's schema by name, with optional case-insensitivity.
+     *
+     * Returns `null` if the table doesn't exist.
+     */
+    fun findTableMetadata(bindingName: BindingName): TableMetadata? =
+        tables.firstOrNull { bindingName.isEquivalentTo(it.metadata.name) }?.metadata
+
+    /**
+     * Returns a table's metadata, given it's UUID. If no table with the given UUID exists, an an exception is
```
alancai98, line 31:
```suggestion
     * Returns a table's metadata, given its UUID. If no table with the given UUID exists, an exception is
```

lang/test/org/partiql/lang/planner/memorydb/QueryEngine.kt
```diff
@@ -0,0 +1,211 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+
+private fun PartiqlPhysical.Plan.toPrettyString(): String =
+    SexpAstPrettyPrinter.format(toIonElement().asAnyElement().toIonValue(ION))
```
alancai98, line 211: Unused function here

lang/test/org/partiql/lang/planner/memorydb/QueryEngine.kt
```diff
@@ -0,0 +1,211 @@
...
+import java.util.UUID
+
+// The name of the database in the context variable.
+internal const val DB_CONTEXT_VAR = "in-memory-database"
+
+/**
+ * This class is a demonstration of how to integrate a storage layer with PartiQL's query planner.
+ */
+class QueryEngine(val db: MemoryDatabase) {
+    var enableDebugOutput = false
```
alancai98, line 37: nit: could include `enableDebugOutput` as an argument to the function and make it immutable since it's not changed in the existing code

lang/test/org/partiql/lang/planner/memorydb/QueryEngine.kt
```diff
@@ -0,0 +1,211 @@
...
+        override fun get(bindingName: BindingName): ExprValue {
+            // TODO: PartiQL may need some additional cleanup here because, perhaps very confusingly, the bindingName
+            // passed here contains the UUID of the table and *not* its name.  It should also always specify a
+            // case-sensitive binding name.  Really, we should reconsider if [PlannerPipeline] should use
+            // [Bindings<T>] at all, perhaps another interface that's more narrow should be used instead.
+            // Another difference in how PlannerPipeline uses Bindings<ExprValue> (and argument for using a new
+            // interface) is that the bindingName here is guaranteed to be valid because otherwise planning would have
+            // been aborted.  If the lookup fails for some reason it would mean the plan is invalid.  But with
+            // the [CompilerPipeline] these checks don't exist and therefore the [Bindings<T>] implementation is
+            // expected to throw if the variable does not exist.
```
alancai98, line 81: Does this cleanup apply to PartiQL as a whole or just the planner's use of `BindingName`? If it applies to PartiQL as a whole, it may be better to capture this comment in a GitHub issue rather than buried in a TODO of a test class.

alancai98 (commented):
Left some nits and a question about whether a TODO should be made more visible through a GitHub issue. Otherwise, the refactor looks reasonable to me.

lang/test/org/partiql/lang/planner/memorydb/QueryEngine.kt
dlurton: I would prefer to leave this public and mutable since it spits out hundreds of lines per query its mutability allows me to enable it only for one query out of a sequence.

lang/test/org/partiql/lang/planner/memorydb/QueryEngine.kt
dlurton: I should probably be more specific in this comment--it applies to only `PlannerPipeline` and not the `CompilerPipeline`.

dlurton: Will reword.

lang/test/org/partiql/lang/planner/memorydb/MemoryDatabase.kt
dlurton: fixed

lang/test/org/partiql/lang/planner/memorydb/QueryEngine.kt
dlurton: Removing.

alancai98 (approved):

PR #687, opened 2022-07-26
Refactors bag operators to single IR node

RCHowell:
*Description of changes:*

Simplification of our IR by combining set operators. This reduces our code and will simplify the evaluator and planning compiler. I think we need to revisit our AST and Query pretty printers. They are _very_ verbose and are quite a headache.

Yes. This PR is not backwards incompatible because it removes the `union`, `intersect`, and `except` IR nodes. These nodes have been replaced with the `bag_op` node which is parameterized on the operator type. This choice was made to simplify the IR and remove duplicated logic.

```
```


lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -126,13 +126,8 @@ may then be further optimized by selecting better implementations of each operat
             (date year::int month::int day::int)
             (lit_time value::time_value)
 
-            // Set operators
-            (union setq::set_quantifier operands::(* expr 2))
-            (except setq::set_quantifier operands::(* expr 2))
-            (intersect setq::set_quantifier operands::(* expr 2))
-            (outer_union setq::set_quantifier operands::(* expr 2))
-            (outer_except setq::set_quantifier operands::(* expr 2))
-            (outer_intersect setq::set_quantifier operands::(* expr 2))
```
alancai98, line 135: I think these changes are backwards-incompatible and warrants calling out in the `CHANGELOG.MD` since users using the "unrefactored" AST may need to change code on their end.

Also will need to update the PR description accordingly.

alancai98 (commented):
Left a nit related to a comment and a comment that this PR's change is backwards-incompatible and requires an update to the `CHANGELOG.md` (and PR description).

alancai98 (approved):

PR #690, opened 2022-07-27
Adds OUTER bag operators [RFC-0007]

RCHowell:
*Issue #, if available:*

https://github.com/partiql/partiql-lang-kotlin/issues/184

*Description of changes:*

This PR implements OUTER bag operators. These operators are the multi-set operators we know and love from math. The full specification is defined in [RFC-0007 ](https://github.com/partiql/partiql-docs/blob/main/RFCs/0007-rfc-bag-operators.md)

*Have you updated the `Unreleased` section of `CHANGELOG.md` with your changes? (y/n), If not, please explain why:* N

I will update CHANGELOG once set operators are added. This is just the OUTER operators.

 also mention any other alternatives you've considered and the reason they've been discarded?:* N

```
```

also mention any other alternatives you've considered and the reason they've been discarded?:* N

### DIFF
--- a/lang/src/org/partiql/lang/eval/BagOp.kt
+++ b/lang/src/org/partiql/lang/eval/BagOp.kt
@@ -4,7 +4,9 @@ import com.amazon.ion.IonContainer
 import com.amazon.ion.IonStruct
 import com.amazon.ionelement.api.MetaContainer
 import org.partiql.lang.domains.PartiqlAst
+import org.partiql.lang.domains.PartiqlPhysical
 import org.partiql.lang.errors.ErrorCode
+import org.partiql.pig.runtime.DomainNode
 
 fun interface ExprValueBagOp {
     fun eval(lhs: ExprValue, rhs: ExprValue): Sequence<ExprValue> {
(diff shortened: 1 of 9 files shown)


dlurton (commented):
What is the plan to update the physical plan's `PhysicalExprToThunkConverterImpl`, which is the planner's equivalent of `EvaluatingCompiler`?  I would think that new features should at least be added to `PhysicalExprToThunkConverterImpl`, and that `EvaluatingCompiler` should only be modified if there's a customer which cannot migrate to the new planner for some reason.

RCHowell:
The plan is either
1. Hold off on adding to planning compiler until we have a formalization of PartiQL's relation algebra.
2. Add to existing planner, but inevitably refactor.

The planner is subject to change after a formalization of PartiQL's relational algebra. These operators touch on a formally undefined area of relational/binding expr vs a value expr. The OUTER operators work with expr value whereas the SQL conformant bag operators would be a bexpr.

I can add these to the planner tomorrow if you want, but I'm a bit more interested in revisiting the planner design before adding more to it.

RCHowell:
Unfortunately I had just joined the team when the planner PR was created, so I hadn't had a chance to provide feedback. I can take a guess at some design decisions, which I should clarify with you, but I am interested in simplifying what was merged while its still experimental

dlurton:
My concern with not implementing in the `PlannerPipeline` for (value expr) is that if you don't, deprecating the CompilerPipeline and `EvaluatingCompiler` will in favor of PlannerPipeline will be unnecessarily hard.  I don't know if it even makes sense to implement these operators for `bexpr`, but that is a decision we can make later.  In the mean time, I don't want to make the existing feature disparity between `CompilerPipeline` supported features and `PlannerPipeline` supported features any worse.

am357 (commented):
Overall LGTM—added a comment about `kdoc`.

In addition, since adding the feature to `Planner` is minimal effort in this case, I don't see an issue w/ merging this change; for future evaluator changes we've already considered a conversation to define our approach moving forward for updating `EvaluatingCompiler` and/or `Planner`. 

Sice

lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
```diff
@@ -684,3 +685,23 @@ internal fun createUniqueExprValueFilter(): (ExprValue) -> Boolean {
...
+    }
+}
+
+fun Sequence<ExprValue>.multiplicities(): TreeMap<ExprValue, Int> {
+    val multiplicities: TreeMap<ExprValue, Int> = TreeMap(DEFAULT_COMPARATOR)
+    this.forEach {
+        multiplicities.compute(it) { _, v -> (v ?: 0) + 1 }
+    }
+    return multiplicities
+}
```
alancai98, line 707: Are these intentionally public? If so, may need some kdoc describing these extension functions.

lang/src/org/partiql/lang/eval/BagOp.kt
```diff
@@ -0,0 +1,113 @@
...
+
+import com.amazon.ion.IonContainer
+import com.amazon.ion.IonStruct
+import com.amazon.ionelement.api.MetaContainer
+import org.partiql.lang.domains.PartiqlAst
+import org.partiql.lang.domains.PartiqlPhysical
+import org.partiql.lang.errors.ErrorCode
+import org.partiql.pig.runtime.DomainNode
+
+fun interface ExprValueBagOp {
```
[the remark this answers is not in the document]
RCHowell: I'm not sure this needs a kdoc since it's not a public API, but I've included it. It can be an internal class, but the package structure of `eval` is non-existent. Once refactored, this will cease to exist or be moved depending on the expr abstraction. Kind of redundant and cluttering to define such obvious comments on abstractions, but the RFC link is worthwhile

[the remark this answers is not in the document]
am357: Fair point, thanks for explaining.

am357 (approved):

lang/test/org/partiql/lang/eval/EvaluatorTestSuite.kt
```diff
@@ -1398,4 +1415,16 @@ internal val EVALUATOR_TEST_SUITE: IonResultTestSuite = defineTestSuite {
...
     }
+    group("bagOperators") {
+        test("outerUnionDistinct", "<< 1, 2, 2, 3, 3, 3 >> OUTER UNION << 1, 2, 3, 3 >>", "$partiql_bag::[1, 2, 3]")
+        test("outerUnionAll", "<< 1, 2, 2, 3, 3, 3 >> OUTER UNION ALL << 1, 2, 3, 3 >>", "$partiql_bag::[1, 2, 2, 3, 3, 3]")
+        test("outerIntersectDistinct", "<< 1, 2, 2, 3, 3, 3 >> OUTER INTERSECT << 1, 2, 3, 3 >>", "$partiql_bag::[1, 2, 3]")
+        test("outerIntersectAll", "<< 1, 2, 2, 3, 3, 3 >> OUTER INTERSECT ALL << 1, 2, 3, 3 >>", "$partiql_bag::[1, 2, 3, 3]")
+        test("outerExceptDistinct", "<< 1, 1, 1, 2 >> OUTER EXCEPT << 1 >>", "$partiql_bag::[1, 2]")
+        test("outerExceptAll", "<< 1, 1, 1, 2 >> OUTER EXCEPT ALL << 1 >>", "$partiql_bag::[1, 1, 2]")
+        test("outerUnionCoerceScalar", "1 OUTER UNION 2", "$partiql_bag::[1, 2]")
+        test("outerUnionCoerceStruct", "{'a': 1} OUTER UNION {'b': 2}", "$partiql_bag::[ {'a': 1}, {'b': 2} ]")
```
alancai98, line 1426: Could add a couple more test cases:
- OUTER UNION w/ struct of same key but different value (i.e.` {'a': 1} OUTER UNION {'a': 2}`)
- Using Ion constructors

alancai98 (commented):
Looks good. Implementation looks very clean w/ how you've factored it. Left a couple minor comments.

lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
RCHowell: `distinct()` yes, but `multiplicities()` could be private in `ExprValBagOp`. I can do the little refactor in a different PR

Issue #513, opened 2022-01-26
Negation of the minimum INT8 value allows overflow without exception


alancai98:
This issue is referenced near a commented-out test case for unary negation in `EvaluatingCompilerNAryIntOverflowTests`.

This expression: 

```
-(-9223372036854775808)
```

...should result in an overflow exception or `MISSING` value. 

Instead, the result is returned without the negation applied `-9223372036854775808`.  This is because on the JVM, negating `Long.MIN_VALUE` overflows and the result is also `Long.MIN_VALUE`.

This currently only true for the `INT8` and "unconstrained" `INT` types since the result of negating the integer types `INT2` and `INT4` can be checked for overflow *after* the negation has happened.  

However, for 8-byte integers, the overflow occurs before the check and therefore cannot be detected by the same overflow checking.

Commented out test can be found in `EvaluatingCompilerNAryIntOverflowTests`:
https://github.com/partiql/partiql-lang-kotlin/blob/fb7992468158d947557dc979f48a5f41ce83aa43/lang/test/org/partiql/lang/eval/EvaluatingCompilerNAryIntOverflowTests.kt#L243-L244

PR #691, opened 2022-07-29
Fix neg overflow

yliuuuu:
*Issue #, if available:* [#513](https://github.com/partiql/partiql-lang-kotlin/issues/513)

*Description of changes:*
1. fix the bug `neg(LONG.min_value) = LONG.min_value`
2. uncomment the test case
 https://github.com/partiql/partiql-lang-kotlin/blob/cba305370b8cf78772c77808056cecec7a898620/lang/test/org/partiql/lang/eval/EvaluatingCompilerNAryIntOverflowTests.kt#L222
3. fix some typos

Yes. 

```
```

Yes, because now `-(-9223372036854775808)` will throw an error, and before it returns itself.

No.

### DIFF
--- a/lang/src/org/partiql/lang/util/NumberExtensions.kt
+++ b/lang/src/org/partiql/lang/util/NumberExtensions.kt
@@ -111,7 +111,8 @@ fun Number.ionValue(ion: IonSystem): IonValue = when (this) {
 
 operator fun Number.unaryMinus(): Number {
     return when (this) {
-        is Long -> -this
+        // - LONG.MIN_VALUE will result in LONG.MIN_VALUE in JVM because LONG is a signed two's-complement integers
+        is Long -> if (this == Long.MIN_VALUE) BigInteger.valueOf(Long.MAX_VALUE).add(BigInteger.ONE) else -this
         is BigInteger -> this.negate()
         is Double -> -this
         is BigDecimal -> if (this.isZero()) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ stage in the `PlannerPipeline` and to generate performance metrics for the indiv
 ### Fixed
 - Codecov report uploads in GitHub Actions workflow
 - GitHub Actions capability to run on forks
+- Negation overflow caused by minimum INT8
 
 ### Removed
 - README.md badge for travisci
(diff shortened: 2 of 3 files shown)


RCHowell (approved):
Nice!

alancai98 (approved):
Thanks for the fix!

lang/src/org/partiql/lang/util/NumberExtensions.kt
```diff
@@ -111,7 +111,8 @@ fun Number.ionValue(ion: IonSystem): IonValue = when (this) {
 
 operator fun Number.unaryMinus(): Number {
     return when (this) {
-        is Long -> -this
+        // - LONG.MIN_VALUE will result in LONG.MIN_VALUE in JVM because LONG is a signed two's-complement integers
+        is Long -> if (this == Long.MIN_VALUE) BigInteger.valueOf(Long.MAX_VALUE).add(BigInteger.ONE) else -this
```
vgapeyev, line 115: Please be kind to educate me here! 
 
For negating the Long.MIN_VALUE, this returns an integer that is 1 more than is representable for Long, so the result is of type BigInteger, not Long as was the input.  (Smells fishy, but ok). 

But that's a number as the result.  How does then the newly-uncommented test on Line 222 manage to succeed?  It seems to expect MISSING, not a proper number!

am357: @vgapeyev hope I've understood your question correctly: since this is being tested under _permissive_ mode upon receiving [the exception](https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt#L175) `Thunk` returns the _MISSING_ value, see [here](https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/src/org/partiql/lang/eval/Thunk.kt#L708). For the other point regarding BigInt unfortunately, I haven't been able to review this PR so can't really comment on that at the moment.

yliuuuu: Also, exception, if my understanding is correct, happens during "evaluation time". 

https://github.com/partiql/partiql-lang-kotlin/blob/0b4540f474caff7dba1d6c327e0e85012402829b/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt#L550-L575

vgapeyev: Yep, the permissive mode would explain this, thanks!

PR #696, opened 2022-08-02
fix Type Inferencer arithmetic operations error

### LINKED ISSUE #695: resolveIntConstraint causes Error when StaticType is mixed with IntType and non-IntType
Problem: 
When evaluating with `Type Inferencer`,
`SELECT VALUE v + 1 FROM <<1, 2.0, 3e0, 4, 5d0 >> AS v` would failed with Error Message: 

`
Caused by: org.partiql.lang.eval.EvaluationException: Internal error, The expression's static type was supposed to be INT but instead it was DECIMALThis may indicate the presence of a bug in the type inferencer.
`

For details on current Implementation, [check here](https://github.com/partiql/partiql-lang-kotlin/blob/13d434178c20374629f61e96304cf7e66ed76b9d/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt#L530-L575).

Root Cause: 
The current implementation for function `resolveIntConstraint` will check if the given `ExprValue` is a Int in the given type if its StaticType can **potentially** be `Int`, meaning if the function will be called even if the `ExprValue` is a non-Int value if the given operation has a staticType that contains `IntType`. 

Consider the above example: 
`v+1` has a staticType of `unionOf(INT, Decimal, Float)`. 
When `v` is bind to `2.0`, the arithmetic operation's result will be a decimal, invoking IntConstraint check causes the error.

DoD:
- Add test cases to EvaluatorTestSuit. 
- Fix the logic in function resolveIntConstraint.

yliuuuu:
*Issue #, if available:* Closes #695

*Description of changes:* 
1. Change the logic of function `resolveIntConstraint`. 
- If StaticType only contains IntType, check int constraint,
- If no IntType in StaticType, bypass the check. 
- If StaticType contains a mixture of IntType and non IntType, check int constraint if the result is Int(runtime)
 2. Add additional Test Cases in `EvaluatorTestSuit`

Yes. 

```
```

No. 

No.


lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -548,26 +548,52 @@ internal class EvaluatingCompiler(
...
+                            )
+                        }
+                        when (nonIntType) {
+                            // only intType exists
+                            null -> {
+                                // biggestIntegerType can not be null in this case
+                                val validator = integerValueValidator(biggestIntegerType!!.rangeConstraint.validRange)
                                 thunkFactory.thunkEnv(metas) { env ->
                                     val naryResult = computeThunk(env)
-                                    errorSignaler.errorIf(
```
alancai98, line 560: I feel like the fix can be much simpler than what's currently in this PR. If I understand the original issue, this `errorSignaler` is giving an error when the resulting evaluation type is a non-Int (e.g. decimal) when the `validator` expects an Int. A simple fix should be to case on when the resulting `naryResult`'s type is an Int:

```Kotlin
when (naryResult.type) {
    ExprValueType.INT -> errorSignaler.errorIf(
        !validator(naryResult),
        ErrorCode.EVALUATOR_INTEGER_OVERFLOW,
        { ErrorDetails(metas, "Integer overflow", errorContextFrom(metas)) },
        { naryResult }
    )
    else -> naryResult
}
```

lang/test/org/partiql/lang/eval/EvaluatorTestSuite.kt
```diff
@@ -1398,4 +1398,38 @@ internal val EVALUATOR_TEST_SUITE: IonResultTestSuite = defineTestSuite {
             "$partiql_bag::[[1968, 4, 3, 12, 31, 59]]"
         )
     }
+    group("arithmetic with mixed type") {
```
alancai98, line 1401: Looking at the other uses of `resolveIntConstraint`, could add tests cases for unary plus, unary minus, modulo, and function calls.

CHANGELOG.md
```diff
@@ -64,6 +64,7 @@ stage in the `PlannerPipeline` and to generate performance metrics for the indiv
 - Codecov report uploads in GitHub Actions workflow
 - GitHub Actions capability to run on forks
 - Negation overflow caused by minimum INT8
+- Type Inferencer Internal error caused by arithmetic operations with mixed types.
```
alancai98, line 67: nit: the error is more so related to the evaluator's integer overflow checking not the type inferencer

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
yliuuuu: If I understand you statement correctly, you are suggesting that the logic should be somewhat like below:
1. If no `IntType` in StaticType, return the computeThunk directly.
2. If there is at least one `IntType` in StaticType, then we first check if the `naryResult` is Int, if so, we run the validator, if not we return `naryResult` directly.  

If my understanding is correct, then there could potentially be an edge case, in which case the StaticType contains only `IntType`, and the result is somehow `Decimal`, which truly indicate an error in Type Inferencer. 

The slightly more complicated logic is then used to catch this scenario. 

Let me know if I am overcomplicating the problem here. If so, more than happy to simplify the logic.

alancai98: Ok, it wasn't clear to me of that edge case (i.e. when the inferencer expects just `IntType`s and the evaluation result isn't an Int) from the original PR. I think using the `integerValueValidator` to also check for the static type matching the evaluator behavior is confusing. Perhaps the following factoring is more clear:

```Kotlin
when (biggestIntegerType) {
    is IntType -> {
        val validator = integerValueValidator(biggestIntegerType.rangeConstraint.validRange)
        thunkFactory.thunkEnv(metas) { env ->
            val naryResult = computeThunk(env)
            when (naryResult.type) {
                ExprValueType.INT, ExprValueType.MISSING, ExprValueType.NULL  -> errorSignaler.errorIf(
                    !validator(naryResult),
                    ErrorCode.EVALUATOR_INTEGER_OVERFLOW,
                    { ErrorDetails(metas, "Integer overflow", errorContextFrom(metas)) },
                    { naryResult }
                )
                else -> {
                    if (staticTypes.all { it is IntType }) {
                        error("...") // <- here use the error that was previously part of `integerValueValidator`
                    } else {
                        naryResult
                    }
                }
            }
        }
    }
    // If there is no IntType StaticType, can't validate the integer size either.
    null -> computeThunk
    else -> computeThunk
}
```

Then for `integerValueValidator`, provide a different error such as saying the function only validates int, missing, and null types.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -538,22 +537,38 @@ internal class EvaluatingCompiler(
...
-
                                 thunkFactory.thunkEnv(metas) { env ->
                                     val naryResult = computeThunk(env)
-                                    errorSignaler.errorIf(
-                                        !validator(naryResult),
-                                        ErrorCode.EVALUATOR_INTEGER_OVERFLOW,
-                                        { ErrorDetails(metas, "Integer overflow", errorContextFrom(metas)) },
-                                        { naryResult }
-                                    )
+                                    // validation shall only happen when the result is INT/MISS/NULL
```
alancai98, line 551:
```suggestion
                                    // validation shall only happen when the result is INT/MISSING/NULL
```

alancai98 (dismissed):
Fix looks good. Spotted one minor typo

alancai98 (approved):

PR #699, opened 2022-08-04
Type inferencer call agg

### LINKED ISSUE #502: [inferencer] Add support for `CallAgg` to static type inferencer
`CallAgg` (calls to aggregation functions) is currently unimplemented in the static type inferencer.

See these failing tests in `EvaluatorStaticTypeTests`:
https://github.com/partiql/partiql-lang-kotlin/blob/fb7992468158d947557dc979f48a5f41ce83aa43/lang/test/org/partiql/lang/eval/EvaluatorStaticTypeTests.kt#L174-L177

yliuuuu:
*Issue #, if available:* Closes #502

*Description of changes:*
- Add support for CallAgg In Type Inferencer.
- Remove 23 aggregate related tests from `FAILING_TESTS` in EvaluatorStaticTypeTests. 

Yes.

```
```
No.

No.


lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -1464,7 +1552,9 @@ internal class StaticTypeInferenceVisitorTransform(
             var currentType = path.root.getStaticType()
             val newComponents = path.steps.map { pathComponent ->
                 currentType = when (pathComponent) {
-                    is PartiqlAst.PathStep.PathExpr -> inferPathComponentExprType(currentType, pathComponent)
+                    is PartiqlAst.PathStep.PathExpr -> {
+                        inferPathComponentExprType(currentType, pathComponent)
+                    }
```
yliuuuu, line 1557:
```suggestion
                    is PartiqlAst.PathStep.PathExpr -> inferPathComponentExprType(currentType, pathComponent)
```

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -524,6 +525,93 @@ internal class StaticTypeInferenceVisitorTransform(
...
+                it.isSubTypeOf(StaticType.unionOf(StaticType.MISSING, StaticType.NULL, StaticType.NUMERIC))
+            }
+            return when (aggFunc.text) {
+                // current implementation of count will always return a long
+                "count" -> StaticType.INT8
+                // max/min supports all type and the result depends on comparison.
+                "max", "min" -> argType
+                // current implementation of avg always return a decimal.
+                "avg" -> {
+                    if (argTypeCheck(argType)) StaticType.DECIMAL else error("Aggregate function AVG has incompatible input type")
```
alancai98, line 554: Don't think it always returns a decimal. E.g.:

```
PartiQL> AVG([1, 1]);
AVG([1, 1])==='
1
---
OK!
```

Perhaps it should return the union of all number types?

lang/test/org/partiql/lang/eval/EvaluatorStaticTypeTests.kt
```diff
@@ -137,28 +137,6 @@ class EvaluatorStaticTypeTests {
             "pathUnpivotEmptyStruct3",
             "projectOfUnpivotPath",
 
-            // Aggregates not supported by STIR
```
alancai98, line 140: While these tests are now succeeding, also need some tests in [StaticTypeInferenceVisitorTransformTest.kt](https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/test/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransformTest.kt).

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -524,6 +525,93 @@ internal class StaticTypeInferenceVisitorTransform(
...
+                }
+                "sum" -> {
```
alancai98, line 556: For computing `SUM`'s return type, could we reuse the `computeBinaryArithmeticResultType` function? Or is the logic slightly different?

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -524,6 +525,93 @@ internal class StaticTypeInferenceVisitorTransform(
...
+                                    error(
+                                        "Internal Error: SUM function only support Number Type. " +
+                                            "This probably indicates an bug in type inferencer."
+                                    )
+                            }
+                        }
+                    } else {
+                        error("Aggregate function SUM has incompatible input type")
+                    }
+                }
```
alancai98, line 609: For the data type 'error' conditions in the static type inferencer (e.g. when a non-numeric type is passed to an aggregation function that expects a numeric type), we've previously added the error to the `ProblemHandler` and defined a continuation behavior. This allows us to find other inferencing errors in the PartiQL statement rather than failing fast on the first data type mismatch.

For example, in an expression such as `'foo' + 'bar' + 'baz'`, the inferencer would identify a data type mismatch at both of the `+` operations since it expects a numeric type. We defined a continuation behavior for the `+` (and other arithmetic operations) to return the union of numeric types ([source](https://github.com/partiql/partiql-lang-kotlin/blob/d03098b5dfd86e69cdc2faf05ba0da96b61aa9d3/lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt#L701-L702)).

I think a similar logic should apply to this case for `SUM`.

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -524,6 +525,93 @@ internal class StaticTypeInferenceVisitorTransform(
...
+
+            return nAry.withStaticType(computeReturnTypeForAggFunc(funcName, argType))
+        }
+
+        private fun computeReturnTypeForAggFunc(aggFunc: SymbolPrimitive, argType: StaticType): StaticType {
+            // nested type is not supported by agg functions.
+            // i.e. sum(1,2,[3,4]) is not supported
+            val argTypeCheck: (StaticType) -> Boolean = {
+                it.isSubTypeOf(StaticType.unionOf(StaticType.MISSING, StaticType.NULL, StaticType.NUMERIC))
+            }
```
alancai98, line 546: Could be named something more specific to what the function actually checks.

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
yliuuuu: The logic is slightly different in how aggregate functions handles `MISSING` and `NULL`.

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
yliuuuu: https://github.com/partiql/partiql-lang-kotlin/blob/f96186e435f65c982372de17b6a05adbf049ab83/lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt#L243-L260

Let me know if my understanding is wrong, but I think the current implementation for `AVG` indeed always return decimal(or `null`). The `1` in you example should be equivalent to `1d0`. 

Now, perhaps in your example the query should return an integer...

alancai98: Ah I see. The decimal is not showing up in the REPL for some reason. You're correct that the return type is numeric/decimal:

https://github.com/partiql/partiql-lang-kotlin/blob/f96186e435f65c982372de17b6a05adbf049ab83/lang/test/org/partiql/lang/eval/EvaluatorTestSuite.kt#L946-L951

lang/test/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransformTest.kt
```diff
@@ -7402,6 +7407,99 @@ class StaticTypeInferenceVisitorTransformTest : VisitorTransformTestBase() {
 
             return incompatibleTypeForIndex + validIndexType
         }
+        private fun createAggFunctionValidTests(
+            functionName: String,
+            inputTypes: StaticType,
+            expectedType: StaticType,
+            isTopLevel: Boolean
```
alancai98, line 7414: Just curious does being a top level value matter for aggregation functions? When I looked at the tests, the results look the same.

lang/test/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransformTest.kt
```diff
@@ -7402,6 +7407,99 @@ class StaticTypeInferenceVisitorTransformTest : VisitorTransformTestBase() {
...
+                    handler = expectQueryOutputType(BagType(StructType(mapOf("result" to expectedType), contentClosed = true)))
+                )
+            }
+
+        @JvmStatic
+        @Suppress("unused")
+        fun parametersForAggFunctionTests() =
+            // valid tests
+            listOf(
+                // count
```
alancai98, line 7444: For COUNT, MIN, MAX, could have some tests with unknown values (e.g. `NULL`, `MISSING`, null/missing with a non-unknown type)

lang/test/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransformTest.kt
```diff
@@ -7402,6 +7407,99 @@ class StaticTypeInferenceVisitorTransformTest : VisitorTransformTestBase() {
...
+                createAggFunctionValidTests("MAX", unionOf(INT, DECIMAL, FLOAT, STRING), unionOf(INT, DECIMAL, FLOAT, STRING), true),
+                // avg
+                createAggFunctionValidTests("AVG", MISSING, NULL, false),
+                createAggFunctionValidTests("AVG", MISSING, NULL, true),
+                createAggFunctionValidTests("AVG", unionOf(MISSING, NULL), NULL, false),
+                createAggFunctionValidTests("AVG", unionOf(MISSING, NULL), NULL, true),
+                createAggFunctionValidTests("AVG", unionOf(INT, DECIMAL, FLOAT), DECIMAL, false),
+                createAggFunctionValidTests("AVG", unionOf(INT, DECIMAL, FLOAT), DECIMAL, true),
+
+                // SUM
```
alancai98, line 7461: Nice set of tests!

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -524,6 +525,127 @@ internal class StaticTypeInferenceVisitorTransform(
...
+                                else ->
+                                    // this should not be reached, only exists to segment the logic.
+                                    error(
+                                        "Internal Error: SUM function only support Number Type. " +
+                                            "This probably indicates an bug in type inferencer."
+                                    )
+                            }
+                        }
+                    }
+                    // continuation in case of type check failed
```
alancai98, line 637:
```suggestion
                    // continuation in case of data type mismatch
```

alancai98 (commented):
Changes look good. Left some nits and some additional tests that could be added.

lang/test/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransformTest.kt
```diff
@@ -7402,6 +7407,92 @@ class StaticTypeInferenceVisitorTransformTest : VisitorTransformTestBase() {
...
+        private fun createAggFunctionValidTests(
+            functionName: String,
+            inputTypes: StaticType,
+            expectedType: StaticType
+        ): TestCase =
+            // testing toplevel aggregated function
+            // testing sql function(t)
+            // global environment here is a bag type
+            TestCase(
+                name = "top level $functionName(t)",
```
alancai98, line 7419:
```suggestion
                name = "top level $functionName($inputTypes) -> $expectedType",
```

lang/test/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransformTest.kt
```diff
@@ -7402,6 +7407,92 @@ class StaticTypeInferenceVisitorTransformTest : VisitorTransformTestBase() {
...
+                createAggFunctionValidTests("SUM", unionOf(MISSING, NULL), NULL),
+                createAggFunctionValidTests("SUM", unionOf(MISSING, NULL, INT2), INT2),
+                createAggFunctionValidTests("SUM", unionOf(INT2, INT4), INT4),
+                createAggFunctionValidTests("SUM", unionOf(INT2, INT4, INT8), INT8),
+                createAggFunctionValidTests("SUM", unionOf(INT2, INT4, INT8, FLOAT), FLOAT),
+                createAggFunctionValidTests("SUM", unionOf(INT2, INT4, INT8, FLOAT, DECIMAL), DECIMAL),
+            ) +
+                // sum input type not compatible
+                TestCase(
+                    name = "SUM(t)",
```
alancai98, line 7466:
```suggestion
                    name = "data type mismatch SUM(STRING)",
```

lang/test/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransformTest.kt
```diff
@@ -7402,6 +7407,92 @@ class StaticTypeInferenceVisitorTransformTest : VisitorTransformTestBase() {
...
+                                functionName = "sum",
+                                expectedArgType = unionOf(MISSING, NULL, NUMERIC),
+                                actualType = STRING
+                            )
+                        )
+                    )
+                ) +
+                // avg input type not compatible
+                TestCase(
+                    name = "AVG(t)",
```
alancai98, line 7482:
```suggestion
                    name = "data type mismatch AVG(STRING)",
```

alancai98 (commented):
Sorry missed this in the prior review. The callAgg tests will have the same names consider changing to the suggested test names.

Otherwise looks good

alancai98 (approved):
🚢

PR #703, opened 2022-08-11
Refactor static scalar types

lziq:
*Motivation*
This PR aims to pull the definition of all the static scalar types out of `org.partiql.lang` package and use a uniform interface to define them. 

*Description of changes:*
1. Created `ScalarType` interface and use it to re-write all the static scalar types. 
2. Removed original static scalar types in `StaticType.kt` file, and created the new `StaticScalarType` as a wrapper for all the interface-defined static scalar types. 
3. Refactored code in other files so now they depend on the `StaticScalarType`-wrapped static scalar types.


lang/src/org/partiql/lang/ots/plugins/standard/types/TimeType.kt
```diff
@@ -0,0 +1,19 @@
+package org.partiql.lang.ots.plugins.standard.types
+
+import org.partiql.lang.eval.ExprValueType
+import org.partiql.lang.ots.interfaces.ScalarType
+
+data class TimeType(
+    val withTimeZone: Boolean = false
+) : ScalarType {
```
lziq, line 8: We may later separate `TIME` & `TIME_WITH_TIME_ZONE` as 2 types

lang/src/org/partiql/lang/types/StaticType.kt
```diff
@@ -11,10 +11,24 @@ import org.partiql.lang.ast.passes.inference.isUnknown
...
+import org.partiql.lang.ots.plugins.standard.types.DecimalType
+import org.partiql.lang.ots.plugins.standard.types.FloatType
+import org.partiql.lang.ots.plugins.standard.types.Int2Type
+import org.partiql.lang.ots.plugins.standard.types.Int4Type
+import org.partiql.lang.ots.plugins.standard.types.Int8Type
+import org.partiql.lang.ots.plugins.standard.types.IntType
+import org.partiql.lang.ots.plugins.standard.types.StringType
+import org.partiql.lang.ots.plugins.standard.types.SymbolType
+import org.partiql.lang.ots.plugins.standard.types.TimeStampType
+import org.partiql.lang.ots.plugins.standard.types.TimeType
```
vgapeyev, line 31: To confirm, are we on the same page that we are moving towards a design where the above references to the specific scalar types (BlobType, ..., TimeType) do not occur in this file?

lziq: Yes. the work of designing API between scalar types interfaces & compile pipeline in `org.partiql.lang` will come later

vgapeyev: I'd stress again that it's important to start doing this sooner rather than later. I believe doing that would give more insight into what the plugin APIs should be than thinking ahead what are the obstacles to doing that or looking at Trino for inspiration.

vgapeyev (approved):

PR #706, opened 2022-08-15
Pulled scalar CAST operator out

lziq:
*Description of changes:*
1. Pulled code of type inference & evaluation of CAST operator on scalar types out of `org.partiql.lang` package
2. Added scalar CAST & plugin interface
3. Added standard scalar CAST & standard plugin
4. Added `ScalarTypeSystem` as API between plugins & core. 
5. Pulled `TypedOpBehavior` out of `org.partiql.lang` package and refactored it as an option of the standard plugin.
6. Partly pulled `defaultTimeZoneOffeset` out of `org.partiql.package` and refactored it as an option of the standard plugin.

*Note*
1. Let's not worry about imports from `org.partiql.lang.util` for now, as this folder stores only constants and common util functions.
2. Let's not worry about `ots_work` package's dependencies on error handling from core. We can later setup a specific PR to deal with it. 
3. Also, Let's not worry about dependencies in test files for now.


lang/src/org/partiql/lang/CompilerPipeline.kt
```diff
@@ -77,6 +80,8 @@ typealias ProcessingStep = (PartiqlAst.Statement, StepContext) -> PartiqlAst.Sta
  * per thread.
  */
 interface CompilerPipeline {
+    val scalarTypeSystem: ScalarTypeSystem
```
lziq, line 83: We consider a scalar type system as one configuration option of the compiler pipeline.

lang/src/org/partiql/lang/ast/passes/inference/StaticTypeExtensions.kt
```diff
@@ -34,22 +33,13 @@ internal fun StaticType.isNumeric(): Boolean = this is StaticScalarType && (scal
...
 
-/**
- * Returns the maximum number of digits a decimal can hold after reserving digits for scale
- *
- * For example: The maximum value a DECIMAL(5,2) can represent is 999.99, therefore the maximum
- *  number of digits it can hold is 3 (i.e up to 999).
- */
-// TODO: What's PartiQL's max allowed precision?
-private fun DecimalType.maxDigits(parameters: TypeParameters): Int = (parameters[0] ?: Int.MAX_VALUE) - parameters[1]!!
-
```
lziq, line 45: I forgot to remove this snippet of code in the last PR

vgapeyev:
I see a new folder `ots_work`, while the old `ots` is still there, so it's not just a rename. What's the significance/purpose of this?

lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
```diff
@@ -286,350 +249,25 @@ fun ExprValue.cast(
...
-                    else -> error("Unsupported type: ${targetType.scalarType}")
-                }
-                is ListType -> if (type.isSequence) return valueFactory.newList(asSequence())
-                is SexpType -> if (type.isSequence) return valueFactory.newSexp(asSequence())
-                is BagType -> if (type.isSequence) return valueFactory.newBag(asSequence())
-                // no support for anything else
-                else -> {}
-            }
+    when (targetType) {
+        is StaticScalarType -> scalarTypeSystem.invokeCastOp(this, targetType.toCompileTimeType(), locationMeta).let {
```
vgapeyev, line 263: Here is what I expect should be happening here: 
 * Prerequisite: `ExprValue` has been refactored, similar to what you have done with `StaticType`, so that specific scalar classes like `DecimalExprValue` are now in the plugin and there is a subclass of `ExprValue` - say, `ScalarExprValue`, that is an "envelope" for their instances -- similar in role to `StaticScalarType`.
 * Then, here, `this` (which is an `ExprValue`) is to be analyzed to see if it is a `ScalarExprValue` and if so, its scalar value is extracted and *then* a call is made to a function in a plugin, with this extracted scalar value and the scalar type similarly extracted from `targetType`. 

Right now, this does make a call to the plugin, `scalarTypeSystem.invokeCastOp`, but this call does the extraction, which is too much for the plugin to handle, and it manifests in the wrong references (to core PartiQL) appearing in the plugin's interfaces -- which I complain about there. 

I recall you saying something like "we will pull scalar ExprValue into the plugin later", but it is hard for me to imagine what other meaningful refactoring work can proceed without this ExprValue work being done first. It could even be that some of the complexity in the current plugin interfaces is needlessly caused by holding this ExprValue work back.

lang/src/org/partiql/lang/ots_work/interfaces/ScalarType.kt
```diff
@@ -1,4 +1,4 @@
-package org.partiql.lang.ots.interfaces
+package org.partiql.lang.ots_work.interfaces
```
vgapeyev, line 1: Below, there are references to `ExprValueType` and `ExprValue`, which are classes in core PartiQL.  These should not be here. 

Either indicate this is a TODO, or we must discuss why not.

lang/src/org/partiql/lang/types/StaticType.kt
```diff
@@ -13,22 +13,23 @@ import org.partiql.lang.eval.ExprValueType
...
+import org.partiql.lang.ots_work.plugins.standard.types.DecimalType
+import org.partiql.lang.ots_work.plugins.standard.types.FloatType
+import org.partiql.lang.ots_work.plugins.standard.types.Int2Type
+import org.partiql.lang.ots_work.plugins.standard.types.Int4Type
+import org.partiql.lang.ots_work.plugins.standard.types.Int8Type
+import org.partiql.lang.ots_work.plugins.standard.types.IntType
+import org.partiql.lang.ots_work.plugins.standard.types.StringType
+import org.partiql.lang.ots_work.plugins.standard.types.SymbolType
+import org.partiql.lang.ots_work.plugins.standard.types.TimeStampType
+import org.partiql.lang.ots_work.plugins.standard.types.TimeType
```
vgapeyev, line 32: The above type imports, as well as constants like INT4, DATE, STRING defined on lines 65-80 below refer to things in the plugin instance (not plugin interface) and, therefore, should not be here.  

Either indicate this is a TODO, or we must discuss why not.

vgapeyev (changes requested):
This is nothing like a comprehensive review, but I think I found a spot  where the refactoring should proceed next. Without refactoring of `ExprValue`, it is hard to see this changeset as meaningful, while after some work on that, I believe, plugin interfaces and implementations would get simpler.

lang/src/org/partiql/lang/eval/physical/PhysicalExprToThunkConverterImpl.kt
```diff
@@ -144,7 +146,11 @@ internal class PhysicalExprToThunkConverterImpl(
     private val procedures: Map<String, StoredProcedure>,
     private val evaluatorOptions: EvaluatorOptions = EvaluatorOptions.standard(),
     private val bexperConverter: PhysicalBexprToThunkConverter,
+    private val scalarTypeSystem: ScalarTypeSystem
 ) : PhysicalExprToThunkConverter {
+    // TODO: remove the following hard-coded variable later
+    private val typedOpBehavior = (scalarTypeSystem.plugin as StandardPlugin).typedOpBehavior
```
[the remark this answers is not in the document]
lziq: Yeah. The `typedOpBehavior` is only used in compiling `IS`, `CAN_CAST` & `CAN_LOSSLESS_CAST` operators, and I will remove it and refactor these operators with their functionality unchanged in the next PR.

lang/src/org/partiql/lang/types/StaticType.kt
lziq: I can mark them as TODO

lang/src/org/partiql/lang/ots_work/interfaces/ScalarType.kt
lziq: This is a TODO work. We will decouple scalar run time types from `ExprValueType` later, and migrate them into the plugin interfaces package.

lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
lziq: My current opinion is that, like what you said, we need decouple scalar run-time values from `ExprValue` and decouple scalar run-time types from `ExprValueType`, and migrate them into the OTS interfaces package (probably called `ScalarExprValue` & `ScalarExprValueType`). And ideally, this block code should check if the value is a scalar value. If it is, this block of code transforms it into `ScalarExprValue` first, and then, makes a call to the `CAST` invoke defined in the standard plugin.

lziq: Does this sound like a feasible solution?

vgapeyev: Yes, we both write pretty much the same thing here.

My point is that it is logical to do this work ahead of other things.  Continuing within this PR would make more sense to me than merging it and then doing another one, because the plugin interfaces should greatly improve and seeing that within one change set will be more understandable and satisfying overall.

lziq:
Refer to https://github.com/partiql/partiql-lang-kotlin/pull/734 for the complete work

Issue #664, opened 2022-07-07
Convert Lexer/Parser into ANTLR Grammar


johnedquinn:
## Motivation

- There isn't yet a formal grammar for PartiQL
- The handwritten Lexer/Parser is difficult for developers (our customers) to understand and contribute to
- ANTLR is an industry-standard parser generator and is widely used
- ANTLR plugins are used in several IDEs
- With a specified grammar, it is easier to create language-specific tools such as syntax highlighters, auto-completers, and more.

## Goal of Changes

- Convert the existing SqlLexer and SqlParser into a generated ANTLR Parser using the ANTLR G4 files.
- Then, we need to create an ANTLR visitor to convert the ANTLR ParseTree into a PartiQL AST.
- The PartiQL ASTs for any particular query (using the new ANTLR visitor and the old SqlLexer/SqlParser) need to be identical.

## Description of Required Changes
- [x] Initialize a new branch, [antlr](/partiql/partiql-lang-kotlin/tree/antlr)
- [x] Ceate the new Lexer Grammar to produce the same tokens as SqlLexer
- [x] Pass the existing SqlLexerTests
- [x] Configure ANTLR v4.10+ to work in PartiQL (versioning, GitHub Actions, Java 8 vs 11, etc)
- [x] Create the new Parser Grammar to produce a similar AST (ANTLR ParseTree) as before
- [x] Create the ANTLR Visitor to convert the ParseTree into a PartiQL AST
- [x] Create a `PartiQLParser` that uses the generated Lexer/Parser and the ANTLR Visitor to take in a query and output the correct AST
- [x] Pass the existing SqlParserTests using the new PartiQLParser
- [x] Rework the directory structure (ANTLR vs Kotlin source files -- talk with Robert about this)
- [x] Set the default Parser to be the PartiQLParser in the CompilerPipeline
- [x] Make sure all tests across the project pass
- [ ] Deprecate the SqlLexer and SqlParser

## Relevant Links

- [PartiQL Rust Grammar](https://github.com/partiql/partiql-lang-rust/blob/main/partiql-parser/src/parse/partiql.lalrpop)
- [GPL Parser Prototype](https://github.com/OlofMorra/GQL-parser/blob/main/src/main/java/antlr/GqlParser.g4)
[...]

Issue #710, opened 2022-08-17
SqlParser needs to throw a ParserException on StackOverflow


johnedquinn:
## Description

With the nature of recursive descent parsers, it's common to hit a StackOverflowError upon parsing large queries. Similar to Trino, the SqlParser needs to catch StackOverflowErrors and throw an explicit ParserException.

## To Reproduce
1000 nested parenthesis:
```
[...]

Issue #709, opened 2022-08-17
Allowing multiple MATCHES


johnedquinn:
## Description

SqlParser should not allow the following query:

```
PartiQL> SELECT p FROM g2 MATCH (c) -> (d), (e) -> (f)
!!
```

## More Information

There should be a parenthesis around multiple matches: AKA: `((c) -> (d), (e) -> (f))`

I've fixed this in PartiQLParser (experimental in branch `antler`)

vgapeyev:
@johnedquinn, I'd think this working as described with the ANTLR parser is sufficient for closing this issue? (Assign yourself first? ;-)

johnedquinn:
Thanks !

johnedquinn:
Resolved by #711

Issue #708, opened 2022-08-17
Ambiguous Returning Clause for INSERT


johnedquinn:
## Description

The current SqlParser allows an extraneous (unused) returning clause for INSERT.

## To Reproduce
```
PartiQL> UPDATE a INSERT INTO b VALUE c RETURNING ALL NEW d RETURNING ALL NEW e
!!
```

Output:
```
(
  dml
  (
    operations
    (
      dml_op_list
      (
        insert_value
        (
          id
          b
          (
            case_insensitive
          )
          (
            unqualified
          )
        )
        (
          id
          c
          (
            case_insensitive
          )
          (
            unqualified
          )
        )
        null
        null
      )
    )
  )
  (
    from
    (
      scan
      (
        id
        a
        (
          case_insensitive
        )
        (
          unqualified
        )
      )
      null
      null
      null
    )
  )
  (
    returning
    (
      returning_expr
      (
        returning_elem
        (
          all_new
        )
        (
          returning_column
          (
            id
            e
            (
              case_insensitive
            )
            (
              unqualified
            )
          )
        )
      )
    )
  )
)
```

## More Information

Other DML Base commands do NOT allow returning clauses. Only DML commands beginning with Update/From allow them. So, when we have an insert command after an update, it chooses the LAST returning clause. Note that the output above doesn't reference column `d`.

I'm leaving a note in PartiQL Parser (experimental in branch `antlr`) that mentions this issue.

PR #711, opened 2022-08-18
Initializes the PartiQLParser using ANTLR

johnedquinn:
## Relevant Issues
- Closes Issue #664

## General Description
- My recommendation for anyone reviewing this PR is to actually pull the `antlr` branch locally and build. That way, you can more easily click around and view all changes (compare with this PR).
- The new PartiQLParser leverages ANTLR to generate a Lexer and Parser to convert a query into an ANTLR ParseTree. From there, the PartiQLParser uses an implementation of ANTLR Visitor (PartiQLVisitor) to visit each node of the ParseTree and convert it into a PartiqlAst.PartiqlAstNode.
- This PR does NOT make the PartiQLParser the default for the evaluating compiler.

## Detailed Description
- Creates a module `partiql-grammar` which currently holds the ANTLR token and grammar files. This module uses the ANTLR Gradle plugin to generate a parser.
  - Adds a dependency on ANTLR v4.10 -- which requires our project to be built with Java 11+
  - Our project currently forces the `src` and `test` directories using the root `build.gradle`, and, therefore, I needed to provide a workaround to mimic a normal project structure (which the ANTLR Gradle Plugin requires). See the root `build.gradle` for more information.
  - The tokens were grabbed from LexerConstants
- `:lang` imports `:partiql-grammar` by Gradle `implementation` to hide the generated parser. Then, it takes the G4 resource files and publishes them in `:lang`.
- Adds to the `lang` module a `PartiQLParser` and a `PartiQLVisitor`
  - The parser uses the generated ANTLR parser to convert text into a ParseTree
  - The PartiQLVisitor visits each node of ParseTree and converts to a PartiQL AST
- Adds an `AntlrUtilities` file to help convert between Antlr tokens and PartiQL Tokens. Also has some helper functions for exception throwing.
- Adds 2 new metas (for use while parsing -- see their respective Java Docs)
- Adds a CLI flag (-l, --parser) to enable the use of PartiQLParser. It also sets the PartiQLParser to be the default for the CLI.
- Catches a StackOverflowError on ParseAstStatement and throws a ParserException (related to issues of stack usage for both SqlParser and PartiQLParser).

## PR Code Breakdown
- I broke down the PR into two PRs for ease of review.
  - PR #717
  - PR #718 

## PR Quantitative Breakdown
- Approximately 2K lines added in `lang/src` (kotlin source code)
- Approximately 1.1K lines added in module `partiql-grammar` (all ANTLR tokens and grammar)
- Approximately 3.4K lines modified in `lang/test` (kotlin test code)

## Testing Breakdown
- I modified the `SqlParserTestBase` to have PartiQLParser as the default (to be used by evaluator, etc) -- but I also added `targetParsers` to each assertion test for parsing, and the default is BOTH parsers. This is to continue testing the SqlParser while it is still not deprecated.
- With the above in mind, the PartiQLParser is passing all SqlParser tests (precedence, match, general, join, datetime, etc tests)
- As mentioned above, added a `targetParsers` so that we can continue testing SqlParser as it is not being deprecated in this PR.
- I searched through all of :lang/test for 'SqlParser(' to find any instantiations of the old parser, and there is only a single test. A test dedicated to thread interrupt testing with a parseExpression function, which is not a public API for Parser.

## More Information
- Backward compatible
- Adds an ANTLR dependency
- I will squash and merge all commits

## Issues Fixed
The PartiQL Parser fixes several issues that dealt with SqlParser, namely:
- #707
- #683 
- #709
- #708 
- #730 
- #710

## Post Merge
- After merging this PR, subsequent PRs should:
  - Deprecate `SqlParser` and `SqlLexer`
  - Deprecate `LexerConstants`
  - Deprecate `SqlDataType`
  - Rename all `SqlParser*Tests.kt` to be `Parser*Tests.kt`
  - Remove all the above deprecated items


partiql-grammar/src/main/antlr/org/partiql/grammar/PartiQL.g4
```diff
@@ -0,0 +1,614 @@
...
+    | dateFunction               # ExprPrimaryBase
+    | aggregate                  # ExprPrimaryBase
+    | trimFunction               # ExprPrimaryBase
+    | functionCall               # ExprPrimaryBase
+    | nullIf                     # ExprPrimaryBase
+    | exprPrimary pathStep+      # ExprPrimaryPath
+    | caseExpr                   # ExprPrimaryBase
+    | valueList                  # ExprPrimaryBase
+    | values                     # ExprPrimaryBase
+    | exprTerm                   # ExprPrimaryBase
```
RCHowell, line 480: You do not need the alternative labels here

partiql-grammar/src/main/antlr/org/partiql/grammar/PartiQL.g4
```diff
@@ -0,0 +1,614 @@
...
+    ;
+
+nullIf
+    : NULLIF PAREN_LEFT expr COMMA expr PAREN_RIGHT;
+
+coalesce
+    : COALESCE PAREN_LEFT expr ( COMMA expr )* PAREN_RIGHT;
+
+caseExpr
+    : CASE case=expr? (WHEN when=expr THEN then=expr)+ (ELSE expr)? END;
```
RCHowell, line 505: Perhaps label the "else" expr to clarify the visitor

lang/src/org/partiql/lang/visitors/PartiQLVisitor.kt
```diff
@@ -0,0 +1,1561 @@
...
+            val whenExpr = visitExpr(ctx.expr(i))
+            val thenExpr = visitExpr(ctx.expr(i + 1))
+            exprPairList.add(exprPair(whenExpr, thenExpr))
+        }
+        val elseExpr = if (ctx.ELSE() != null) visitExpr(ctx.expr(end)) else null
+        val caseMeta = ctx.CASE().getSourceMetaContainer()
+        when (ctx.case_) {
+            null -> searchedCase(exprPairList(exprPairList), elseExpr, metas = caseMeta)
+            else -> simpleCase(visitExpr(ctx.case_), exprPairList(exprPairList), elseExpr, metas = caseMeta)
+        }
```
RCHowell, line 927: See comment in PartiQL.g4 regarding cleaning this up

partiql-grammar/src/main/antlr/org/partiql/grammar/PartiQL.g4
RCHowell: I see your skip 2, and raise you ANTLR list labels.

```antlr
caseExpr
    : CASE case=expr? (WHEN conds+=expr THEN thens+=expr)+ (ELSE else=expr)? END;
```

```kotlin
override fun visitCaseExpr(ctx: PartiQLParser.CaseExprContext) = PartiqlAst.build {
        val pairs = ctx.conds.indicies.map { i -> exprPair(
            visitExpr(ctx.conds(i)),
            visitExpr(ctx.thens(i))
        )}
        val elseExpr = if (ctx.else_ != null) visitExpr(ctx.else_) else null
        val caseMeta = ctx.CASE().getSourceMetaContainer()
        when (ctx.case_) {
            null -> searchedCase(exprPairList(exprPairList), elseExpr, metas = caseMeta)
            else -> simpleCase(visitExpr(ctx.case_), exprPairList(exprPairList), elseExpr, metas = caseMeta)
        }
}
```

partiql-grammar/src/main/antlr/org/partiql/grammar/PartiQL.g4
johnedquinn: Note the ExprPrimaryPath alternative label

partiql-grammar/build.gradle
```diff
@@ -0,0 +1,46 @@
...
+    arguments += ["-visitor", "-long-messages", "-package", "org.partiql.grammar.parser.generated", "-Xexact-output-dir" ]
+    outputDirectory = ANTLR_OUTPUT_DIR
+}
+
+// ANTLR Source
+sourceSets {
+    main {
+        antlr.srcDirs = [ANTLR_INPUT_SOURCE_ROOT]
+    }
+}
```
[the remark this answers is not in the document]
johnedquinn: You're totally right. Thanks!

cli/src/org/partiql/cli/main.kt
```diff
@@ -223,7 +242,7 @@ fun main(args: Array<String>) = try {
     exitProcess(1)
 }
 
-private fun runShell(environment: Bindings<ExprValue>, optionSet: OptionSet, compilerPipeline: CompilerPipeline) {
+private fun runShell(environment: Bindings<ExprValue>, optionSet: OptionSet, compilerPipeline: CompilerPipeline, parser: Parser) {
```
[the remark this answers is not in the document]
johnedquinn: Before, the parser was a global variable. So, we configure the compiler pipeline to use whatever instance of parser the user specifies, but the compiler pipeline sets the parser to be private.

Nothing really changes here -- the Shell() constructor took on a parser prior to this PR, but now the parser object is no longer global, so we need to allow the helper method `runShell()` to get the parser as input.

lang/src/org/partiql/lang/visitors/PartiQLVisitor.kt
```diff
@@ -0,0 +1,1561 @@
...
+
+    override fun visitDropIndex(ctx: PartiQLParser.DropIndexContext) = PartiqlAst.build {
+        val id = visitSymbolPrimitive(ctx.target)
+        val key = visitSymbolPrimitive(ctx.on)
+        dropIndex(key.toIdentifier(), id.toIdentifier())
+    }
+
+    override fun visitCreateTable(ctx: PartiQLParser.CreateTableContext) = PartiqlAst.build {
+        val name = visitSymbolPrimitive(ctx.symbolPrimitive()).name
+        createTable_(name)
```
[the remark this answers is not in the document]
johnedquinn: PIG usually adds 2 builder functions whenever there is a SymbolPrimitive. The underscored functions contain the ability to pass in the SymbolPrimitive immediately (if it's already created). The non-underscored functions take on a string, and it internally will create the symbol primitive. So, I use the underscored ones whenever I already have symbol primitives.

partiql-grammar/src/main/antlr/org/partiql/grammar/PartiQL.g4
johnedquinn: This is great. Really like this idea. Thanks !

lang/src/org/partiql/lang/visitors/PartiQLVisitor.kt
```diff
@@ -0,0 +1,1561 @@
...
+        val metas = metaContainerOf(IsImplictJoinMeta.instance)
+        when (ctx.rhs) {
+            is PartiQLParser.JoinRhsTableJoinedContext -> join(joinType, rhs, lhs, metas = metas)
+            else -> join(joinType, lhs, rhs, metas = metas)
+        }
+    }
+
+    /**
+     * Note: Similar to the old SqlParser, we have an odd condition (if the RHS is a nested join), where we flip
+     * the LHS and RHS operands.
```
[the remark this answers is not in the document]
johnedquinn: I actually think you're totally right. I was matching the existing functionality of SqlParser, but I spoke with Alan, and the existing functionality is a bug. I cut an issue #730 . Anyways, updating this.

cli/src/org/partiql/cli/main.kt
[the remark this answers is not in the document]
RCHowell: I see, this is for syntax highlighting

lang/src/org/partiql/lang/visitors/PartiQLVisitor.kt
```diff
@@ -700,34 +701,20 @@ internal class PartiQLVisitor(val ion: IonSystem, val customTypes: List<CustomTy
...
 
-    /**
-     * Note: Similar to the old SqlParser, we have an odd condition (if the RHS is a nested join), where we flip
-     * the LHS and RHS operands.
-     */
     override fun visitTableQualifiedJoin(ctx: PartiQLParser.TableQualifiedJoinContext) = PartiqlAst.build {
         val lhs = visit(ctx.lhs, PartiqlAst.FromSource::class)
         val joinType = visitJoinType(ctx.joinType())
         val rhs = visit(ctx.rhs, PartiqlAst.FromSource::class)
         val predicate = visitOrNull(ctx.joinSpec(), PartiqlAst.Expr::class)
```
RCHowell, line 716: nit: rename predicate to condition

CHANGELOG.md
```diff
@@ -32,6 +32,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Replaces `union`, `intersect`, `except` IR nodes with common `bag_op` IR node
 - Add support for CallAgg in Type Inferencer.
 - A GitHub Workflow to automatically sync the `docs` directory with the GitHub Wiki
+- Introduces the `PartiQLParser`, an implementation of `Parser` using `ANTLR`
+  - Matches the functionality of the existing `SqlParser`
+  - Now catches a StackOverflowError and throws a ParserException
+  - Support for DQL, DDL, DML, GPML, and EXEC
+  - Handles consistency and precedence issues seen in SqlParser
+    - See GitHub Issues [#709](https://github.com/partiql/partiql-lang-kotlin/issues/709), [#708](https://github.com/partiql/partiql-lang-kotlin/issues/708),
```
jpschorr, line 40: You'll want to add these issues to the PR for tracking

johnedquinn: Added to the PR description. Thx !

lang/src/org/partiql/lang/syntax/PartiQLParser.kt
```diff
@@ -0,0 +1,162 @@
...
+        parser.addErrorListener(handler)
+        return parser
+    }
+
+    /**
+     * Create a map where the key is the index of a '?' token relative to all tokens, and the value is the index of a
+     * '?' token relative to all other '?' tokens (starting at index 1). This is used for visiting.
+     * NOTE: This needs to create its own lexer. Cannot share with others due to consumption of token stream.
+     */
+    private fun getNumberOfParameters(query: String): Map<Int, Int> {
```
jpschorr, line 94: s/getNumberOfParameters/calculateTokenToParmeterOrdinals/ or similar?

lang/src/org/partiql/lang/syntax/PartiQLParser.kt
```diff
@@ -0,0 +1,162 @@
...
+        val tokens = CommonTokenStream(lexer)
+        val parser = GeneratedParser(tokens)
+        val handler = ParseErrorListener(ion)
+        parser.removeErrorListeners()
+        parser.addErrorListener(handler)
+        return parser
+    }
+
+    /**
+     * Create a map where the key is the index of a '?' token relative to all tokens, and the value is the index of a
```
jpschorr, line 90: Probably a good idea to discuss 'why' this is needed.

I.e., the visitor visits the params in 2-dimensional tree traversal order, but the ordinal of the parameters is implicit in the 1-dimensional traversal of the token stream.


It might also be possibly to "wrap" the `GeneratedLexer` in your own `TokenSource`-derivative that delegates to to all other methods, but "wraps" `nextToken` to count tokens as they are emitted by the Lexer. The downside is more complicated lifetime of shared objects, but it would only require a single lexer pass through the input.

jpschorr (dismissed):
Some minor cleanup/rename stuff that you can feel free to take or leave.

Only real substantive comment is regarding lexing the input twice in `PartiQLParser`; though even that could be handled as a follow-on issue.

johnedquinn:
Rebased to fix merge conflicts. Fixed token typos. Added Robert's suggestion. Leaving the double pass of tokenizer for a future GH issue.

am357 (approved):

PR #718, opened 2022-08-23
Initializes the PartiQLParser (Tests Only)

johnedquinn:
## Description
- Modifies tests pertatining to operator/predicate precedence (see document on PartiQL Operator Precedence)
  - Adds a section containing SqlParser-specific tests and PartiQLParser-specific tests.
- For common use-case tests, adds the ability to test both SqlParser and PartiQLParser using `targetParsers`
- Since the PartiQLParser is powered by ANTLR, ANTLR receives a syntax error and throws a ParserException (always with the error code: PARSE_UNEXPECTED_TOKEN). Therefore, for all tests testing the error code, this PR modifies the ErrorCodes to be PARSE_UNEXPECTED_TOKEN. There have been NO removals of any tests -- just a modification of existing tests to suit the PartiQLParser.
- Updated the MATCH test because USER is a SQL keyword.
- Since the parser matches tokens only -- all of the unary negative/positive tests now expect the pos and neg AST nodes.

## More Information
- This PR will not be merged. It is solely to help split up the PR into reviewable chunks. Please see #711 for approvals.
- This PR focuses on the updated test suite for the parser.
- I searched through all of `:lang/test` for 'SqlParser(' to find any instantiations of the old parser, and there is only a single test. A test dedicated to thread interrupt testing with a `parseExpression` function, which is not a public API for Parser.


lang/test/org/partiql/lang/errors/ParserErrorsTest.kt
```diff
@@ -831,10 +1427,23 @@ class ParserErrorsTest : SqlParserTestBase() {
...
+                Property.LINE_NUMBER to 1L,
+                Property.COLUMN_NUMBER to 5L,
+                Property.TOKEN_TYPE to TokenType.STAR,
+                Property.TOKEN_VALUE to ion.newSymbol("*")
+            ),
+            targetParsers = setOf(ParserTypes.PARTIQL_PARSER)
         )
     }
 
+    // FIXME: PartiQL Parser says the error location is at the parenthesis, when the real error token is `*`
```
am357, line 1446: Do we need to cut an issue for this?

lang/test/org/partiql/lang/syntax/SqlParserTestBase.kt
```diff
@@ -227,17 +247,20 @@ abstract class SqlParserTestBase : TestBase() {
...
-            try {
-                parser.parseAstStatement(input)
-                fail("Expected ParserException but there was no Exception")
-            } catch (pex: ParserException) {
-                checkErrorAndErrorContext(errorCode, pex, expectErrorContextValues)
-            } catch (ex: Exception) {
-                fail("Expected ParserException but a different exception was thrown \n\t  $ex")
+            targetParsers.forEach {
+                try {
+                    it.parser.parseAstStatement(input)
```
am357, line 257: Considering introducing the new builder, do we need to change this?

am357 (commented):
Considering the one to one addition of test that are passing for `SQLParser` (except those that are should not pass b/c of a bug), had a cursory review of the tests in general. Overall LGTM; added couple minor comments.

lang/test/org/partiql/lang/syntax/SqlParserTestBase.kt
johnedquinn: No, because we directly use the internal implementations in the tests. Then, we can call parseAstStatement directly on the implemented `Parser`s.

lang/test/org/partiql/lang/errors/ParserErrorsTest.kt
johnedquinn: Yeah, Trino actually does something interesting (that I tried to replicate). ANTLR essentially throws exceptions at the last evaluated rule, but it doesn't throw the exception in the "deepest" rule. So, Trino, upon receiving a SyntaxError, creates some state machine to try to figure out the "deepest" rule to find the most precise location where the syntax error is really coming from.

I tried to replicate their state machine, but it didn't end up working for me. Decided to also just punt it because it seemed that it would add complexity to this PR when what we really need is to understand parse errors happening as a whole.

So, yeah, I'll create an issue and reference this and the Trino implementation. Linking soon.

johnedquinn: See #731

PR #721, opened 2022-08-25
Restructures documentation and adds a GH Workflow to sync with the GH Wiki

johnedquinn:
## Description
- Restructures documentation by removing unneccessary files pertaining to pandoc, docker, etc
- Cleans up the `docs` directory
- Adds a GH Workflow to sync with the GH Wiki

## To View

These changes were tested on a personal repository. It can be viewed [here](https://github.com/johnedquinn/wiki-test/wiki).

To review this PR, I would largely look at the link above, as it holds all the same information.

This PR's `docs` directory holds the same data as the `docs` directory in the test repository. If you'd like to check that the content is the same, clone the repo at the same level as `partiql-lang-kotlin`, and go into the parent directory and run:
```shell
diff -qr partiql-lang-kotlin/docs wiki-test/docs
diff -qr partiql-lang-kotlin/.github/workflows/docs-to-gh-wiki.yml wiki-test/.github/workflows/docs-to-gh-wiki.yml
```

You will get 1 file that is different, `docs/_Sidebar.md` which holds hyperlinks to the different files. These needed to be updated to reflect the correct repository.

## To Review

Understanding that we are removing 11,000+ lines of code, I'd first go to `File changed` and minimize the directories. I'd take a good look at `.github/workflows`, `README`, `CHANGELOG`, and `docker-entrypoint.sh`. Then, IMHO, there are so many unneccessary files deleted in `docs` that it makes it difficult to understand what's going on. I would personally skim through the changes and mostly look at the linked repository wiki above.

## Other Notes
- Updated CHANGELOG
- No backward incompatibility changes
- An "external dependency" was added for the GH Workflow. See `.github/workflows/docs-to-gh-wiki.yml`
- As this is largely a PR to restructure the documentation, this PR does NOT address actual content of the documentation. It does NOT aim to update out-dated documentation.


RCHowell:
The "CLI Versions" link is broken in the sidebar

docs/_Sidebar.md
```diff
@@ -0,0 +1,26 @@
+* General
+  * [Home](https://github.com/partiql/partiql-lang-kotlin/wiki/Home)
+* Tutorials
+  * [General Tutorial](https://github.com/partiql/partiql-lang-kotlin/wiki/Tutorial)
+  * [Command Line Tutorial](https://github.com/partiql/partiql-lang-kotlin/wiki/Command-Line-Tutorial)
```
alancai98, line 5: nit: I think this doc is more of a guide/reference rather than a tutorial

docker-entrypoint.sh
```diff
@@ -1,8 +0,0 @@
-#!/bin/bash
```
alancai98, line 1: I'm fine with removing all the pandoc/docker-related documentation files since I don't recall it being used anywhere. Could we also delete the [gh-pages branch](https://github.com/partiql/partiql-lang-kotlin/tree/gh-pages) (or create an issue to track deleting it) so that it's clearer we aren't supporting the pandoc/docker-related documentation.

alancai98 (commented):
Thanks for going through this effort! Hopefully the GitHub Actions workflow can make our documentation more maintainable and usable by our users. Mostly left some minor comments.

docs/design/Abstract Syntax Tree.md
```diff
@@ -0,0 +1,1136 @@
+# Introduction to the ExprNode AST
```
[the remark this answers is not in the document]
johnedquinn: Created #723

alancai98 (approved):

PR #725, opened 2022-08-29
Update dependencies

popematt:
*Issue #, if available:*

None

*Description of changes:*

ion-element: v0.2.0 -> v1.0.0
partiql-ir-generator: v0.5.0 -> v0.6.0
partiql-ir-generator-runtime: v0.5.0 -> v0.6.0
partiql-isl-kotlin: v0.1.0 -> v0.2.0

y

y

It's a backwards incompatible dependency update to a stable release of `ion-element-kotlin`.

n

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -6,5 +6,5 @@ repositories {
 dependencies {
     // Adding this dependency here in buildSrc exposes it to all build.gradle files in this project
     // which allows PIG to be invoked by the :lang project at build time.
-    implementation 'org.partiql:partiql-ir-generator:0.5.0'
+    implementation 'org.partiql:partiql-ir-generator:0.6.0'
 }
--- a/lang/build.gradle
+++ b/lang/build.gradle
@@ -49,9 +49,9 @@ dependencies {
     implementation 'org.antlr:antlr4-runtime:4.10.1'
 
     api 'com.amazon.ion:ion-java:1.9.0'
-    api 'com.amazon.ion:ion-element:0.2.0'
-    api 'org.partiql:partiql-ir-generator-runtime:0.5.0'
-    api 'org.partiql:partiql-isl-kotlin:0.1.0'
+    api 'com.amazon.ion:ion-element:1.0.0'
+    api 'org.partiql:partiql-ir-generator-runtime:0.6.0'
+    api 'org.partiql:partiql-isl-kotlin:0.2.0'
 
     implementation 'org.apache.commons:commons-csv:1.8'
 
@@ -76,9 +76,8 @@ task generatePigDomains {
 
 
     def typeUniverse = new File(projectDir, "resources/org/partiql/type-domains/partiql.ion")
-    def targetLanguage = new TargetLanguage.Kotlin("org.partiql.lang.domains", PIG_OUTPUT_DIR)
-    def cmd = new Command.Generate(typeUniverse , targetLanguage)
-
+    def targetLanguage = new TargetLanguage.Kotlin("org.partiql.lang.domains", PIG_OUTPUT_DIR, null)
+    def cmd = new Command.Generate(typeUniverse, targetLanguage)
 
     // Tell gradle about the input and output of this task so that it only invokes pig when
     // the input file has changed or the output files are missing.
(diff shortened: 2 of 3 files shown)


buildSrc/build.gradle
```diff
@@ -6,5 +6,5 @@ repositories {
 dependencies {
     // Adding this dependency here in buildSrc exposes it to all build.gradle files in this project
     // which allows PIG to be invoked by the :lang project at build time.
-    implementation 'org.partiql:partiql-ir-generator:0.5.0'
+    implementation 'org.partiql:partiql-ir-generator:0.5.1'
```
RCHowell, line 9: Thank you for the PR. This will need to be 0.6.0

lang/build.gradle
```diff
@@ -43,8 +43,8 @@ dependencies {
     implementation 'org.jetbrains.kotlin:kotlin-reflect'
 
     api 'com.amazon.ion:ion-java:1.9.0'
-    api 'com.amazon.ion:ion-element:0.2.0'
-    api 'org.partiql:partiql-ir-generator-runtime:0.5.0'
+    api 'com.amazon.ion:ion-element:1.0.0'
+    api 'org.partiql:partiql-ir-generator-runtime:0.5.1'
```
RCHowell, line 47: 0.6.0 here as well

lang/build.gradle
```diff
@@ -113,9 +112,8 @@ runKtlintCheckOverMainSourceSet.dependsOn generatePigDomains
 
 ktlint {
     filter {
-        // excluding multiple files from ktlint was somewhat of a challenge using other approaches..
-        // We use the "workaround" described here: https://github.com/JLLeitschuh/ktlint-gradle/issues/222
-        exclude({it.file.absolutePath.contains("/generated-src/")})
+        // Excludes any code in the build directory, which is assumed to be generated.
+        exclude { projectDir.toURI().relativize(it.file.toURI()).path.startsWith("build") }
```
RCHowell, line 116: Is this the best way? Doesn't gradle have `project.buildDir` which we could use?

RCHowell: It seems weird that a linter would ever lint the build directory

RCHowell: The ignore for the generated sources might actually just be copy-paste from the PIG package, and is likely unnecessary in the PartiQL package. Ktlint should ignore the build directory, but [this same configuration was added to the pig-tests package](https://github.com/partiql/partiql-ir-generator/pull/128/files#diff-825d89dbeee13ab092df4a6b3762752aec1d25ed867a797c85a22e4bcd75e214) because that generated source in _not_ in build/

Perhaps we can get away with removing this config altogether.

RCHowell: update: the linter runs for everything in the main sourceSet, and pig sources are explicitly added hence why we need to ignore

RCHowell:
Build is failing

```
Execution failed for task ':lang:compileKotlin'.
> Error while evaluating property 'filteredArgumentsMap' of task ':lang:compileKotlin'
   > Could not resolve all files for configuration ':lang:compileClasspath'.
      > Could not resolve org.partiql:partiql-isl-kotlin:0.2.0.
        Required by:
            project :lang
         > No matching variant of org.partiql:partiql-isl-kotlin:0.2.0 was found. The consumer was configured to find an API of a library compatible with Java 8, preferably in the form of class files, preferably optimized for standard JVMs, and its dependencies declared externally, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'jvm' but:
```

popematt:
Yes, I just noticed that. It seems like `partiql-isl-kotlin` targets JDK 11 compatibility, so when Almann changed the build logic for `partiql-lang-kotlin` it had the effect of preventing us from ever using `partiql-isl-kotlin:0.2.0`. We will need another release of `partiql-isl-kotlin` to get this sorted out.

popematt:
In order to finish this PR, the changes in https://github.com/partiql/partiql-isl-kotlin/pull/12 need to be released.

alancai98:
> In order to finish this PR, the changes in https://github.com/partiql/partiql-isl-kotlin/pull/12 need to be released. 

Hey @popematt, I'm looking into doing another release. Will confirm with the team whether it will be released as `0.2.1` or `0.3.0`.

RCHowell:
Close in favor of https://github.com/partiql/partiql-lang-kotlin/pull/805

Issue #735, opened 2022-09-01
Gradle Build Fails on JDK 17


almann:
If you run a clean build (including cleaning out any pre-existing Gradle cache/config) on JDK 17, we get the following failure:

```
> Task :testscript:compileKotlin FAILED
e: java.lang.NoClassDefFoundError: Could not initialize class org.jetbrains.kotlin.com.intellij.pom.java.LanguageLevel
        at org.jetbrains.kotlin.com.intellij.core.CoreLanguageLevelProjectExtension.<init>(CoreLanguageLevelProjectExtension.java:26)
        at org.jetbrains.kotlin.com.intellij.core.JavaCoreProjectEnvironment.<init>(JavaCoreProjectEnvironment.java:42)
        at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreProjectEnvironment.<init>(KotlinCoreProjectEnvironment.kt:26)
        at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment$ProjectEnvironment.<init>(KotlinCoreEnvironment.kt:121)
        at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment$Companion.createForProduction(KotlinCoreEnvironment.kt:425)
        at org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.createCoreEnvironment(K2JVMCompiler.kt:226)
        at org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.doExecute(K2JVMCompiler.kt:152)
        at org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.doExecute(K2JVMCompiler.kt:52)
        at org.jetbrains.kotlin.cli.common.CLICompiler.execImpl(CLICompiler.kt:88)
        at org.jetbrains.kotlin.cli.common.CLICompiler.execImpl(CLICompiler.kt:44)
        at org.jetbrains.kotlin.cli.common.CLITool.exec(CLITool.kt:98)
        at org.jetbrains.kotlin.incremental.IncrementalJvmCompilerRunner.runCompiler(IncrementalJvmCompilerRunner.kt:386)
        at org.jetbrains.kotlin.incremental.IncrementalJvmCompilerRunner.runCompiler(IncrementalJvmCompilerRunner.kt:110)
        at org.jetbrains.kotlin.incremental.IncrementalCompilerRunner.compileIncrementally(IncrementalCompilerRunner.kt:286)
[...]

PR #736, opened 2022-09-02
Fixes Build on JDK17

### LINKED ISSUE #375: Migrate StatementRedactor to PIG generated visitor

almann:
Fixes #375

*Description of changes:*

* Upgrades Kotlin Gradle plugin to 1.5.31.
* Upgrades JaCoCo to 0.8.8.
* Adds language/lib target to Kotlin 1.4 to ensure compatibility.
* Adds Java 8 compatibility globally to align with Kotlin.
* Adds CI support for JDK 17 and removes non-LTS releases that are past EOL.

The CI change showing build failure with JDK 17 before this change can be seen here[^1].

*Have you updated the `Unreleased` section of `CHANGELOG.md` with your changes? (y/n), If not, please explain why:* No, this change is purely build/CI.

 also mention any other alternatives you've considered and the reason they've been discarded?:* No.

also mention any other alternatives you've considered and the reason they've been discarded?:* No.

[^1]: https://github.com/almann/partiql-lang-kotlin/runs/8145823436?check_suite_focus=true

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,7 +13,7 @@
  */
 
 plugins {
-    id 'org.jetbrains.kotlin.jvm' version '1.4.32' apply false
+    id 'org.jetbrains.kotlin.jvm' version '1.5.31' apply false
     // https://arturbosch.github.io/detekt/groovydsl.html
     id "io.gitlab.arturbosch.detekt" version "1.20.0-RC1" apply false
     id 'org.jlleitschuh.gradle.ktlint' version '10.2.1'
@@ -24,7 +24,7 @@ allprojects {
     apply plugin: 'org.jlleitschuh.gradle.ktlint'
 
     jacoco {
-        toolVersion = '0.8.5'
+        toolVersion = '0.8.8'
         reportsDirectory = file("$buildDir/reports/jacoco/")
     }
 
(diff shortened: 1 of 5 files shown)


am357 (approved):
Thanks for the dive deep—LGTM.

build.gradle
```diff
@@ -88,13 +88,17 @@ subprojects {
...
             main.java.srcDirs = ["src"]
             main.resources.srcDirs = ["resources"]
             test.java.srcDirs = ["test"]
             test.resources.srcDirs = ["test-resources"]
         }
+
+        java {
+            sourceCompatibility = JavaVersion.VERSION_1_8
+            targetCompatibility = JavaVersion.VERSION_1_8
+        }
```
almann, line 101: Without this there is a warning that we may have misalignment on `javac` produced binaries and `kotlinc`

testscript/build.gradle
```diff
@@ -16,8 +16,16 @@ dependencies {
 }
 
 compileKotlin {
-    kotlinOptions.jvmTarget = "1.8"
+    kotlinOptions {
+        jvmTarget = "1.8"
+        apiVersion = "1.4"
+        languageVersion = "1.4"
+    }
```
almann, line 23: Not sure if this should be more global--but basically every Kotlin sub-project has to explicitly set this and this change is consistent with existing package structure.

RCHowell: Once the planner refactor is closed, I will begin the build refactor. These options should be in the buildSrc like was done with the PIG refactor https://github.com/partiql/partiql-ir-generator/blob/main/buildSrc/src/main/kotlin/pig.conventions.gradle.kts#L26-L31

RCHowell: I guess that whole file shows how to fix the issue in one place and apply to all gradle subprojects via a conventions plugin

alancai98:
PR description had slight typo. Should reference https://github.com/partiql/partiql-lang-kotlin/issues/735 rather than #375.

PR #737, opened 2022-09-02
Corrected type inferencer's behavior on aggregate functions

lziq:
*Motivation:*
The type inferencer's behavior on `sum` & `avg` is not modeled correctly. e.g. the type of expression `sum(a)` with binding `'a' to list(union(int2, int4))` should be inferred as `union(int2, int4)`, instead of only `int4`. This is because with `'a' to list(union(in2, int4))` specified, it does not necessarily mean `a` must have 2 elements, with each one's type to be `int2` & `int4`, separately. It is possible that `a` only has one element with type to be either `int2` or `int4`. In this case, the return type is `union(int2, int4)`. 

The semantics of union of types is that, the type of the actual value can be one of the specified types, instead of must be all the specified types.

### DIFF
--- a/lang/src/org/partiql/lang/ast/passes/inference/StaticTypeExtensions.kt
+++ b/lang/src/org/partiql/lang/ast/passes/inference/StaticTypeExtensions.kt
@@ -24,6 +24,7 @@ internal fun StaticType.isNumeric(): Boolean = (this is IntType || this is Float
 internal fun StaticType.isText(): Boolean = (this is SymbolType || this is StringType)
 internal fun StaticType.isLob(): Boolean = (this is BlobType || this is ClobType)
 internal fun StaticType.isUnknown(): Boolean = (this.isNullOrMissing() || this == StaticType.NULL_OR_MISSING)
+internal fun StaticType.isKnown(): Boolean = !isUnknown()
 
 /**
  * Returns the maximum number of digits a decimal can hold after reserving digits for scale
--- a/lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
+++ b/lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
@@ -55,7 +55,6 @@ import org.partiql.lang.types.TypedOpParameter
 import org.partiql.lang.types.UnknownArguments
 import org.partiql.lang.types.toTypedOpParameter
 import org.partiql.lang.util.cartesianProduct
-import org.partiql.pig.runtime.SymbolPrimitive
 
 /**
  * A [PartiqlAst.VisitorTransform] that annotates nodes with their static type.
@@ -530,20 +529,18 @@ internal class StaticTypeInferenceVisitorTransform(
             val nAry = super.transformExprCallAgg(node) as PartiqlAst.Expr.CallAgg
             val funcName = nAry.funcName
             // unwrap the type if this is a collectionType
-            val argType = when (val type = nAry.arg.getStaticType()) {
-                is CollectionType -> type.elementType
-                else -> type
-            }
+            // only collection types work for agg functions, e.g. SUM(1) & SUM(NULL) return NULL
+            val type = nAry.arg.getStaticType() as? CollectionType ?: return nAry.withStaticType(StaticType.NULL)
             val sourceLocation = nAry.getStartingSourceLocationMeta()
-            return nAry.withStaticType(computeReturnTypeForAggFunc(funcName, argType, sourceLocation))
+            return nAry.withStaticType(computeReturnTypeForAggFunc(funcName.text, type.elementType, sourceLocation))
         }
 
-        fun handleInvalidInputTypeForAggFun(sourceLocation: SourceLocationMeta, aggFunc: SymbolPrimitive, actualType: StaticType, expectedType: StaticType) {
+        private fun handleInvalidInputTypeForAggFun(sourceLocation: SourceLocationMeta, funcName: String, actualType: StaticType, expectedType: StaticType) {
             problemHandler.handleProblem(
                 Problem(
                     sourceLocation = sourceLocation,
                     details = SemanticProblemDetails.InvalidArgumentTypeForFunction(
-                        functionName = aggFunc.text,
+                        functionName = funcName,
                         expectedType = expectedType,
                         actualType = actualType
                     )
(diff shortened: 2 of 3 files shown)


lang/test/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransformTest.kt
```diff
@@ -7471,7 +7473,7 @@ class StaticTypeInferenceVisitorTransformTest : VisitorTransformTestBase() {
                             createInvalidArgumentTypeForFunctionError(
                                 sourceLocation = SourceLocationMeta(1L, 1L, 3L),
                                 functionName = "sum",
-                                expectedArgType = unionOf(MISSING, NULL, NUMERIC),
+                                expectedArgType = unionOf(NULL, NUMERIC),
```
lziq, line 7476: `sum` never returns `MISSING`, e.g. `sum([MISSING])` returns `NULL`.

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -530,120 +529,74 @@ internal class StaticTypeInferenceVisitorTransform(
...
             )
         }
 
-        private fun computeReturnTypeForAggFunc(aggFunc: SymbolPrimitive, argType: StaticType, sourceLocation: SourceLocationMeta): StaticType {
-            // nested type is not supported by agg functions.
-            // i.e. sum(1,2,[3,4]) is not supported
-            val isNullOrMissingOrNumeric: (StaticType) -> Boolean = {
-                it.isSubTypeOf(StaticType.unionOf(StaticType.MISSING, StaticType.NULL, StaticType.NUMERIC))
+        private fun computeReturnTypeForAggFunc(funcName: String, elementType: StaticType, sourceLocation: SourceLocationMeta): StaticType {
+            if (funcName == "count") {
```
alancai98, line 555: nit: could keep this within the `when` expression for consistent style

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -530,120 +529,74 @@ internal class StaticTypeInferenceVisitorTransform(
...
+                return StaticType.INT
```
alancai98, line 556: Why does this now return an unconstrained `INT` rather than an `INT8`? The current evaluator returns always returns a long.

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -530,120 +529,74 @@ internal class StaticTypeInferenceVisitorTransform(
...
-                        else -> StaticType.DECIMAL
+            }
+
+            fun StaticType.isUnknownOrNumeric() = isUnknown() || isNumeric()
+
+            return when (funcName) {
+                // In case that any element is MISSING or there is no element, we should return NULL
+                "max", "min" -> StaticType.unionOf(elementTypes.convertMissingToNull())
+                "sum" -> when {
+                    elementTypes.all { !it.isUnknownOrNumeric() } -> {
```
alancai98, line 574: nit: boolean expression could be less confusing with the following:

```suggestion
                    elementTypes.none { it.isUnknownOrNumeric() } -> {
```

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -530,120 +529,74 @@ internal class StaticTypeInferenceVisitorTransform(
...
+                        handleInvalidInputTypeForAggFun(sourceLocation, funcName, elementType, StaticType.unionOf(StaticType.NULL_OR_MISSING, StaticType.NUMERIC).flatten())
+                        StaticType.unionOf(StaticType.NULL, StaticType.NUMERIC)
                     }
+                    elementTypes.any { !it.isUnknownOrNumeric() } -> StaticType.unionOf(
+                        elementTypes.filter { it.isUnknownOrNumeric() }.toMutableSet().apply { add(StaticType.MISSING) }
+                    )
+                    // In case that any element is MISSING or there is no element, we should return NULL
+                    else -> StaticType.unionOf(elementTypes.convertMissingToNull())
```
alancai98, line 582: I'm slightly confused by these two cases. What is the case on L578 describing and why is `MISSING` being added? From the comment on L571, you said the aggregation functions will return `NULL`. Is this for the case of a mistyped argument? If so, some of the comments may need adjusted.

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -530,120 +529,74 @@ internal class StaticTypeInferenceVisitorTransform(
...
-                        }
-                    }
-                    // continuation in case of data type mismatch
-                    else {
-                        val expectedType = StaticType.unionOf(StaticType.MISSING, StaticType.NULL, StaticType.NUMERIC)
-                        handleInvalidInputTypeForAggFun(sourceLocation, aggFunc, argType, expectedType)
-                        expectedType
+                // "avg" returns DECIMAL or NULL
+                "avg" -> when {
+                    elementTypes.all { !it.isUnknownOrNumeric() } -> {
```
alancai98, line 586: nit: can adjust the boolean expression to be less confusing

```suggestion
                    elementTypes.none { it.isUnknownOrNumeric() } -> {
```

lang/src/org/partiql/lang/ast/passes/inference/StaticTypeExtensions.kt
```diff
@@ -24,6 +24,7 @@ internal fun StaticType.isNumeric(): Boolean = (this is IntType || this is Float
 internal fun StaticType.isText(): Boolean = (this is SymbolType || this is StringType)
 internal fun StaticType.isLob(): Boolean = (this is BlobType || this is ClobType)
 internal fun StaticType.isUnknown(): Boolean = (this.isNullOrMissing() || this == StaticType.NULL_OR_MISSING)
+internal fun StaticType.isKnown(): Boolean = !isUnknown()
```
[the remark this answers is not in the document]
lziq: Will apply this comment in the next revision

lang/test/org/partiql/lang/eval/EvaluatorTestSuite.kt
```diff
@@ -1110,12 +1110,6 @@ internal val EVALUATOR_TEST_SUITE: IonResultTestSuite = defineTestSuite {
             """$partiql_bag::[{a: 1}, {a: 2}] """
         )
 
-        test(
-            "selectDistinctWithAggregate",
-            """SELECT SUM(DISTINCT t.a) AS a FROM `[{a:10}, {a:1}, {a:10}, {a:3}]` t""",
-            "$partiql_bag::[{a:14}]"
-        )
```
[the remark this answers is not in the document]
lziq: This test `"selectDistinctWithAggregate"` was known to be failed, https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/test/org/partiql/lang/eval/EvaluatorStaticTypeTests.kt#L25. It is fixed in this PR.

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -530,120 +529,74 @@ internal class StaticTypeInferenceVisitorTransform(
             val nAry = super.transformExprCallAgg(node) as PartiqlAst.Expr.CallAgg
             val funcName = nAry.funcName
             // unwrap the type if this is a collectionType
+            // only collection types work for agg functions, e.g. SUM(1) & SUM(NULL) return NULL
```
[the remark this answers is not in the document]
lziq: Will remove it in the next revision

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
lziq: I'd like to keep it there so these lines won't get executed: 
```
            val elementTypes = elementType.allTypes

            fun List<StaticType>.convertMissingToNull() = toMutableSet().apply {
                if (contains(StaticType.MISSING)) {
                    remove(StaticType.MISSING)
                    add(StaticType.NULL)
                }
            }

            fun StaticType.isUnknownOrNumeric() = isUnknown() || isNumeric()
```

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
lziq: In the current implementation, unconstraint integer happens to have the same constraint as a long (`INT8`), https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/src/org/partiql/lang/types/StaticType.kt#L393-L397. 

So returning long does not necessarily mean it must be `INT8`. It can be unlimited `INT` as well. And my preference is to make it return an unlimited `INT`, since the unlimited `INT` is also used elsewhere as the default integer type, e.g. the type of expression `1` is `INT`, instead of `INT8`.

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
lziq: I believe a general principle the current implementation of type inferencer is that, if by any chance there could be a data type mismatch error, we should add `MISSING` to the result types to indicate it. e.g. the expression `a + b`, with binding of `"a" to INT, "b" to unionOf(INT, STRING)`, returns `unionOf(INT, MISSING)` instead of just `INT`.

yliuuuu (commented):
Resolved concerns on unionOf type and Aggregation function offline. The PR looks good to me.

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
alancai98: Ah ok that makes more sense, I think adding a comment would be helpful for understanding this case. Something like "non-numeric, non-unknown types will result in adding MISSING to return types"

lang/test/org/partiql/lang/eval/EvaluatorTestSuite.kt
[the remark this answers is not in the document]
lziq: I think I am removing it from the failing test set, https://github.com/partiql/partiql-lang-kotlin/pull/737/files#diff-89edc33c1bfb3e3fcda6442eeb6dcf36ed57167826ad75ee3f1ca845228909e8L117. Is this the thing you are mentioning?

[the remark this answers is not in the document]
alancai98: In this PR, you removed the test itself. The `EvaluatorTestSuite` defines the set of tests used by other classes of tests (e.g. `EvaluatorStaticTypeTests`).

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
lziq: Thanks for explaining. I have no strong opinion here. I will apply this comment.

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
lziq: Will apply this comment

lang/test/org/partiql/lang/eval/EvaluatorTestSuite.kt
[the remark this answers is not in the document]
lziq: I got it now. I added it back.

alancai98 (approved):

PR #740, opened 2022-09-06
Adds projection alias support to ORDER BY clause

### LINKED ISSUE #571: Add support for using aliases created in select list in `GROUP BY` and `ORDER BY`
Thanks @alancai98, I can use `Order By` now. I also want to add alias support for `Order By` and `Group By`. Currently, following queries are not working because of missing bindings.
* `SELECT a as b FROM table ORDER BY b`
* `SELECT a as b FROM table GROUP BY b`

Do you mind if I open a new PR to support for aliases, if can implement? or Do I need to open an issue for this?

_Originally posted by @onurkuruu in https://github.com/partiql/partiql-lang-kotlin/issues/554#issuecomment-1092059878_

johnedquinn:
## Related Issues
- Closes #571

## Description
- A PR that adds support for projection aliases in ORDER BY -- it does this by rewriting the query in accordance with PartiQL Spec Section 12.5
- Revisions:
  - 1: Initially created a PR that would essentially compile the projection list twice to evaluate the ORDER BY RHS expression, but after discussing with Yingtao and Robert, I decided to make revision #2
  - 2: Make ORDER BY go AFTER projection. However -- this was *wrong* according to PartiQL Spec Sections 3.3 and 12.5.
    - A breakdown of the situation can be summarized by [this PartiQL Specification discussion](https://github.com/partiql/partiql-spec/discussions/34) where Almann chimed in.
  - 3: I rewrite the query using a Visitor Transform and introduce a Meta to help maintain idempotency. Based on the PartiQL Spec sections mentioned above.

## Good Examples

The following queries are supposed to work, and now they do.
```sql
SELECT b AS "C"
FROM (
    SELECT a AS b, (-1 * a) AS z
    FROM [ { 'a': 2, 'z': 4 }, { 'a': 1, 'z': 2 }, { 'a': -1, 'z': -2 } ]
    ORDER BY b
)
ORDER BY "C"
====
[
  {
    'C': -1
  },
  {
    'C': 1
  },
  {
    'C': 2
  }
]
--- 
```

```sql
SELECT (a * -1) AS b FROM << { 'a': 1 }, { 'a': 2 } >> ORDER BY b;
--- RESULT:
[
  {
    'b': -2
  },
  {
    'b': -1
  }
]
```

```sql
SELECT sum(a) AS c, (a * -1) AS b FROM << { 'a': 1 }, { 'a': 2 } >> GROUP BY a ORDER BY b LIMIT 1;
--- RESULT:
[
  {
    'c': 2,
    'b': -2
  }
]
```

And, here's a very interesting one that shows what it means to modify the environment's bindings.
```sql
SELECT (a * -1) AS a FROM << { 'a': 1 }, { 'a': 2 } >> ORDER BY -a;
-- RESULT:
[
  {
    'a': -1
  },
  {
    'a': -2
  }
]
```

```sql
-- We ran !add_to_global_env { 'c': -1 } before the following
SELECT (a * -1) AS a FROM << { 'a': 1 }, { 'a': 2 } >> ORDER BY a * c;
-- RESULT:
[
  {
    'a': -1
  },
  {
    'a': -2
  }
]
```

## Bad Examples

The following queries should NOT work, and they do NOT.

```sql
SELECT b AS "C"
FROM (
    SELECT a AS b, (-1 * a) AS z
    FROM [ { 'a': 2, 'z': 4 }, { 'a': 1, 'z': 2 }, { 'a': -1, 'z': -2 } ]
    ORDER BY b
)
ORDER BY "c"
```
This failure shows that the `a` binding cannot be used in ORDER BY.
```sql
SELECT (a * -1) AS b FROM << { 'a': 1 }, { 'a': 2 } >> ORDER BY a;
```

## More Information

- Updated CHANGELOG
- No backward incompatible changes
- No external dependencies

### DIFF
--- a/lang/src/org/partiql/lang/eval/visitors/OrderByAliasVisitorTransform.kt
+++ b/lang/src/org/partiql/lang/eval/visitors/OrderByAliasVisitorTransform.kt
@@ -57,7 +57,11 @@ class OrderByAliasVisitorTransform : VisitorTransformBase() {
      * expression using the [projectionAliases]
      */
     override fun transformSortSpec_expr(node: PartiqlAst.SortSpec): PartiqlAst.Expr {
-        return OrderByAliasSupport(projectionAliases).transformSortSpec_expr(node)
+        val newExpr = when (node.expr.metas.containsKey(IsTransformedOrderByAliasMeta.TAG)) {
+            true -> super.transformSortSpec_expr(node)
+            false -> OrderByAliasSupport(projectionAliases).transformSortSpec_expr(node)
+        }
+        return newExpr.copy(metas = newExpr.metas + metaContainerOf(IsTransformedOrderByAliasMeta.instance))
     }
 
     /**
@@ -68,12 +72,10 @@ class OrderByAliasVisitorTransform : VisitorTransformBase() {
     class OrderByAliasSupport(val aliases: Map<String, PartiqlAst.Expr>) : VisitorTransformBase() {
         override fun transformExprId(node: PartiqlAst.Expr.Id): PartiqlAst.Expr {
             val transformedExpr = super.transformExprId(node)
-            if (transformedExpr.metas.containsKey(IsTransformedOrderByAliasMeta.TAG)) return transformedExpr
-            val newExpr = when (node.case) {
+            return when (node.case) {
                 is PartiqlAst.CaseSensitivity.CaseSensitive -> aliases[node.name.text] ?: transformedExpr
                 else -> aliases[node.name.text.toLowerCase()] ?: aliases[node.name.text.toUpperCase()] ?: transformedExpr
             }
-            return newExpr.copy(metas = newExpr.metas + metaContainerOf(IsTransformedOrderByAliasMeta.instance))
         }
     }
 }
--- a/lang/src/org/partiql/lang/eval/visitors/OrderByAliasVisitorTransform.kt
+++ b/lang/src/org/partiql/lang/eval/visitors/OrderByAliasVisitorTransform.kt
@@ -14,12 +14,15 @@
 
 package org.partiql.lang.eval.visitors
 
+import org.partiql.lang.ast.IsTransformedOrderByAliasMeta
 import org.partiql.lang.domains.PartiqlAst
+import org.partiql.lang.domains.metaContainerOf
 import org.partiql.pig.runtime.SymbolPrimitive
 
 /**
  * A [PartiqlAst.VisitorTransform] to replace the [PartiqlAst.SortSpec] of a [PartiqlAst.OrderBy] with a reference to
- * a [PartiqlAst.ProjectItem]'s [PartiqlAst.Expr] if an alias is provided.
+ * a [PartiqlAst.ProjectItem]'s [PartiqlAst.Expr] if an alias is provided. Also utilizes [IsTransformedOrderByAliasMeta]
+ * to enforce idempotency (delivering the same result through multiple passes).
  *
  * Turns:
  *
@@ -58,15 +61,19 @@ class OrderByAliasVisitorTransform : VisitorTransformBase() {
     }
 
     /**
-     * A [PartiqlAst.VisitorTransform] that converts any found Expr.Id's into what it is mapped to in [aliases]
+     * A [PartiqlAst.VisitorTransform] that converts any found Expr.Id's into what it is mapped to in [aliases]. Also
+     * utilizes [IsTransformedOrderByAliasMeta] to ensure idempotency of this visitor (AKA visiting twice will NOT result
+     * in a different AST)
      */
     class OrderByAliasSupport(val aliases: Map<String, PartiqlAst.Expr>) : VisitorTransformBase() {
         override fun transformExprId(node: PartiqlAst.Expr.Id): PartiqlAst.Expr {
             val transformedExpr = super.transformExprId(node)
-            return when (node.case) {
+            if (transformedExpr.metas.containsKey(IsTransformedOrderByAliasMeta.TAG)) return transformedExpr
+            val newExpr = when (node.case) {
                 is PartiqlAst.CaseSensitivity.CaseSensitive -> aliases[node.name.text] ?: transformedExpr
                 else -> aliases[node.name.text.toLowerCase()] ?: aliases[node.name.text.toUpperCase()] ?: transformedExpr
             }
+            return newExpr.copy(metas = newExpr.metas + metaContainerOf(IsTransformedOrderByAliasMeta.instance))
         }
     }
 }
--- a/lang/src/org/partiql/lang/eval/visitors/OrderByAliasVisitorTransform.kt
+++ b/lang/src/org/partiql/lang/eval/visitors/OrderByAliasVisitorTransform.kt
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at:
+ *
+ *      http://aws.amazon.com/apache2.0/
+ *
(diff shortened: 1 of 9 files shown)


yliuuuu (commented):
The core logic here seems to be evaluating the project thunk and therefore creating a new environment for `ORDER BY` so that the alias is bind to a value. Perhaps it is possible to rewrite the query before the compilation? Just so there is no need to create a new environment for `ORDER BY`.  
For example: 
`SELECT a + 1 as b FROM tbl ORDER BY b` can be rewritten to `SELECT a+1 as b FROM tbl ORDER BY a+1`

johnedquinn:
> The core logic here seems to be evaluating the project thunk and therefore creating a new environment for `ORDER BY` so that the alias is bind to a value. Perhaps it is possible to rewrite the query before the compilation? Just so there is no need to create a new environment for `ORDER BY`. For example: `SELECT a + 1 as b FROM tbl ORDER BY b` can be rewritten to `SELECT a+1 as b FROM tbl ORDER BY a+1`

So, I was thinking about this, and it *shouldn't* work for specific scenarios. AKA, if we refer to the order of SQL clause evaluations (loosely):
1. FROM clause
2. WHERE clause
3. GROUP BY clause
4. HAVING clause
5. SELECT clause
6. ORDER BY clause

Then, we can assume that the aliases created during SELECT would be immediately usable as bindings for ORDER BY. So, consider the following query:

```sql
SELECT (a * -1) AS a FROM << { 'a': 1 }, { 'a': 2 } >> ORDER BY a;
```

From my interpretation, the above query should result in:
```ion
<<
  { a: -2 },
  { a: -1 }
>>
```

So, replacing the RHS of ORDER BY to be:
```sql
SELECT (a * -1) AS a FROM << { 'a': 1 }, { 'a': 2 } >> ORDER BY (a * -1);
```
SHOULD result in:
```ion
<<
  { a: -1 },
  { a: -2 }
>>
```

Because the `a` on the RHS of ORDER BY is now the new `a` from the projection (-1, -2), and the tuples will be ordered by their absolute values --> 1, 2.

But the current evaluating compiler has the evaluation order wrong, so your proposal DOES work with the existing compiler, but the compiler is wrong. So, regardless, I really like your comment, because it made me think about this in terms of evaluation logic and relational algebra, and I think I need to modify the order of the thunks rather than do either of our proposals.

FYI, MySQL takes the following query:
```sql
SELECT -id AS id1 FROM test ORDER BY -id1;
-- where test is a single-column table with values 1 and 2
```
and returns:
```
-1
-2
```

yliuuuu:
I see what you mean here. 
Based on the content you have written, I suppose that the last query really should be: 
```sql
-- what your wrote
SELECT -id AS id1 FROM test ORDER BY -id1;
-- what your probably mean
SELECT -id AS id1 FROM test ORDER BY id1;
```

A similar analogy to consider is "Using SELECT alias in GROUP BY` in pure SQL context. 
For example, consider: 
```sql
SELECT a + 1 as group_key from test GROUP BY b
```
Here SQL face the same problem, some how it needs to know what is `b` even though `GROUP BY` get evaluated before `SELECT`. 
Check this [DB-Fiddle link](https://www.db-fiddle.com/f/wavXsKL9eGyxRx5sZHxfsM/2) to see the example query, notice how the two query plan generated by PostgreSQL are exactly the same. 

Now back to the `ORDER BY` problem, I think the biggest confusion here is that: 
Semantically, SQL's `ORDER BY` get evaluated after `SELECT`, but PARTIQL's `ORDER BY` get evaluated before `SELECT`. See Spec section 12.5 for more detail. 

I suspect the spec defined it that way for semantic consistency, but I could be wrong. 

If you main concern is the double evaluation (i.e., evaluate order by twice, once during projection and once during ordering), I suspect this problem can be addressed during the query planning. 

Consider: 
```sql
SELECT <expr> as a FROM tbl ORDER BY a
-- <expr> is some expression
```

I felt like the planer can essentially choose from
```sql
-- double evaluation
SELECT <expr> as a FROM tbl ORDER BY <expr>
```
or 
```sql
-- double scan
SELECT a from (SELECT <expr> as a FROM tbl) ORDER BY a
```
Depending on which operation is cheaper. 

check this[ DB-Fiddle Link](https://www.db-fiddle.com/f/4braioNELpCXWLYEYTFGrg/1) for ORDER BY evaluation in PostgreSQL. 

I could be totally wrong on this, but I am happy to brainstorming more on this topic with you. Hopefully it helps.

johnedquinn:
Marking this as a DRAFT until I get some feedback on [this PartiQL Spec Discussion](https://github.com/partiql/partiql-spec/discussions/34)

johnedquinn:
After the feedback, I updated the PR to rewrite the query.

lang/src/org/partiql/lang/eval/visitors/OrderByAliasVisitorTransform.kt
```diff
@@ -0,0 +1,81 @@
...
+     */
+    class OrderByAliasSupport(val aliases: Map<String, PartiqlAst.Expr>) : VisitorTransformBase() {
+        override fun transformExprId(node: PartiqlAst.Expr.Id): PartiqlAst.Expr {
+            val transformedExpr = super.transformExprId(node)
+            return when (node.case) {
+                is PartiqlAst.CaseSensitivity.CaseSensitive -> aliases[node.name.text] ?: transformedExpr
+                else -> aliases[node.name.text.toLowerCase()] ?: aliases[node.name.text.toUpperCase()] ?: transformedExpr
+            }
+        }
+    }
```
yliuuuu, line 80: Would you please elaborate a little more on why do we need another class here?

lang/src/org/partiql/lang/eval/visitors/OrderByAliasVisitorTransform.kt
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Turns:
+ *
+ * ```SELECT a + 1 AS b FROM c ORDER BY b```
+ *
+ * Into:
+ *
+ * ```SELECT a + 1 AS b FROM c ORDER BY a + 1```
+ */
+class OrderByAliasVisitorTransform : VisitorTransformBase() {
```
yliuuuu, line 35: nit: The name `OrderByAlias` can be somewhat confusing, i.e., one may think of it as `ORDER BY <expr> AS <id>`
Maybe change it to `OrderBySortSpecVisitorTransform`? Since we essentially want to rewrite the SortSpec. 
```suggestion
class OrderBySortSpecVisitorTransform : VisitorTransformBase() {
```

yliuuuu (commented):
Great work overall. Just two small nitpick comments.

lang/src/org/partiql/lang/eval/visitors/OrderByAliasVisitorTransform.kt
johnedquinn: So, take the following query:
```sql
SELECT (a + b) AS b FROM << {'a': 0, 'b': 1}, {'a': 0, 'b': 1} >> ORDER BY b
```

Visitor transforms only allow specifying a single behavior per datatype. So, we use "support" classes to essentially deliver multiple types of behavior. AKA, in this scenario, we only want to replace the `id` in `ORDER BY` with whatever it is aliased to. Essentially, it helps with limiting the scope of rewriting. We might not necessarily want to rewrite ALL `id`s that have value `b`.

So, instead of replacing EVERY instance of `id` (b) with a lookup, we scope it to ORDER BY. That means the example from above gets rewritten to:

```sql
SELECT (a + b) AS b FROM << {'a': 0, 'b': 1}, {'a': 0, 'b': 1} >> ORDER BY (a + b)
```

and not:

```sql
SELECT (a + (a + b)) AS b FROM << {'a': 0, 'b': 1}, {'a': 0, 'b': 1} >> ORDER BY (a + b)
```

I'm not entirely sure if it would get rewritten exactly like this because off the top of my head IDK the order of evaluation of the visitor transforms, but you get the point. We don't want to rewrite *every* encounter of `b` -- we only want to when it is in the sortSpecs of Order By. See `transformSortSpec_expr`.

yliuuuu (approved):
LGTM.

PR #752, opened 2022-09-08
Enables CLI highlighting using any Parser (PartiQLParser Included)

### LINKED ISSUE #748: Create PartiQLParser CLI Highlighter
As `SqlParser` and `SqlLexer` are getting deprecated, it is now relevant that the CLI should use a Syntax Highlighter that uses the new `PartiQLParser` to highlight the input queries.

johnedquinn:
## Relevant Issues
- Closes #748

## Note
- Leaving as draft until I determine whether we should create a public `PartiQLLexerBuilder`, or if we should deprecate `Lexer` and add `tokenize` to Parser public methods since `Lexer` isn't really used. Will consult with team.

## Description
- Modifies internal impl of Parser to also implement Lexer (tokenize method)
- Modifies shell highlighter to take in Lexer and Parser as params
- Modifies shell to use a Shell Highlighter specific to which parser is being used

## Other Information
- No backward incompatible changes
- No need to update changelog. Small changes.
- No external dependency changes

### DIFF
--- a/lang/src/org/partiql/lang/syntax/PartiQLParser.kt
+++ b/lang/src/org/partiql/lang/syntax/PartiQLParser.kt
@@ -17,7 +17,6 @@ package org.partiql.lang.syntax
 import com.amazon.ion.IonSexp
 import com.amazon.ion.IonSystem
 import org.antlr.v4.runtime.BaseErrorListener
-import org.antlr.v4.runtime.CharStreams
 import org.antlr.v4.runtime.CommonTokenStream
 import org.antlr.v4.runtime.Lexer
 import org.antlr.v4.runtime.RecognitionException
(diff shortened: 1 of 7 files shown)


lang/src/org/partiql/lang/syntax/PartiQLParser.kt
```diff
@@ -46,7 +47,7 @@ import org.partiql.grammar.parser.generated.PartiQLTokens as GeneratedLexer
 internal class PartiQLParser(
     private val ion: IonSystem,
     val customTypes: List<CustomType> = listOf()
-) : Parser {
+) : Parser, org.partiql.lang.syntax.Lexer {
```
yliuuuu, line 50: The fact that Parser has to implement both Parser and Lexer seems a bit odd to me. Is it possible to decouple them by creating a `PartiQLLexer` class? I do not have a strong opinion on this.

johnedquinn: Yeah, I was going to leave it to the reviewer's discretion whether we should create a PartiQLLexer, and I like it. It's a small change -- will send it out tomorrow! Thanks for the feedback!

RCHowell:
I'm of the opinion that we should remove the lexer and tokens, as well as not expose a lex method from the parser. We can have the exact same functionality use the generated antlr lexer, and if customers wish to lex, then they can do exactly the same. Our lexer and tokens do not provide any more functionality than the antlr lexer, and they come with (minimal) maintenance cost. Anyways, I suggest we clean up by removing all things lexing because that's something we get for free via ANTLR

johnedquinn:
Closing this -- as I'll be using the generated ANTLR Lexer and Parser for this. Will be a completely different approach.

Issue #744, opened 2022-09-07
Trim Function only accepts lowercase `both | leading | trailing`


yliuuuu:
Currently our trim function signature looks like `TRIM: [ String ] String -> String`

This implies that we are treating the optional first argument `LEADING|TRAILING|BOTH` as string instead of keywords, causing the case sensitivity issue. 

A quick work around will be adding upper case string to our Enum Class. 

https://github.com/partiql/partiql-lang-kotlin/blob/40e957a783f1726c01b512c255fe2943d1c24d06/lang/src/org/partiql/lang/eval/builtins/TrimExprFunction.kt#L161-L170

johnedquinn:
@yliuuuu should this be closed?

yliuuuu:
Closed by [PR#754](https://github.com/partiql/partiql-lang-kotlin/pull/754)

PR #754, opened 2022-09-09
change trim spec to be case insensitive 

yliuuuu:
*Issue #, if available:*
Closes Issue #744 

*Description of changes:*
1. Added support for trim to accept case insensitive [ BOTH | LEADING | TRAILING]
2. Fixed error message in case of Trim Spec mismatch
3. Added an Parser Error to specifically for trim spec mismatch
4. Added test case for upper case trim spec
5. Added test cases for Trim Spec Mismatch

Future to do: 
Decide whether or not Trim Spec mismatch shall be a parser error or an evaluator error, and clean up the code accordingly. 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/lang/src/org/partiql/lang/errors/ErrorCode.kt
+++ b/lang/src/org/partiql/lang/errors/ErrorCode.kt
@@ -531,6 +531,12 @@ enum class ErrorCode(
         "Other expressions may not be present in the select list when '*' is used without dot notation."
     ),
 
+    PARSE_INVALID_ARGUMENTS_FOR_TRIM(
+        ErrorCategory.PARSER,
+        LOC_TOKEN,
+        "Invalid arguments for trim"
+    ),
+
     // Evaluator errors
     // TODO:  replace uses of this with UNIMPLEMENTED_FEATURE
     EVALUATOR_FEATURE_NOT_SUPPORTED_YET(
--- a/lang/src/org/partiql/lang/errors/ErrorCode.kt
+++ b/lang/src/org/partiql/lang/errors/ErrorCode.kt
@@ -531,7 +531,7 @@ enum class ErrorCode(
         "Other expressions may not be present in the select list when '*' is used without dot notation."
     ),
 
-    PARSE_INVALID_ARGUMENTS_FOR_TRIM(
+    PARSE_INVALID_TRIM_SPEC(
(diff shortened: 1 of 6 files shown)


lang/test/org/partiql/lang/errors/ParserErrorsTest.kt
```diff
@@ -1286,6 +1286,36 @@ class ParserErrorsTest : SqlParserTestBase() {
         )
     }
 
+    @Test
+    fun callTrimSpecificationMismatch() {
+        // for SQL parser, the logic seems to be if the first token is not one of { NONE | BOTH | LEADING | TRAILING}
+        // treating the token as identifier and expect the trim expression to end
```
lziq, line 1292: Can you add a `TODO` here saying that this test case will be removed once we remove the legacy `SqlParser`?

lang/src/org/partiql/lang/errors/ErrorCode.kt
```diff
@@ -531,6 +531,12 @@ enum class ErrorCode(
         "Other expressions may not be present in the select list when '*' is used without dot notation."
     ),
 
+    PARSE_INVALID_ARGUMENTS_FOR_TRIM(
```
lziq, line 534: Probably change the name to `PARSE_INVALID_TRIM_SPEC`? The term "arguments" can refer to the string being trimmed.

lziq (changes requested):
Overall looks good to me. I think trim specification mismatch should be considered as a parser error.

lang/src/org/partiql/lang/errors/ErrorCode.kt
yliuuuu: Addressed in the new commit.

lang/test/org/partiql/lang/errors/ParserErrorsTest.kt
yliuuuu: Addressed in the new commit. I will open up another issue on treating specification mismatch as a parser error.

lziq (dismissed):
As what we discussed offline, `trim` spec mismatch error will be considered further in another PR.

Issue #510, opened 2022-01-26
Remove hardcoded types from the PIG domain


alancai98:
There are currently some hardcoded types added to the PIG domain. See `partiql.ion`:
https://github.com/partiql/partiql-lang-kotlin/blob/fb7992468158d947557dc979f48a5f41ce83aa43/lang/resources/org/partiql/type-domains/partiql.ion#L392-L414

These types are temporary and should be removed as soon as possible so users don't take a dependency on these types. Removal of these types may cause downstream changes to [AstDeserialization.kt](https://github.com/partiql/partiql-lang-kotlin/blob/fb7992468158d947557dc979f48a5f41ce83aa43/lang/src/org/partiql/lang/ast/AstDeserialization.kt#L164-L166), [CustomTypeVisitorTransform.kt](https://github.com/partiql/partiql-lang-kotlin/blob/fb7992468158d947557dc979f48a5f41ce83aa43/lang/src/org/partiql/lang/eval/visitors/CustomTypeVisitorTransform.kt#L10), [StatementToExprNode](https://github.com/partiql/partiql-lang-kotlin/blob/fb7992468158d947557dc979f48a5f41ce83aa43/lang/src/org/partiql/lang/ast/StatementToExprNode.kt#L351-L368), [ExprValueType.kt](https://github.com/partiql/partiql-lang-kotlin/blob/fb7992468158d947557dc979f48a5f41ce83aa43/lang/src/org/partiql/lang/eval/ExprValueType.kt#L218), and [SqlParserCustomTypeCatalogTests.kt](https://github.com/partiql/partiql-lang-kotlin/blob/fb7992468158d947557dc979f48a5f41ce83aa43/lang/test/org/partiql/lang/syntax/SqlParserCustomTypeCatalogTests.kt#L15-L18).

There is no support for these types, beyond parsing and (de)serialization, and we are not aware of currently-active users. The effect of these nodes can be achieved with the generic `(custom_type ...)` node.

lziq:
See https://github.com/partiql/partiql-lang-kotlin/pull/755#discussion_r967346752 for the reason to remove.

PR #755, opened 2022-09-09
Removed custom type AST nodes

lziq:
This PR aims to close https://github.com/partiql/partiql-lang-kotlin/issues/510

*Breaking Changes:*
The following custom type AST nodes are removed from `partiql.ion` file: `es_boolean`, `es_integer`, `es_float`, `es_text`, `es_any`, `spark_short`, `spark_integer`, `spark_long`, `spark_double`, `spark_boolean`, `spark_float`, `rs_varchar_max`, `rs_integer`, `rs_bigint`, `rs_boolean`, `rs_real`, `rs_double_precision`. The related visitor transform `CustomTypeVisitorTransform` is also removed


lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -518,30 +518,6 @@ may then be further optimized by selecting better implementations of each operat
             // Special types
             (any_type)
 
-            // TODO: Remove these hardcoded nodes from the PIG domain once [https://github.com/partiql/partiql-lang-kotlin/issues/510]
```
alancai98, line 521: There are still some references to #510 in AstSerialization.kt. Can those references also be removed?

- https://github.com/partiql/partiql-lang-kotlin/blob/342c2d23e933d032e278d4df26891aed78507eb8/lang/src/org/partiql/lang/ast/AstDeserialization.kt#L183-L184
- https://github.com/partiql/partiql-lang-kotlin/blob/342c2d23e933d032e278d4df26891aed78507eb8/lang/src/org/partiql/lang/ast/AstDeserialization.kt#L261-L262
- https://github.com/partiql/partiql-lang-kotlin/blob/342c2d23e933d032e278d4df26891aed78507eb8/lang/src/org/partiql/lang/ast/AstDeserialization.kt#L291-L292

alancai98 (commented):
To my knowledge, we have no customers using these custom type AST nodes, so I'm generally fine with their removal.

Since this affects the PIG domain and some APIs, this PR is considered a breaking change. I think we still need to be dutiful as a library owner by properly documenting the removal of these nodes along with impacted public APIs (i.e. PIG domain, `CustomTypeVisitorTransform`) in the CHANGELOG and the PR description (using the PR template to document the breaking changes).

lang/resources/org/partiql/type-domains/partiql.ion
lziq: I think these references present a generic custom type node (the same as https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/resources/org/partiql/type-domains/partiql.ion#L545), instead of a specific custom type node, such as `rs_varchar`. So I don't think we are able to remove them

RCHowell:
Thank you for the clean up work!

RCHowell (approved):

CHANGELOG.md
```diff
@@ -79,7 +79,7 @@ stage in the `PlannerPipeline` and to generate performance metrics for the indiv
 
 ### Removed
 - README.md badge for travisci
-
+- The following custom type AST nodes are removed from `partiql.ion` file: `es_boolean`, `es_integer`, `es_float`, `es_text`, `es_any`, `spark_short`, `spark_integer`, `spark_long`, `spark_double`, `spark_boolean`, `spark_float`, `rs_varchar_max`, `rs_integer`, `rs_bigint`, `rs_boolean`, `rs_real`, `rs_double_precision`. The related visitor transform `CustomTypeVisitorTransform` is also removed
```
vgapeyev, line 82: Would it make sense to also document rationale and/or consequences for this change?  I did not yet reverse-engineer what they are, but I imagine something like 
"PartiQL did not have custom support for these types, beyond parsing, and we are not aware of currently-active users.  The effect of these nodes can be achieved with the generic (custom_type ...) node".

lziq: I don't know much about the story behind, but I guess that originally, we had these specific custom type AST nodes. Later, we decided to wrap all of them using one generic custom type AST node `custom_type`, and at that moment, we just forgot (or not confident enough) to remove these specific AST nodes. At least, from what I can see, these specific AST nodes are not really useful and they are not used anyway with regard to the current custom type API in our codebase. So I think we can remove them safely.

lziq: I think we should only document the changes in short in the `CHANGELOG.md` file. I also wonder if we should document rationale behind these changes somewhere else.

vgapeyev: I see fairly detailed entries in the current [Unreleased] section.  It does look somewhat of an outlier compared to the earlier, but is that history long / active enough to take as the precedent law?
At the very least, a reference to Issue 510 + adding this explanation there, I think.

alancai98: +1 to what Vladimir had mentioned. At the very least would be helpful to add a link to issue #510's comment

alancai98 (commented):
Looks good other than updating the CHANGELOG with what Vladimir had mentioned.

alancai98 (dismissed):

vgapeyev (changes requested):
After looking into this, I believe it is worth revisiting those 510-related TODOs that Alan pointed out. 

Also, I took the liberty to commit tweaks to the CHANGELOG entry -- it seemed easier than requesting them.

lziq:
> After looking into this, I believe it is worth revisiting those 510-related TODOs that Alan pointed out.
> 
> Also, I took the liberty to commit tweaks to the CHANGELOG entry -- it seemed easier than requesting them.

Thanks for updating the `CHANGELOG` file. 

I wrote the following code to test V0 AST related to custom types: 
```
    val ion = IonSystemBuilder.standard().build()
    val myType = CustomType(
        "my_type",
        TypedOpParameter(
            StaticType.ANY
        )
    )

    val parser = PartiQLParser(ion, listOf(myType))
    val query = "CAST(1 AS my_type)"
    val astStatement = parser.parseAstStatement(query)
    val exprNode = astStatement.toExprNode(ion)
    val v0Ast = AstSerializer.serialize(exprNode, AstVersion.V0, ion)
    println(v0Ast)

    val transformedExprNode = AstDeserializerBuilder(ion, listOf(myType)).build().deserialize(v0Ast, AstVersion.V0)
    println(transformedExprNode)
```

The result I got: 
``` 
(cast (meta (lit 1) {line:1,column:6}) (type my_type))
Exception in thread "main" java.lang.IllegalArgumentException: 'my_type' is not a valid data type
	at org.partiql.lang.ast.AstDeserializationKt.err(AstDeserialization.kt:1308)
	at org.partiql.lang.ast.AstDeserializationKt.access$err(AstDeserialization.kt:1)
	at org.partiql.lang.ast.AstDeserializerInternal$deserializeDataTypeV0$1.invoke(AstDeserialization.kt:1258)
	at org.partiql.lang.ast.AstDeserializerInternal$deserializeDataTypeV0$1.invoke(AstDeserialization.kt:288)
	at org.partiql.lang.ast.AstDeserializerInternal$deserializeSexpMetaOrTerm$1.invoke(AstDeserialization.kt:334)
	at org.partiql.lang.ast.AstDeserializerInternal$deserializeSexpMetaOrTerm$1.invoke(AstDeserialization.kt:288)
	at org.partiql.lang.ast.AstDeserializerInternal.deserializeIonValueMetaOrTerm(AstDeserialization.kt:356)
	at org.partiql.lang.ast.AstDeserializerInternal.deserializeSexpMetaOrTerm(AstDeserialization.kt:333)
	at org.partiql.lang.ast.AstDeserializerInternal.deserializeDataTypeV0(AstDeserialization.kt:1253)
	at org.partiql.lang.ast.AstDeserializerInternal.deserializeDataType(AstDeserialization.kt:1249)
	at org.partiql.lang.ast.AstDeserializerInternal.deserializeTypedCast(AstDeserialization.kt:797)
	at org.partiql.lang.ast.AstDeserializerInternal.access$deserializeTypedCast(AstDeserialization.kt:288)
	at org.partiql.lang.ast.AstDeserializerInternal$deserializeExprNode$1.invoke(AstDeserialization.kt:418)
	at org.partiql.lang.ast.AstDeserializerInternal$deserializeExprNode$1.invoke(AstDeserialization.kt:288)
	at org.partiql.lang.ast.AstDeserializerInternal$deserializeSexpMetaOrTerm$1.invoke(AstDeserialization.kt:334)
	at org.partiql.lang.ast.AstDeserializerInternal$deserializeSexpMetaOrTerm$1.invoke(AstDeserialization.kt:288)
	at org.partiql.lang.ast.AstDeserializerInternal.deserializeIonValueMetaOrTerm(AstDeserialization.kt:356)
	at org.partiql.lang.ast.AstDeserializerInternal.deserializeSexpMetaOrTerm(AstDeserialization.kt:333)
	at org.partiql.lang.ast.AstDeserializerInternal.deserializeExprNode$lang(AstDeserialization.kt:368)
	at org.partiql.lang.ast.AstDeserializerInternal.deserialize(AstDeserialization.kt:303)
	at org.partiql.lang.ast.AstDeserializerBuilder$build$1.deserialize(AstDeserialization.kt:283)
	at TestKt.main(Test.kt:75)
	at TestKt.main(Test.kt)

```

This shows that we can transform a custom type from `ExprNode` into V0 AST, but we cannot perform the reverse transformation, which is weird. 

Also, I found that none of 2 custom type tags in `AstDeserialization.lt`, [`CUSTOM_TYPE`](https://github.com/partiql/partiql-lang-kotlin/blob/342c2d23e933d032e278d4df26891aed78507eb8/lang/src/org/partiql/lang/ast/AstDeserialization.kt#L181) & [`CUSTOM_TYPE_FOR_OLDER_PIG_ASTS`](https://github.com/partiql/partiql-lang-kotlin/blob/342c2d23e933d032e278d4df26891aed78507eb8/lang/src/org/partiql/lang/ast/AstDeserialization.kt#L185], are actually being used. When deserialize a V0 custom type AST node, [this snippet of code](https://github.com/partiql/partiql-lang-kotlin/blob/342c2d23e933d032e278d4df26891aed78507eb8/lang/src/org/partiql/lang/ast/AstDeserialization.kt#L1262-L1270) treats it as a built-in type AST node and throw an error. 

Anyway, I think that custom type in V0 AST is not modeled really well, but I don't think we should invest more time to investigate or even maintain it.

lziq:
Since V0 AST is removed in this PR, https://github.com/partiql/partiql-lang-kotlin/pull/863, the block we were discussing is not there any more. I think we can unblock this PR.

lang/resources/org/partiql/type-domains/partiql.ion
vgapeyev: Agreeing with @lziq that any objection above is now more than mute, after AstDeserialization.kt is now completely gone, in #863.

vgapeyev (approved):

PR #761, opened 2022-09-09
Optimize GH Actions Build by removing documentation builds

johnedquinn:
## Relevant Issues
- None

## Description
- Modifies build GH actions to not build when updating documentation.
- See the commit history to see the types of changes and the build types.

## Other Information
  - Small change.

## License Information

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,15 @@
 name: "Build and Report Generation"
-on: [push, pull_request]
+on: 
+  push:
+    paths:
+      - '**'
+      - '!docs/**'
+      - '!*.md'
+  pull_request:
+    paths:
+      - '**'
+      - '!docs/**'
+      - '!*.md'
 
 jobs:
   test:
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -1,5 +1,15 @@
 name: "Build and Report Generation"
-on: [push, pull_request]
+on: 
+  push:
+    paths:
+      - '**'
+      - '!docs/**'
```
yliuuuu, line 6: I am not sure if the files under docs/assets/code have any automatic testing process associated with them. i.e., if there is an examination process to make sure that `q1.sql`, when supplied with `q1.env` produce the same output as `q1.output`. I do not see this in the build config, but I will double check to make sure. Otherwise love the idea!

johnedquinn: Yeah, the code is only there for reference -- not used in any build. The only place that has any reference to that folder is, [Tutorial.md](https://github.com/partiql/partiql-lang-kotlin/blob/11c265e08f45abb7be7ebaa5f51fe58aa4bc0eb3/docs/tutorials/Tutorial.md#partiql-queries-are-sql-compatible) -- just pointing to the larger .env file.

yliuuuu (approved):
LGTM

PR #779, opened 2022-09-14
Parallelizes subset of tests by splitting test into many

### LINKED ISSUE #778: Frozen Tests
## Description
- If you haven't noticed, when we test the project -- it always freezes around test # 77,000 for several minutes. Doesn't give you any status update, which is quite annoying because developers don't know where they are in the tests.
- See below:

https://user-images.githubusercontent.com/40360967/190221147-1eccb8c9-d9c2-4edd-9209-fb82caed8165.mov

## To Reproduce
1. ```./gradlew clean build```

## Expected Behavior
- Tests running without stopping.

johnedquinn:
## Relevant Issues
- Closes #778 

## Description
- There was 4-5 tests performing 100,000 iterations each of an operation
- This caused an annoying test experience -- as the build would stop at around 77,000 tests, and it wouldn't give you an update for several minutes
- I took the input data and wrapped it to pass it to a ParameterizedTest so that JUnit could parallelize the test runs
- This way, the tests never freeze (YAY !), we can always see our progress, and I saw preliminary results of significantly faster tests.

## Testing Results
Ran tests over two days (seems like Day 1 my computer was more friendly). Removed the build cache before each run and ran `./gradlew clean build` -- below are the test times.
- Branch `main`
  - 8m 46s (Day 1)
  - 8m 59s (Day 1)
  - 10m 6s (Day 2)
  - 9m 19s (Day 2)
  - **AVERAGE: 9m 17.5s**
- Branch `fix-stopped-tests`
  - 4m 58s (Day 1)
  - 4m 50s (Day 1)
  - 7m 21s (Day 2)
  - 6m 55s (Day 2)
  - **AVERAGE: 06m 01s**

From the tests, there is approximately a speedup of 54.4% for local builds.

## Other Information
- Another thing is that, before, we used to have between 80-90K tests. Now, since I've split up these massive tests, we now have around 490K tests
- Also, according to the [Gradle documentation](https://docs.gradle.org/current/userguide/performance.html#parallel_test_execution), we should probably add the following to our Gradle tests config at some point:
```
tasks.withType(Test).configureEach {
    maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
}
```

## Other Information
  - No -- it's a testing change

## License Information

### DIFF
--- a/lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
+++ b/lang/test/org/partiql/lang/eval/EvaluatingCompilerDateTimeTests.kt
@@ -19,7 +19,67 @@ import kotlin.random.Random
 
 class EvaluatingCompilerDateTimeTests : EvaluatorTestBase() {
 
-    private val randomTestsSize = 50000
+    abstract class RandomTestsProvider() : ArgumentsProviderBase() {
+        private val randomTestsSize = 50000
+        private val randomGenerator = generateRandomSeed()
+
+        private fun generateRandomSeed(): Random {
+            val seed = Random.nextInt()
+            println("Randomly generated seed is $seed. Use this to reproduce failures in dev environment.")
+            return Random(seed)
+        }
+
+        private fun Random.nextTime(withPrecision: Boolean = false, withTimezone: Boolean = false): TimeForValidation {
+            val hour = nextInt(24)
+            val minute = nextInt(60)
+            val second = nextInt(60)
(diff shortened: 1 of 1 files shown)


RCHowell:
Did we ever settle on plans to move this to fuzz testing as a workflow?

RCHowell (approved):
Looks good for now. Follow up item — https://github.com/partiql/partiql-lang-kotlin/issues/787

PR #780, opened 2022-09-14
Initializes non-reserved PartiQL keywords

### LINKED ISSUE #719: Ability to modify SQL keyword list
I know keywrods come from the spec, however the popular words like `domain`, `user`, `public` is not actually useful in this library, not used in any way but we need to escape them as `SELECT * FROM "user"` which does not look nice. Would you consider the ability to customize that list as an advanced, somewhat hidden feature? 

```
Parser Error: at line 1, column 8: unexpected keyword found, KEYWORD : public
```

johnedquinn:
## Relevant Issues
- Closes #719

## Description
- Adds non-reserved keywords to grammar files, allowing the use of several keywords as identifiers.
  - We do this by creating a `nonreservedKeywords` rule in ANTLR that is a fallback for `symbolPrimitive`
- This realistically does not make many changes -- since we treat the majority of nonreservedKeywords as non-reserved anyways. We initially did this by allowing identifiers in their place, then, in the PartiQLVisitor, we would check that they conform to whatever keyword we expected.
  - For example, the `TRIM` function allowed an optional IDENTIFIER as the first argument -- then, in the PartiQLVisitor, we would check that the identifier was either `leading`, `trailing`, or `both`. Now, we have formalized a way to still use `leading`, `trailing`, or `both` as identifiers while still enforcing that the first optional parameter must be one of the 3.
  - Similarly, take pattern restrictors for example. In GPML, the pattern restrictors (`trail`, `simple`, and `acyclic`) were parsed as IDENTIFIERS, but upon visiting, we would check that the identifiers are one of the allowed 3. It's more ideal that we have a formal rule -- that way, writing grammars is easier, and we don't need to write checks in the visitor.
  - Now, we do add PUBLIC, DOMAIN, and USER as non-reserved keywords -- since we don't use them at all in our grammar.
- We also initialized a Keywords Wiki document that specifies all of our keywords and distinguishes between reserved and non-reserved. See the doc for more information.

## Other Information
  - Not yet.

## License Information


partiql-grammar/src/main/antlr/org/partiql/grammar/PartiQL.g4
```diff
@@ -539,7 +553,7 @@ extract
     : EXTRACT PAREN_LEFT IDENTIFIER FROM rhs=expr PAREN_RIGHT;
 
 trimFunction
-    : func=TRIM PAREN_LEFT ( mod=IDENTIFIER? sub=expr? FROM )? target=expr PAREN_RIGHT;
+    : func=TRIM PAREN_LEFT ( mod=(BOTH|LEADING|TRAILING)? sub=expr? FROM )? target=expr PAREN_RIGHT;
```
RCHowell, line 556: Is pattern restrictor its own AST node? Also I was just reading through the grammar and noticed that we don't need many of the element labels. The element and list element labels are useful when the rule has more than one element with the same type. Unless that's the case, they're clutter

For example, we don't need any of the element labels here. It makes it harder to read
Ex: https://github.com/partiql/partiql-lang-kotlin/blob/main/partiql-grammar/src/main/antlr/org/partiql/grammar/PartiQL.g4#L329-L337

johnedquinn: For this example, are you talking about labels `sub` and `target`? Only did that because `sub` is optional.

Or `func`? I added the `func` label just because it made the PartiQLVisitor code look uniform. AKA DateFunction, and TrimFunction, and others result in:

```kotlin
call(ctx.func.text.toLowerCase(), args, metas)
```

But doesn't matter to me.

And yes, PatternRestrictor is its own AST node.

PR #782, opened 2022-09-16
Refactor STIVT Code Style

lziq:
Refactored some code in `StaticTypeInferenceVisitorTransform.kt` file:
1. Formatted the file 
2. Changed variable name from `nary` to `processedNode`
3. Changed `if` statement to `when` statement. 

This PR does not have any changes behavior of type inferencer.

### DIFF
--- a/lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
+++ b/lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
@@ -225,7 +225,7 @@ internal class StaticTypeInferenceVisitorTransform(
             if (existing != null) {
                 TODO(
                     "A variable named '$name' was already defined in this scope. " +
-                            "This wouldn't be the case if StaticTypeVisitorTransform was executed first."
+                        "This wouldn't be the case if StaticTypeVisitorTransform was executed first."
                 )
             }
             localsMap[name] = type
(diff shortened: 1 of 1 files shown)


lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -361,180 +361,168 @@ internal class StaticTypeInferenceVisitorTransform(
...
-            return if (hasValidOperandTypes(args, { it is BoolType }, "NOT", nAry.metas)) {
-            transformNAry(nAry, args) { recurseForNAryOperations(nAry, it, ::getTypeForNAryLogicalOperations) }
-        } else {
-                nAry.withStaticType(StaticType.BOOL)
-            }
+            val processedNode = super.transformExprNot(node) as PartiqlAst.Expr.Not
+            val args = listOf(processedNode.expr)
+            return when (hasValidOperandTypes(args, { it is BoolType }, "NOT", processedNode.metas)) {
+            true -> transformNAry(processedNode, args) { recurseForNAryOperations(processedNode, it, ::getTypeForNAryLogicalOperations) }
+            false -> processedNode.withStaticType(StaticType.BOOL)
```
alancai98, line 426: Just an observation (won't block on this behavior by ktlint): not sure why ktlint chose to format the `when` expression this way (I would have expected the formatting that's on L414 - 417). It seems like the original `if`/`else` expression also had this formatting. Similarly for other later `when` expressions.

alancai98 (approved):

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
lziq: I don't know either. It's wired.

PR #791, opened 2022-09-20
Refactor stivt code

lziq:
*This PR aims to refactor code style of operators handling in `StaticTypeInferenceVisitorTransform.kt` for better readability. There is no behavior change in this PR*

##Changes:
1. Now the code of operators handling are structurally classified into the following 4 categories:
- Unary operators: `NOT`, `POS` & `NEG`
- Logical NAry operators: `AND` & `OR`
- Common NAry operators: `PLUS`, `MINUS`, `TIMES`, `DIVIDE`, `MODULO`, `CONCAT`, `EQ`, `NE`, LT`, `LTE`, `GT` & `GTE`
- Other special NAry operators: `LIKE`, `BETWEEN` &` IN`
Each category has similar coding structure and calls similar helper methods. e.g. NAry Arithmetic operators & comparator operators now both call `computeReturnTypeForNAry`

### DIFF
--- a/lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
+++ b/lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
@@ -271,16 +271,6 @@ internal class StaticTypeInferenceVisitorTransform(
             return node.withStaticType(foundType)
         }
 
-        private fun transformNAry(
-            expr: PartiqlAst.Expr,
-            operands: List<PartiqlAst.Expr>,
-            compute: (List<StaticType>) -> StaticType
-        ): PartiqlAst.Expr {
-            val argTypes = operands.getStaticType()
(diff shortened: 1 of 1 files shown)


lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -269,64 +271,36 @@ internal class StaticTypeInferenceVisitorTransform(
             return node.withStaticType(foundType)
         }
 
-        private fun transformNAry(
```
alancai98, line 272: The diff for this PR is pretty hard to read because there's both refactoring of logic within functions and rearrangement of function order. One way to make the diff somewhat easier for reviewers to review could be to have two distinct commits: 1. performs logic changes/inner function refactoring. 2. rearranges the functions

alancai98 (approved):

lziq:
thx

PR #793, opened 2022-09-20
Adds sort operator to planner

### LINKED ISSUE #781: Add support for ORDER BY to Planner
## Relevant Issue/Bug
- Reference Issue #758 

## Requested Feature
- To reach feature parity with the evaluating compiler, the new planner needs to add support to ORDER BY.

## Describe Alternatives
- NA.

## Additional Context
Need to:
- [ ] Add to Logical PIG domain
- [ ] Modify AstToLogicalVisitorTransform to transform ORDER BY
- [ ] Possibly include Visitor Transform included in #740
- [ ] Modify PhysicalExprToThunkConverterImpl to support ordering.

## DoD (Definition of Done)
- ORDER BY working for planner

johnedquinn:
## Relevant Issues
- Closes #781 

## Description
- Adds a `SORT` node to `logical`, `resolved`, and `physical` plans (domains)
- Adds appropriate transforms
- Adds an abstract class `SortOperatorFactory` implementing a `RelationalOperatorFactory`
- Adds a simple `InMemorySortFactory` that takes in a `RelationExpression` and outputs a sorted `RelationExpression`. It is the default `SORT` implementation.
  - It works by evaluating the source relation expression, copying the state's registers (for the purpose of sorting), evaluating the ORDER BY sort specs using the passed state's registers, and then modifying the state's registers to output the correct state order.
- Removes the check for the existence of `ORDER BY` -- it used to throw intentional "unimplemented" exceptions
- Adds tests to the three types of transforms, and enables the evaluator tests for ORDER BY to test using the new planner

## Other Information
  - No. Only adding PartiQL domain nodes.

## License Information

### DIFF
--- a/lang/src/org/partiql/lang/eval/physical/operators/SortOperatorFactory.kt
+++ b/lang/src/org/partiql/lang/eval/physical/operators/SortOperatorFactory.kt
@@ -6,22 +6,21 @@ import org.partiql.lang.eval.ExprValue
 import org.partiql.lang.eval.NaturalExprValueComparators
 import org.partiql.lang.eval.err
 import org.partiql.lang.eval.physical.EvaluatorState
-import org.partiql.lang.eval.physical.transferState
 import org.partiql.lang.eval.relation.RelationType
 import org.partiql.lang.eval.relation.relation
 import org.partiql.lang.util.interruptibleFold
 
 /** Provides an implementation of the [PartiqlPhysical.Bexpr.Order] operator.*/
-internal abstract class SortOperatorFactory(name: String) : RelationalOperatorFactory {
-    final override val key: RelationalOperatorFactoryKey = RelationalOperatorFactoryKey(RelationalOperatorKind.SORT, name)
-    internal abstract fun create(
+public abstract class SortOperatorFactory(name: String) : RelationalOperatorFactory {
+    public final override val key: RelationalOperatorFactoryKey = RelationalOperatorFactoryKey(RelationalOperatorKind.SORT, name)
+    public abstract fun create(
         impl: PartiqlPhysical.Impl,
         sortKeys: List<CompiledSortKey>,
         sourceRelation: RelationExpression
     ): RelationExpression
 }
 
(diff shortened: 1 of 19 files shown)


lang/src/org/partiql/lang/eval/physical/operators/SortOperatorFactory.kt
```diff
@@ -0,0 +1,75 @@
...
+        if (intermediate == null) {
+            return@interruptibleFold compareBy<Array<ExprValue>, ExprValue>(sortKey.comparator) { row ->
+                transferState(state, row)
+                sortKey.value(state)
+            }
+        }
+        return@interruptibleFold intermediate.thenBy(sortKey.comparator) { row ->
+            transferState(state, row)
+            sortKey.value(state)
+        }
```
RCHowell, line 68: Could you clarify this? 

We are comparing a row to a row (Array<ExprValue>) by producing a chain of comparators (one for each sort key). How does a single comparator work?

lang/src/org/partiql/lang/planner/transforms/LogicalToLogicalResolvedVisitorTransform.kt
```diff
@@ -374,6 +374,19 @@ internal data class LogicalToLogicalResolvedVisitorTransform(
         }
     }
 
+    override fun transformSortSpec(node: PartiqlLogical.SortSpec): PartiqlLogicalResolved.SortSpec {
+        return super.transformSortSpec(node)
+    }
```
RCHowell, line 379: remove?

lang/src/org/partiql/lang/eval/physical/operators/SortOperatorFactory.kt
johnedquinn: If we're using a single ordering key, say `ORDER BY a DESC NULLS FIRST`:
- The fold will never reach `thenBy`. So we will only reach `compareBy. That's when we load in the registers to evaluate and use the NaturalExprValueComparator of NULLS_FIRST_DESC to compare each row by their value of `a`.

LMK if that answers your question.

RCHowell (approved):

PR #797, opened 2022-09-23
Switched from MATCH-as-FROM-source to MATCH-as-expression

vgapeyev:
## Description

In a graph query like 
```
    SELECT * FROM MyGraph MATCH (a)-[b]-(c)
```
the construct involving MATCH, instead of being understood as a special kind of a FROM source, 
is now understood as a new expression form. 
That is, `MyGraph MATCH (a)-[b]-(c)` is an expression that, given a graph-yielding expression and a graph pattern, produces a bag of structs (a "table") with attributes `a`, `b`, `c`. 

With this change, a MATCH expression can be a source in a FROM clause, as any other table-producing expression.  This happens to give variable-binding behavior that is different than under the earlier MATCH design. In particular, it is possible to write 
```
     SELECT x.* FROM MyGraph MATCH (a)-[b]-(c) AS x
```
which produces a table with columns `a`, `b`, `c`.

Only the ANTLR-based parser is properly supported in this PR.  
The old SqlParser got minimal changes to make it compile and still contains most of the earlier code for the original design of MATCH.  However, SqlParser now breaks when given MATCH-containing source, because the target PIG AST has changed.
*Please advise whether it is worth bringing SqlParser up to date.*

There are a few code comments marked with `PR-COMMENT`.  These highlight spots worthy of attention and  are meant to be either removed or converted to regular comments prior to merging.

## Other Information
  - Modified the earlier entry about the experimental graph query feature. 
  - The version-controlled behavior does change in a breaking way w.r.t. to prior commits, but they have not yet been released and the overall graph feature is experimental anyway.

## License Information

### DIFF
--- a/lang/src/org/partiql/lang/prettyprint/ASTPrettyPrinter.kt
+++ b/lang/src/org/partiql/lang/prettyprint/ASTPrettyPrinter.kt
@@ -557,6 +557,8 @@ class ASTPrettyPrinter {
                     if (node.offset == null) it else (it.plusElement(toRecursionTree(node.offset, "offset")))
                 }
             )
+            // PR-COMMENT This is on par with the TODO that was in toRecursionTree(PartiqlAst.FromSource) overload.
+            is PartiqlAst.Expr.GraphMatch -> TODO("Unsupported GraphMatch AST node")
         }
 
     private fun toRecursionTreeList(nodes: List<PartiqlAst.Expr>, attrOfParent: String? = null): List<RecursionTree> =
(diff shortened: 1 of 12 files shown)


lang/src/org/partiql/lang/prettyprint/ASTPrettyPrinter.kt
```diff
@@ -557,6 +557,8 @@ class ASTPrettyPrinter {
                     if (node.offset == null) it else (it.plusElement(toRecursionTree(node.offset, "offset")))
                 }
             )
+            // PR-COMMENT This is on par with the TODO that was in toRecursionTree(PartiqlAst.FromSource) overload.
```
jpschorr, line 560: Can remove the `PR-COMMENT` line

jpschorr (approved):

Issue #799, opened 2022-09-23
Upgrade to ion-element 1.0.0


RCHowell:
We need to upgrade to the first major version of ion-element. In order to do so, we must upgrade pig to 0.6.0

- [x] #798 
- [x] https://github.com/partiql/partiql-isl-kotlin/issues/13
- [x] https://github.com/partiql/partiql-isl-kotlin/commit/234d23af318028ea487bd17330bce498e8a2ba55

Issue #798, opened 2022-09-23
Upgrade to partiql-ir-generator 0.6.0


RCHowell:
Upgrade to latest version of partiql-ir-generator which has ion-element 1.0

PR #805, opened 2022-09-24
Moves partiql-isl-kotlin to a library of partiql-lang-kotlin

RCHowell:
## Relevant Issues
- https://github.com/partiql/partiql-lang-kotlin/pull/725
- #799 
- #798 

## Description
This PR moves the partiql-isl-kotlin library from its own package to a package within partiql-lang-kotlin. It remains consumable as before, but including it as a gradle subproject enables easy simultaneous releases and consistent versioning across partiql-lang-kotlin and its dependencies owned by the partiql team.

## Other Information

No, this has no affect on the partiql-lang-kotlin artifact.

No

No

## License Information


lib/partiql-isl/build.gradle
```diff
@@ -0,0 +1,159 @@
+import org.partiql.pig.cmdline.Command
```
alancai98, line 1: Should we also add a task to the [GH actions workflow](https://github.com/partiql/partiql-lang-kotlin/blob/96d00f0cacaec80abc880e0ce49077fb6a7cbb48/.github/workflows/build.yml) to test `partiql-isl-kotlin`?

lib/partiql-isl/README.md
```diff
@@ -0,0 +1,93 @@
+## PartiQL ISL Kotlin
```
alancai98, line 1: For consistency, should the subproject still be named `partiql-isl-kotlin` rather than `partiql-isl`? The [maven](https://mvnrepository.com/artifact/org.partiql/partiql-isl-kotlin) artifact is still called `partiql-isl-kotlin`

alancai98 (commented):
A couple minor questions related to CI builds and directory naming. Otherwise looks good

lib/partiql-isl/build.gradle
alancai98: I see. Did we also want to move the [tests](https://github.com/partiql/partiql-isl-kotlin/tree/main/test) from the `partiql-isl-repo` to `partiql-lang-kotlin`?

lib/partiql-isl/README.md
RCHowell: I didn't want to suffix kotlin on modules within a kotlin package. Seems redundant.

For example, `lang` has artifact`partiql-lang-kotlin`, so do we rename the module to `partiql-lang-kotlin`?

lib/partiql-isl/build.gradle
RCHowell: Ah that's my miss. Turns out all the sources are in the `src` directory haha

lib/partiql-isl/README.md
alancai98: I get your point. Adding the kotlin suffix is somewhat redundant since it's already within a kotlin package. Fine w/ the current naming

lib/partiql-isl/build.gradle
RCHowell: Included

alancai98 (approved):

Issue #772, opened 2022-09-13
PSTS: Pull Specific Scalar Types Out Of Lexer & Parser


lziq:
Prerequisite refactoring is done in the following PRs: 
1. https://github.com/partiql/partiql-lang-kotlin/pull/685
2. https://github.com/partiql/partiql-lang-kotlin/pull/753 

This work covers 2 parts: 
1. Migrate type aliases out of lexer & parser (1 SDE week)
2. Remove `ExprValueType`'s dependency on type aliases (1 SDE day), https://github.com/partiql/partiql-lang-kotlin/pull/808

PR #808, opened 2022-09-25
Removed type names from ExprValueType

lziq:
## Background
In legacy parser, type names are considered as keywords and stored [here](https://github.com/partiql/partiql-lang-kotlin/blob/2e19331afc8e6b4bdfced0ec78dcdc81553e2e9d/lang/src/org/partiql/lang/syntax/LexerConstants.kt). The property [`ExprValueType.typeNames`](https://github.com/partiql/partiql-lang-kotlin/blob/2e19331afc8e6b4bdfced0ec78dcdc81553e2e9d/lang/src/org/partiql/lang/eval/ExprValueType.kt#L31) duplicates those type names and [this part of code](https://github.com/partiql/partiql-lang-kotlin/blob/2e19331afc8e6b4bdfced0ec78dcdc81553e2e9d/lang/src/org/partiql/lang/eval/ExprValueType.kt#L122-L139) ensures type names defined in lexer and `ExprValueType` are the same. 

## Motivation
As we are going to remove legacy parser, also, since we want to make type names (type aliases) user-defined in OTS, there is a need to remove the property `ExprValueType.typeNames` and the related hard-coded type names. 

## Changes
1. **Breaking Change**: Removed type names stored in `ExprValueType`. 
2. Defined `typeToTypeAliases` table locally in `CastTestBase.kt`

### DIFF
--- a/lang/test/org/partiql/lang/eval/CastTestBase.kt
+++ b/lang/test/org/partiql/lang/eval/CastTestBase.kt
@@ -42,25 +42,25 @@ data class NotImplemented(val expectedQuality: CastQuality) : CastQualityStatus(
  */
 data class FixSemantics(val expectedQuality: CastQuality) : CastQualityStatus()
 
-private val typeToTypeAliases = mapOf(
-    ExprValueType.MISSING to listOf("missing"),
-    ExprValueType.NULL to listOf("null"),
-    ExprValueType.BOOL to listOf("bool", "boolean"),
-    ExprValueType.INT to listOf("int", "smallint", "integer2", "int2", "integer", "integer4", "int4", "integer8", "int8", "bigint"),
-    ExprValueType.FLOAT to listOf("float", "real", "double precision"),
-    ExprValueType.DECIMAL to listOf("dec", "decimal", "numeric"),
-    ExprValueType.DATE to listOf("date"),
-    ExprValueType.TIMESTAMP to listOf("timestamp"),
-    ExprValueType.TIME to listOf("time"),
-    ExprValueType.SYMBOL to listOf("symbol"),
-    ExprValueType.STRING to listOf("string", "char", "varchar", "character", "character varying"),
-    ExprValueType.CLOB to listOf("clob"),
-    ExprValueType.BLOB to listOf("blob"),
-    ExprValueType.LIST to listOf("list"),
-    ExprValueType.SEXP to listOf("sexp"),
-    ExprValueType.STRUCT to listOf("struct", "tuple"),
-    ExprValueType.BAG to listOf("bag")
-)
+private fun ExprValueType.typeAliases(): List<String> = when (this) {
+    ExprValueType.MISSING -> listOf("missing")
+    ExprValueType.NULL -> listOf("null")
+    ExprValueType.BOOL -> listOf("bool", "boolean")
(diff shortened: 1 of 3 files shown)


lang/test/org/partiql/lang/eval/CastTestBase.kt
```diff
@@ -55,6 +42,26 @@ data class NotImplemented(val expectedQuality: CastQuality) : CastQualityStatus(
  */
 data class FixSemantics(val expectedQuality: CastQuality) : CastQualityStatus()
 
+private val typeToTypeAliases = mapOf(
```
alancai98, line 45: Rather than return a map, could this be modeled as a function that takes an `ExprValueType` and returns a list of strings? Not sure of the benefit of returning a map.

If you model it as a function, you could have a `when` expression over all the `ExprValueType`s, so if new `ExprValueType`s are added in the future, we can make sure the corresponding branch in the `when` expression is updated.

alancai98 (commented):
Could you provide some more context regarding why this change is needed and how it relates to the open type system work? I saw this PR is part of a broader issue of https://github.com/partiql/partiql-lang-kotlin/issues/772, but I'm having trouble understanding how it fits into the rest of changes.

lang/test/org/partiql/lang/eval/CastTestBase.kt
lziq: > Could you provide some more context regarding why this change is needed and how it relates to the open type system work? I saw this PR is part of a broader issue of #772, but I'm having trouble understanding how it fits into the rest of changes.
> 
> Also, why is this change being pushed to `main` rather than to the OTS feature branch?

In OTS, we are going to enable users to define and customize type aliases for a specific type. The code here hard-codes all the type aliases, which is contradicting against what we want in OTS. 

As for why we want to merge this PR to `main`, this is because I still consider this as a refactoring to the existing code. And in general, I want to push such changes to `main` as many as possible since it reduces workload to maintain the OTS feature branch.

lziq:
> Could you provide some more context regarding why this change is needed and how it relates to the open type system work? I saw this PR is part of a broader issue of #772, but I'm having trouble understanding how it fits into the rest of changes.
> 
> Also, why is this change being pushed to `main` rather than to the OTS feature branch?

In OTS, we are going to enable users to define and customize type aliases for a specific type. The code here hard-codes all the type aliases, which is contradicting against what we want in OTS. 

As for why we want to merge this PR to `main`, this is because I still consider this as a refactoring to the existing code. And in general, I want to push such changes to `main` as many as possible since it reduces workload to maintain the OTS feature branch.

lang/test/org/partiql/lang/eval/CastTestBase.kt
lziq: I will apply this comment.

alancai98 (dismissed):
Looks good. Need to call out somewhere (PR description and CHANGELOG) that this is a breaking change since `ExprValueType` and its property `typeNames` were public APIs

alancai98:
Slightly unrelated, but in the PR description, your links point to code currently `main`. After this PR is merged, some of those links will not contain the relevant code artifacts. To keep the previous context, it may be helpful to link code with a specific code reference. This can be done by going to the most recent commit. E.g. 
1. Go to https://github.com/partiql/partiql-lang-kotlin/commit/2e19331afc8e6b4bdfced0ec78dcdc81553e2e9d
2. Browse files
3. Link corresponding file or code artifact

alancai98 (approved):

PR #812, opened 2022-09-27
Follow up on PR 774 comments

lziq:


lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -414,20 +403,14 @@ internal class StaticTypeInferenceVisitorTransform(
...
-                Int4Type,
-                Int8Type,
-                IntType,
-                FloatType -> type
+        private fun getUnaryArithmeticOpInferencer(scalarOp: ScalarOp) = { type: SingleType ->
+            when {
+                // Propagate NULL or MISSING
+                type is MissingType -> StaticType.MISSING
+                type is NullType -> StaticType.NULL
+                type is StaticScalarType -> scalarOp.inferReturnType(listOf(type.toCompileTimeType())).toStaticType()
```
lziq, line 411: I forgot to cover these changes in the previous PR.

lang/src/OTS/ITF/org/partiql/ots/operator/ScalarOp.kt
```diff
@@ -1,12 +1,40 @@
...
+abstract class BinaryPlusOp : ScalarOp
+abstract class BinaryMinusOp : ScalarOp
+abstract class BinaryTimesOp : ScalarOp
+abstract class BinaryDivideOp : ScalarOp
+abstract class BinaryModuloOp : ScalarOp
+abstract class BinaryConcatOp : ScalarOp
+abstract class LikeOp : ScalarOp {
+    override val defaultReturnTypes: List<CompileTimeType>
+        get() = listOf(BoolType.compileTimeType)
 }
```
vgapeyev, line 40: I might overlook something, but these abstract classes inheriting from `ScalarOp` appear useless.  I believe, all their uses can be replaced with `ScalarOp` itself.  

[For another PR]
Now, if the above succeeds for the operators, it would be worthwhile to explore whether the types in the plugin can be architected similarly: 

- Note that singleton objects in `...ots.legacy.types` already inherit from a single interface `ScalarType`, without abstract types in between.  So, they are "ahead" of the operators in this regard. 

- However, they are "behind", in that `Plugin` and `StandardPlugin` do not list them as value fields, as they do for the operators.  I hope this can be done, which should eliminate the inappropriate references to them from core PartiQL (see, say, the import of Int2Type in StaticTypeInferenceVisitorTransform).

lang/src/OTS/IMP/org/partiql/ots/legacy/plugin/StandardPlugin.kt
```diff
@@ -22,16 +42,14 @@ import OTS.ITF.org.partiql.ots.operator.LikeOp
 import OTS.ITF.org.partiql.ots.operator.NegOp
 import OTS.ITF.org.partiql.ots.operator.NotOp
 import OTS.ITF.org.partiql.ots.operator.PosOp
+import OTS.ITF.org.partiql.ots.type.BoolType
 import com.amazon.ion.Timestamp
 import java.time.ZoneOffset
 
 data class StandardPlugin(
```
vgapeyev, line 49: Separately, let's switch from the "standard" to the "legacy" lexeme here as well.

lang/src/OTS/IMP/org/partiql/ots/legacy/operators/StandardBinaryConcatOp.kt
```diff
@@ -17,7 +17,12 @@ object StandardBinaryConcatOp : BinaryConcatOp() {
     override val defaultReturnTypes: List<CompileTimeType> =
```
vgapeyev, line 17: This is getting into matters of taste, but I wouldn't bother neither about "Standard" nor "Legacy" in the names of the operator implementations (`StandardBinaryConcatOp` here) -- they are scoped inside the legacy plugin instance, which should be enough.

vgapeyev (approved):
I feel this corner of the plugin interface is not starting to take expected shape! 

If you understand and agree with the suggestions in the review, feel free to merge after making the corresponding changes. (Except for the one marked for another PR.)

lang/src/OTS/ITF/org/partiql/ots/operator/ScalarOp.kt
lziq: Do you think we can add `operatorId` to `ScalarOp` and remove these abstract classes?

lang/src/OTS/ITF/org/partiql/ots/operator/ScalarOpId.kt
```diff
@@ -0,0 +1,14 @@
...
+    POS,
+    NEG,
+    PLUS,
+    MINUS,
+    TIMES,
+    DIVIDE,
+    MODULO,
+    CONCAT,
+    LIKE
+}
```
vgapeyev, line 14: @lziq , this is probably an omission after we discussed this. Please take this out. It is not being used anywhere.  To experiment with this idea, please use a separate branch/PR.

vgapeyev: No rush.  It is ok to do as part of follow-up work in either PR816 or 818.

lziq: Removed it here, https://github.com/partiql/partiql-lang-kotlin/pull/818/files

PR #816, opened 2022-09-28
Re-model type parameters

lziq:
*Changes:*
1. Changed the semantics of `TypeParameter`. Previously, `null` in the list indicates that a parameter is not explicitly specified. Now, the list just represents whatever the type parameters are in the original query. e.g.

| Type | Previously | Now |
|---|---|---|
| `DECIMAL` | `[nul, null]` | `[]` |
| `DECIMAL(1)` | `[1, null]` | `[1]` |
| `DECIMAL(2, 1)` | `[2, 1]` | `[2, 1]` |

2. Also added some helper classes, .e.g. `DecimalTypeParameters`, to help initialize type parameters in legacy plugin package. 

*For Review:*
1. Don't review this PR until https://github.com/partiql/partiql-lang-kotlin/pull/812 is merged
2. A good starting point is to good at `TypeParameter` first, then helper classes, e.g. `DecimalTypeParameters`, defined in the specific scalar types. Other changes are just to make it compile and make tests pass, which also show how to use those type parameter helper classes.


vgapeyev:
Re: "Don't review this PR until https://github.com/partiql/partiql-lang-kotlin/pull/812 is merged", for future situations like this: 

If this PR816 is developed on top of PR812, then, to avoid the suspense, it (PR816) should be opened w.r.t. the latter's branch.  That is, PR816 should be w.r.t. `pr-774-followup` branch, rather than `new-PSTS-FB`. 

However, I see that PR816's branch, `remodel-type-parameters` and `pr-774-followup` have diverged slightly.  This can be helped by merging `pr-774-followup` into `remodel-type-parameters`. As long as merge commit is used for the manipulation, this would merely assure that `remodel-type-parameters` is a successor of `pr-774-followup`, as intended.

Now, that PR812 might be close to be merged, the above does not matter as much, but I'd recommend to do these manipulations, just for fun and experience, prior to addressing the comments in PR812.

lang/src/OTS/IMP/org/partiql/ots/legacy/operators/LegacyBinaryConcatOp.kt
```diff
@@ -33,8 +33,15 @@ object LegacyBinaryConcatOp : ScalarOp {
...
-        val rightLength = rType.parameters[0]!!
+        fun checkUnlimitedVarchar(type: CompileTimeType) =
+            type.scalarType === VarcharType && type.parameters.isEmpty()
+
+        if (checkUnlimitedVarchar(lType) || checkUnlimitedVarchar(rType)) {
+            return Successful(StringType.compileTimeType)
+        }
+
+        val leftLength = lType.parameters.getOrElse(0) { 1 }
+        val rightLength = rType.parameters.getOrElse(0) { 1 }
```
lziq, line 44: Let's not focus on code quality in each operator/function invoker, since anyway it will be refactored once we are working on supporting function overloading

lziq: This comment applies to other operator/function invoker as well

vgapeyev: My bigger concern here that this might be introducing new logic that differs from what is on the main branch.

lziq: I think there is no logic change here

lziq: Line 43 & 44 specify the default type parameter of `CharType` to be 1, which is the same as `main` branch.

lziq: The only difference is that in `main`, there is no unlimited `VarcharType`. Unlimited `VarcharType` there is considered as `StringType`.

lziq: So something like `CAST(1 AS VARCHAR)` is actually `CAST(1 AS STRING)` in `main`.

vgapeyev (approved):
The treatment of parameterized types is one of the areas where we did not succeed in properly examining the OTS-side decisions, sadly.  I fear that new designs and logic were being introduced that differed from those on the mainline-side, while I did not get convinced those changes were necessary for the OTS-1st-phase refactoring per se. 
(For example, I just checked with mainline's StaticType -- it does not have CharType as a subclass.)

So, for now, it's up to you what's happening here, but I'd prefer this (as any other!) aspect of "language concepts" on the OTS side to move in the direction of restoring "sameness" with the mainline-side rather than improving, however flawed, designs there. (Though, if the sameness is restored by making the improvements on mainline instead -- that's fair game.)

The improvement task would be another phase of the OTS project, where we'd implement a different, "better" plugin as an alternative to the "legacy" plugin being pulled out currently.

lziq:
And I admit that probably I was thinking too much on refactoring static string types & static integer types. The work there is a must for the function overloading goal, not pluggability goal. But since we already merged the changes into `new-PSTS-FB` branch, I will see if there is chance we can merge https://github.com/partiql/partiql-lang-kotlin/pull/693 & https://github.com/partiql/partiql-lang-kotlin/pull/701 to `main`

Issue #800, opened 2022-09-23
Add support for PIVOT to planner


RCHowell:
## Relevant Issue/Bug
#758

## Requested Solution/Feature
The planner needs to support the PIVOT clause.

## Describe Alternatives
NA

## Additional Context
...

## DoD (Definition of Done)
All evaluator tests with a PIVOT clause are passing

PR #817, opened 2022-09-28
Adds PIVOT projection to PartiQLPlanner

RCHowell:
## Relevant Issues
#800 

## Description
Adds PIVOT projection expression to the logical domain

## Other Information
Yes. PIVOT support in the planner is a new, unreleased change so it has been included in the CHANGELOG.md

No

No

## License Information

### DIFF
--- a/lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
+++ b/lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
@@ -266,6 +266,7 @@ internal class PhysicalPlanCompilerImpl(
             // bag operators
             is PartiqlPhysical.Expr.BagOp -> compileBagOp(expr, metas)
             is PartiqlPhysical.Expr.BindingsToValues -> compileBindingsToValues(expr)
+            is PartiqlPhysical.Expr.Pivot -> compilePivot(expr, metas)
         }
     }
 
@@ -1851,6 +1852,26 @@ internal class PhysicalPlanCompilerImpl(
                 is PartiqlPhysical.SetQuantifier.Distinct -> op.eval(l, r).distinct()
             }
             valueFactory.newBag(result)
+        }
+    }
+
+    private fun compilePivot(expr: PartiqlPhysical.Expr.Pivot, metas: MetaContainer): PhysicalPlanThunk {
+        val inputBExpr: RelationThunkEnv = bexperConverter.convert(expr.input)
+        // The names are intentionally flipped for clarity; consider fixing this in the AST
+        val valueExpr = compileAstExpr(expr.key)
+        val keyExpr = compileAstExpr(expr.value)
+        return thunkFactory.thunkEnv(metas) { env ->
+            val attributes: Sequence<ExprValue> = sequence {
+                val relation = inputBExpr(env)
+                while (relation.nextRow()) {
+                    val key = keyExpr.invoke(env)
+                    if (key.type.isText) {
+                        val value = valueExpr.invoke(env)
+                        yield(value.namedValue(key))
+                    }
+                }
+            }
+            valueFactory.newStruct(attributes, StructOrdering.UNORDERED)
         }
     }
 
--- a/lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
+++ b/lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
@@ -59,51 +59,46 @@ internal class AstToLogicalVisitorTransform(
             PartiqlLogical.build { limit(transformExpr(it), algebra, node.limit.metas) }
         } ?: algebra
 
-        return convertProjectionToBindingsToValues(node, algebra)
+        val expr = transformProjection(node, algebra)
+
+        // SELECT DISTINCT ...
+        if (node.setq != null && node.setq is PartiqlAst.SetQuantifier.Distinct) {
+            return PartiqlLogical.build { call("filter_distinct", expr) }
+        }
+
+        return expr
     }
 
-    private fun convertProjectionToBindingsToValues(node: PartiqlAst.Expr.Select, algebra: PartiqlLogical.Bexpr) =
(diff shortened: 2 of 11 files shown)


lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
```diff
@@ -1847,6 +1848,23 @@ internal class PhysicalPlanCompilerImpl(
...
+        val inputBExpr: RelationThunkEnv = bexperConverter.convert(expr.input)
+        val keyExpr = compileAstExpr(expr.key)
+        val valueExpr = compileAstExpr(expr.value)
+        return thunkFactory.thunkEnv(metas) { env ->
+            val attributes: Sequence<ExprValue> = sequence {
+                val relation = inputBExpr(env)
+                while (relation.nextRow()) {
+                    val key = keyExpr.invoke(env)
+                    val value = valueExpr.invoke(env)
+                    yield(key.namedValue(value))
```
yliuuuu, line 1861: Per Spec Section 14:
```
PIVOT v AT a
....
The expression a is expected to evaluate into a string value. In strict mode, it is an error if this evaluates to a
non-string value. In permissive mode, the attribute is considered MISSING and does not appear in the output. The
expression v can be any PartiQL value, but if it is MISSING it will not be generated in the resulting tuple.
```

Consider: 
`PIVOT x.v AT i FROM [ { 'a' : 'first', 'v' : 'John' }, {'a' : 'last', 'v' : 'doe' } ] AS x AT i` 
Two things to note here: 
1. The current evaluating compiler returns an empty struct in `LEGACY` mode, which arguably should be an error. 
2. The current logic in the PR produces `{'null': 'John', 'null': 'doe'}`. 

Some additional logic may be need here, something like
```kotliin
if (key.type.isText) {
      val value = valueExpr.invoke(env)
      yield(key.namedValue(value))
}
else{
     // yield missing to conform with the current behavior
     yield(valueFactory.MissingValue)
}
```

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -54,51 +54,46 @@ internal class AstToLogicalVisitorTransform(
...
+                is PartiqlAst.Projection.ProjectStar -> {
+                    // `SELECT * FROM bar AS b` is rewritten to `SELECT b.* FROM bar as b` by
+                    // [SelectStarVisitorTransform]. Therefore, there is no need to support `SELECT *` here.
+                    errAstNotNormalized("Expected SELECT * to be removed")
+                }
+                is PartiqlAst.Projection.ProjectPivot -> {
+                    pivot(
+                        input = algebra,
+                        key = transformExpr(project.key),
+                        value = transformExpr(project.value),
```
yliuuuu, line 93: The choice of key and value seems a bit unfortunate for me, but you don't have to change it.

RCHowell: I agree, the naming is backwards. The unfortunate naming broke my brain a little haha, but unfortunately changing it is a breaking change.

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -54,51 +54,46 @@ internal class AstToLogicalVisitorTransform(
...
+            return PartiqlLogical.build { call("filter_distinct", expr) }
+        } else expr
+
+        return expr
     }
 
-    private fun convertProjectionToBindingsToValues(node: PartiqlAst.Expr.Select, algebra: PartiqlLogical.Bexpr) =
+    private fun transformProjection(node: PartiqlAst.Expr.Select, algebra: PartiqlLogical.Bexpr) =
         PartiqlLogical.build {
-            bindingsToValues(
```
johnedquinn, line 62: Ty very much for cleaning this function up.

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -54,51 +54,46 @@ internal class AstToLogicalVisitorTransform(
             PartiqlLogical.build { limit(transformExpr(it), algebra, node.limit.metas) }
         } ?: algebra
 
-        return convertProjectionToBindingsToValues(node, algebra)
+        var expr = transformProjection(node, algebra)
+
+        // SELECT DISTINCT ...
+        expr = if (node.setq != null && node.setq is PartiqlAst.SetQuantifier.Distinct) {
+            return PartiqlLogical.build { call("filter_distinct", expr) }
+        } else expr
```
johnedquinn, line 62: This is a bit odd.

```suggestion
        expr = if (node.setq != null && node.setq is PartiqlAst.SetQuantifier.Distinct) {
            PartiqlLogical.build { call("filter_distinct", expr) }
        } else expr
```

OR

```suggestion
        if (node.setq != null && node.setq is PartiqlAst.SetQuantifier.Distinct) {
            return PartiqlLogical.build { call("filter_distinct", expr) }
        }
```

johnedquinn (commented):
Overall very clean. Just echoing @yliuuuu's comment regarding PIVOT keys as string values.

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
RCHowell: I don't know what I was thinking hahaha thanks

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
RCHowell: https://i.imgur.com/tAgcJBD.jpg

yliuuuu (dismissed):
LGTM!

yliuuuu (approved):

PR #818, opened 2022-09-28
Refactor `inferPathComponentExprType` to remove dependency on specific integer types in type inferencer

lziq:
Don't review this PR until https://github.com/partiql/partiql-lang-kotlin/pull/812 is merged


vgapeyev:
See comment in PR816 on choosing the base branch for a PR. 

To follow up on the suggestion there about dealing with divergence, I guess some of it can be dealt with by rebasing and force-pushing as well.  I can confirm that I have not yet checked out local branches for the newer PRs, 816 and 818, so that should not affect me.

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -1463,8 +1459,9 @@ internal class StaticTypeInferenceVisitorTransform(
                 is ListType,
                 is SexpType -> {
                     val previous = previousComponentType as CollectionType // help Kotlin's type inference to be more specific
-                    val staticType = currentPathComponent.index.getStaticType()
-                    if (staticType is StaticScalarType && (staticType.scalarType in listOf(Int2Type, Int4Type, Int8Type, IntType))) {
+                    val staticType = currentPathComponent.index.getStaticType() as SingleType
+                    // As long as the expression surrounded by brackets can be evaluated as an integer scalar, we don't care which type it is at compile time
```
lziq, line 1463: Here, "brackets" mean the bracket notation we use to access an item in a list/sexp, e.g. `sexp(1, 2, 3)[0]`

vgapeyev: It's worth folding this explanation into the code comment.

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -4,10 +4,6 @@
 
 package org.partiql.lang.eval.visitors
 
-import OTS.IMP.org.partiql.ots.legacy.types.Int2Type
-import OTS.IMP.org.partiql.ots.legacy.types.Int4Type
-import OTS.IMP.org.partiql.ots.legacy.types.Int8Type
-import OTS.IMP.org.partiql.ots.legacy.types.IntType
```
vgapeyev, line 10: If the driving purpose of this PR was to eliminate the above "inappropriate" references to `OTS.IMP`, then note that, after PR812, this should no longer be a problem: it should be possible to refer to these "it type descriptors" as fields in `Plugin` interface. 

I did not dig in to really understand the solution here or whether there was a reason for it different from the above.  But among the two, I'd prefer the solution that is closer to the corresponding design on mainline. 

(On the face of it, the solution here is design-wise different from the one in the previous commit: switching from a "static type" to a "runtime type"; I do not yet understand the codebase enough to judge whether this matters.  And I do not know which of these two is more in agreement with the mainline.)

vgapeyev (approved):
Up to you whether to change this before merging, based on the comment.

lang/src/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
lziq: My opinion for brackets notation to access an element in a list is that the integer value it takes is different from the data of integer types we are querying against. It is something similar to type parameters, which we require to be integer literals syntactically. I don't think we can require it to be also integer literals syntactically, but it is something special, and I don't think data types should be involved with it. 

The reason I did this refactoring is simply that we need to remove specific scalar types out of type inferencer and I think this refactoring of semantics does not hurt anything as long as it works. After all, I think nobody would care which static type it is for the expression inside the brackets notation.

lziq: We might need to add a new static type `StaticType.INTEGER_LITERAL` which works as the static type of a PartiQL-native integer literal. We use it on, e.g. return type of `COUNT`, inferred type of the variable in `AT` clause, integer literals written in queries, etc., and we here we require the expression inside the bracket notation to be `StaticType.INTEGER_LITERAL`. And also, in the plugin interface, we require user to provide the corresponding scalar type of it (which scalar type `StaticType.INTEGER_LITERAL` should be transformed as).

PR #821, opened 2022-09-30
Adds support for aggregations and grouping to the planner

### LINKED ISSUE #794: Add support for GROUP BY and AGGs to Planner
## Relevant Issue/Bug
- Reference Issue #758 

## Requested Feature
- To reach feature parity with the evaluating compiler, the new planner needs to add support for GROUP BY and AGGREGATIONS

## Description
- I'm using this issue as a means to collect my own thoughts (and the thoughts of others) before implementing a solution.
- Regardless, these features are required to reach feature parity.

## Additional Context
Need to:
- [ ] Figure out the best way to implement

## DoD (Definition of Done)
- GROUP BY and aggregations working for planner
- All existing GROUP BY tests passing

johnedquinn:
## Relevant Issues
- Closes #794

## TLDR
- Adds support for an `AGGREGATE` relational operator to support GROUP BY, GROUP AS, and Aggregations in Projections.

## New Relational Operator
- We introduce an `AGGREGATE` relational operator that takes in a few things, namely:
  - Group Keys: groups the aggregations on these keys and outputs the keys to the VarDecls at the end.
  - Aggregation Functions: performs partial and full aggregations and places the results in VarDecls

## Transforms
- The AST transformation goes something like this:
  - Example Query: `SELECT k, MAX(b), g FROM <<{'a': 0, 'b': 1}>> t GROUP BY t.a AS k GROUP AS g`
  - We check by the existence of GROUP or aggregation functions in the projection list
  - The GroupKeys get appended to them a meta containing their unique name (something like `$partiql_group_0`)
  - The projections that are not in aggregation functions get transformed to reference the group keys new unique names
  - The aggregation functions are gathered to be replaced with references to VarDecls
  - Then, the AST is transformed into a Logical Plan
    - The AST projection items are now only references (IDs) that point to the output of the new `AGGREGATION` operator.
    - We take the VarDecls of the GroupKeys, GroupAs function, and other aggregation functions and create a `AGGREGATE` operator.

## Other Information
  - Not yet.
- I haven't yet added tests for each class -- but I made sure everything was passing in the existing GroupByTests.
- Also, **NOTE**: since we are actually aggregating via accumulation, a query like `AVG([1, 2, 3])` doesn't work. I can work on a PR or add the scope to this one.

## License Information


lang/test/org/partiql/lang/eval/EvaluatingCompilerGroupByTest.kt
```diff
@@ -103,6 +102,16 @@ class EvaluatingCompilerGroupByTest : EvaluatorTestBase() {
             val compilationOptions: List<CompOptions> = CompOptions.values().toList()
         )
 
+        /**
+         * The [EvaluatorTestTarget.PLANNER_PIPELINE] does NOT support [UndefinedVariableBehavior.MISSING], so if the
+         * [compOptions] includes the [UndefinedVariableBehavior], we should use the [EvaluatorTestTarget.COMPILER_PIPELINE]
+         */
+        private fun getTestTarget(compOptions: CompOptions, default: EvaluatorTestTarget): EvaluatorTestTarget = when (compOptions) {
```
johnedquinn, line 109: See [the PartiQLCompilerPipelineFactory's note on this](https://github.com/partiql/partiql-lang-kotlin/blob/11c265e08f45abb7be7ebaa5f51fe58aa4bc0eb3/lang/test/org/partiql/lang/eval/evaluatortestframework/PartiQLCompilerPipelineFactory.kt#L39).

johnedquinn:
Seems like the recent addition of the SORT operator evaluated the query twice -- so I added a `IsOrderedMeta` to prevent that. Caught it after my tests started failing.

johnedquinn:
Rebased ^

lang/src/org/partiql/lang/eval/physical/operators/AggregateOperatorFactory.kt
```diff
@@ -0,0 +1,172 @@
...
+        RelationExpression { state ->
+            val sourceIter = source.evaluate(state)
+
+            val aggregationMap = AggregationMap()
+            val seen = List(functions.size) { TreeSet(DEFAULT_COMPARATOR) }
+            val registerSize = state.registers.size
+            val variables = functions.flatMap { function -> function.variables } + keys.map { key -> key.variable }
+            val variablesIndexes = variables.map { variable -> variable.index.value.toIntExact() }.toSet()
+
+            while (sourceIter.nextRow()) {
```
yliuuuu, line 73: I suspect the "double evaluation" and other troubles are caused here.

Consider: 
```
internal class DefaultFooOperator(name: string): FooOperatorFactory(name){
       override fun create(
              impl: PartiqlPhysical.Impl,
             source: RelationExpression
        ) : RelationExpression = {
        // recall that relationExpression. eval -> RelationIterator
                 RelationExpression { state -> 
                       // preparation work
                       val sourceIter = source.evalute(state)
                       val someMaterializedColl = while(sourceIter.nextRow()){
                        // .... do some materialization and preparation work
                       }
                       // the relation iterator produced by operatorFoo, calling eval on operate won't going into the RelationScope defined
                       relation(sourceIter.Type){
                                val iter = someMaterializedColl.iterator()
                                while(iter.hasNext()){
                                       val row = iter.next()
                                       // processing and set up state
                                       yield()
                                }
                       }

                 }
        }
}
```

One observation is that in line 277 of the original code, we call `relationTypeThunk = bexprThunk(env)`. Later, we have another call `val relItr = bexprThunk(env)` (Line 281 in the original Commit). 

For the `foo` operator, it essentially materializes the source twice. I am not sure how to reproduce the exact issue( since the original commit is missing for some reasons?), but I suspect this is the root cause of the issue. 

So one simple fix can be, move the preparation work into the relation scope. 

Let me know if this is helpful.

yliuuuu (commented):
Leave some note on the double evaluation issue, will need more time to review the end-to-end logic here. Let me know if the comment is helpful.

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -636,6 +636,26 @@ may then be further optimized by selecting better implementations of each operat
...
+            )
+            (product aggregate_function
+                quantifier::set_quantifier
+                name::symbol
+                arg::expr
+                as_vars::(* var_decl 1)
+            )
+            (product aggregate_function_list
+                functions::(* aggregate_function 1)
+            )
```
yliuuuu, line 656: Can we remove this so we don't need the nested `functions.functions`

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -668,6 +688,13 @@ may then be further optimized by selecting better implementations of each operat
                 // rows for which `predicate` returns true.
                 (filter predicate::expr source::bexpr)
 
+                (aggregate
```
yliuuuu, line 691: Worth to leave some comments. I think the point that we treat `GROUP AS` as a special aggregate function is quite a big change and may worth to explain here.

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -668,6 +688,13 @@ may then be further optimized by selecting better implementations of each operat
...
+                     source::bexpr
+                     strategy::grouping_strategy
+                     groups::group_key_list
+                     functions::aggregate_function_list
```
yliuuuu, line 695:
```suggestion
                     functions::(* aggregate_function 1)
```

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -874,6 +900,15 @@ may then be further optimized by selecting better implementations of each operat
                     predicate::(? expr))
 
                 (sort i::impl source::bexpr sort_specs::(* sort_spec 1))
+
+                (aggregate
+                    i::impl
+                    source::bexpr
+                    strategy::grouping_strategy
+                    groups::group_key_list
+                    functions::aggregate_function_list
```
yliuuuu, line 909:
```suggestion
                    functions::(* aggregate_function 1)
```

lang/src/org/partiql/lang/eval/physical/operators/AggregateOperatorFactory.kt
```diff
@@ -0,0 +1,172 @@
...
+        functions: List<CompiledAggregateFunction>
+    ): RelationExpression
+
+    public abstract fun getFunctionFactory(
+        operator: AggregationKind,
+        name: String
+    ): AggregationFunctionFactory
+}
+
+internal class FullAggregateFactory(name: String) : AggregateOperatorFactory(name) {
```
yliuuuu, line 42: Seems like a hash based approach? Mind explain a little more on the meaning of `FullAggregate`?

lang/src/org/partiql/lang/eval/physical/operators/AggregateOperatorFactory.kt
```diff
@@ -0,0 +1,172 @@
...
+
+    private val aggregationFunctionFactories = mapOf(
+        AggregationFunctionFactoryKey(AggregationKind.MAX, "MAX") to MaxFactory("MAX"),
+        AggregationFunctionFactoryKey(AggregationKind.MIN, "MIN") to MinFactory("MIN"),
+        AggregationFunctionFactoryKey(AggregationKind.COUNT, "COUNT") to CountFactory("COUNT"),
+        AggregationFunctionFactoryKey(AggregationKind.SUM, "SUM") to SumFactory("SUM"),
+        AggregationFunctionFactoryKey(AggregationKind.GROUP_AS, "GROUP_AS") to GroupAsFactory("GROUP_AS"),
+        AggregationFunctionFactoryKey(AggregationKind.AVG, "AVG") to AverageFactory("AVG"),
+    )
```
yliuuuu, line 51: This is interesting. 
In my opinion, this approach is similar to our aggregate function implementation in evaluating compiler. Only that now with the planner, we can have multiple `aggregate` operator, i.e., sort based aggregate, hash based aggregate, and we have to map those function to each one of the operator implementation. 

It will be interesting to design a factory so that we can essentially pass in the factory and called `.createFunction()`. This is actually something that I am experimenting with the window functions. 

Anyways, just want to share some of my thoughts here. I am good with this implementation so far. No blocker.

lang/src/org/partiql/lang/eval/physical/operators/AggregateOperatorFactory.kt
```diff
@@ -0,0 +1,172 @@
...
+
+                // Initialize the AggregationMap
+                val isMissingArray = Array(keys.size) { false }
+                val evaluatedGroupByKeys = keys.mapIndexed { index, groupByKey ->
+                    val ionValue = groupByKey.value.invoke(state).ionValue
+                    if (ionValue.isNullValue) {
+                        isMissingArray[index] = ionValue.hasTypeAnnotation(MISSING_ANNOTATION)
+                        ionValue.removeTypeAnnotation(MISSING_ANNOTATION)
+                    }
+                    ionValue
```
yliuuuu, line 83: Could you please explain a little more on why do we need to use ionValue here? (as opposite to a `sexp ExprValue`)

lang/src/org/partiql/lang/eval/physical/operators/AggregateOperatorFactory.kt
```diff
@@ -0,0 +1,172 @@
...
+                val groupMapValue = aggregationMap.getOrPut(evaluatedGroupByKeys) {
+                    GroupMapValue(
+                        state = EvaluatorState(
+                            state.session,
+                            state.valueFactory,
+                            Array(registerSize) { index ->
+                                when (variablesIndexes.contains(index)) {
+                                    true -> state.valueFactory.missingValue
+                                    false -> state.registers[index]
+                                }
```
yliuuuu, line 94: I am not sure I follow the logic here, mind explain a little more?

lang/src/org/partiql/lang/eval/physical/operators/AggregateOperatorFactory.kt
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * A [LinkedHashMap] indexed by a [List] of [IonValue]s which implement [equals] and [hashCode]. This allows us to index
+ * into the map after evaluating the [CompiledGroupKey]. Then, the [GroupMapValue] is a representation of the total
+ * aggregations, and it holds metadata to help differentiate between [ExprValueType.MISSING] and [ExprValueType.NULL]
+ */
+private typealias AggregationMap = LinkedHashMap<List<IonValue>, GroupMapValue>
+
+public class CompiledGroupKey(
+    val function: SetVariableFunc,
```
yliuuuu, line 156:
```suggestion
    val setGroupKeyVal: SetVariableFunc,
```

lang/src/org/partiql/lang/eval/physical/operators/AggregateOperatorFactory.kt
```diff
@@ -0,0 +1,172 @@
...
+    val value: ValueExpression,
+    val variable: PartiqlPhysical.VarDecl
+)
+
+public class CompiledAggregateFunction(
+    val function: SetVariableFunc,
```
yliuuuu, line 162:
```suggestion
    val setAggregateVal: SetVariableFunc,
```

function.function is quite difficult to read.

lang/src/org/partiql/lang/eval/physical/operators/AggregationFunctionFactory.kt
```diff
@@ -0,0 +1,203 @@
...
+                initializeMissingRegisterWithNull(state, indexAvg)
+                return@thunk
+            }
+            checkIsNumberType(funcName = "AVG", value = value)
+            conditionallyInitializeRegister(state, indexAvg, state.valueFactory.newInt(0))
+            conditionallyInitializeRegister(state, indexSum, state.valueFactory.newInt(0))
+            conditionallyInitializeRegister(state, indexCount, state.valueFactory.newInt(0))
+            state.registers[indexSum] = (state.registers[indexSum].numberValue() + value.numberValue()).exprValue(state.valueFactory)
+            state.registers[indexCount] = (state.registers[indexCount].numberValue() + 1L).exprValue(state.valueFactory)
+            state.registers[indexAvg] = (state.registers[indexSum].numberValue() / bigDecimalOf(state.registers[indexCount].numberValue())).exprValue(state.valueFactory)
```
yliuuuu, line 82: A little bit counter-intuitive as we store intermediate result into the registers, and even more so with `AVG`...
As a by-product, this is a lot of local variables created. 
That said, of the top of my head, I can't really think of another way to do so without a large refactor... 
May be worth to add some comments explains this approach?

lang/src/org/partiql/lang/eval/physical/operators/AggregateOperatorFactory.kt
johnedquinn: Well, instead of performing a partial aggregation and then a final aggregation, I opted to create a "Full Aggregation" as it's easier to implement. Just some naming thing, but I'm not pressed on keeping the name.

lang/src/org/partiql/lang/eval/physical/operators/AggregateOperatorFactory.kt
johnedquinn: IonValue implements equals and hashcode, so we can use them as keys within a map.

lang/src/org/partiql/lang/eval/physical/operators/AggregateOperatorFactory.kt
johnedquinn: Eh it might actually not be necessary anymore -- but I essentially only was initializing the output variable declarations with MISSING, but I can check if that's still necessary.

lang/src/org/partiql/lang/eval/physical/operators/AggregationFunctionFactory.kt
johnedquinn: Yeah, this is related to Partial vs Final vs Full Aggregation. I'll leave a comment.

lang/resources/org/partiql/type-domains/partiql.ion
johnedquinn: Modified so that (aggregate) uses `functionList` instead to keep it consistent with group keys. LMK if that's sufficient. Now, it should be `functionList.functions`

lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
```diff
@@ -274,10 +275,12 @@ internal class PhysicalPlanCompilerImpl(
         val mapThunk = compileAstExpr(expr.exp)
         val bexprThunk: RelationThunkEnv = bexperConverter.convert(expr.query)
 
-        return thunkFactory.thunkEnv(expr.metas) { env ->
-            val relationTypeThunk = bexprThunk(env)
-            val relationType: RelationType = relationTypeThunk.relType
+        val relationType = when (expr.metas.containsKey(IsOrderedMeta.TAG)) {
```
yliuuuu, line 278: Could you please confirm that 
```
SELECT * FROM (SELECT a FROM <<{'a':1}>> as tbl ORDER BY tbl.a) as tbl2 
```
Returns a `BAG`. 

I wasn't too sure if the what the `isOrderedMeta` is going to be outside of the subquery.

lang/src/org/partiql/lang/eval/physical/operators/Accumulator.kt
```diff
@@ -0,0 +1,184 @@
...
+    override fun nextValue(value: ExprValue) {
+        checkIsNumberType(funcName = "SUM", value = value)
+        if (sum == null) sum = 0.0
+        this.sum = value.numberValue() + this.sum!!
+    }
+
+    override fun compute(): ExprValue {
+        val totalSum = sum?.let { bigDecimalOf(it) }
+        return totalSum?.exprValue(valueFactory) ?: valueFactory.nullValue
+    }
```
yliuuuu, line 77: Nit: It seems like the sum function will always return `Decimal` Type. In the evaluating compiler, sum may return an integer or float type.

lang/src/org/partiql/lang/eval/physical/operators/Accumulator.kt
```diff
@@ -0,0 +1,184 @@
...
+private fun comparisonAccumulator(comparator: NaturalExprValueComparators): (ExprValue?, ExprValue) -> ExprValue =
+    { left, right ->
+        when {
+            left == null || comparator.compare(left, right) > 0 -> right
+            else -> left
+        }
+    }
+
+internal fun checkIsNumberType(funcName: String, value: ExprValue) {
+    if (!value.type.isNumber) {
```
yliuuuu, line 165: Should `NULL` or `MISSING` be allowed here as well?

yliuuuu (commented):
The core logic seems good to me. Nice job refactoring the aggregate functions. 
Just a couple Nit comments on the aggregation behaviors, otherwise great job!

lang/src/org/partiql/lang/eval/physical/operators/Accumulator.kt
johnedquinn: No, see `Accumulator.next()`. We essentially filter out `NULL`, `MISSING`, and non-distinct values if requested.

lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
johnedquinn: Yes, the query returns a bag. Here's the output:
```
<<
  {
    'a': 1
  }
>>
```
Here's another query showing that the ordering is confined in scope:

```sql
SELECT (SELECT a FROM <<{'a': 2}, {'a': 1}>> ORDER BY a) AS subQuery FROM [0]
```
returns:
```
<<
  {
    'subQuery': [
      {
        'a': 1
      },
      {
        'a': 2
      }
    ]
  }
>>
```

While `[0]` is sorted, the result is still a bag, while the projection item `subQuery` is a list because it uses ORDER BY.

lang/src/org/partiql/lang/eval/physical/operators/Accumulator.kt
johnedquinn: Thanks! Looking into this!

johnedquinn: You're totally right -- submitting a revision once tests show passing.

lang/src/org/partiql/lang/eval/physical/operators/Accumulator.kt
yliuuuu: I see. Sorry I missed that.

lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
```diff
@@ -274,10 +275,12 @@ internal class PhysicalPlanCompilerImpl(
...
+            true -> RelationType.LIST
+            false -> RelationType.BAG
+        }
```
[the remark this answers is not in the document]
RCHowell: Edit, currently debugging RelType reference

lang/src/org/partiql/lang/ast/IsOrderedMeta.kt
```diff
@@ -0,0 +1,29 @@
...
+ *  or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+ *  language governing permissions and limitations under the License.
+ */
+package org.partiql.lang.ast
+
+/**
+ * To reduce any extraneous passes over data, this [Meta] indicates whether the associated BindingsToValues Physical
+ * expression should be an ordered list or a bag.
+ */
```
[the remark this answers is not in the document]
johnedquinn: Removed `IsOrderedMeta` and got the `RelType` reference to work without it. LMK if you'd want to include the above changes in this PR, or if the `RelType` fix is sufficient. See the `PhysicalPlanCompiler.compileBindingsToValues()`

lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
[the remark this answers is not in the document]
johnedquinn: Fixed ^

lang/src/org/partiql/lang/eval/physical/operators/Accumulator.kt
```diff
@@ -0,0 +1,183 @@
...
+
+internal class AccumulatorCount(
+    internal val valueFactory: ExprValueFactory,
+    internal override val filter: (ExprValue) -> Boolean
+) : Accumulator(filter = filter) {
+
+    var count: Long = 0L
+
+    override fun nextValue(value: ExprValue) {
+        this.count += 1L
```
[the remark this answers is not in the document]
johnedquinn: At least for nulls, Section 11.2 says: "The group-by and aggregation of PartiQL is backwards compatible to SQL" -- where SQL filters nulls with the exception of COUNT(*). We still follow this -- see `Accumulator.next()`

As for MISSING, it doesn't seem the spec elaborates on NULLS and MISSING specifically for aggregate functions. It does talk about it in the context of grouping (related): "Notice that PartiQL will group together the NULL and the MISSING grouping expressions, since any grouping expression resulting to MISSING has been coerced into NULL before eqg does comparisons for grouping."

Beyond that, the Evaluating Compiler filters `NULL` and `MISSING` values (similar to the above mentioned `Accumulator.next()`) while accumulating.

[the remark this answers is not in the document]
johnedquinn: Also, to note: for typical SQL, `COUNT(*)` is exempt from NULL values and accumulates for every row encountered. This is still kept with the PR implementation since `COUNT(*)` is parsed into `call('count', lit(1))` -- where the value expression will never evaluate to `NULL`

lang/src/org/partiql/lang/eval/physical/operators/AggregateOperatorFactory.kt
```diff
@@ -0,0 +1,21 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+ *  language governing permissions and limitations under the License.
+ */
+package org.partiql.lang.eval.physical.operators
+
+import org.partiql.lang.domains.PartiqlPhysical
+
+public abstract class AggregateOperatorFactory(name: String) : RelationalOperatorFactory {
+    public final override val key: RelationalOperatorFactoryKey = RelationalOperatorFactoryKey(RelationalOperatorKind.AGGREGATE, name)
+    public abstract fun create(impl: PartiqlPhysical.Impl): AggregateOperator
```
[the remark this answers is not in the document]
johnedquinn: Did something essentially very similar, but the AggregateOperator is now removed, since it's the same thing as a RelationExpression. Looks similar to the rest of the implementations.

lang/src/org/partiql/lang/eval/physical/operators/AggregateOperatorFactoryDefault.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import org.partiql.lang.domains.PartiqlPhysical
+import org.partiql.lang.planner.transforms.DEFAULT_IMPL_NAME
+
+internal class AggregateOperatorFactoryDefault : AggregateOperatorFactory(DEFAULT_IMPL_NAME) {
+    override fun create(impl: PartiqlPhysical.Impl): AggregateOperator = when (val operator = impl.name.text.toLowerCase()) {
+        DEFAULT_IMPL_NAME -> AggregateOperatorDefault()
+        else -> throw IllegalArgumentException("Unknown implementation of the Aggregate Operator: $operator")
+    }
+}
```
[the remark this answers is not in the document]
johnedquinn: Did something very similar -- but I actually like your use of `object` here. Submitting a rev to make it an object.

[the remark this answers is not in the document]
johnedquinn: See most recent commit.

johnedquinn:
Needed to rebase the MATCH PR changes ^

lang/src/org/partiql/lang/planner/PartiQLPlannerDefault.kt
```diff
@@ -103,6 +106,9 @@ internal class PartiQLPlannerDefault(
             SelectListItemAliasVisitorTransform(),
             FromSourceAliasVisitorTransform(),
             OrderBySortSpecVisitorTransform(),
+            GroupByItemAliasVisitorTransform(),
+            GroupByPathExpressionVisitorTransform(),
+            GroupKeyReferencesVisitorTransform(),
```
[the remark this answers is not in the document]
johnedquinn: Discussed offline -- resolving.

lang/src/org/partiql/lang/eval/physical/PhysicalBexprToThunkConverter.kt
```diff
@@ -71,6 +74,27 @@ internal class PhysicalBexprToThunkConverter(
...
+        }
+        val compiledKeys = node.groupList.keys.map { key ->
+            val value = exprConverter.convert(key.expr).toValueExpr(key.expr.metas.sourceLocationMeta)
+            val function = key.asVar.toSetVariableFunc()
+            CompiledGroupKey(function, value, key.asVar)
+        }
+
+        // Get Implementation
+        val factory = findOperatorFactory<AggregateOperatorFactory>(RelationalOperatorKind.AGGREGATE, node.i.name.text)
+        val relationExpression = factory.create(source, compiledKeys, compiledFunctions)
```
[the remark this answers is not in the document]
johnedquinn: Adding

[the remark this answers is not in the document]
johnedquinn: Added to factory.create() method. Thanks !

PR #822, opened 2022-10-01
Pulled scalar types out of lexer & parser

lziq:
*Changes*
1. Added a `typeAliases` property to `ScalarType` interface to define type aliases of a scalar type, and migrate specific type aliases from the core package to the legacy plugin package. 
2. Added a `validateParameters` method to validate type parameters of a a scalar type, and migrate specific scalar types’ validation from the core package to the legacy plugin package. 
3. Now any type alias is considered as valid at parsing time. The pass `PartiqlAstSanityValidator` checks whether a type alias is registered, and validates its type parameters. 
4. Added `RealType` & `DoubleType` in legacy plugin package as needed. 

*For review:*
1. Ignore imports of error handling from the core package to the legacy plugin package. We will deal with error handling as part of the pluggability goal later. 
2. The removal of `ScalarTypeExtension.kt` might be a good start to get through the code changes in this PR.


partiql-grammar/src/main/antlr/org/partiql/grammar/PartiQL.g4
```diff
@@ -662,5 +662,5 @@ type
     | CHARACTER VARYING ( PAREN_LEFT arg0=LITERAL_INTEGER PAREN_RIGHT )?                                               # TypeVarChar
     | datatype=(DECIMAL|DEC|NUMERIC) ( PAREN_LEFT arg0=LITERAL_INTEGER ( COMMA arg1=LITERAL_INTEGER )? PAREN_RIGHT )?  # TypeArgDouble
     | TIME ( PAREN_LEFT precision=LITERAL_INTEGER PAREN_RIGHT )? (WITH TIME ZONE)?                                     # TypeTimeZone
-    | symbolPrimitive                                                                                                  # TypeCustom
+    | symbolPrimitive ( PAREN_LEFT LITERAL_INTEGER ( COMMA LITERAL_INTEGER )* PAREN_RIGHT )?                           # GenericType
```
lziq, line 665: Any type alias with any int type parameter is considered as valid during parsing now.

lang/src/org/partiql/lang/visitors/PartiQLVisitor.kt
```diff
@@ -1287,14 +1287,18 @@ internal class PartiQLVisitor(val ion: IonSystem, val customTypes: List<CustomTy
...
+            in customKeywords, in customTypeAliases.keys -> {
+                val customTypeAlias = customTypeAliases.getOrDefault(name, name)
+                customType_(SymbolPrimitive(customTypeAlias, metas), metas)
+            }
+            else -> {
+                val typeParameters = ctx.LITERAL_INTEGER().map { it.text.toLong() }
+                scalarType(name, typeParameters, metas)
+            }
         }
-        customType_(SymbolPrimitive(customName, metas), metas)
```
lziq, line 1297: Here we deal with custom types come first, since we still want to support the legacy `CustomType` API. The legacy `CustomType` API will be removed later as a separate goal of OTS, other than the pluggability goal.

lang/src/org/partiql/lang/types/ScalarTypeExtensions.kt
```diff
@@ -1,33 +0,0 @@
...
-    STRING(listOf("string"), 0..0),
-    SYMBOL(listOf("symbol"), 0..0),
-    TIME(listOf("time"), 0..1),
-    TIME_WITH_TIME_ZONE(listOf("time_with_time_zone"), 0..1),
-    TIMESTAMP(listOf("timestamp"), 0..0);
-}
-
-@JvmField internal val TYPE_ALIAS_TO_SCALAR_TYPE = BuiltInScalarType.values().flatMap { scalarType ->
-    scalarType.typeAliases.map { typeAlias -> typeAlias to scalarType }
-}.associate { it.first to it.second }
```
lziq, line 33: This file is introduced in https://github.com/partiql/partiql-lang-kotlin/pull/753/files#diff-c3e6084bad083065b5a40e9feb4a8d84bcdf86289ae0d42c3628eb09c7339ac9 temporarily, to summarize all the type aliases and type arity defined here in `main`. branch, https://github.com/partiql/partiql-lang-kotlin/blob/a0209df1b22f831a754eb11b070d0b7f9231c78a/lang/src/org/partiql/lang/syntax/LexerConstants.kt.

lziq: Since now we are exposing interfaces to allow users to define type aliases & type validation for a certain scalar type, we are removing it.

lang/src/org/partiql/lang/types/StaticType.kt
```diff
@@ -594,7 +594,10 @@ data class StaticScalarType(
     override val allTypes: List<StaticType>
         get() = listOf(this)
 
-    override fun toString(): String = scalarType.id
+    override fun toString(): String = scalarType.typeName + when {
+        parameters.isEmpty() -> ""
+        else -> parameters.toString()
+    }
```
lziq, line 600: Minor unrelated change to the purpose of this PR.

lang/test/org/partiql/lang/eval/CastTestBase.kt
```diff
@@ -1240,7 +1240,7 @@ abstract class CastTestBase : EvaluatorTestBase() {
             ).types(ExprValueType.DECIMAL.typeAliases().map { "$it(4,4)" }),
             // DECIMAL(2, 4) is a compilation failure in this mode
             listOf(
-                case("1", ErrorCode.SEMANTIC_INVALID_DECIMAL_ARGUMENTS)
+                case("1", ErrorCode.SEMANTIC_INVALID_DECIMAL_PARAMETERS)
```
lziq, line 1243: Minor unrelated change to the purpose of this PR.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -1221,23 +1221,10 @@ internal class EvaluatingCompiler(
...
-            TYPE_ALIAS_TO_SCALAR_TYPE[expr.type.alias.text] === BuiltInScalarType.FLOAT &&
-            expr.type.parameters.isNotEmpty() // if precision of FLOAT is explicitly specified in the original query
-        ) {
-            err(
-                "FLOAT precision parameter is unsupported",
-                ErrorCode.SEMANTIC_FLOAT_PRECISION_UNSUPPORTED,
-                errorContextFrom(expr.type.metas),
-                internal = false
-            )
-        }
```
lziq, line 1240: This is validation of `FLOAT` type parameters, and we are moving it to the `FloatType` defined  legacy plugin package.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -1265,23 +1252,10 @@ internal class EvaluatingCompiler(
...
-            TYPE_ALIAS_TO_SCALAR_TYPE[asType.alias.text] === BuiltInScalarType.FLOAT &&
-            asType.parameters.isNotEmpty() // if precision of FLOAT is explicitly specified in the original query
-        ) {
-            err(
-                "FLOAT precision parameter is unsupported",
-                ErrorCode.SEMANTIC_FLOAT_PRECISION_UNSUPPORTED,
-                errorContextFrom(asType.metas),
-                internal = false
-            )
-        }
```
lziq, line 1284: This is validation of `FLOAT` type parameters, and we are moving it to the `FloatType` defined  legacy plugin package.

lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
```diff
@@ -990,23 +990,10 @@ internal class PhysicalPlanCompilerImpl(
...
-            TYPE_ALIAS_TO_SCALAR_TYPE[expr.type.alias.text] === BuiltInScalarType.FLOAT &&
-            expr.type.parameters.isNotEmpty() // if precision of FLOAT is explicitly specified in the original query
-        ) {
-            err(
-                "FLOAT precision parameter is unsupported",
-                ErrorCode.SEMANTIC_FLOAT_PRECISION_UNSUPPORTED,
-                errorContextFrom(expr.type.metas),
-                internal = false
-            )
-        }
```
lziq, line 1009: This is validation of `FLOAT` type parameters, and we are moving it to the `FloatType` defined  legacy plugin package.

lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
```diff
@@ -1034,23 +1021,10 @@ internal class PhysicalPlanCompilerImpl(
...
-            TYPE_ALIAS_TO_SCALAR_TYPE[asType.alias.text] === BuiltInScalarType.FLOAT &&
-            asType.parameters.isNotEmpty() // if precision of FLOAT is explicitly specified in the original query
-        ) {
-            err(
-                "FLOAT precision parameter is unsupported",
-                ErrorCode.SEMANTIC_FLOAT_PRECISION_UNSUPPORTED,
-                errorContextFrom(asType.metas),
-                internal = false
-            )
-        }
```
lziq, line 1053: This is validation of `FLOAT` type parameters, and we are moving it to the `FloatType` defined  legacy plugin package.

lang/src/org/partiql/lang/eval/visitors/PartiqlAstSanityValidator.kt
```diff
@@ -67,25 +73,12 @@ class PartiqlAstSanityValidator : PartiqlAst.Visitor() {
...
-                err(
-                    "Scale $scale should be between 0 and precision $precision",
-                    errorCode = ErrorCode.SEMANTIC_INVALID_DECIMAL_ARGUMENTS,
-                    errorContext = errorContextFrom(metas),
-                    internal = false
-                )
-            }
-        }
-    }
-
```
lziq, line 82: This is validation of `DECIMAL` type parameters, and we are moving it to the `FloatType` defined  legacy plugin package.

lang/src/org/partiql/lang/eval/visitors/PartiqlPhysicalSanityValidator.kt
```diff
@@ -61,25 +62,12 @@ class PartiqlPhysicalSanityValidator(private val evaluatorOptions: EvaluatorOpti
...
-                err(
-                    "Scale $scale should be between 0 and precision $precision",
-                    errorCode = ErrorCode.SEMANTIC_INVALID_DECIMAL_ARGUMENTS,
-                    errorContext = errorContextFrom(metas),
-                    internal = false
-                )
-            }
-        }
-    }
-
```
lziq, line 76: This is validation of `DECIMAL` type parameters, and we are moving it to the `DecimalType` defined  legacy plugin package.

lang/src/OTS/ITF/org/partiql/ots/Plugin.kt
```diff
@@ -1,11 +1,17 @@
...
  */
 interface Plugin {
+    /**
+     * All the scalar types
+     */
+    val scalarTypes: List<ScalarType>
+
     val posOp: ScalarOp
     val negOp: ScalarOp
     val binaryPlusOp: ScalarOp
```
RCHowell, line 17: We need to review the "Plugin" interface based on what it brings to the engine.

- Plugin name spaces
- Scalar Types
- Scalar Functions
- Aggregate Functions
- Plugin specific configuration
- Operator implementations

Are the plugins like Trino's with data sources, udf's, and schema providers? Or are these plugins just for types and functions?

lang/src/org/partiql/lang/CompilerPipeline.kt
```diff
@@ -264,6 +267,15 @@ internal class CompilerPipelineImpl(
...
 ) : CompilerPipeline {
 
+    // Initialize a map from a type alias to a scalar type, as part of work for PartiQL compile pipeline to install the plugin
+    private val aliasesToScalarType: Map<String, ScalarType>
+
+    init {
+        plugin.validate()
+
+        aliasesToScalarType = plugin.mapAliasToScalarType()
+    }
```
RCHowell, line 277: We should not be assuming there is only one plugin. We should also namespace any identifiers named within a plugin

lang/src/org/partiql/lang/CompilerPipeline.kt
```diff
@@ -273,7 +285,8 @@ internal class CompilerPipelineImpl(
             }
         }.flatten().toMap(),
         procedures,
-        compileOptions
+        compileOptions,
+        aliasesToScalarType
```
RCHowell, line 289: I suggest not using raw types like Map and List. Wrapping types like this in simple holder classes introduces indirection which is helpful for future changes

lang/src/org/partiql/lang/compiler/PartiQLCompilerBuilder.kt
```diff
@@ -51,6 +53,7 @@ class PartiQLCompilerBuilder private constructor() {
     private var customFunctions: List<ExprFunction> = emptyList()
     private var customProcedures: List<StoredProcedure> = emptyList()
     private var customOperatorFactories: List<RelationalOperatorFactory> = emptyList()
+    private var plugin: Plugin = LegacyPlugin()
```
RCHowell, line 56: Why is it called LegacyPlugin? Also, we shouldn't have just one plugin

lang/src/org/partiql/lang/compiler/PartiQLCompilerDefault.kt
```diff
@@ -56,14 +61,22 @@ internal class PartiQLCompilerDefault(
...
+
         exprConverter = PhysicalPlanCompilerImpl(
             valueFactory = valueFactory,
             functions = functions,
             customTypedOpParameters = customTypedOpParameters,
             procedures = procedures,
             evaluatorOptions = evaluatorOptions,
-            bexperConverter = bexprConverter
+            bexperConverter = bexprConverter,
+            aliasToScalarType = aliasToScalarType
```
RCHowell, line 79: Same as before. We don't have to pass around List<T> and Map<K, V> everywhere. It's generally better to wrap these things to encapsulate the underlying data structure — which we are then free to change.

Think of the type identifiers to type names mappings. If we have `Map<String, String>` then we are limited to just looking up the id of given a name. We can do this and more by wrapping. Call it the TypeRegistry or something (don't use this name, this is just an example for illustrative purposes).

```
class TypeRegistry(private val types: List<StaticType>) {

  private lateinit var namesToType: Map<String, StaticType>
  private lateinit var idToType: Map<String, StaticType>

  init {
    // populate the maps
  }

  // now we control access to the underlying data structures
  fun getTypeByName(name: String): StaticType = namesToType[name]!! // something better in practice
 
  // contrived example
  fun getTypeById(id: String): StaticType = idsToType[id]!! 

 // can be extended and different underlying data structures changed
}
```

lang/src/org/partiql/lang/prettyprint/QueryPrettyPrinter.kt
```diff
@@ -785,29 +783,7 @@ class QueryPrettyPrinter {
...
-                BuiltInScalarType.STRING -> sb.append("STRING")
-                BuiltInScalarType.SYMBOL -> sb.append("SYMBOL")
-                BuiltInScalarType.BLOB -> sb.append("BLOB")
-                BuiltInScalarType.CLOB -> sb.append("CLOB")
-                BuiltInScalarType.DATE -> sb.append("DATE")
-                BuiltInScalarType.TIME -> sb.append("TIME")
-                BuiltInScalarType.TIME_WITH_TIME_ZONE -> sb.append("TIME WITH TIME ZONE")
-                else -> error("Unrecognized scalar type ID")
-            }
+            is PartiqlAst.Type.ScalarType -> sb.append(node.alias.text.toUpperCase())
```
RCHowell, line 786: You need to replace "_" with " "

RCHowell (changes requested):

lang/src/org/partiql/lang/ast/passes/inference/StaticTypeInferencer.kt
```diff
@@ -30,6 +33,15 @@ class StaticTypeInferencer(
...
     private val plugin: Plugin
 ) {
+    // Initialize a map from a type alias to a scalar type, as part of work for PartiQL inferencer to install the plugin
+    private val aliasToScalarType: Map<String, ScalarType>
+
+    init {
+        plugin.validate()
+
+        aliasToScalarType = plugin.mapAliasToScalarType()
+    }
```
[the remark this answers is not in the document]
lziq: For now we only consider there is only one plugin for approaching pluggability of OTS.

lang/src/OTS/ITF/org/partiql/ots/Plugin.kt
lziq: For now we only consider there is only one plugin for approaching pluggability of OTS.

lang/src/org/partiql/lang/CompilerPipeline.kt
lziq: For now we only consider there is only one plugin for approaching pluggability of OTS.

lang/src/org/partiql/lang/compiler/PartiQLCompilerBuilder.kt
lziq: Currently, it still contains legacy types, functions & operators. Later when we start to work on function overloading, and finish the work on rewriting each function/operator specific to its arguments, we can call it `StandardPlugin`.

lang/src/OTS/ITF/org/partiql/ots/type/ScalarType.kt
```diff
@@ -9,9 +9,39 @@ import org.partiql.lang.eval.ExprValueType
...
+     *      2. DOUBLE PRECISION -> DOUBLE_PRECISION
+     *      3. TIME WITH TIME ZONE -> TIME_WITH_TIME_ZONE
+     *
+     * One type can have multiple type aliases
+     *
+     * Types cannot share the same type alias
+     *
+     * Type alias cannot be a PartiQL keywords e.g. 'select'
+     */
+    val aliases: List<String>
```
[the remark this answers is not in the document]
lziq: Will apply this comment

lang/src/OTS/IMP/org/partiql/ots/legacy/types/CharType.kt
```diff
@@ -6,8 +6,17 @@ import org.partiql.lang.eval.ExprValue
...
-        get() = "character"
+    override val typeName = "character"
+
+    override val aliases = listOf("character", "char")
+
+    override fun validateParameters(typeParameters: TypeParameters) {
+        when (typeParameters.size) {
+            0 -> {}
+            1 -> require(typeParameters[0] > 0)
+            2 -> error("$typeName type requires at most 1 parameter")
```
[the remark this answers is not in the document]
lziq: Will apply this comment

lang/src/org/partiql/lang/prettyprint/QueryPrettyPrinter.kt
lziq: Actually, we don't if a user really wants "_" in the original type name or not. e.g. If I want to define a new type, whose type name is `MY_TYPE`, the transformation to `MY TYPE` can cause parsing error. 

Probably we can consider multiple [identifier](https://github.com/partiql/partiql-lang-kotlin/blob/a9e5a4ef090ffa73caa2d85771c4322acbed5e4d/partiql-grammar/src/main/antlr/org/partiql/grammar/PartiQLTokens.g4#L352)s with spaces in between as valid type name, and require no underscore in type names lexically?

lziq: If we can make such change in partiql grammar, then lexically a space corresponds to an underscore defined here for type names. Then, the problem is lexically where do we put type parameters for those types with name to be multiple identifiers. Probably `TIME (2) WITH TIME ZONE` is a good example.

lziq: This might be a good design, but as of now, the use cases are still kind of ambiguous. So I think it's fine to just keep this line as is?

lang/src/org/partiql/lang/CompilerPipeline.kt
lziq: Will apply this comment

RCHowell (changes requested):

lang/src/OTS/ITF/org/partiql/ots/type/ScalarType.kt
```diff
@@ -9,10 +8,40 @@ import org.partiql.lang.eval.ExprValueType
...
+     * `CAST(1 AS <type_names>)`, the target type of the CAST operator corresponds to the scalar type whose
+     * [names] has `<type_name>`
+     *
+     * Note that spaces are not allowed in type aliases. For special type aliases with space in standard sql, e.g.
+     * `TIME WITH TIME ZONE` in the query `CAST(a AS TIME WITH TIME ZONE)`, the spaces there are replaced with
+     * underscore '_' during parsing, so the query is actually transformed as `CAST(a AS TIME_WITH_TIME_ZONE)`.
+     * All the special type aliases are:
+     *      1. CHARACTER VARYING -> CHARACTER_VARYING
+     *      2. DOUBLE PRECISION -> DOUBLE_PRECISION
+     *      3. TIME WITH TIME ZONE -> TIME_WITH_TIME_ZONE
```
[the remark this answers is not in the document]
lziq: Because currently, PartiQL syntax does not allow spaces in type name, except for special type names, such as `TIME WITH TIME ZONE`.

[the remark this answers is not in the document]
lziq: like `CAST(1 AS MY TYPE)` throws a syntax on `MY TYPE`

lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
```diff
@@ -174,7 +174,7 @@ internal class PhysicalPlanCompilerImpl(
      * hope that long-running compilations may be aborted by the caller.
      */
     fun compile(plan: PartiqlPhysical.Plan): Expression {
-        PartiqlPhysicalSanityValidator(evaluatorOptions).walkPlan(plan)
+        PartiqlPhysicalSanityValidator(evaluatorOptions, typeRegistry).walkPlan(plan)
```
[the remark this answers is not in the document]
lziq: Do you think we can move line 177 to `AstToLogicalVisitorTransform`, checking if it is a valid logical AST after the transformation?

lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
```diff
@@ -1165,7 +1139,7 @@ internal class PhysicalPlanCompilerImpl(
     }
 
     private fun compileCanLosslessCast(expr: PartiqlPhysical.Expr.CanLosslessCast, metas: MetaContainer): PhysicalPlanThunk {
-        val typedOpParameter = expr.asType.toTypedOpParameter(customTypedOpParameters)
+        val typedOpParameter = expr.asType.toTypedOpParameter(customTypedOpParameters, typeRegistry)
```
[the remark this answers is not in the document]
lziq: Do you think we can change `PartiqlAst.Type.toTypedOpParameter` to be an internal function and leave it as is? If I understand correctly, we will remove the function anyway as we approach PartiQL 1.0

lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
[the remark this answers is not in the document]
lziq: To my knowledge, the transformation from "logical" to "logical resolved", and to "physical plan" handles variables and bindings explicitly, which is orthogonal to the general AST validation in my opinion.

lang/src/org/partiql/lang/util/PluginExtensions.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import OTS.ITF.org.partiql.ots.Plugin
+import OTS.ITF.org.partiql.ots.type.ScalarType
+
+class TypeRegistry(types: List<ScalarType>) {
+    private var namesToType: Map<String, ScalarType> = types.flatMap { scalarType ->
+        scalarType.names.map { typeAlias -> typeAlias to scalarType }
+    }.associate { it.first to it.second }
+
+    fun getTypeByName(name: String): ScalarType? = namesToType[name]
```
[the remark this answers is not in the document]
lziq: It should be a private/internal class. However, currently the fact that the public class `PartiqlPhysicalSanityValidator`'s constructor takes `TypeRegistry` as one argument requires `TypeRegistry` to be public. The 2 solutions I can come up with are: 
1. Use a `Plugin` to construct `PartiqlPhysicalSanityValidator`, and generate `TypeRegistry` from the `Plugin` once again. 
2. Make `PartiqlPhysicalSanityValidator` internal and use it to validate an AST right after parsing (instead of right before compiling). 

I personally prefer the second solution. What do you think?

lang/src/org/partiql/lang/prettyprint/QueryPrettyPrinter.kt
RCHowell: If you do not replace "_" with " ", then the pretty printer will produce unparse-able PartiQL. Because the current design replaces " " with "_" in the lexer, you need to undo this when printing PartiQL text back.

For example,

```sql
--original
SELECT CAST(x AS character varying) FROM Foo
```

The AST is something like
```
query(
  select(
    project(
      cast(
         id(x)
        type(character_varying)
      )
    )
  )
  from(
    scan(
      id(Foo)
    )
)
```

Then pretty printed never removes the "_" inserted by the " ".
```sql
--pretty
SELECT CAST(x AS character_varying) FROM Foo
```

I posted this question elsewhere, but why are we inserted "_" anyways? This seems like a hold-over from the handwritten lexer's multi-lexeme tables. So now if a custom explicitly wants "_" then we can't distinguish between a "_" inserted by the customer or inserted by the lexer. So why insert them in the first place?


Then when you pretty

lang/src/OTS/ITF/org/partiql/ots/type/ScalarType.kt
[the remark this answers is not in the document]
RCHowell: Ok, so the validation should assert that names do not included spaces and our parser should be able to handle the special cases. By introducing the rewrite of " " to "_" then we can't reliably biject since there is no distinction between an explicity "_" or a lexer/parser inserted "_"

lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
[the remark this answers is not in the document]
RCHowell: AST validation should be done on the AST. I would think type resolution should be done on the logical plan. I don't think I understand why the physical plan needs to the type registry.

lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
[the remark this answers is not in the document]
RCHowell: Ok, but this is poor style; hence why you had to change it in so many places. So when we remove it in PartiQL 1.0, we have to change all those places again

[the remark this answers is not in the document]
RCHowell: It's unnecessary parameter juggling, but sure a temporary blemish

lang/src/org/partiql/lang/util/PluginExtensions.kt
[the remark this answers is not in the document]
RCHowell: I prefer 2 as well. The validator shouldnt have been made public. Also, why are we validating the physical plan? Any validation should be done on the logical plan, and production of a physical plan from a logical plan is a minimal operation

lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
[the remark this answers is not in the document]
lziq: The funciton `PartiqlAst.Type.toTypedOpParameter()` is also used in `StaticTypeInferenceVisitorTransform`. If we change it as private and remove parameters, we have to re-write it twice (in compiler & inferencer, separately), and maintain it at 2 places.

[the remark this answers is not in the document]
lziq: Actually, I think it is a good suggestion and I will apply it.

lang/src/org/partiql/lang/prettyprint/QueryPrettyPrinter.kt
lziq: I agree with you that "_" insertion is unnecessary.

lang/src/OTS/ITF/org/partiql/ots/type/ScalarType.kt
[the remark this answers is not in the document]
lziq: I agree with you that "_" insertion is unnecessary.

lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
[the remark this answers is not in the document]
lziq: If type resolution is considered as part of work in compiling, I think physical plan needs that. But if it is not, do you think we should have an explicit pass for it in AST transformations?

lang/src/org/partiql/lang/util/PluginExtensions.kt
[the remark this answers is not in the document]
lziq: Also, as for whether we should throw error in case that we cannot find one entry in the type registry, I don't have strong opinion. Because in case we want to handle this case other than throwing an error, null handling or error handling seems identical to me.

lang/src/org/partiql/lang/types/PartiqlAstTypeExtensions.kt
```diff
@@ -16,5 +17,5 @@ fun PartiqlAst.Type.toTypedOpParameter(customTypedOpParameters: Map<String, Type
     // to keep this function around.
     val sexp = this.toIonElement()
     val physicalType = PartiqlPhysical.transform(sexp) as PartiqlPhysical.Type
-    return physicalType.toTypedOpParameter(customTypedOpParameters)
+    return physicalType.toTypedOpParameter(customTypedOpParameters, typeRegistry)
```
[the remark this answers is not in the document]
lziq: Resolved in another PR: https://github.com/partiql/partiql-lang-kotlin/commit/05c17854653ddda54a15db8d1e3814c5b6450459#diff-a3cc371d2ab940b9857af140becf807e82d5c7428a74f853050ea97095c0b224

lziq:
The above several commits applies changes in PR https://github.com/partiql/partiql-lang-kotlin/commit/05c17854653ddda54a15db8d1e3814c5b6450459

lang/src/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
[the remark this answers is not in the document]
lziq: See https://github.com/partiql/partiql-lang-kotlin/issues/848 for more details

lang/src/OTS/IMP/org/partiql/ots/legacy/types/TimeStampType.kt
```diff
@@ -1,11 +1,12 @@
 package OTS.IMP.org.partiql.ots.legacy.types
 
-import OTS.ITF.org.partiql.ots.type.ScalarType
+import OTS.ITF.org.partiql.ots.type.NonParametricType
 import org.partiql.lang.eval.ExprValueType
 
-object TimeStampType : ScalarType {
-    override val id: String
-        get() = "timestamp"
+object TimeStampType : NonParametricType() {
```
RCHowell, line 6: TimeStampType -> TimestampType

lang/src/OTS/IMP/org/partiql/ots/legacy/types/TimeType.kt
```diff
@@ -1,14 +1,37 @@
...
+            false -> listOf("time")
+        }
+
+    override fun validateParameters(typeParameters: TypeParameters) {
+        when (typeParameters.size) {
+            0 -> {}
+            1 -> require(typeParameters[0] > 0)
+            2 -> error("$id type requires at most 1 parameter")
+        }
+    }
```
RCHowell, line 34: Opinion, but I think it would be cleaner to just have two different classes.

TimeType and TimeWithTimeZoneType or something like that.

lang/src/OTS/IMP/org/partiql/ots/legacy/types/VarcharType.kt
```diff
@@ -1,13 +1,22 @@
...
 import org.partiql.lang.eval.ExprValueType
 
 object VarcharType : ScalarType {
-    override val id: String
-        get() = "varying_character"
+    override val id = "character_varying"
+
+    override val names = listOf("varchar", "character varying")
+
+    override fun validateParameters(typeParameters: TypeParameters) {
```
RCHowell, line 13: This is interesting. Opinion, but I think it would be cleaner to have an interface for ParametricType rather than NonParametricType. My opinion, but it seems a little backwards to have the "simple" non-parameterized types to be declared specially - inheriting from an abstract class, rather than treating the parameterized "special" types as special.

Here's a simple example of what I'm suggesting

```kotlin
interface ScalarType {

    fun validateParameters(): Boolean = true

}

interface ParametricType : ScalarType {

    // Declare override without a body; forcing implementation
    override fun validateParameters(): Boolean
}

// No additional classification
class IntType : ScalarType


class VarcharType : ParametricType {
    
    override fun validateParameters(): Boolean {
        TODO()
    }

}

```

RCHowell (commented):
Looking really close now. Some minor comments.

lang/src/OTS/IMP/org/partiql/ots/legacy/types/TimeType.kt
lziq: I think this is a good idea, but this is a refactoring to `Time` type. Do you think we can setup another PR for it, if needed?

RCHowell (approved):

PR #824, opened 2022-10-03
Window operator

yliuuuu:
This is the second step of a series of PR focusing on experimental window function.

In the PR, a window operator is created for handling window function. 

The logical window operator contains two parts, window specification(for now it is a list of window partition expression and a list of window sort spec) and window expression (for now it is the lag and lead function). 

The window operator takes in a binding collection and add the result of window expression to the binding collection. 

One window function in the `SELECT` clause corresponding to one window operator, regardless of whether the window specification is the same. (This is an opportunity for optimization here, we want the window operator to contain a list of window function in the future.) 

The default window operator operates via a sort based approach. We first sort the entire "table" by partition expression firstly then by order spec secondly. Then we create a list of partitions using by sequentially scan the ordered table and comparing the partition expression. The created partition helps us limited the scope of each window function. 

The evaluation of `lag`/`lead` is relatively straight-forward, since the partition is materialized, we can access the target row using index, and evaluated the result by setting the state's register to the target row. Then we reset the state back to current row, and appending the result to the register. 

Notice that the code is deliberately keep in a very "raw" stage. No abstraction/optimization is done at this point. 

This is because, our implementation is partial at this stage. Incorporating in missing parts such as `frame` may change the abstraction design significantly. 


partiql-grammar/src/main/gen/PartiQLTokens.interp
```diff
@@ -0,0 +1,957 @@
+token literal names:
```
johnedquinn, line 1: BTW these files weren't supposed to be committed. @RCHowell moved the grammar to `:lang`.

lang/src/org/partiql/lang/eval/physical/operators/DefaultOperatorFactories.kt
```diff
@@ -24,9 +24,11 @@ import org.partiql.lang.planner.transforms.DEFAULT_IMPL_NAME
  * @see [org.partiql.lang.planner.PlannerPipeline.Builder.addRelationalOperatorFactory]
  */
 internal val DEFAULT_RELATIONAL_OPERATOR_FACTORIES = listOf(
+
     AggregateOperatorFactoryDefault,
     SortOperatorFactoryDefault,
     UnpivotOperatorFactoryDefault,
+    SortBasedWindowOperator(DEFAULT_IMPL_NAME),
```
johnedquinn, line 31: Had a conversation with @RCHowell regarding how the factories should be created. We can discuss later -- but also see the 3 above for more preliminary info.

johnedquinn (commented):
Left small preliminary comments so you can address sooner rather than later.

lang/src/org/partiql/lang/eval/physical/operators/DefaultOperatorFactories.kt
yliuuuu: I saw the changes, I like the new way. If you agree, I would like to delay the change to the next PR though, trying to reduce the overhead on this one.

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -81,6 +86,62 @@ internal class AstToLogicalVisitorTransform(
...
+            is PartiqlAst.Projection.ProjectValue -> return algebra
+            is PartiqlAst.Projection.ProjectList -> {
+                project.projectItems.asSequence().filterNot {
+                    it is PartiqlAst.ProjectItem.ProjectAll
+                }.filter {
+                    (it as PartiqlAst.ProjectItem.ProjectExpr).expr is PartiqlAst.Expr.CallWindow
+                }.map {
+                    (((it as PartiqlAst.ProjectItem.ProjectExpr).expr) as PartiqlAst.Expr.CallWindow)
+                }
+            }
```
johnedquinn, line 114: Currently trying to get the semantics down, but is there any way where the ProjectList expression will ever hold anything else?

I know the the changes in `PartiQL.g4` aren't part of this PR -- but they're relevant -- `windowFunction` is an `exprPrimary` and therefore can be expressed as a sub-expression. So, a query like:
```sql
SELECT 5 + LAG(t.co) OVER() FROM t
```

Could theoretically be possible, unless a visitor pass throws an exception on this -- but if there wasn't some sort of pass on this, then this would be allowed, and the lines referenced above would not be able to extract the window function. You'd unfortunately need to create a visitor to grab them.

johnedquinn: BTW the above is allowed in Postgres too -- just checked to make sure

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -81,6 +86,62 @@ internal class AstToLogicalVisitorTransform(
...
+            is PartiqlAst.Projection.ProjectStar -> return algebra
+        }
+
+        // todo: if multiple window functions are operating on the same window, we can potentially add them in a single window operator to increase performance
+        var modifiedAlgebra = algebra
+        windowExpressions.forEachIndexed { index, callWindow ->
+            modifiedAlgebra = callWindow.let { callWindowNode ->
+                PartiqlLogical.build {
+                    window(
+                        algebra,
```
johnedquinn, line 126: Is this supposed to be `modifiedAlgebra`? It seems like we only really grab the last window expression.

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -81,6 +86,62 @@ internal class AstToLogicalVisitorTransform(
...
+            }
+
+            is PartiqlAst.Projection.ProjectPivot -> return algebra
+            is PartiqlAst.Projection.ProjectStar -> return algebra
+        }
+
+        // todo: if multiple window functions are operating on the same window, we can potentially add them in a single window operator to increase performance
+        var modifiedAlgebra = algebra
+        windowExpressions.forEachIndexed { index, callWindow ->
+            modifiedAlgebra = callWindow.let { callWindowNode ->
```
johnedquinn, line 123: We don't need the let here, right?

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -433,6 +494,16 @@ internal class AstToLogicalVisitorTransform(
             )
         }
 
+    override fun transformExprCallWindow(node: PartiqlAst.Expr.CallWindow): PartiqlLogical.Expr =
+        PartiqlLogical.build {
+            id(
+                getUniqueWindowName(node,),
+                caseInsensitive(),
```
johnedquinn, line 501: This can likely be case sensitive.

lang/src/org/partiql/lang/planner/transforms/LogicalToLogicalResolvedVisitorTransform.kt
```diff
@@ -244,6 +244,26 @@ internal data class LogicalToLogicalResolvedVisitorTransform(
         }
     }
 
+    // TODO: Remove from experimental once https://github.com/partiql/partiql-docs/issues/31 is resolved and a RFC is approved
+    override fun transformBexprWindow_windowSpecification(node: PartiqlLogical.Bexpr.Window): PartiqlLogicalResolved.Over {
+        val bindings = getOutputScope(node)
```
johnedquinn, line 249:
```suggestion
        val bindings = getOutputScope(node).concatenate(this.inputScope)
```
We *may* need to concatenate `this.inputScope`. Also, this is applicable to both added transforms.

Is the following possible?
```sql
SELECT
  (SELECT LAG(t.a) OVER(PARTITION BY t.b) FROM <<0>>)
FROM t
```

lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -940,6 +944,13 @@ may then be further optimized by selecting better implementations of each operat
                 (offset i::impl row_count::expr source::bexpr)
                 (limit i::impl row_count::expr source::bexpr)
                 (let i::impl source::bexpr bindings::(* let_binding 1))
+
+                (window
+                    i::impl
+                    source:: bexpr
+                    window_specification:: over
+                    window_expression:: window_expression
+                )
```
johnedquinn, line 953: No action here -- just mental notes.
```sql
SELECT
  LAG(t.a) OVER(),
  LEAD(t.b) OVER(),
  LAG(t.c) OVER()
FROM t
WHERE t.d > 0
```

Will produce:
```
SCAN -> FILTER -> WINDOW -> WINDOW -> WINDOW -> PROJ
```

johnedquinn (commented):
Overall some really great work -- you're putting a ton of code together for this, and it looks great. I do have a question about what APIs we **know** will be changing in the future (specifically with the window bexpr) and what APIs **might** be changing.

The remainder of the comments are quite small.

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
yliuuuu: Spot on! Working on this now!

lang/src/org/partiql/lang/eval/physical/operators/WindowRelationalOperatorFactory.kt
```diff
@@ -0,0 +1,204 @@
...
+        val (target, offset, default) = when (arguments.size) {
+            1 -> listOf(arguments[0], null, null)
+
+            2 -> listOf(arguments[0], arguments[1], null)
+
+            3 -> listOf(arguments[0], arguments[1], arguments[2])
+
+            else -> error("Wrong number of Parameter for Lead Function")
+        }
+        return relation(RelationType.BAG) {
```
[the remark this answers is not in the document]
yliuuuu: I think so, window operator get evaluated before Sorting.

lang/src/org/partiql/lang/eval/physical/operators/WindowRelationalOperatorFactory.kt
```diff
@@ -0,0 +1,204 @@
...
+        if (windowExpression.funcName.text.toLowerCase() == "lag") {
+            LagFunction(windowExpression, partition, windowFunctionParameter, state).eval()
+        } else {
+            LeadFunction(windowExpression, partition, windowFunctionParameter, state).eval()
+        }
+    }
+}
+
+// TODO : The next PR will focus on modeling the window function.
+internal class LeadFunction(val windowExpression: PartiqlPhysical.WindowExpression, val partition: MutableList<List<Array<ExprValue>>>, val arguments: List<ValueExpression>, val state: EvaluatorState) {
```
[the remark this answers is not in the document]
yliuuuu: Like a lagLeadCommon. I tried something similar, decided to not using this approach for now since it makes the code less readable. Maybe discuss more on this matter in the next PR?

lang/src/org/partiql/lang/eval/physical/operators/WindowRelationalOperatorFactory.kt
```diff
@@ -0,0 +1,204 @@
...
+        }
+        // need to be partitioned
+        else {
+            val iter = sortedRegisters.iterator()
+            var rowInPartition = mutableListOf<Array<ExprValue>>()
+            var previousPartition: ExprValue? = null
+            while (iter.hasNext()) {
+                val currentRow = iter.next()
+                state.load(currentRow)
+                val currentPartition = state.valueFactory.newSexp(
```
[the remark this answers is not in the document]
yliuuuu: Hmm, I wasn't think that much. Is there any benefit we can get from using the list?

lang/src/org/partiql/lang/eval/physical/operators/WindowRelationalOperatorFactory.kt
```diff
@@ -0,0 +1,204 @@
...
+                            numberValue
+                        } else {
+                            error("offset need to be non-negative integer")
+                        }
+                    } ?: 1L // default offset is one
+                    // We leave the checking mechanism for type mismatch out for now.
+                    val defaultValue = default?.invoke(state) ?: state.valueFactory.nullValue
+                    val targetIndex = index + offsetValue.toLong()
+                    // if targetRow is within partition
+                    if (targetIndex >= 0 && targetIndex <= rowsInPartition.size - 1) {
```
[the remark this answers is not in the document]
yliuuuu: This is right for Lead. :) Smart!

lang/src/org/partiql/lang/planner/transforms/LogicalToLogicalResolvedVisitorTransform.kt
yliuuuu: Spot on. Fixed!

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
yliuuuu: It can, but should we specify this to be case Sensitive breaks the semantics a little doesn't it?

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
yliuuuu: Came up with a Visitor Transform that is inspired by your group by transformer. See the new commit.

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -494,15 +506,15 @@ internal class AstToLogicalVisitorTransform(
...
-        }
+    override fun transformExprCallWindow(node: PartiqlAst.Expr.CallWindow): PartiqlLogical.Expr = error("call transformer")
+//        PartiqlLogical.build {
+//            id(
+//                getUniqueWindowName(node),
+//                caseInsensitive(),
+//                unqualified(),
+//                metas = node.metas
+//            )
+//        }
```
johnedquinn, line 517: Remove comments here. Perhaps a better error message as well.

johnedquinn (commented):
Great job with the visitor transform overall. I'm going to continue looking at this PR, but in the meantime, I wanted you to look into this:
```sql
SELECT leader FROM (
  SELECT LEAD(t.a) OVER(ORDER BY t.a) AS leader
  FROM <<{'a':1}, {'a': 2}>> AS t
)
```

Results in:
```
org.partiql.lang.eval.EvaluationException: No such binding: t.
	Evaluator Error: at line 1, column 52: Binding 't' does not exist
```

Decided to look into this because something seemed slightly off in the `CallWindowFunctionProjectionReplacer`. Similar to the `CallAggregationsProjectionReplacer`. Seems like you need to make sure you don't recurse into further SELECTs. Take a look at the other visitor transform at `transformExprSelect` -- but that should fix it.

Without that, the window functions get transformed outside of the projection list at the top-most level -- never at any level below.

lang/src/org/partiql/lang/eval/physical/PhysicalBexprToThunkConverter.kt
```diff
@@ -291,6 +292,34 @@ internal class PhysicalBexprToThunkConverter(
...
+
+        val compiledWindowFunctionParameter = node.windowExpression.args.map {
+            exprConverter.convert(it).toValueExpr(it.metas.sourceLocationMeta)
+        }
+
+        // locate operator factory
+        val factory = findOperatorFactory<WindowRelationalOperatorFactory>(RelationalOperatorKind.WINDOW, node.i.name.text)
+
+        // create operator implementation
+        val bindingsExpr = factory.create(node.i, source, compiledPartitionBy, compiledOrderBy, node.windowExpression, compiledWindowFunctionParameter)
```
johnedquinn, line 318: Ideally we'd want some sort ofobject representing the compiled window expression / function -- but I know you're working on the representation of the function -- so I'll just wait for those changes in the next PR.

lang/src/org/partiql/lang/eval/physical/operators/WindowRelationalOperatorFactory.kt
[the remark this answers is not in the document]
johnedquinn: Nothing comes off the top of my head -- more just curiosity.

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -81,6 +87,73 @@ internal class AstToLogicalVisitorTransform(
...
+     *          LEAD
+     *          t.a
+     * ```
+     */
+    private fun transformWindowFunctions(node: PartiqlAst.Expr.Select, algebra: PartiqlLogical.Bexpr): Pair<PartiqlAst.Expr.Select, PartiqlLogical.Bexpr>? {
+        val windowReplacer = CallWindowFunctionProjectionReplacer()
+
+        val transformedNode = windowReplacer.transformExprSelect(node) as PartiqlAst.Expr.Select
+
+        val windowExpressions = windowReplacer.getWindowFuncs()
```
johnedquinn, line 126: This is essentially where the scoping issue mentioned before gets messed up. We've accidentally recursed too far, so window functions are created at the top-level, then we recurse and the LAG/LEAD is gone and the ID has taken its place.

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -507,6 +587,56 @@ private class FromSourceToBexpr(
...
+private class CallWindowFunctionProjectionReplacer() : VisitorTransformBase() {
+    val callWindowFunctionVisitorTransform = CallWindowFunctionReplacer()
+
+    override fun transformProjectItemProjectExpr_expr(node: PartiqlAst.ProjectItem.ProjectExpr): PartiqlAst.Expr {
+        return callWindowFunctionVisitorTransform.transformExpr(node.expr)
+    }
+
+    override fun transformProjectionProjectValue_value(node: PartiqlAst.Projection.ProjectValue): PartiqlAst.Expr {
+        return callWindowFunctionVisitorTransform.transformExpr(node.value)
+    }
```
johnedquinn, line 603: As mentioned in other comment, this is most likely where we need to make sure we don't recurse into exprSelect.

johnedquinn (approved):
Overall great work. Thanks for addressing comments so quickly.

I'll just echo things that will eventually need to be addressed prior to `main`:
- factory clean-up
  - re-review visibility modifiers (adding this so we don't forget)
- compiled window function representation
- taking additional time to look into the partiql.ion representation of the window operator.

Again, great work!

PR #830, opened 2022-10-05
Window refactor

yliuuuu:
This is a third PR of the experimental `LAG` and `LEAD` implementation. 

In this PR, I have refactored the raw window function design. The change are following: 
1. a window function interface: 
This is a bare-bone of window function, the interface contains window function signature, which for now only has function name. `reset` which should be call when entering a new partition. Notice that `processRow` will undergo some major changes in the future, once we factor in the concept of frame and peer group. 
```
interface WindowFunction {
     val signature,
     val reset(partition: Array<List<ExprValue>>),
     val processRow(state: EvaluatorState, arguments: List<ValueExpression>, windowVarDecl: PartiqlPhysical.VarDecl) 
}
```
2. a navigation window function abstract class:
Such class hold some common logic for navigation window function. i.e., `LAG`, `LEAD`, and in future `FIRST_VALUE`, `LAST_VALUE` etc. 
In the future we can create other "sub-interface" such as ranking window function interface, aggregation window function interface, etc. 
3. `Lag`, `Lead` function class : 
Notice that we did not further abstract `Lag`, `Lead` into a `LagLeadCommon`, for two reasons, 1. we use function name to get function object in the physical operator class, 2) such abstraction, although can remove some duplicated code, arguable make the logic less "clear cut" 


lang/src/org/partiql/lang/eval/physical/window/WindowFunctionSignature.kt
```diff
@@ -0,0 +1,12 @@
+package org.partiql.lang.eval.physical.window
+
+/**
+ * Window function signature
+ *
+ * For now just holding function name
+ *
+ * In the future we can add additional parameter such as frame information here.
+ */
+class WindowFunctionSignature(
```
johnedquinn, line 10: Will there be an eventual use case for the signature? Would it be of use to make this a data class?

lang/src/org/partiql/lang/eval/physical/window/NavigationWindowFunction.kt
```diff
@@ -0,0 +1,37 @@
...
+import org.partiql.lang.eval.physical.EvaluatorState
+import org.partiql.lang.eval.physical.SetVariableFunc
+import org.partiql.lang.eval.physical.operators.ValueExpression
+
+/**
+ * This abstract class holds some common logic for navigation window function, i.e., LAG, LEAD
+ *
+ * TODO: When we support FIRST_VALUE, etc, we probably need to modify the process row function, since those function requires frame
+ */
+abstract class NavigationWindowFunction() : WindowFunction {
```
johnedquinn, line 13: I'm curious whether NavigationWindowFunction and Lag/Lead need to be public.

lang/src/org/partiql/lang/eval/physical/operators/WindowOperatorFactoryDefault.kt
```diff
@@ -0,0 +1,111 @@
...
+
+        return relation(RelationType.BAG) {
+            partition.forEach { rowsInPartition ->
+                val windowFunc = compiledWindowFunction.func
+                // set the window function partition to the current partition
+                windowFunc.reset(rowsInPartition)
+
+                rowsInPartition.forEach { row ->
+                    // reset state
+                    state.load(row)
```
johnedquinn, line 100: With the implementation of NavigationWindowFunction, I don't think this load is necessary.

lang/src/org/partiql/lang/eval/physical/PhysicalBexprToThunkConverter.kt
```diff
@@ -311,12 +313,26 @@ internal class PhysicalBexprToThunkConverter(
...
+        }
+
+        // We need compiled window function here.
+        val compiledWindowFunction = CompiledWindowFunction(
+            builtinWindowFunctionsMap[node.windowExpression.funcName.text] ?: error("window function not supported yet"),
+            compiledWindowFunctionParameter,
+            node.windowExpression.decl.toSetVariableFunc()
+        )
+
+        // val allWindowFunctionsMap = builtinWindowFunctionsMap
```
johnedquinn, line 329: Can remove

lang/src/org/partiql/lang/eval/physical/window/Lead.kt
```diff
@@ -0,0 +1,44 @@
...
+    override fun processRow(state: EvaluatorState, arguments: List<ValueExpression>, currentPos: Int): ExprValue {
+        val (target, offset, default) = when (arguments.size) {
+            1 -> listOf(arguments[0], null, null)
+
+            2 -> listOf(arguments[0], arguments[1], null)
+
+            3 -> listOf(arguments[0], arguments[1], arguments[2])
+
+            else -> error("Wrong number of Parameter for Lag Function")
+        }
```
johnedquinn, line 23: nit -- excessive spacing

lang/src/org/partiql/lang/eval/physical/PhysicalBexprToThunkConverter.kt
```diff
@@ -311,12 +313,26 @@ internal class PhysicalBexprToThunkConverter(
             exprConverter.convert(it).toValueExpr(it.metas.sourceLocationMeta)
         }
 
+        // window function map
+        val builtinWindowFunctions = createBuiltinWindowFunctions()
+        val builtinWindowFunctionsMap = builtinWindowFunctions.associateBy {
+            it.signature.name
+        }
```
johnedquinn, line 320: I'm curious if there's a better way to create compiled window functions. Seemingly, for several nested window operators, we'll be creating all builtin functions even when we don't use them. I'm also curious about the modeling of CompiledWindowFunctions vs WindowFunctions.

johnedquinn (commented):
Overall some really great work! This PR really ties in some of the questions I had on the last PR. My comments really revolve around:
- visibility modifiers
- the need for one or more window functions per operator
- and how CompiledWindowFunctions, WindowFunctions, and WindowFunctionSignatures will all play unique roles in the evaluation.

Oh, and total side note -- I know you left some comments regarding experimental changes -- but we don't really expose that as part of API documentation. It could be nice to see our first @RequiresOptIn come from these PRs if we will be treating the public APIs as truly experimental (obviously not including the generated code since we currently don't have that freedom).

Great work.

lang/src/org/partiql/lang/eval/physical/window/NavigationWindowFunction.kt
yliuuuu: NavigationWindowFunction may or may not be. The idea is that if a customer want to create their own "custom window function", he or she could just pick an abstract class as the base class and implement their own codes. Kind of similar to our `ExprFunc`. `Lag/Lead` do not need to be public. Good call.

lang/src/org/partiql/lang/eval/physical/window/WindowFunctionSignature.kt
yliuuuu: Potentially we might be able to put the frame clause here in the future. If so, we can potentially group the window functions by partition, i.e., each window operator will have a list of window function that shares the same window partition but may differ at window frame. This is an optimization work so we don't need to recalculate the partition multiple times.

lang/src/org/partiql/lang/eval/physical/window/NavigationWindowFunction.kt
```diff
@@ -23,13 +25,16 @@ abstract class NavigationWindowFunction() : WindowFunction {
...
     ) {
         state.load(currentPartition[currentPos])
         val value = processRow(state, arguments, currentPos)
         // before we declare the window function result, we need to go back to the current row
         state.load(currentPartition[currentPos])
-        windowVarDecl(state, value)
+        windowVarDecl.toSetVariableFunc().invoke(state, value)
+        // make sure the change of state is reflected in the partition
+        // so the result of the current window function won't get removed by the time we process the next window function at the same row level.
+        currentPartition[currentPos][windowVarDecl.index.value.toInt()] = state.registers[windowVarDecl.index.value.toInt()]
```
johnedquinn, line 37: I left a note within `WindowFunction` regarding this exact line. If we were to go about creating the `partitions` to be a list of `EvaluatorState`, then the suggestion would actually be:
```suggestion
        setWindowVar.invoke(state, value)
        setWindowVar.invoke(currentPartition[currentPos], value)
```

Assuming `setWindowVar` is a `SetVariableFunc`, and `currentPartition[currentPos]` is an `EvaluatorState`.

lang/src/org/partiql/lang/eval/physical/operators/WindowRelationalOperatorFactory.kt
```diff
@@ -1,28 +1,27 @@
...
     /** Creates a [RelationExpression] instance for [PartiqlPhysical.Bexpr.Window]. */
     abstract fun create(
         source: RelationExpression,
-
         windowPartitionList: List<ValueExpression>,
-
         windowSortSpecList: List<CompiledSortKey>,
-
-        windowFunction: CompiledWindowFunction
+        windowFunction: List<CompiledWindowFunction>
```
johnedquinn, line 17:
```suggestion
        windowFunctions: List<CompiledWindowFunction>
        // or
        compiledWindowFunctions: List<CompiledWindowFunction> // Similar to the default impl
```

johnedquinn (commented):
Minor comments -- main one is about VarDecl vs SetVariableFunc. Open for a discussion regarding which one to use. But beyond that, this PR is great. Thanks for addressing comments.

lang/src/org/partiql/lang/eval/physical/window/NavigationWindowFunction.kt
yliuuuu: Love the comment here. I actually modeled a few different way:
1. Each partition is a array of `register`, and we use SetVariableFunc:
    Doing so we loss the ability to get the index of the global register the result of the given window function should go to, the solution will be to loop and rewrite the state's register. Now if the state contains a large number of element, the overhead is large. 
2. Each partition is a array of `register` and we use `VarDecl`:  (This is the current approach)
    As you pointed out, this way is not uniform with other operator, and directly access and modifying the "register alike data" is indeed an unsafe operation in that, if there were to be a bug, it would be difficult to debug... But this does reduce the overhead. 

Please correct me if I am wrong but you seem to be suggesting a third way: which is 
Each partition is a array of `EvalutorState` and we use `SetVariableFunc`:
I have not yet modeled this approach, but my first concern is that if we create a evaluatorState for essentially each row in the table, the overhead is quite large because for each of those we have to carry a reference of session data.

lang/src/org/partiql/lang/eval/physical/window/ExperimentalWindowFunc.kt
```diff
@@ -0,0 +1,5 @@
+package org.partiql.lang.eval.physical.window
+
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION, AnnotationTarget.LOCAL_VARIABLE)
+@RequiresOptIn(message = "This API is experimental. It may be changed in the future without notice.", level = RequiresOptIn.Level.ERROR)
+annotation class ExperimentalWindowFunc
```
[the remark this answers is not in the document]
yliuuuu: If necessary, we can remove this when merging the feature branch to main.
But for now, if you agree, I would leave it as is since we are only merging this to the feature branch.

lang/src/org/partiql/lang/eval/physical/window/NavigationWindowFunction.kt
johnedquinn: Yeah, I'm not excited about the state either. Just wanted to bring it up as something the other operators have used. Anyways, since we have yet to fully decide on how we're handling state, we'll leave this as a TODO to come back to.

lang/src/org/partiql/lang/planner/transforms/optimizations/ConcatWindowFunction.kt
```diff
@@ -0,0 +1,92 @@
+package org.partiql.lang.planner.transforms.optimizations
+
+import org.partiql.lang.domains.PartiqlPhysical
+import org.partiql.lang.errors.ProblemHandler
+import org.partiql.lang.planner.PartiqlPhysicalPass
+
+fun createConcatWindowFunctionPass(): PartiqlPhysicalPass =
```
johnedquinn, line 7: Can you make this internal?

yliuuuu, line 7:
```suggestion
internal fun createConcatWindowFunctionPass(): PartiqlPhysicalPass =
```

johnedquinn (approved):
Great work! Before merging to main, we'll take a look @ the experimental annotation and the use of VarDecl vs SetVariableFunc. Really great work!

PR #831, opened 2022-10-05
Final changes for REPLACE/UPSERT/INSERT EXCLUDED

### LINKED ISSUE #704: Request to support `INSERT` DML with `ON CONFLICT DO REPLACE EXCLUDED` clause
My team could really use support for the `ON CONFLICT DO REPLACE EXCLUDED` clause from the [INSERT, UPSERT, and REPLACE](https://github.com/partiql/partiql-docs/pull/11/files) RFC PR.  At the moment we wish to support "put semantics", that is, if a record already exists with the specified primary key then the entire record is replaced, otherwise a new record is inserted.

As of today, the parser has only limited support for the `ON CONFLICT` clause which does not include `DO REPLACE EXCLUDED`, and then only when `INSERT INTO ... VALUE` used (which is not even part of that RFC).

Examples:

```
INSERT into foo <<{'id': 1, 'name':'bob'}>> 
```

```
INSERT into foo SELECT bar.id, bar.name  FROM bar
```

Success criteria:

- Parser can parse the examples above.
- Modification of any planner APIs related to DML so that the embedding application can check if `ON CONFLICT DO REPLACE EXCLUDED` was specified or not. (i.e. `DmlCommand`).

Note that my team does not require any other syntax proposed in the RFC not shown above.

am357:
## Relevant Issues
- Closes #704 

## Description
- This PR includes the following:
 - Support for `INSERT INTO tbl ... ON CONFLICT DO REPLACE EXCLUDED` and `REPLACE INTO tbl ...` evaluation.
 - Support for `INSERT INTO tbl ... ON CONFLICT DO UPDATE EXCLUDED` logical planning.
    - The reason that evaluation is excluded is b/c the logic is for update evaluation requires implementing merge with existing values which makes it more completed; since it's not part of the related ask (attached issue) we defer the implementation for now.
- Adds parsing support for `AS alias` for `UPSERT INTO` and `REPLACE INTO` statements.

## Other Information
Yes.

No.

No.

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Parsing of `INSERT` DML with `ON CONFLICT DO REPLACE EXCLUDED` based on [RFC-0011](https://github.com/partiql/partiql-docs/blob/main/RFCs/0011-partiql-insert.md)
 - Adds experimental parsing of `REPLACE INTO` and `UPSERT INTO` DML commands pending approval of the following RFC for moving out of experimental:
   - https://github.com/partiql/partiql-docs/issues/27
-- Logical plan representation of `INSERT` DML with `ON CONFLICT DO REPLACE EXCLUDED` based on [RFC-0011](https://github.com/partiql/partiql-docs/blob/main/RFCs/0011-partiql-insert.md)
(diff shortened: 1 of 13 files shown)


RCHowell (dismissed):

dlurton (approved):
A high level review LGTM.

PR #836, opened 2022-10-10
Applied comments in PR 822

lziq:
Some comments original from https://github.com/partiql/partiql-lang-kotlin/pull/822 are for `main` branch instead of OTS feature branch. This PR aims to apply those comments to `main`. The changes include: 
1. Make function `PartiqlPhysical.Type.toTypedOpParameter()` internal
2. Remove arguments from `PartiqlAst.Type.toTypedOpParameter()` and make it private 
3. Make `PartiqlAstSanityValidator` internal. 
4. Removed `PartiqlPhysicalSanityValidator`, and moved the validation there to `AstToLogicalVisitorTransform`

Note that don't worry too much about if passing argument of `TypedOpParameters` is correct, since we will deprecate and remove it as we approach PartiQL 1.0


lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -326,6 +365,55 @@ internal class AstToLogicalVisitorTransform(
...
+            if (setQuantifier is PartiqlAst.SetQuantifier.Distinct && metas.containsKey(IsCountStarMeta.TAG)) {
+                err(
+                    "COUNT(DISTINCT *) is not supported",
+                    ErrorCode.EVALUATOR_COUNT_DISTINCT_STAR,
+                    errorContextFrom(metas),
+                    internal = false
+                )
+            }
+        }
 }
```
RCHowell, line 417: Validation should be its own pass using Visitor. It's not a transform

lang/src/org/partiql/lang/planner/transforms/LogicalToLogicalResolvedVisitorTransform.kt
```diff
@@ -195,6 +198,22 @@ internal data class LogicalToLogicalResolvedVisitorTransform(
...
+            locals.forEachIndexed { idx, it ->
+                if (it.registerIndex.value != idx.toLong()) {
+                    throw EvaluationException(
+                        message = "Variable index must match ordinal position of variable",
+                        errorCode = ErrorCode.INTERNAL_ERROR,
+                        errorContext = propertyValueMapOf(),
+                        internal = true
+                    )
+                }
+            }
```
RCHowell, line 216: again, validation doesn't belong in transform passes. Were these changes meant to be included?

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
lziq: Will apply this comment

lang/src/org/partiql/lang/planner/transforms/LogicalToLogicalResolvedVisitorTransform.kt
lziq: I agree that validation should have its own pass. But do you think we really need the check here?

lziq: It is trying to ensure the code we are writing is correct

RCHowell (approved):

PR #839, opened 2022-10-13
Cleans up SORT and UNPIVOT factories

johnedquinn:
## Relevant Issues
- Related To #821

## Description
- While deliberating over operator factories, decided to update the other recently implemented factories to match the same style.
- Totally just an aesthetic thing. Not pressed to merge this -- just creating this PR so that others can discuss if we'd like to adopt the same style. If so, we can think about merging this.
- TLDR:
  - Each factory just returns a single implementation
  - Therefore, each factory doesn't need an Impl for input
  - Making each factory an object
  - Each uses the DEFAULT_IMPL_NAME

## Other Information
- Updated Unreleased Section in CHANGELOG: No, changing names.
  - No, we haven't released these factories to the public yet.

## License Information


RCHowell (approved):

RCHowell:
This is spot on as discussed. Nice work

PR #845, opened 2022-10-18
Bumps partiql-isl-kotlin to 0.2.2

RCHowell:
## Relevant Issues
Version bump with no changes simplifies 0.8.0 release. Starting at 1.0.0 we should consider having vended libs share the library version as this is avoid these bumps and is common elsewhere

## License Information


johnedquinn (approved):
LGTM

Issue #616, opened 2022-05-27
Change `errorContextArg` of `SqlException```to non-nullable.


dlurton:
In a comment in #612:

> It [errorContextArg] is nullable because it always has been nullable.  Changing this to an argument and defaulting the `errorContext` property to an empty map if `errorContextArg` is null was a way of simplifying error handling code that reads context properties without changing all of the call sites of this constructor (a few  direct, but many indirect) to conform its new signature, which would have resulted in many more noisy diffs in this series of PRs.  If you want, I can undo this change in another commit against this PR, but I would consider making this argument non-nullable to be out of scope of this work.

_Originally posted by @dlurton in https://github.com/partiql/partiql-lang-kotlin/pull/612#discussion_r883929394_

vgapeyev:
Addressed in PR#847.

PR #847, opened 2022-10-18
Makes SqlException.errorContext non-nullable, resolving Issue #616.

vgapeyev:
## Relevant Issues
- Closes Issue #616

## Description

This is a simple PR: There was an in-code TODO at SqlException, pointing at Issue #616.
The suggestion was to change the field to non-nullable, switching from null to an empty PropertyValueMap as the declared default.  It turned out that many use sites have already provided non-null actuals, so even a default is not necessary in a few places.  The only subclass that had use sites with null actuals (coming through the default) was EvaluationException, so its default did get changed rather than removed. 

## Other Information
  - This is an internal change.

## License Information

### DIFF
--- a/lang/src/org/partiql/lang/SqlException.kt
+++ b/lang/src/org/partiql/lang/SqlException.kt
@@ -18,7 +18,6 @@ import org.partiql.lang.errors.ErrorCode
 import org.partiql.lang.errors.Property
 import org.partiql.lang.errors.PropertyValueMap
 import org.partiql.lang.errors.UNKNOWN
-import org.partiql.lang.util.propertyValueMapOf
 
 /**
  * General exception class for the interpreter.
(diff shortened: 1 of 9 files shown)


RCHowell:
Looks good to me, but I'm curious why the partiql-tests git module was included? Perhaps you can do `git submodule update --init` and update the PR?

vgapeyev:
> 

Not sure what is going on with the submodule.  If it has been introduced recently, I must have missed the memo on how to deal with it.  

Here is what happened on my side, if anyone can do the forensics: I started this PR off main, before 55de0988 came in (which had a change in test/partiql-tests). After that, I pulled myself up on main, branched and committed.  Looking at the history, this commit reverted what 55de0988 did in test/partiql-tests. Why? -- someone with better understanding of git is needed. With "regular" files the above workflow did not give me grief before.

At this point, the suggested `git submodule update --init` did not have a detectable effect.  
I think I'll just revert the revert that I see in test/partiql-tests and hope for the better.

vgapeyev:
Well `git restore --source 55de0988 test/partiql-tests` did not have a visible effect either.  

@alancai98 do you have any suggestions?

Otherwise, not understanding what is gong on, I'd just cancel this PR and do a new one with the same delta.

alancai98:
> Well `git restore --source 55de0988 test/partiql-tests` did not have a visible effect either.  
@alancai98 do you have any suggestions?
Otherwise, not understanding what is gong on, I'd just cancel this PR and do a new one with the same delta. 

I'm not sure why Robert's suggested command doesn't work to update the git submodule. An alternative is to manually go to the submodule and pull in the latest changes: i.e.

1. `cd test/partiql-tests`
2. `git pull`

Alternatively from the root directory, I think you can run `git submodule foreach git pull origin main` to have each submodule update with what's in `main`.

vgapeyev:
@RCHowell , @alancai98 thanks for the catch and for the assistance.  The diff looks as expected now.

alancai98 (approved):

PR #854, opened 2022-10-26
Updates the CLI Shell Highlighter to use ANTLR-generated lexer/parser

### LINKED ISSUE #748: Create PartiQLParser CLI Highlighter

johnedquinn:
## Relevant Issues
- Closes #748

## Description
- Updates the Shell Highlighter to use ANTLR generated lexer/parser instead of SqlParser
- There are several new additions to the PartiQLParser (DML and more) that are not able to be highlighted by the existing Shell Highlighter
- Changes made:
  - Created an UNRECOGNIZED token (which grabs anything not matching)
  - Pushed comments and whitespace to hidden channel (so we can reproduce in CLI)
  - Since we add to hidden channel, I needed to update the way we get parameter indexes because it uses a method that only tacks ON channel tokens.
  - Also found a bug with `visitLiteralIon` in `PartiQLVisitor` -- if Ion can't parse the value, it throws an unhandled exception, and we need to catch it (addressed in this PR as well)

## Other Information

## License Information

### DIFF
--- a/cli/src/org/partiql/shell/ShellHighlighter.kt
+++ b/cli/src/org/partiql/shell/ShellHighlighter.kt
@@ -28,19 +28,21 @@ import org.partiql.lang.syntax.antlr.PartiQLTokens
 import java.nio.charset.StandardCharsets
 import java.util.regex.Pattern
 
-private const val ADD_TO_GLOBAL_ENV_STR = "!add_to_global_env"
-private val ALLOWED_SUFFIXES = setOf("!!")
-
-private val STYLE_COMMAND = AttributedStyle.DEFAULT.foreground(AttributedStyle.CYAN)
-private val STYLE_KEYWORD = AttributedStyle.BOLD.foreground(AttributedStyle.CYAN).bold()
-private val STYLE_DATATYPE = AttributedStyle.DEFAULT.foreground(AttributedStyle.GREEN)
-private val STYLE_IDENTIFIER = AttributedStyle.DEFAULT.foreground(AttributedStyle.BRIGHT)
-private val STYLE_STRING = AttributedStyle.DEFAULT.foreground(AttributedStyle.YELLOW)
-private val STYLE_NUMBER = AttributedStyle.DEFAULT.foreground(AttributedStyle.BLUE)
-private val STYLE_COMMENT = AttributedStyle.DEFAULT.foreground(AttributedStyle.BRIGHT).italic()
-private val STYLE_ERROR = AttributedStyle.DEFAULT.foreground(AttributedStyle.RED)
-
 internal class ShellHighlighter : Highlighter {
+
+    companion object {
+        private const val ADD_TO_GLOBAL_ENV_STR = "!add_to_global_env"
+        private val ALLOWED_SUFFIXES = setOf("!!")
+
+        private val STYLE_COMMAND = AttributedStyle.DEFAULT.foreground(AttributedStyle.CYAN)
+        private val STYLE_KEYWORD = AttributedStyle.BOLD.foreground(AttributedStyle.CYAN).bold()
+        private val STYLE_DATATYPE = AttributedStyle.DEFAULT.foreground(AttributedStyle.GREEN)
+        private val STYLE_IDENTIFIER = AttributedStyle.DEFAULT.foreground(AttributedStyle.BRIGHT)
+        private val STYLE_STRING = AttributedStyle.DEFAULT.foreground(AttributedStyle.YELLOW)
+        private val STYLE_NUMBER = AttributedStyle.DEFAULT.foreground(AttributedStyle.BLUE)
+        private val STYLE_COMMENT = AttributedStyle.DEFAULT.foreground(AttributedStyle.BRIGHT).italic()
+        private val STYLE_ERROR = AttributedStyle.DEFAULT.foreground(AttributedStyle.RED)
+    }
 
     override fun highlight(reader: LineReader, line: String): AttributedString {
 
@@ -67,20 +69,13 @@ internal class ShellHighlighter : Highlighter {
         val usableInput = input.substring(0, lastValidQueryIndex)
 
         // Build Token Colors (Last Token is EOF)
-        val stream = getTokenStream(usableInput)
-        stream.fill()
-        val tokenIter = stream.tokens.iterator()
+        val tokenIter = getTokenStream(usableInput).also { it.fill() }.tokens.iterator()
         var builder = AttributedStringBuilder()
-        var hasError = false
         while (tokenIter.hasNext()) {
             val token = tokenIter.next()
-            val type = token.type
-            val text = token.text
+            val (type, text) = token.type to token.text
             when {
-                isError(hasError, type) -> {
-                    hasError = true
-                    builder.styled(STYLE_ERROR, text.removeSuffix("<EOF>"))
-                }
+                isUnrecognized(type) -> builder.styled(STYLE_ERROR, text)
                 isDatatype(type) -> builder.styled(STYLE_DATATYPE, text)
                 isIdentifier(type) -> builder.styled(STYLE_IDENTIFIER, text)
                 isString(type) -> builder.styled(STYLE_STRING, text)
@@ -88,7 +83,7 @@ internal class ShellHighlighter : Highlighter {
                 isComment(type) -> builder.styled(STYLE_COMMENT, text)
                 isIonMode(type, text) -> builder.append(text)
                 isKeyword(type, text) -> builder.styled(STYLE_KEYWORD, text)
-                isEOF(type) -> { /* Do nothing */ }
+                isEOF(type) -> builder.styled(STYLE_ERROR, text.removeSuffix("<EOF>"))
                 else -> builder.append(text)
             }
         }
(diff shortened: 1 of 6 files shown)


lang/antlr/PartiQLTokens.g4
```diff
@@ -362,13 +362,16 @@ IDENTIFIER_QUOTED
...
 COMMENT_SINGLE
-    : '--' ~[\r\n]* '\r'? '\n'? -> skip;
+    : '--' ~[\r\n]* '\r'? '\n'? -> channel(HIDDEN);
 
 COMMENT_BLOCK
-    : '/*' .*? '*/' -> skip;
+    : '/*' .*? '*/' -> channel(HIDDEN);
+
+UNRECOGNIZED
+    : . ;
```
yliuuuu, line 374: Could you please explain a little more on what does this means? 

The parser now could not parse Ion Literal, i.e., 1d0. After some experiment, removing this. fixed the issue. Not exactly sure why since I am not that familiar with ANTLR, but hopefully this can help you debug.

lang/src/org/partiql/lang/syntax/PartiQLParser.kt
```diff
@@ -95,9 +95,12 @@ internal class PartiQLParser(
         val tokenIndexToParameterIndex = mutableMapOf<Int, Int>()
         var parametersFound = 0
         val tokens = CommonTokenStream(lexer)
```
yliuuuu, line 97:
```suggestion
        val stream = CommonTokenStream(lexer).fill()
```

lang/src/org/partiql/lang/syntax/PartiQLParser.kt
```diff
@@ -95,9 +95,12 @@ internal class PartiQLParser(
...
-        for (i in 0 until tokens.numberOfOnChannelTokens) {
-            if (tokens[i].type == GeneratedParser.QUESTION_MARK) {
-                tokenIndexToParameterIndex[tokens[i].tokenIndex] = ++parametersFound
+        tokens.fill()
+        val tokenIter = tokens.tokens.iterator()
```
yliuuuu, line 99:
```suggestion
        val tokenIter = stream.tokens.iterator()
```

lang/antlr/PartiQLTokens.g4
yliuuuu: See here
```
 Parser Error: unexpected token found, IDENTIFIER : '`1d0`'
```

cli/src/org/partiql/shell/ShellHighlighter.kt
```diff
@@ -13,130 +13,114 @@
...
-                }
-            } else {
-                error
-            }
-        }
+        val usableInput = input.substring(0, lastValidQueryIndex)
 
         // Build Token Colors (Last Token is EOF)
+        val stream = getTokenStream(usableInput)
+        stream.fill()
```
yliuuuu, line 71:
```suggestion
        val stream = getTokenStream(usableInput).fill()
```

cli/src/org/partiql/shell/ShellHighlighter.kt
```diff
@@ -13,130 +13,114 @@
...
+                    builder.styled(STYLE_ERROR, text.removeSuffix("<EOF>"))
                 }
+                isDatatype(type) -> builder.styled(STYLE_DATATYPE, text)
+                isIdentifier(type) -> builder.styled(STYLE_IDENTIFIER, text)
+                isString(type) -> builder.styled(STYLE_STRING, text)
+                isLiteral(type) -> builder.styled(STYLE_NUMBER, text)
+                isComment(type) -> builder.styled(STYLE_COMMENT, text)
+                isIonMode(type, text) -> builder.append(text)
+                isKeyword(type, text) -> builder.styled(STYLE_KEYWORD, text)
+                isEOF(type) -> { /* Do nothing */ }
```
yliuuuu, line 91: Is it possible (or necessary) to model those better? Like an Enum or a map?

cli/src/org/partiql/shell/ShellHighlighter.kt
```diff
@@ -148,97 +132,101 @@ internal class ShellHighlighter() : Highlighter {
...
+    PartiQLTokens.NULL,
+    PartiQLTokens.MISSING,
+    PartiQLTokens.BOOL,
+    PartiQLTokens.BOOLEAN,
+    PartiQLTokens.ANY -> true
+    else -> false
 }
 
-private fun ansi(string: String, style: AttributedStyle) = AttributedString(string, style).toAnsi()
+private fun isIonMode(type: Int, text: String) = type == PartiQLTokens.EOF && text.contains("<EOF>").not()
```
yliuuuu, line 204: Not sure I follow the logic here?

cli/src/org/partiql/shell/ShellHighlighter.kt
```diff
@@ -148,97 +132,101 @@ internal class ShellHighlighter() : Highlighter {
...
+    PartiQLTokens.FLOAT,
+    PartiQLTokens.CHAR,
+    PartiQLTokens.CHARACTER,
+    PartiQLTokens.VARYING,
+    PartiQLTokens.VARCHAR,
+    PartiQLTokens.NULL,
+    PartiQLTokens.MISSING,
+    PartiQLTokens.BOOL,
+    PartiQLTokens.BOOLEAN,
+    PartiQLTokens.ANY -> true
```
yliuuuu, line 200: Less new line if possible?

yliuuuu (changes requested):
Thank you for putting this together. It seems like the change on PartiQLTokens somehow affect our ability to parse Ion literal. Other than that just some small comments.

lang/antlr/PartiQLTokens.g4
johnedquinn: So, the Gradle ANTLR plugin loves to use the Gradle cache.

My output:
```
PartiQL> `1d0`
   | 
===' 
1
--- 
OK!
```

To remove your Gradle cache and run the CLI, run:
```
./gradlew clean                                 && \
rm -rf ~/.gradle/caches/build-cache-1/          && \
./gradlew clean assemble                        && \
./cli/shell.sh
```

johnedquinn: BTW, whenever the PartiQLTokens file has been updated and you see an error like the one you've seen, it's likely due to the Gradle cache.

lang/src/org/partiql/lang/syntax/PartiQLParser.kt
johnedquinn: `.fill()` is a void-returning function. But, agreed, it's worth changing to be more aesthetic.

Proposing the following instead:
```
val tokenStream = CommonTokenStream(lexer).also { it.fill() }
```

Committing in next revision.

cli/src/org/partiql/shell/ShellHighlighter.kt
johnedquinn: As mentioned above, `.fill()` is a void-returning function. But changing to use `.also()`.

lang/antlr/PartiQLTokens.g4
johnedquinn: Also, for your first question -- since the token `UNRECOGNIZED` is at the bottom of the grammar (excluding ION mode), it's a catch-all for syntactic errors. If, in the middle of lexing, the lexer throws a Recognition Exception, we can't recover a single token. So, in the middle of typing something like:
```
1 + 'hello
```

without the second quotation, then we don't have access to any of the tokens, because it can't tokenize `'hello`.

So, the simplest way to get around this is an `UNRECOGNIZED` token that isn't used in the Parser. It just allows us to tokenize (then we only need to parse, which will throw a location exposing error). Another interesting thing about ANTLR is that the implementation will actually catch RecognitionExceptions (for lexing) and then just not consume the faulty token, and it'll just swallow the exception. So, it intentionally skips them. You can work around this by throwing an exception that isn't RecognitionException -- but then we're left with the inability to get tokens that can lex.

cli/src/org/partiql/shell/ShellHighlighter.kt
johnedquinn: I actually prefer this way -- specifically because some of these functions are more than just `when` statements. Like, to remove verbosity, I opted to have a `when` statement for `isDatatype` which intentionally evaluates before `isKeyword`. And the reason for this is because `isKeyword` will actually take the input string and look into the PartiQLTokens vocabulary to see if it matches the keywords there (and datatypes) are keywords too.

Anyways, with this way, instead of having to write out every single keyword into an enum or map, I only need to write out the data types.

cli/src/org/partiql/shell/ShellHighlighter.kt
johnedquinn: Leaving a note there. Essentially, ANTLR treats island modes differently.

johnedquinn:
> Thank you for putting this together. It seems like the change on PartiQLTokens somehow affect our ability to parse Ion literal. Other than that just some small comments.

Yeah! For the Ion literals, you just need to clear the Gradle cache as mentioned in other comment. The ANTLR Gradle plugin loves to use the cache. Submitting a revision soon.

cli/src/org/partiql/shell/ShellHighlighter.kt
yliuuuu: No objection.

yliuuuu (approved):
Discussed some concerns offline. Great job!

PR #856, opened 2022-10-26
Deprecates Lexer, Token, TokenType, SourcePosition, and SourceSpan

johnedquinn:
## Relevant Issues
- None

## Description
- Deprecates Lexer, Token, TokenType, SourcePosition, and SourceSpan
- There isn't any major advantage for exposing the above mentioned classes. As we expose the ANTLR files, any customer applications can resort to using the generated files if they so desire.
- This will overall greatly simplify the scope of our public APIs in regards to syntax.

## Other Information
  - Upon review, I can update the CHANGELOG to be:
    - `Deprecates Lexer, Token, TokenType, SourcePosition, and SourceSpan`
  - These are deprecated. Not deleted. Will be removed after v0.9.0+.

## License Information


lang/src/org/partiql/lang/syntax/Lexer.kt
```diff
@@ -12,13 +12,19 @@
...
+
 package org.partiql.lang.syntax
 
 /**
  * Tokenizes a [String] representation of a query into a [List] of [Token].
  *
  * Implementations must be thread-safe.
  */
+@Deprecated(
+    message = "This will be removed after v0.8.1+.",
```
RCHowell, line 25:
```suggestion
    message = "This will be removed before v0.9.0",
```

lang/src/org/partiql/lang/syntax/SourcePosition.kt
```diff
@@ -15,6 +15,10 @@
 package org.partiql.lang.syntax
 
 /** Simple source line/column value. */
+@Deprecated(
+    message = "This will be removed after v0.8.1+.",
```
RCHowell, line 19: Same suggestion here. Following similar wording as Calcite https://github.com/apache/calcite/search?q=deprecated

RCHowell (approved):

PR #868, opened 2022-11-02
POC: Removed Factory Pattern

lziq:
A POC PR for removing factory pattern to construct `ExprValue`


lziq:
Closed with https://github.com/partiql/partiql-lang-kotlin/pull/922/files published

PR #870, opened 2022-11-02
Adds collection aggregation functions

### LINKED ISSUE #837: Add collection aggregate functions
## Relevant Issue/Bug
- Somewhat relevant to #821

## Requested Solution/Feature
- We are initializing the `aggregate` relalg operator in the above linked PR
- However, in PartiQL Eval Compiler, we also support using `aggregate` functions on collections of scalars.
- We also mention its use within the specification -- search 'COLL_AVG'
- So, we should add support for `COLL_{AGG}` and also rewrite any `{AGG}` that actually is performing an aggregation on a collection of scalars.

## Describe Alternatives
- A clear and concise description of any alternative solutions or features you've considered.

## Additional Context
- I actually have a version of this working locally, but as we're waiting for #821 to be merged, I'm creating this issue to keep track.
- My local changes can correctly evaluate the following query:
  - Code:
    ```
    PartiQL> SELECT g, SUM(b) AS sum_b, COLL_SUM('all', a) AS coll_sum_a, COLL_AVG('all', a) AS coll_avg_a
       |   FROM << {'a': [0, 1], 'b': 3}, {'a': [10, 11], 'b': 4} >> AS t
       |   WHERE AVG(a) > 0.3
       |   GROUP BY a GROUP AS g
       |   ORDER BY SUM(b) DESC
    ```
  - Analysis:
    - All `COLL_SUM`'s in projection list are an explicit call to aggregate what `a` currently is at that moment (not an actual aggregation) -- AKA `a` is a list for both tuples.
    - The `SUM` in the projection list is an aggregation function. Needs to act on `b` which is a scalar binding.
    - The `AVG` in the `WHERE` clauses gets internally converted into a `COLL_AVG`
    - The `SUM` in the `ORDER BY` clause is an aggregate `SUM`

## DoD (Definition of Done)
- Something similar to the above additional context

johnedquinn:
## Relevant Issues
- Closes #837 

## Description
- Adds collection aggregation functions (COLL_SUM, COLL_AVG, COLL_MIN, etc)
- Creates a sealed class `CollectionAggregationFunction`, where each implementation just leverages the aggregation logic used by the aggregate operator. See Accumulator. We also add these to the built-in functions.
- Modifies the AstToLogical:
  - In the normal flow, all CallAggs in a SFW query will be sent to the aggregate operator and replaced with identifiers
  - After this happens, we just need to replace any remaining aggregation functions with the equivalent collection aggregation function.
  - Similarly, we also need to replace any CallAggs that are outside of the SFW query. So, this PR also does that.

## Other Information
- Any backward-incompatible changes? **YES** -- CallAgg is removed from Logical/Physical Plan

## License Information

### DIFF
--- a/lang/resources/org/partiql/type-domains/partiql.ion
+++ b/lang/resources/org/partiql/type-domains/partiql.ion
@@ -589,6 +589,10 @@ may then be further optimized by selecting better implementations of each operat
         (with expr
             // Remove the select and struct node from the `expr` sum type, which will be replaced below.
             (exclude select struct)
+              
+            // CallAgg's such as SUM, MIN, MAX, etc. either become calls to built-in-functions (example: call COLL_SUM)
+            // or aggregate functions within the logical aggregate operator.
+            (exclude call_agg)
 
             (include
                 // Invokes `exp` once in the context of every binding tuple returned by `query`, returning a
--- a/lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
+++ b/lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
@@ -1,12 +1,14 @@
 package org.partiql.lang.planner.transforms
 
 import com.amazon.ionelement.api.emptyMetaContainer
+import com.amazon.ionelement.api.ionString
 import com.amazon.ionelement.api.ionSymbol
 import org.partiql.lang.domains.PartiqlAst
 import org.partiql.lang.domains.PartiqlAstToPartiqlLogicalVisitorTransform
 import org.partiql.lang.domains.PartiqlLogical
 import org.partiql.lang.errors.Problem
 import org.partiql.lang.errors.ProblemHandler
+import org.partiql.lang.eval.builtins.CollectionAggregationFunction
 import org.partiql.lang.eval.physical.sourceLocationMetaOrUnknown
 import org.partiql.lang.eval.visitors.VisitorTransformBase
 import org.partiql.lang.planner.PlanningProblemDetails
@@ -65,6 +67,18 @@ internal class AstToLogicalVisitorTransform(
             is PartiqlAst.SetQuantifier.Distinct -> call("filter_distinct", expr)
             else -> expr
         }
+    }
+
+    // This transformation is used for top-level expression transformations and for the SFW clauses prior to the
+    //  `aggregation` relational algebra operator.
+    override fun transformExprCallAgg(node: PartiqlAst.Expr.CallAgg): PartiqlLogical.Expr = PartiqlLogical.build {
+        call(
+            "${CollectionAggregationFunction.collectionAggregationPrefix}${node.funcName.text}",
+            listOf(
+                lit(ionString(node.setq.javaClass.simpleName.toLowerCase())),
+                transformExpr(node.arg)
+            )
+        )
     }
 
     /**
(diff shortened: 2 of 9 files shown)


lang/resources/org/partiql/type-domains/partiql.ion
```diff
@@ -589,6 +589,9 @@ may then be further optimized by selecting better implementations of each operat
         (with expr
             // Remove the select and struct node from the `expr` sum type, which will be replaced below.
             (exclude select struct)
+              
+            // Call Aggs either become calls to built-in-functions are aggregate functions
```
RCHowell, line 593:
```suggestion
            // A call_agg node represents either a collection aggregation function or an aggregation function of the  grouping operator
```

RCHowell: This is strange actually. Why would we conflict these two? Any of the collection aggregation functions like "COLL_" should be represented by the normal function IR. By representing both of these things by `call_agg` requires splitting them later. It's over-complicating things

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -58,6 +62,8 @@ internal class AstToLogicalVisitorTransform(
 
         var (select, algebraAfterAggregation) = transformAggregations(node, algebra) ?: node to algebra
 
+        select = removeCallAggregations(callAggregationReplacer, select)
```
RCHowell, line 65: The "Call Aggregations" should never be considered "grouping operator aggregation functions". Their related insofar as they have a similar name and computation, but a "Call Aggregation" shouldn't be conflated with traditional relation-level aggs

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -504,6 +526,31 @@ private class FromSourceToBexpr(
...
 }
 
+/**
+ * This class is used to transform all instances of [PartiqlAst.Expr.CallAgg] into function calls denoted with
+ * COLL_${AGG}, where AGG represents the aggregation call being transformed. This class is designed to be used
+ * AFTER replacing all aggregation functions via [AstToLogicalVisitorTransform.transformAggregations].
+ *
+ * This transformer is designed to be called directly on a [PartiqlAst.Expr.Select] and does NOT recurse into
+ * encountered [PartiqlAst.Expr.Select]'s.
+ */
```
RCHowell, line 536: I don't believe this is necessary if "COLL_*" are just represented as function calls to begin with. We know they're different when we parse hence the name

lang/src/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -67,6 +69,18 @@ internal class AstToLogicalVisitorTransform(
         }
     }
 
+    // This transformation is used for top-level expression transformations and for the SFW clauses prior to the
+    //  `aggregation` relational algebra operator.
+    override fun transformExprCallAgg(node: PartiqlAst.Expr.CallAgg): PartiqlLogical.Expr = PartiqlLogical.build {
+        call(
+            "${CollectionAggregationFunction.collectionAggregationPrefix}${node.funcName.text}",
```
RCHowell, line 76: Fine for now, but assume "builtins" is not treated specially. That is, in a world where functions are pluggable, we shouldn't have compiler logic dependent on a particular function library.

RCHowell (approved):

Issue #682, opened 2022-07-22
Remove + delete `ExprNode`


alancai98:
`ExprNode` has been deprecated since [v0.6.0](https://github.com/partiql/partiql-lang-kotlin/releases/tag/v0.6.0-alpha). The parser, evaluator, and AST rewriting have migrated to use PIG-generated `PartiqlAst`. By the v1.0.0 release of `partiql-lang-kotlin`, we should remove and delete `ExprNode` and its related files (e.g. [ast.kt](https://github.com/partiql/partiql-lang-kotlin/blob/ff041b28c6db53853c3b370690246bc01913a400/lang/src/org/partiql/lang/ast/ast.kt)).

yliuuuu:
Using this comment to track the progress on removing the ExprNode in our codebase: 
Step one: Removal of V0 ast ( Meta related changes are excluded in this step). PR #863. 
Step two: Meta related clean up, focusing on removal of the deprecated `MetaContainer`. PR #869

yliuuuu:
Step Three: Remove rewriter, ast walker/ visitor. 
Step Four: Remove the dependencies on ExprNode in our code base.
Issues left to address: issue #878 , issue #879

PR #872, opened 2022-11-03
removal rewriter/AstWalker/AstVisitor

yliuuuu:
## Relevant Issues
- [Related To] Issue #682 

## Description
- Removed:
   - AstRewriter Interface & AstRewriterBase class
   - AstVisitor Interface & AstVisitorBase class
   - AstWalker class
   - MetaStrippingRewriter class
   - AstRewriterBaseTest
   - AstWalkerTest
   - RewriterTestBase
   - Tests in ThreadInterruptedTests that tests on AstVisitor and AstWalker
   - AstRewriterBaseTestAdapter
   - ExprNodeTestCase
   - EvaluatorTestCasesAsExprNodeTestCases
   - PartiqlAstToExprNodeRoundTripTests (This is removed here because it has dependency on EvaluatorTestCasesAsExprNodeTestCases)

Note that to achieve the same functionality in testing, the followings are added:
- EvaluatorTestCasesAsStatementTestCases
- StatementTestCases
- VisitorTransformTestAdapter
- VisitorTransformBaseTest

## Other Information
  - Yes
  - Yes. The deprecated rewriter, astWalker, and astVisitor will be removed from our code base. 

  - No

## License Information


lang/test/org/partiql/lang/ast/passes/AstRewriterBaseTest.kt
```diff
@@ -1,26 +0,0 @@
...
-class AstRewriterBaseTest {
-    @Suppress("DEPRECATION")
-    private val defaultRewriter = AstRewriterBase()
-    @ParameterizedTest
-    @ArgumentsSource(EvaluatorTestCasesAsExprNodeTestCases::class)
-    fun identityRewriteTest(tc: ExprNodeTestCase) {
-        val rewritten = defaultRewriter.rewriteExprNode(tc.expr)
-        tc.assertEquals(rewritten)
-    }
-}
```
lziq, line 26: Do we also want to remove https://github.com/partiql/partiql-lang-kotlin/blob/removal-AstRewriter/lang/test/org/partiql/lang/util/testdsl/ExprNodeTestCase.kt?

CHANGELOG.md
```diff
@@ -41,6 +41,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - MetaContainerImpl
   - MetaDeserializer
   - MemoizedMetaDeserializer
+- Removes the deprecated Rewriter/AstWalker/AstVisitor in the code base, removed interfaces and classes include:
+  - AstRewriter Interface & AstRewriterBase class
+  - AstVisitor Interface & AstVisitorBase class
+  - AstWalker class
+  - MetaStrippingRewriter class
```
lziq, line 48: I think it's better to use brackets to surround the classes/interfaces referenced from the code, e.g., `AstWalker` -> `[AstWalker]`

lziq (commented):
Overall looks good to me. Only a few comments left.

lziq:
Another thing is that instead of deleting `ExprNode` & related API, it's better to first work on removing PartiQL compiler's APIs which are involved with `ExprNode`, e.g., `SqlParser.parseExprNode()`, so only APIs related to conversion between PIG AST & `ExprNode` exist. But I won't block this PR because of this.

yliuuuu:
> Another thing is that instead of deleting `ExprNode` & related API, it's better to first work on removing PartiQL compiler's APIs which are involved with `ExprNode`, e.g., `SqlParser.parseExprNode()`, so only APIs related to conversion between PIG AST & `ExprNode` exist. But I won't block this PR because of this.

I tried this before. Unfortunately, such top down approach, even though it is conceptually clear, leads to a huge diff and making the PR less reviewable IMO.

lziq (approved):

Issue #873, opened 2022-11-04
Remove `partiql_` prefix as annotation of `IonValue`


lziq:
## Description
In terms of representing a PartiQL value using `IonValue`, we are using `IonList` with annotation of `partiql_bag` to represent a partiql bag: https://github.com/partiql/partiql-lang-kotlin/blob/3640876a40cb42ce8b9d5ab54ac5d786513ebaf3/lang/src/org/partiql/lang/eval/ExprValueFactory.kt#L49 However, as what we discussed for PartiQL conformance test suite, the `partiql_` prefix in `partiql_bag` as annotation is not needed. Same for `partiql_missing`, `partiql_date` & `partiql_time`. 

## Expected Behavior
`partiql_` prefix should be removed from `partiql_bag`, `partiql_missing`, `partiql_date` & `partiql_time` annotation

PR #874, opened 2022-11-07
Remove `partiql_`  prefix

lziq:
## Description
Resolving https://github.com/partiql/partiql-lang-kotlin/issues/873

## Breaking Changes
- PartiQL MISSING in ION representation now becomes ion null with annotation of `$missing`, instead of `$partiql_missing`
- PartiQL BAG in ION representation now becomes ion list with annotation of `$bag`, instead of `$partiql_bag`
- PartiQL DATE in ION representation now becomes ion timestamp with annotation of `$date`, instead of `$partiql_date`
- PartiQL TIME in ION representation now becomes ion struct with annotation of `$time`, instead of `$partiql_time`


CHANGELOG.md
```diff
@@ -30,6 +30,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Now `CompileOption` uses `TypedOpParameter.HONOR_PARAMETERS` as default.
 - Updates the CLI Shell Highlighter to use the ANTLR generated lexer/parser for highlighting user queries
+- PartiQL MISSING in ION representation now becomes ion null with annotation of `$missing`, instead of `$partiql_missing`
+- PartiQL BAG in ION representation now becomes ion list with annotation of `$bag`, instead of `$partiql_bag`
+- PartiQL DATE in ION representation now becomes ion timestamp with annotation of `$date`, instead of `$partiql_date`
+- PartiQL TIME in ION representation now becomes ion struct with annotation of `$time`, instead of `$partiql_time`
```
alancai98, line 36: It's "Ion" not "ION`

```suggestion
- PartiQL MISSING in Ion representation now becomes ion null with annotation of `$missing`, instead of `$partiql_missing`
- PartiQL BAG in Ion representation now becomes ion list with annotation of `$bag`, instead of `$partiql_bag`
- PartiQL DATE in Ion representation now becomes ion timestamp with annotation of `$date`, instead of `$partiql_date`
- PartiQL TIME in Ion representation now becomes ion struct with annotation of `$time`, instead of `$partiql_time`
```

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
```diff
@@ -45,10 +45,10 @@ import java.math.BigInteger
 import java.time.LocalDate
 import kotlin.collections.asSequence
 
-internal const val MISSING_ANNOTATION = "\$partiql_missing"
-internal const val BAG_ANNOTATION = "\$partiql_bag"
-internal const val DATE_ANNOTATION = "\$partiql_date"
-internal const val TIME_ANNOTATION = "\$partiql_time"
+const val MISSING_ANNOTATION = "\$missing"
```
alancai98, line 48: Does this need to be public? It was previously internal?

cli/test/org/partiql/cli/CliTest.kt
```diff
@@ -35,7 +37,7 @@ class CliTest {
     private val ion = IonSystemBuilder.standard().build()
     private val output = ByteArrayOutputStream()
     private val testFile = File("test.ion")
-    private val partiqlBagAnnotation = "\$partiql_bag::"
```
alancai98, line 38: nit: is this variable necessary? Could we just remove instances of `partiqlBagAnnotation` with `BAG_ANNOTATION`?

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/Util.kt
```diff
@@ -7,17 +7,16 @@ import com.amazon.ion.IonSexp
...
+import org.partiql.lang.eval.MISSING_ANNOTATION
 import org.partiql.lang.eval.StructOrdering
 import org.partiql.lang.eval.name
 import org.partiql.lang.eval.namedValue
 import org.partiql.lang.eval.stringValue
 
-const val BAG_ANNOTATION = "\$bag"
-const val MISSING_ANNOTATION = "\$missing"
-
 /**
```
alancai98, line 20: Could we actually delete these extension functions and use the default `ExprValue` to `IonValue` conversion function? These extension functions only existed previously because the annotations differed between the conformance tests and the default Kotlin tests using Ion to represent PartiQL values.

cli/test/org/partiql/cli/CliTest.kt
alancai98: `MISSING_ANNOTATION` is used directly rather than using another variable.

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
lziq: I think it should be public

cli/test/org/partiql/cli/CliTest.kt
lziq: Will update

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/Util.kt
lziq: Yeah, I will do this in the PR to remove `ionValue` property in `ExprValue` interface

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
RCHowell: Could you please elaborate why? I would think the annotations are either all public, or all internal.

lziq: They are all public

lziq: PartiQL data should be able to be represented by `IonValue`. The annotation itself is considered the part of this representation, otherwise, we won't be able to differentiate 2 types. e.g. the only way to differentiate PartiQL BAG & LIST types in representation of `IonValue` is to check the annotation of the `IonList`. If it has `$bag` there, that means it is a BAG, otherwise, it is a LIST. This representation should be publicly visible to users.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/Util.kt
alancai98: Ok, can you cut an issue to track?

lang/src/org/partiql/lang/eval/ExprValueFactory.kt
alancai98: I agree that the representation should probably be public. I was just curious as to why these variables were internal previously and in this PR we had to make them public. I had thought the intent of this PR was to just remove the `partiql_` prefix from the Ion representation.

For Robert's comment, think there might be some confusion over the lines I selected for the PR comment. All the annotations were made public in this PR, but my comment only highlighted the first annotation.

alancai98 (approved):

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/Util.kt
lziq: https://github.com/partiql/partiql-lang-kotlin/issues/875

PR #877, opened 2022-11-09
Modified custom procedure example to remove the use of `ionValue`

lziq:
## Description
- Remove the dependency on `ionValue` property in the custom procedure example

### DIFF
--- a/examples/src/kotlin/org/partiql/examples/CustomProceduresExample.kt
+++ b/examples/src/kotlin/org/partiql/examples/CustomProceduresExample.kt
@@ -1,7 +1,5 @@
 package org.partiql.examples
 
-import com.amazon.ion.IonDecimal
-import com.amazon.ion.IonStruct
 import com.amazon.ion.system.IonSystemBuilder
 import org.partiql.examples.util.Example
 import org.partiql.lang.CompilerPipeline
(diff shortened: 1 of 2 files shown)


examples/src/kotlin/org/partiql/examples/CustomProceduresExample.kt
```diff
@@ -71,16 +68,23 @@ class CalculateCrewMoonWeight(private val valueFactory: ExprValueFactory) : Stor
         // Now that we've confirmed the given `crewName` is in the session's global bindings, we calculate and store
         // the moon weight for each crewmate in the crew.
         // In addition, we keep a running a tally of how many crewmates we do this for.
```
alancai98, line 70: Comments need updated

examples/src/kotlin/org/partiql/examples/CustomProceduresExample.kt
```diff
@@ -26,12 +27,8 @@ import java.math.RoundingMode
 private val ion = IonSystemBuilder.standard().build()
 
 /**
- * A simple custom stored procedure that calculates the moon weight for each crewmate of the given crew, storing the
- * moon weight in the [EvaluationSession] global bindings. This procedure also returns the number of crewmates we
- * calculated the moon weight for, returning -1 if no crew is found.
- *
- * This example demonstrates how to create a custom stored procedure, check argument types, and modify the
- * [EvaluationSession].
```
alancai98, line 34: Think we can include more of these comments. Everything but the last part about modifying the `EvaluationSession` seems the same.

alancai98 (approved):

PR #881, opened 2022-11-11
Deprecate `ExprValue.ionValue`

lziq:
## Description
- Deprecates `ionValue` property in `ExprValue` interface. 
- Adds Kotlin extension function `ExprValue.toIonValue(ion: IonSystem)` for converting an `ExprValue` to an `IonValue`. 
- Replaces dependency on `ExprValue.ionValue` with `ExprValue`'s native APIs in this repo. Now we transform `ExprValue` to an `IonValue` only in case that we want to print the result in `IonValue` form or we want to the get `IonValue` result from querying an Ion file. 

## Other Information
  - Yes
  - No


extensions/src/org/partiql/extensions/cli/functions/BaseFunction.kt
```diff
@@ -1,39 +0,0 @@
...
- */
-package org.partiql.extensions.cli.functions
-
-import com.amazon.ion.IonStruct
-import org.partiql.lang.eval.ExprFunction
-import org.partiql.lang.eval.ExprValue
-import org.partiql.lang.eval.ExprValueFactory
-
-abstract class BaseFunction(val valueFactory: ExprValueFactory) : ExprFunction {
-    protected fun optionsStruct(
```
lziq, line 22: This function is not used anywhere actually

lang/src/main/kotlin/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -1599,25 +1585,17 @@ internal class EvaluatingCompiler(
...
-        // We throw an exception here if the value exceeds the supported range (say if we change that
-        // restriction or if a custom [ExprValue] is provided which exceeds that value).
-        val limitIonValue = limitExprValue.ionValue as IonInt
-        if (limitIonValue.integerSize == IntegerSize.BIG_INTEGER) {
+        val originalLimitValue = limitExprValue.numberValue()
+        val limitValue = originalLimitValue.toLong()
+        if (originalLimitValue != limitValue as Number) { // Make sure `Number.toLong()` is a lossless transformation
             err(
-                "IntegerSize.BIG_INTEGER not supported for LIMIT values",
+                "Too large integer provided as LIMIT value",
```
RCHowell, line 1592: Can we say something more like `LIMIT value exceeds Long.MAX_VALUE` or something? I'm not sure what "too large" means without the context of the previous comment that was here.

lang/src/main/kotlin/org/partiql/lang/eval/ExprValue.kt
```diff
@@ -56,3 +78,114 @@ interface ExprValue : Iterable<ExprValue>, Faceted {
...
+    ionStruct.asSequence().map {
+        it.toExprValue().namedValue(valueFactory.newString(it.fieldName))
+    }
+) {
+    override val bindings: Bindings<ExprValue> =
+        IonStructBindings(valueFactory, ionStruct)
+
+    override val ionValue: IonValue
+        get() = ionStruct
+}
```
RCHowell, line 191: What was this created for? Just asking out of curiosity. It's a little difficult for me to pick out since this is a large PR. Is this for an explicit struct instead of partiql bindings?

lang/src/main/kotlin/org/partiql/lang/eval/ExprValueFactory.kt
```diff
@@ -184,11 +167,11 @@ private class ExprValueFactoryImpl(override val ion: IonSystem) : ExprValueFacto
...
 
-    override val missingValue = MissingExprValue(ion.newNull())
-    override val nullValue = NullExprValue(ion.newNull())
+    override val missingValue = MissingExprValue(ion)
+    override val nullValue = NullExprValue(ion)
 
-    private val trueValue = TrueBoolExprValue(ion.newBool(true))
-    private val falseValue = FalseBoolExprValue(ion.newBool(false))
+    private val trueValue = TrueBoolExprValue(ion)
+    private val falseValue = FalseBoolExprValue(ion)
```
RCHowell, line 174: I guess this a question for a different place, but why do we have two different True/False bool values? Is this because they're "atomic" and every "false" is the same "false"? Seems weird this way rather than a traditional constants in the top-level (BoolExprValue) class.

```kotlin
// private constructor, control instantiation
class BoolExprValue(private val value: Boolean) private constructor()

     companion object {
   
         // you can still instantiate from a primitive, but the authors have control
          fun of(value: Boolean) = when (value) {
              true -> BoolExprValue.TRUE
              false -> BoolExprValue.FALSE
          }
    }

// ... naming(?)
BoolExprValue.TRUE     // BoolExprValue(true)
BoolExprValue.FALSE    // BoolExprValue(false)
```

lang/src/main/kotlin/org/partiql/lang/eval/physical/operators/LimitRelationalOperatorFactory.kt
```diff
@@ -80,23 +78,18 @@ internal class LimitOperator(
...
-        // restriction or if a custom [ExprValue] is provided which exceeds that value).
-        val limitIonValue = limitExprValue.ionValue as IonInt
-        if (limitIonValue.integerSize == IntegerSize.BIG_INTEGER) {
+
+        val originalLimitValue = limitExprValue.numberValue()
+        val limitValue = originalLimitValue.toLong()
+        if (originalLimitValue != limitValue as Number) { // Make sure `Number.toLong()` is a lossless transformation
             err(
-                "IntegerSize.BIG_INTEGER not supported for LIMIT values",
+                "Too large integer provided as LIMIT value",
```
RCHowell, line 86: Same comment as the other LIMIT. We should provide the user with what "too large" means

RCHowell (changes requested):
Overall looks nice. I am looking forward to seeing this combined with removing the ExprValueFactory. Some minor comments

lang/src/main/kotlin/org/partiql/lang/eval/ExprValue.kt
lziq: See https://github.com/partiql/partiql-lang-kotlin/issues/908

lziq: This will be removed eventually. Ion structs will be the same as PartiQL structs.

lang/src/main/kotlin/org/partiql/lang/eval/ExprValueFactory.kt
lziq: Why is this different? The public API for constructing boolean values is `ExprValueFactory.newBool(value: Boolean)`

RCHowell (approved):

RCHowell (approved):

PR #882, opened 2022-11-12
Separate randomized tests 

yliuuuu:
## Relevant Issues
- N/A

## Description
- Create an additional gradle task that runs the fuzz tests. ( Randomized tests)
    - Before: 
        -  `./gradlew build` builds and runs all the tests. 
    - With this PR:
        - `./gradlew build` builds and runs the deterministic unit test only, enable a faster build time while developing new features.
        - `./gradlew :test:partiql-randomized-tests:test -PrandomTests` builds and runs the randomized test only. 
- Github workflow: 
   - Decoupling the unit test and the fuzz tests: See below: 
   - image

  - Ideally, while developing new feature, one would not need to run the randomized tests in the local environment. 
  - Enables gradle cache during the work flow build process. 
  - The workflow will not be trigger when pushing to a branch other than `main`. (We should prohibit such behavior at some points). This way we avoid double building (one on push and the other on pull request)

## Other Information
  - No
  - No.
  - No.

## License Information


RCHowell:
I'd suggest randomized testing is not included in the normal build. I'm curious if these tests need to be packaged within unit tests? Perhaps we set it up like so,

We may be able to place these tests into `test/partiql-randomized-tests`.

```shell
./gradlew build         # does not execute tests from partiql-randomized-tests
./gradlew lang:build    # only builds `lang` since randomized testing is separated
./gradlew partiql-randomized-tests:run  # execute randomized tests, to be used for Github workflow
```

lziq:
Sorry misclicked

lziq:
When I click the green triangle button in IntelliJ to test `PartiQL.lang.test`, I hope it does not cover the randomized tests. Is there a way to do so?

yliuuuu:
> When I click the green triangle button in IntelliJ to test `PartiQL.lang.test`, I hope it does not cover the randomized tests. Is there a way to do so?

It is now possible.

test/partiql-randomized-tests/build.gradle
```diff
@@ -0,0 +1,40 @@
...
+    testImplementation 'org.junit.vintage:junit-vintage-engine:5.7.0'
+    testImplementation 'org.junit.jupiter:junit-jupiter-params:5.7.0'
+}
+
+project(':test:partiql-randomized-tests') {
+    test {
+        onlyIf {
+            project.hasProperty("randomTests")
+        }
+    }
```
RCHowell, line 39: 👍

RCHowell (approved):

Issue #702, opened 2022-08-09
SELECT DISTINCT returns empty bag with type check enabled


yliuuuu:
**Problem:** 
https://github.com/partiql/partiql-lang-kotlin/blob/e54cf53d6f329bbd56b642d2c4af029e3ab73aca/lang/test/org/partiql/lang/eval/EvaluatorStaticTypeTests.kt#L58-L69

**Root Cause:** 
The root cause for this problem is this line in `evaluatingCompiler.kt`
``` kotlin
val quantifiedRows = when (selectExpr.setq ?: PartiqlAst.SetQuantifier.All()) {
       // wrap the ExprValue to use ExprValue.equals as the equality
        is PartiqlAst.SetQuantifier.Distinct -> projectedRows.filter(createUniqueExprValueFilter())  // <- Problem
        is PartiqlAst.SetQuantifier.All -> projectedRows
}.let { rowsWithOffsetAndLimit(it, env) }
```

`quantifiedRows` is a sequence, which is a data structure that allows lazy evaluation. With type checking enabled, the `createUniqueExprValueFilter()` will be triggered during the evaluation of `thunk`, adding elements in `quantifiedRows` to `seen`, a data structure we use to record visited elements. i.e., by the time the evaluation of `quantifiedRows` is trigger, seen has already recorded all the (distinct) elements in `quantifiedRows`, making the output to be an empty bag. 

``` kotlin
    private fun checkEvaluationTimeType(thunkResult: ExprValue, metas: MetaContainer): ExprValue {
       ... ...
        if (!staticType.isInstance(thunkResult)) {     // <- invoke the lazy evaluation
            ...
        }
        return thunkResult
    }
```

**Additional Implication:** 
Similar issue can be reconstructed without type checking enabled,

``` kotlin
        val pipeline = CompilerPipeline.build(IonSystemBuilder.standard().build()) {
        compileOptions {
            typingMode(TypingMode.PERMISSIVE)
            thunkOptions { }
        }
    }
        val expr = pipeline.compile("SELECT DISTINCT t.a FROM <<{'a': 1}, {'a': 2}, {'a': 1}>> as t")
        val result = expr.eval(EvaluationSession.standard())
        println("result is $result") 
        println("result.ion value is ${result.ionValue}")
```
The output will be: 
```
result is <<{'a': 1}, {'a': 2}>>
result.ion value is $partiql_bag::[]
```

**Potential Fix:** 
[...]

PR #887, opened 2022-11-16
fix distinct issue

yliuuuu:
## Relevant Issues
- [Closes] Issue #702 

## Description
- With this quick fix, we are able to run distinct on the same `sequence<ExprValue>` multiple times.
- A particular use case is running the query with type inferencer enabled. 
- 7 failed type inferencer test cases are passing now. 

## Other Information
  - No, minor bug fix. 
  - No.
  - No.

## License Information

### DIFF
--- a/lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
+++ b/lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
@@ -701,16 +701,14 @@ internal fun createUniqueExprValueFilter(): (ExprValue) -> Boolean {
 }
 
 fun Sequence<ExprValue>.distinct(): Sequence<ExprValue> {
-    val seen = TreeSet(DEFAULT_COMPARATOR)
     return sequence {
+        val seen = TreeSet(DEFAULT_COMPARATOR)
         this@distinct.forEach {
             if (!seen.contains(it)) {
                 seen.add(it.unnamedValue())
                 yield(it)
             }
         }
-        // clear out the seen set in case we need to call distinct multiple times, i.e. Type Inferencer
-        seen.clear()
     }
 }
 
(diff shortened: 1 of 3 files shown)


lang/src/org/partiql/lang/eval/ExprValueExtensions.kt
```diff
@@ -709,6 +709,8 @@ fun Sequence<ExprValue>.distinct(): Sequence<ExprValue> {
                 yield(it)
             }
         }
+        // clear out the seen set in case we need to call distinct multiple times, i.e. Type Inferencer
+        seen.clear()
```
RCHowell, line 713: An extension function becomes a static function in Java, so the TreeSet was being instantiated in the static class which caused an unintentional global. It should be moved inside the sequence creation

```suggestion
fun Sequence<ExprValue>.distinct() = sequence {
    val seen = TreeSet(DEFAULT_COMPARATOR)
    this@distinct.forEach {
        if (!seen.contains(it)) {
            seen.add(it.unnamedValue())
            yield(it)
        }
    }
}
```

RCHowell: The nearby multiplicities function will also need to be fixed

yliuuuu: The multiplicities function should work fine based on the examination of the compiled java code. 

```
   public static final TreeMap multiplicities(@NotNull Sequence $this$multiplicities) {
      Intrinsics.checkNotNullParameter($this$multiplicities, "$this$multiplicities");
      // Seems like the a new treeMap is constructed every time the method got called. 
      TreeMap multiplicities = new TreeMap((Comparator)DEFAULT_COMPARATOR);
      int $i$f$forEach = false;
      Iterator var4 = $this$multiplicities.iterator();

      while(var4.hasNext()) {
         Object element$iv = var4.next();
         ExprValue it = (ExprValue)element$iv;
         int var7 = false;
         multiplicities.compute(it, (BiFunction)ExprValueExtensionsKt$multiplicities$1$1.INSTANCE);
      }

      return multiplicities;
   }
```

RCHowell (approved):

PR #890, opened 2022-11-17
Removes use of passed context to improve latency of parse

johnedquinn:
## Relevant Issues
- None

## Description
- Removes use of passed context to improve latency of parse
- While evaluating the effectiveness of SLL vs LL parsing, I noticed that almost all queries were seeing dramatically decreased latency. By examining the grammar, `PartiQL.g4`, I noted and tested several rules that would break the rules of SLL:
  - there are no empty right hand sides
  - each alternative of a non-terminal symbol begins with a unique terminal symbol
  - the grammar is not left recursive
- Beyond this, I was looking for oddly formulated rule definitions.
- There was a single rule that takes advantage of ANTLR's context-passing feature. I tested the feature using the following query: `SELECT a FROM a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p`. See the results below:

| Name | Mode | Count | Score | +/- Error | Units |
| -- | -- | -- | -- | -- | -- |
| PartiQLParser | avgt | 20 | 94598.178 | 7220.772 | microsec / op |
- After making the changes in this PR:

| Name | Mode | Count | Score | +/- Error | Units |
| -- | -- | -- | -- | -- | -- |
| PartiQLParser | avgt | 20 | 69.896 | 3.403 | microsec / op |

## Other Information

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - PartiQL DATE in Ion representation now becomes ion timestamp with annotation of `$date`, instead of `$partiql_date`
 - PartiQL TIME in Ion representation now becomes ion struct with annotation of `$time`, instead of `$partiql_time`
 - Simplifies the aggregation operator in the experimental planner by removing the use of metas
+- Increases the performance of the PartiQLParser by changing the parsing strategy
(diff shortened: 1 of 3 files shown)


alancai98 (approved):
Looks good! Thanks for the deep dive into the SLL parsing strategy, and it's great that you found a fix in the grammar

PR #905, opened 2022-11-29
Added `ExpectedResultFormat.STRICT` and migrated `PARTIQL`-format tests to it

lziq:
Adds `PARTIQL_STRICT` test format for testing `ExprValue` in strict mode.


lang/src/test/kotlin/org/partiql/lang/ExprValueStrictEqualsTest.kt
```diff
@@ -0,0 +1,152 @@
+package org.partiql.lang
```
lziq, line 1: Actually, I don't know if the packaging of `ExprValueStrictEquals.kt` & `ExprValueStrictEqualsTest.kt` is correct. The former is a tool used in test, and the latter is test to this tool.

lang/src/main/kotlin/org/partiql/lang/util/ExprValueFormatter.kt
```diff
@@ -42,7 +43,15 @@ class ConfigurableExprValueFormatter(private val config: Configuration) : ExprVa
...
-                ExprValueType.INT, ExprValueType.DECIMAL -> out.append(value.scalar.numberValue().toString())
+                ExprValueType.INT -> out.append(value.scalar.numberValue().toString())
+                ExprValueType.DECIMAL -> {
+                    val decimalValue = value.scalar.numberValue() as? BigDecimal
+                    out.append(decimalValue.toString())
+                    // If this is a decimal with scale to be 0, we need to add a decimal point to differentiate it from an integer
+                    if (decimalValue != null && decimalValue.scale() == 0) {
+                        out.append(".")
+                    }
+                }
```
lziq, line 54: Originally, a decimal with scale to be 0 is printed the same as an integer. e.g. a decimal with value `1.0` is printed as `1`. The change here makes it printed as `1.`, which is a decimal in PartiQL syntax.

lziq: This change is related to this PR, since people normally write a decimal with scale of 0 as integer. e.g. if the result is `1.0`, they can probably write it as `1`. But this is not correct since the data type does not match. The change applied here prints the actual value as `1.`, instead of just `1`, which reminds people of the data type.

lang/src/test/kotlin/org/partiql/lang/eval/EvaluatingCompilerCollectionAggregationsTest.kt
```diff
@@ -83,7 +83,7 @@ internal class EvaluatingCompilerCollectionAggregationsTest : EvaluatorTestBase(
                     + COLL_MAX('all', [1, 1, 2])     -- + 2
                     + COLL_COUNT('all', [1, 1, 2])   -- + 3 = 12
                 """,
-                expectedResult = "12"
+                expectedResult = "12."
```
lziq, line 86: Aggregate function `avg()` returns decimal, instead of integer

lang/src/test/kotlin/org/partiql/lang/eval/EvaluatingCompilerOrderByTests.kt
```diff
@@ -193,7 +193,7 @@ class EvaluatingCompilerOrderByTests : EvaluatorTestBase() {
             // text types compared by lexicographical ordering of Unicode scalar (ASC)
             EvaluatorTestCase(
                 """SELECT * FROM [{ 'a': `'\uD83D\uDCA9'`}, { 'a': 'Z'}, { 'a': '9' }, { 'a': 'A'}, { 'a': `"\U0001F4A9"`}, { 'a': 'a'}, { 'a': 'z'}, { 'a': '0' }] ORDER BY a""",
-                """[{'a': '0'}, {'a': '9'}, {'a': 'A'}, {'a': 'Z'}, {'a': 'a'}, {'a': 'z'}, {'a': `"\U0001F4A9"`}, {'a': `'\uD83D\uDCA9'`}]"""
+                """[{'a': '0'}, {'a': '9'}, {'a': 'A'}, {'a': 'Z'}, {'a': 'a'}, {'a': 'z'}, {'a': `'\uD83D\uDCA9'`}, {'a': `"\U0001F4A9"`}]"""
```
lziq, line 196: The order of 2 values needs to be swapped, since it was not right originally. But originally, we did not detect it, since for `ExprValue.exprEquals()`, the 2 values are considered as equal.

lang/src/test/kotlin/org/partiql/lang/eval/builtins/windowFunctions/WindowFunctionTests.kt
```diff
@@ -49,12 +49,12 @@ class WindowFunctionTests : EvaluatorTestBase() {
                     FROM stock_price as sp
                 """,
                 expectedResult = """<<
-                    { 'date': `2022-09-30`, 'ticker': 'AMZN', 'current_price': 113.00, 'previous_price': NULL},
```
lziq, line 52: The value of `ticker` in the input is written as an ion symbol, instead of a string.

lang/src/test/kotlin/org/partiql/lang/eval/builtins/windowFunctions/WindowFunctionTests.kt
```diff
@@ -303,10 +303,10 @@ class WindowFunctionTests : EvaluatorTestBase() {
                     GROUP BY EXTRACT(MONTH FROM sp."date") as month, sp.ticker as ticker GROUP AS g
                 """,
                 expectedResult = """<<
-                    { 'current_month': 9, 'ticker': 'AMZN', 'current_month_average': 113.00, 'next_month_avg': 118.485},
-                    { 'current_month': 10, 'ticker': 'AMZN', 'current_month_average': 118.485, 'next_month_avg': NULL},
-                    { 'current_month': 9, 'ticker': 'GOOG', 'current_month_average': 96.15, 'next_month_avg': 100.17},
-                    { 'current_month': 10, 'ticker': 'GOOG', 'current_month_average': 100.17, 'next_month_avg': NULL}
+                    { 'current_month': 9., 'ticker': `AMZN`, 'current_month_average': 113.00, 'next_month_avg': 118.485},
```
lziq, line 306: The `EXTRACT(MONTH FROM sp."date")` also returns decimal in the current implementation of PartiQL

lziq: I don't know if we want to change the evaluation logic in `EXTRACT`, so it outputs an integer for month

lang/src/test/kotlin/org/partiql/lang/eval/evaluatortestframework/ExprValueStrictEqualsTest.kt
```diff
@@ -0,0 +1,153 @@
...
+import com.amazon.ion.system.IonSystemBuilder
+import org.junit.Assert
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ArgumentsSource
+import org.partiql.lang.CompilerPipeline
+import org.partiql.lang.eval.EvaluationSession
+import org.partiql.lang.util.ArgumentsProviderBase
+
+class ExprValueStrictEqualsTest {
+    data class TestCase(val q1: String, val q2: String)
```
vgapeyev, line 12: I think this tests collection will read more convincingly if equality and inequality tests where coming together for each expression kind.

lang/src/test/kotlin/org/partiql/lang/eval/evaluatortestframework/ExprValueStrictEqualsTest.kt
```diff
@@ -0,0 +1,153 @@
...
+            TestCase("[{'a': 1, 'b': 2}, {'a': 3, 'b': 4}]", "[{'b': 2, 'a': 1}, {'b': 4, 'a': 3}]"),
+            // SEXP
+            TestCase("sexp(1,'2',TRUE)", "sexp(1,'2',TRUE)"),
+            TestCase("sexp(1,'2',<<1,'2',TRUE>>)", "sexp(1,'2',<<1, '2', TRUE>>)"),
+            TestCase("sexp({'a': 1, 'b': 2}, {'a': 3, 'b': 4})", "sexp({'b': 2, 'a': 1}, {'b': 4, 'a': 3})"),
+            // BAG
+            TestCase("<<1,'2',TRUE>>", "<<1,'2',TRUE>>"),
+            TestCase("<<TRUE,1,'2'>>", "<<TRUE,1,'2'>>"),
+            TestCase("<<1,'2',[1,'2',TRUE]>>", "<<[1,'2',TRUE],1,'2'>>"),
+            TestCase("<<{'a': 1, 'b': 2}, {'a': 3, 'b': 4}>>", "<<{'b': 4, 'a': 3}, {'b': 2, 'a': 1}>>"),
```
vgapeyev, line 71: How about some tests where number of occurrences plays out?  E.g. `<<1, 2, 1>>` and `<<1, 1, 2>>` being equal.  Even more interestingly, for structs.

lang/src/test/kotlin/org/partiql/lang/eval/evaluatortestframework/ExprValueStrictEquals.kt
```diff
@@ -0,0 +1,146 @@
...
+import org.partiql.lang.eval.numberValue
+import org.partiql.lang.eval.stringValue
+import org.partiql.lang.eval.timeValue
+import org.partiql.lang.eval.timestampValue
+
+/**
+ * Only used to verify the actual `ExprValue` is the same as the expected `ExprValue` in test
+ *
+ * Ensures 2 [ExprValue]s to have the same type & value
+ */
```
vgapeyev, line 20: I believe the doc comments could give a tad more insight into what you have figured out.  
What follows are some suggestions.  Fell free to modify, esp. make sure they don't say something that's actually wrong. 

/**
Strict equality that captures equality in the PartQL data model.
That is, two differently-constructed [ExprValue] values are never equal, unless they are bags or structs, where ordering of members is ignored. 
In particular, values are never equal if they belong to different types (such as numerics INT and FLOAT), making this equality different from [ExprValue.exprEquals].  

This implementation is currently only intended for use in tests, as it has not been validated for efficiency.   
*/

lang/src/test/kotlin/org/partiql/lang/eval/evaluatortestframework/ExprValueStrictEquals.kt
```diff
@@ -0,0 +1,146 @@
...
+fun ExprValue.strictEquals(other: ExprValue): Boolean =
+    ExprValueStrictComparator.compare(this, other) == 0
+
+/**
+ * We implement this comparator since to compare 2 bags, we sort them first and then compare each item.
+ *
+ * The order defined in this comparator of [ExprValue] is not related to the order defined by [ExprValue.exprEquals()]
+ */
```
vgapeyev, line 28: /** 
A comparator defining an (arbitrary) total order on [ExprValue].

This comparator is a private implementation detail of [strictEquals]. 
It does not have any interesting semantic meaning, except for being able to compare any two values in a manner consistent with their strictEquals identity.  This is essential for ignoring element order when comparing bags and structs.
*/

lang/src/test/kotlin/org/partiql/lang/eval/evaluatortestframework/ExprValueStrictEquals.kt
```diff
@@ -14,17 +14,26 @@ import org.partiql.lang.eval.timeValue
...
+ * Strict equality that captures equality in the PartQL data model. That is, for 2 scalars, they are considered as equal
+ * if they have the same type & value. For 2 collections, each element in one collection must be strictly equal to
+ * an element in the other collection. The equality of collections respects elements order for sexp & list, and not for
+ * bag & struct.
  *
- * Ensures 2 [ExprValue]s to have the same type & value
+ * Note that, the equality we define here is different from [ExprValue.exprEquals]. That one is used for PartiQL equals
+ * operator & ORDER BY clause. The equality there ignores data types in some cases, e.g., the partiql expressions
+ * `1.0 = 1` returns true (`1.0` is a decimal, while `1` is an integer).
+ *
```
vgapeyev, line 25: "ignores data types in some cases" --> "equates values of distinct data types in some cases"

And, as we are here, a few more nits I'd might let pass otherwise: 

- & on line 23 is particularly egregious; do "and" at least here; 
- exprEquals is not used in ORDER BY -- a comparison consistent with it is what is used there; I'd reword as "... different from [ExprValue.exprEquals], which is based on the comparison in PartiQL ORDER BY clause. The comparison there ignores/equates ..."  
- "the partiql expressions `1.0 = 1`" --> "the PartiQL expression `1.0 = 1`"
- ... --> "even though 1.0 is a decimal"

lang/src/test/kotlin/org/partiql/lang/eval/evaluatortestframework/ExprValueStrictEqualsTest.kt
```diff
@@ -9,145 +9,116 @@ import org.partiql.lang.eval.EvaluationSession
...
     @ParameterizedTest
-    @ArgumentsSource(EqualCases::class)
-    fun equalityTests(testCase: TestCase) {
+    @ArgumentsSource(StrictEqualCases::class)
+    fun strictEqualsTests(testCase: TestCase) {
         val v1 = pipeline.compile(testCase.q1).eval(session)
         val v2 = pipeline.compile(testCase.q2).eval(session)
 
-        Assert.assertEquals(true, v1.strictEquals(v2))
+        Assert.assertEquals(testCase.equals, v1.strictEquals(v2))
```
vgapeyev, line 24: My IntelliJ complains about this line, which looks like a complaint about Junit4 and Junit5 being mixed in this file, which does look to me like a legit complaint.

lang/src/test/kotlin/org/partiql/lang/eval/evaluatortestframework/ExprValueStrictEqualsTest.kt
```diff
@@ -9,145 +9,116 @@ import org.partiql.lang.eval.EvaluationSession
...
     }
 
-    class EqualCases : ArgumentsProviderBase() {
+    class StrictEqualCases : ArgumentsProviderBase() {
         override fun getParameters(): List<Any> = listOf(
             // MISSING
-            TestCase("MISSING", "MISSING"),
+            TestCase("MISSING", "MISSING", true),
+            TestCase("MISSING", "1", false),
```
vgapeyev, line 31: I guess I would have preferred "self-describing" `EqTest`, `NeqTest` instead of `TestCase`, but if going with a true/false flag, maybe at least rename to `EqTest`?

vgapeyev (changes requested):

vgapeyev (approved):
Ok, we can move on from here, I think.  Thanks for bearing with all the comments.  
I took liberty of changing the PR title to better reflect what it ended up being -- in the hope this will become the git blame message after the squish-merge.

lziq:
Thanks!

Issue #895, opened 2022-11-19
Add Performance Benchmarking to GH Actions


johnedquinn:
## Relevant Issue/Bug
- Related to #886 -- increases performance of PartiQLParser
- Related to #890 -- increases performance of PartiQLParser
- Related to #891 -- requests tracing in PartiQL

## Requested Solution/Feature
- It would be beneficial to have performance benchmarking as part of our CI (GitHub Actions).
- This needs research into how we can compare between either the latest version of `main`, or the last published version of `partiql-lang-kotlin` -- or both.
- We currently have performance benchmarks within `lang/jmh`, but this should be expanded to test all of the components vended by PartiQL.

## Describe Alternatives
- None at the moment.

## Additional Context
- The PartiQLParser saw significantly increased latency, and therefore, it has become one of PartiQL's priorities to focus on performance during code releases.

## DoD (Definition of Done)
- Comprehensive performance benchmarking added to all major components of PartiQL that gets triggered before every PR and/or release.

johnedquinn:
This is closed by #921

PR #921, opened 2022-12-05
Adds performance benchmarking to GitHub actions

johnedquinn:
## Relevant Issues
- Related To #895 

## Description
- Adds parser benchmarks
- Adds continuous JMH benchmarking via GH Actions for the existing benchmarks
- On pushes to main:
  - runs JMH benchmarks
  - alerts users / repo-owner on commit if latency breaches double the previous latency (by tagging them in the comment)
  - publishes the results to GH pages
- On manual triggers:
  - runs JMH benchmarks
  - alerts users (NOT repo-owner) on commit if latency breaches double the previous latency (by tagging them in the comment)
  - does NOT publish the results to GH pages
- The fork has its [GH pages](https://johnedquinn.github.io/partiql-lang-kotlin/) published. See the link for how the results look

## Other Information
- Any new external dependencies? **YES**
  - Yes, but the external dependency is NOT added to our library. It's added to our CI.

## License Information


lang/src/jmh/kotlin/org/partiql/jmh/benchmarks/ParserBenchmark.kt
```diff
@@ -0,0 +1,1368 @@
...
+            businessPageCategoryId as pagefeedcategoryid,(25) as defaultCategoryType from businessPageFeedInfo 
+            UNION 
+            select newsfeedId as feedid,(NULL) as title,description,(NULL) as shortDescription,(NULL) as newsUrl,
+            (NULL) as externalUrlOption, categoryMappingId as pagefeedcategoryid,
+            (20) as defaultCategoryType from mediaPageFeedInfo 
+            ) page 
+            ON newsfeed.id = page.feedId WHERE privacy != 10 
+        """.trimIndent()
+
+        val queries = listOf(
```
yliuuuu, line 1325: Nit: May be consider making this a map, so instead of using index, one can use variable name, which is actually meaningful here in this context.

lang/src/jmh/kotlin/org/partiql/jmh/benchmarks/ParserBenchmark.kt
```diff
@@ -0,0 +1,1368 @@
...
+            veryLongQuery01, // 32
+        )
+
+        val queriesFailingHalfway = queries.map { query ->
+            val splitQuery = query.split("\\s".toRegex()).toMutableList()
+            val index = (splitQuery.lastIndex * .50).toInt()
+            splitQuery.add(index, ";")
+            splitQuery.joinToString(separator = " ")
+        }
+    }
```
yliuuuu, line 1367: Curious on the location of the error chosen here.  It seems like it is always going to be at the end of the query, but it may not necessary be the deepest level of the AST. Will this matter?

lang/src/jmh/kotlin/org/partiql/jmh/utils/BenchmarkConstants.kt
```diff
@@ -0,0 +1,21 @@
...
+ * language governing permissions and limitations under the License.
+ */
+
+package org.partiql.jmh.utils
+
+internal const val FORK_VALUE_RECOMMENDED: Int = 2
+internal const val MEASUREMENT_ITERATION_VALUE_RECOMMENDED: Int = 10
+internal const val MEASUREMENT_TIME_VALUE_RECOMMENDED: Int = 1
+internal const val WARMUP_ITERATION_VALUE_RECOMMENDED: Int = 5
+internal const val WARMUP_TIME_VALUE_RECOMMENDED: Int = 1
```
yliuuuu, line 21: I see a quite large fluctuation in the  example page, especially for commit 6d22a1
image.

I could not quite related the jump to the code changed, maybe play around with the parameter here to see if the result can be stabilized a little more? 

See here: https://stackoverflow.com/questions/43763530/how-to-choose-jmh-configuration-parameters

yliuuuu (approved):
Thanks for putting together. My primary concerns is that fluctuation of the result set, other than that just some minor comments.

yliuuuu (changes requested):
Wrong option clicked, sorry...

CHANGELOG.md
```diff
@@ -31,6 +31,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   counterparts (in the experimental planner)
 - Experimental implementation for window function `Lag` and `Lead`.
 - Adds support for EXPLAIN
+- Adds continuous performance benchmarking to the CI for existing JMH benchmarks
+  - Benchmark results can be seen on the project's GitHub Pages site
```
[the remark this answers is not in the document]
johnedquinn: Unsure -- but I saw previous CHANGELOG updates regarding Codecov reports and README badges. Doesn't matter to me.

.github/workflows/benchmark-main.yml
```diff
@@ -0,0 +1,64 @@
...
+jobs:
+  benchmark:
+    name: Run JMH Benchmarks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'corretto'
+          java-version: '11'
```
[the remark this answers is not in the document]
johnedquinn: I don't think it's necessary IMO. We're just looking to compare performance between commits.

lang/src/jmh/kotlin/org/partiql/jmh/benchmarks/ParserBenchmark.kt
johnedquinn: As long as we're consistent with where we are throwing errors. I did it halfway through the query.

johnedquinn:
Alright, I've updated the workflow to use 2 forks, 10 iterations, and 5 warmup iterations -- only on pushes to main.

The workflow will no longer run on PRs -- as the workflow takes a substantial amount of time. However, it is up to the PR author's discretion (or the PR reviewers' discretion) whether to manually trigger a benchmark run (especially if their changes may have an effect on performance).

The [benchmarks](https://johnedquinn.github.io/partiql-lang-kotlin/dev/bench/) displayed in the fork have largely stabilized, though the fluctuation can be as much as 160%. I've increased the threshold to 250%. With this, we'll have pretty good knowledge about our performance changes over time (with associated commit numbers) -- and we'll be notified if any major performance degradations occur.

yliuuuu (approved):
Thank you for putting this together. Looks good to me.

PR #922, opened 2022-12-06
Deprecate facotry pattern

lziq:
## Description
- Currently, we are using the factory pattern [`ExprValueFactory`](https://github.com/partiql/partiql-lang-kotlin/blob/3d0c596d5fa390d8e0a9c98becc1540ddd8c7bea/lang/src/main/kotlin/org/partiql/lang/eval/ExprValueFactory.kt#L42) to construct `ExprValue`. This was necessary since we wanted to guarantee the `IonSystem` in each `ExprValue` is the same. However, since we are moving towards decoupling `ExprValue` from `IonValue` (Deprecation done in https://github.com/partiql/partiql-lang-kotlin/pull/881), we won't have this requirement any more. Thus, we will be able to remove `IonSystem` property in the `ExprValueFactory`. Based on this fact, the same design of `ExprValueFacotry` will complicate construction of `ExprValue` (since users need to initialize one `ExprValueFacotry`, then call construction methods from there). So, in this PR, we are working towards deprecating it and implement new public methods for `ExprValue` construction, instead. 

## Other Information
  - Yes
  - Yes, we are deprecating `ExprValueFactory` in this PR
  - Yes, new public methods for `ExprValue` construction are added in package `org.partiql.lang.eval`: 
    - `exprMissing(): ExprValue`
    - `exprNull(ionType: IonType = IonType.NULL): ExprValue`
    - `exprBoolean(value: Boolean): ExprValue`
    - `exprString(value: String): ExprValue`
    - `exprSymbol(value: String): ExprValue`
    - `exprInt(value: Long): ExprValue`
    - `exprInt(value: Int): ExprValue`
    - `exprFloat(value: Double): ExprValue`
    - `exprDecimal(value: BigDecimal): ExprValue`
    - `exprDecimal(value: Int): ExprValue`
    - `exprDecimal(value: Long): ExprValue`
    - `exprDate(value: LocalDate): ExprValue`
    - `exprDate(year: Int, month: Int, day: Int): ExprValue`
    - `exprDate(value: String): ExprValue`
    - `exprTimestamp(value: Timestamp): ExprValue`
    - `exprTime(value: Time): ExprValue`
    - `exprClob(value: ByteArray): ExprValue`
    - `exprBlob(value: ByteArray): ExprValue`
    - `exprList(values: Sequence<ExprValue>): ExprValue`
    - `exprList(values: Iterable<ExprValue>): ExprValue`
    - `emptyExprList(): ExprValue`
    - `exprBag(values: Sequence<ExprValue>): ExprValue`
    - `exprBag(values: Iterable<ExprValue>): ExprValue`
    - `emptyExprBag(): ExprValue`
    - `exprSexp(values: Sequence<ExprValue>): ExprValue`
    - `exprSexp(values: Iterable<ExprValue>): ExprValue`
    - `emptyExprSexp(): ExprValue`
    - `exprStruct(values: Sequence<ExprValue>, ordering: StructOrdering): ExprValue`
    - `exprStruct(values: Iterable<ExprValue>, ordering: StructOrdering): ExprValue`
    - `emptyExprStruct(): ExprValue`
    - `exprValue(ion: IonSystem, reader: IonReader): ExprValue`


vgapeyev:
Is this supposed to be marked as a draft? 
Otherwise, it needs a PR description: what are the core changes; why; anything to pay attention to? Point at the "factory pattern"; what replaces its functionality?

lziq:
> Is this supposed to be marked as a draft? Otherwise, it needs a PR description: what are the core changes; why; anything to pay attention to? Point at the "factory pattern"; what replaces its functionality?

Yeah, I am adding some documentations now.

lang/src/main/kotlin/org/partiql/lang/eval/ExprValueType.kt
```diff
@@ -31,21 +31,22 @@ enum class ExprValueType(
...
+    INT(isNumber = true, isScalar = true),
+    FLOAT(isNumber = true, isScalar = true),
+    DECIMAL(isNumber = true, isScalar = true),
+    DATE(isScalar = true),
+    TIMESTAMP(isScalar = true),
+    TIME(isScalar = true),
+    SYMBOL(isText = true, isScalar = true),
+    STRING(isText = true, isScalar = true),
+    CLOB(isLob = true, isScalar = true),
+    BLOB(isLob = true, isScalar = true),
```
lziq, line 49: The change here is for the change in `BagOp.kt`. See the comment there for more infor.

lang/src/main/kotlin/org/partiql/lang/eval/BagOp.kt
```diff
@@ -70,7 +70,7 @@ fun interface ExprValueBagOp {
  */
 internal fun ExprValue.coerceToBag(): Sequence<ExprValue> = when {
     isUnknown() -> emptySequence()
-    this is StructExprValue || this is Scalar -> sequenceOf(this)
+    type === ExprValueType.STRUCT || type.isScalar -> sequenceOf(this)
```
lziq, line 73: The original way to check a scalar depends on certain implementation. Scalar `ExprValue` does not have to extend `Scalar` interface. So for better maintainability, we use its type to check if it is a scalar.

lziq: To be more specific, the implementation of scalar `ExprValue` does not depend on `Scalar` interface.

lang/src/main/kotlin/org/partiql/lang/eval/builtins/MathFunctions.kt
```diff
@@ -79,16 +70,3 @@ private fun transformIntType(n: BigInteger): Number = when (n) {
...
-    is Int -> valueFactory.newInt(this)
-    is Long -> valueFactory.newInt(this)
-    is Double, is Float -> valueFactory.newFloat(this.toDouble())
-    is BigDecimal -> valueFactory.newDecimal(this)
-    else -> errNoContext(
-        "Cannot convert number to expression value: $this",
-        errorCode = ErrorCode.EVALUATOR_INVALID_CONVERSION,
-        internal = true
-    )
-}
```
lziq, line 94: This helper method was migrated to `NumberExtension.kt`

lang/src/main/kotlin/org/partiql/lang/eval/physical/operators/Accumulator.kt
```diff
@@ -169,15 +164,3 @@ internal fun checkIsNumberType(funcName: String, value: ExprValue) {
...
-    is Int -> valueFactory.newInt(this)
-    is Long -> valueFactory.newInt(this)
-    is Double -> valueFactory.newFloat(this)
-    is BigDecimal -> valueFactory.newDecimal(this)
-    else -> errNoContext(
-        "Cannot convert number to expression value: $this",
-        errorCode = ErrorCode.EVALUATOR_INVALID_CONVERSION,
-        internal = true
-    )
-}
```
lziq, line 183: This helper method is migrated to `NumberExtensions.kt`.

lang/src/main/kotlin/org/partiql/lang/util/NumberExtensions.kt
```diff
@@ -76,10 +82,11 @@ internal fun Number.isZero() = when (this) {
...
 }
 
+@Suppress("UNCHECKED_CAST")
 /** Provides a narrowing or widening operator on supported numbers. */
-fun Number.coerce(type: Class<out Number>): Number {
+fun <T> Number.coerce(type: Class<T>): T where T : Number {
     val conv = CONVERTERS[type] ?: throw IllegalArgumentException("No converter for $type")
-    return conv(this)
+    return conv(this) as T
 }
```
lziq, line 90: The change here optimizes this helper method, so we don't need to cast manually to a specific subtype of `Number` after calling this method.

lziq: The compiler recognize the class of the argument `type` and infer the return type of this method as it.

lziq:
One thing to note in this PR is that, the `ExprValue` in evaluation is still generated by `ExprValueFactory`, instead of the new public constructor methods. This is because here we only deprecate `ExprValueFactory` and do not remove it from evaluation.

lang/src/main/kotlin/org/partiql/lang/eval/ExprValueFactory.kt
```diff
@@ -390,6 +438,103 @@ private class BlobExprValue(private val ion: IonSystem, val value: ByteArray) :
     override val ionValue by lazy { toIonValue(ion) }
 }
 
+/**
+ * Core [ExprValue] over an [IonValue].
+ */
+internal class IonExprValue(private val valueFactory: ExprValueFactory, override val ionValue: IonValue) : BaseExprValue() {
```
lziq, line 444: This internal was removed in https://github.com/partiql/partiql-lang-kotlin/pull/881/files, but now I decided to add it back. Since at that moment, the implementation of `ExprValueFactory.newFromIonValue(ionValue: IonValue)` was set to depend on `IonValue.toExprValue()`, and `IonValue.toExprValue()` uses factory pattern to create new `ExprValue`. However, in this PR, we changed `IonValue.toExprValue()` to use the new `ExprValue` construction methods. As a result, we need to add this class back so `ExprValueFactory.newFromIonValue(ionValue: IonValue)` method still use factory pattern to construct new `ExprValue`.  

To summarize, the old API is only being deprecated in this PR, not removed. So it should keep the same.

vgapeyev:
> One thing to note in this PR is that, the `ExprValue` in evaluation is still generated by `ExprValueFactory`, instead of the new public constructor methods. This is because here we only deprecate `ExprValueFactory` and do not remove it from evaluation.

It's not clear from this why are you holding back from transitioning to the new way in evaluation as well?  In principle, it is desirable to do.  If there is a reason that makes the transition impossible or difficult at this point, then it's something to discuss!

vgapeyev (changes requested):
Overall, this PR carries out a valuable simplification as a follow-up to the removal of ExprValue.ionValue.  I appreciate how many places there are where instances of ExprValueFactory and IonSystem are no longer needed.  

However, I think the way of achieving this should be more conservative.  
The central comments for my viewpoint on this are at CHANGELOG.txt and ExprValueUtil. 
I think a separate PR, off main, for doing the alternative would be best, so we could compare the outcomes, but any clarifying discussion can happen in this PR or in person.

lang/src/main/kotlin/org/partiql/lang/util/NumberExtensions.kt
lziq: `out Number` refers to a class which extends `Number`

lziq: Now I changed it to generic type `T`, and the `where T : Number` clause specifies that `T` extends `Number`. Since now we have the reference `T` to the type, we can specify the return type of `Number.coerce()` to be `T`

lziq:
Continue the work in https://github.com/partiql/partiql-lang-kotlin/pull/944

PR #923, opened 2022-12-06
Marks GPML and WINDOW AST nodes as Experimental

RCHowell:
## Description
 - Marks GPML and WINDOW AST nodes as Experimental
 - Updates CHANGELOG

## Other Information
No

~Yes. Usage of the WINDOW and GPML nodes will now produce _warnings_ without explicit opt-in

No

## License Information

### DIFF
--- a/lang/src/main/pig/partiql.ion
+++ b/lang/src/main/pig/partiql.ion
@@ -671,7 +671,7 @@ may then be further optimized by selecting better implementations of each operat
             )
         )
 
-        (include (product window_expression decl::var_decl func_name::symbol args::(* expr 1)))
+        (include experimental::(product window_expression decl::var_decl func_name::symbol args::(* expr 1)))
 
         // Now we include new stuff, including PartiQL's relational algebra.
         (include
--- a/lang/src/main/pig/partiql.ion
+++ b/lang/src/main/pig/partiql.ion
@@ -142,14 +142,16 @@ may then be further optimized by selecting better implementations of each operat
             (bag_op op::bag_op_type quantifier::set_quantifier operands::(* expr 2))
 
             // GPML graph pattern match:  <expr> MATCH <gpml_pattern>
-            (graph_match expr::expr gpml_pattern::gpml_pattern)
+            experimental::(graph_match expr::expr gpml_pattern::gpml_pattern)
 
             // Other expression types
             (path root::expr steps::(* path_step 1))
             (call func_name::symbol args::(* expr 1))
             (call_agg setq::set_quantifier func_name::symbol arg::expr)
-            // TODO: In the feature, when we allow use of aggregation function as window function, we need to consider incorporate in setq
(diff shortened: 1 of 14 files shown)


yliuuuu (commented):
In partiql_logical domain, there is a node called `window_expression` (line 674), we could also mark this as experimental.

johnedquinn:
Curious if experimental can also be applied to the domain level AKA all plans.

yliuuuu (approved):
Window function parts look good to me. 
May be also consult with Arash on DML and Vladimir on GPML before merge?

RCHowell:
Arash suggested that DDL/DML are experimental, but INSERT, UPSERT INTO, and REPLACE INTO are not.

lang/src/main/pig/partiql.ion
```diff
@@ -142,14 +142,16 @@ may then be further optimized by selecting better implementations of each operat
             (bag_op op::bag_op_type quantifier::set_quantifier operands::(* expr 2))
 
             // GPML graph pattern match:  <expr> MATCH <gpml_pattern>
-            (graph_match expr::expr gpml_pattern::gpml_pattern)
+            experimental::(graph_match expr::expr gpml_pattern::gpml_pattern)
```
vgapeyev, line 145: The intent of this annotation is that people developing w.r.t. partiql-kotlin-lang as a library are informed that the (PIG-generated) AST node GraphMatch is experimental. 

However, the informing is done by showing the `org.partiql.pig.runtime.Experimental` annotation, which comes from PIG.  It feels like that should be an annotation coming from PartiQL, such as `org.partiql.annotations.PartiQLExperimental`. That is, we should shield the user from the PIG implementation detail.  
I realize that informing PIG of a specific PartiQL-defined annotation we want to use on the nodes PIG generates would require mutual dependency between PIG and PartiQL, which we can't/don't want to introduce.

I also realize that this might not be that visible to the user, since errors and warnings only show the message specified in the annotation. (Though, I am not certain the class name is opaque in all contexts.)  

So, with the opaqueness, I might be talked into leaving this as is, but there is a more serious problem that will transpire in due time: we really need an annotation per feature (in this PR, one for GPML and another for the Window feature), since these features will graduate on different timelines (see https://kotlinlang.org/docs/opt-in-requirements.html#opt-in-requirements-for-pre-stable-apis).  I don't see how this can be done without communicating these multiple experimental annotations to PIG -- running into the mutual dependency obstacle above. 

So, unless someone has a clever idea how to deal with this, I am inclined to advocate **against** making PIG put `RequiresOptIn` annotations on PIG-generated code.

vgapeyev (commented):
I realize I am late to the PR approvals party, as well as that I was among those okay-ing doing this PR (by saying "yes, GPML stuff is experimental"), but it looks like there is a matter serious enough to bring up -- see the included comment.

PR #925, opened 2022-12-07
Adds the ListImplicit node to the AST

johnedquinn:
## Relevant Issues
- None

## Background
- There is a customer use-case to differentiate between lists that are created using `[ , , , ]` and `( , , , )`.
- In our AST, we currently do not differentiate between the two.

## Description
- Adds a `ListImplicit` AST node (we can change the name -- I haven't thought much about it)
  - Might be a `Row` or `RowValue`, see the conversation below. The naming is subject to change.
- Modifies our tests to handle the `ListImplicit` node.
- Evaluation is the same as `List`

## Other Information
  - Not yet. Need to get feedback from others first.

## License Information

### DIFF
--- a/lang/src/main/kotlin/org/partiql/lang/ast/passes/StatementRedactor.kt
+++ b/lang/src/main/kotlin/org/partiql/lang/ast/passes/StatementRedactor.kt
@@ -174,6 +174,7 @@ private class StatementRedactionVisitor(
         } else when (node) {
             is PartiqlAst.Expr.Lit -> redactLiteral(node)
             is PartiqlAst.Expr.List -> redactSeq(node)
+            is PartiqlAst.Expr.ListImplicit -> redactSeq(node)
             is PartiqlAst.Expr.Sexp -> redactSeq(node)
             is PartiqlAst.Expr.Bag -> redactSeq(node)
             is PartiqlAst.Expr.Struct -> redactStruct(node)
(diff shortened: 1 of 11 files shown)


johnedquinn:
I'm not marking this as Ready for Review until I've thought about this a bit more. The reason is that I'm trying to figure out what the difference is between a PartiQL Array (see the PartiQL Specification) which uses notation `[ , , ]` and a SQL collection (list?) (see the [99 Spec](https://web.cecs.pdx.edu/~len/sql1999.pdf)).

## SQL Spec Research
- See Section X.X for Collection Types
  - It doesn't provide any information except for XX
- See Section 7.2 for:
```ebnf
<row value expression> ::=
    <row value special case>
    | <row value constructor>
    ;

<row value special case> ::=
    <value specification>
    | <value expression>
    ;
```
- See Section 7.1 for the \<row value constructor\>:
```ebnf
<row value constructor> ::=
    <row value constructor element>
    | [ ROW ] <left paren> <row value constructor element list> <right paren>
    | <row subquery>
    ;
```
- See Section 8.4 for the \<in predicate value\>:
```ebnf
<in predicate value> ::=
    <table subquery>
    | <left paren> <in value list> <right paren>
    ;

<in value list> ::=
    <row value expression> { <comma> <row value expression> }...
    ;
```
- Section 6.23 doesn't seem to give me a me any insight into a \<collection value expression\>.
  - It has more information on \<array value expressions\>, but it's not useful.
```ebnf
<collection value expression> ::=
    <value expression primary>
    ;
```
- SqlLite refers to them as [Row Values](https://www.sqlite.org/rowvalue.html)

johnedquinn:
If this requires an extensive amount of research before making these changes, I can create another PR to add a public meta (maybe `IsListParenthesized`) to aid in differentiating the use of `[ , , ]` versus `( , , )` in our AST. This would deliver the same results with added functionality for the customer use-case without interrupting the current APIs.

johnedquinn:
Discussed with @almann, and I'm closing this. As arrays/lists are essentially treated the same, there isn't a need to make our AST concrete. Also, to avoid breaking the public API, I'm going to go ahead with using metas to accomplish this. See #926.

PR #926, opened 2022-12-07
Adds the IsListParenthesizedMeta to aid in differentiating list syntax's

johnedquinn:
## Relevant Issues
- Relevant to #925

## Description
- There is a customer use-case to differentiate between lists that are created using `[ , , , ]` and `( , , , )`.
- In our AST, we currently do not differentiate between the two.
- This adds a meta and adds it to our parsing.

## Other Information

## License Information

### DIFF
--- a/lang/src/main/kotlin/org/partiql/lang/ast/IsListParenthesizedMeta.kt
+++ b/lang/src/main/kotlin/org/partiql/lang/ast/IsListParenthesizedMeta.kt
@@ -1,0 +1,20 @@
+package org.partiql.lang.ast
+
+/**
+ * The [IsListParenthesizedMeta] is used to distinguish the two types of [org.partiql.lang.domains.PartiqlAst.Expr.List]
+ * syntax's:
+ *  1. [[ a, b, c ]]
+ *  2. ( a, b, c )
+ *
+ * The existence of this meta on a [org.partiql.lang.domains.PartiqlAst.Expr.List] node indicates that the list uses
+ * the second syntax above (using parenthesis).
+ */
+public class IsListParenthesizedMeta private constructor() : Meta {
+
+    override val tag = TAG
+
+    companion object {
+        const val TAG = "\$is_list_parenthesized"
+        val instance = IsListParenthesizedMeta()
+    }
+}
--- a/lang/src/main/kotlin/org/partiql/lang/ast/IsListParenthesizedMeta.kt
+++ b/lang/src/main/kotlin/org/partiql/lang/ast/IsListParenthesizedMeta.kt
@@ -9,12 +9,6 @@ package org.partiql.lang.ast
  * The existence of this meta on a [org.partiql.lang.domains.PartiqlAst.Expr.List] node indicates that the list uses
  * the second syntax above (using parenthesis).
  */
-public class IsListParenthesizedMeta private constructor() : Meta {
-
-    override val tag = TAG
-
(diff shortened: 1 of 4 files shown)


lang/src/main/kotlin/org/partiql/lang/ast/IsListParenthesizedMeta.kt
```diff
@@ -0,0 +1,20 @@
+package org.partiql.lang.ast
+
+/**
+ * The [IsListParenthesizedMeta] is used to distinguish the two types of [org.partiql.lang.domains.PartiqlAst.Expr.List]
+ * syntax's:
+ *  1. [[ a, b, c ]]
```
RCHowell, line 6: Why the double `[[` and `]]`?

lang/src/main/kotlin/org/partiql/lang/ast/IsListParenthesizedMeta.kt
```diff
@@ -0,0 +1,20 @@
...
+ */
+public class IsListParenthesizedMeta private constructor() : Meta {
+
+    override val tag = TAG
+
+    companion object {
+        const val TAG = "\$is_list_parenthesized"
+        val instance = IsListParenthesizedMeta()
+    }
+}
```
RCHowell, line 20: I noticed all the metas are this way ... we're writing Kotlin not Java; let's use objects
```
object IsListMeta : Meta {
    override val tag: String = "\$is_list_parenthesized"
}
```

The generated Java is doing exactly what we are doing by hand
```
public final class IsListMeta implements Meta {

   @NotNull
   private static final String tag;

   // Here's your static instance...
   @NotNull
   public static final IsListMeta INSTANCE;

   @NotNull
   public String getTag() {
      return tag;
   }
   
   // Here's your "companion object"...
   static {
      IsListMeta var0 = new IsListMeta();
      INSTANCE = var0;
      tag = "$is_list_parenthesized";
   }

   // Here's your private constructor...
   private IsListMeta() {
   }

  // redacted
}
```

RCHowell (changes requested):

lang/src/main/kotlin/org/partiql/lang/ast/IsListParenthesizedMeta.kt
johnedquinn: It escapes the reserved square brackets -- here's a snippet of the generated HTML:
```
syntax's:</p><ul><li><p>[ a, b, c ]</p></li><li><p>( a, b, c )</p></li></ul><p>
```

lang/src/test/kotlin/org/partiql/lang/syntax/PartiQLParserMetaTests.kt
```diff
@@ -51,6 +51,6 @@ internal class PartiQLParserMetaTests : PartiQLParserTestBase() {
         val inCollection = ast.expr as PartiqlAst.Expr.InCollection
         val list = inCollection.operands[1] as PartiqlAst.Expr.List
 
-        assert(list.metas.containsKey(IsListParenthesizedMeta.TAG))
```
RCHowell: metas.containsKey is kind of funny. You'd think there would be a metas.contains(meta) which sure might be backed by the string lookup. Why define an interface if you aren't going to use it 🤷

RCHowell (approved):

Issue #833, opened 2022-10-07
BindingsToValues causes issue referencing ExprValues (Planner)


johnedquinn:
## Description
- The `PhysicalPlanCompiler` method of `BindingsToValues` has a bug where the `ExprValue` that we are yielding is a reference to the `ExprValue` in the global state registers. Therefore, if we print the result of a SFW, we will see the right results -- but if we attempt to access the resulting bag's elements (which are not the last element) the reference will point to the latest results in the global state registers (and you'll see wrong results).
- It doesn't cause issues when we print the result sequentially, since for each tuple, when we materialize the result, it's in-line with the current state registers. It DOES cause an issue when we attempt to reference elements after materializing other elements. So, in instances like sorting.

## To Reproduce
1. Example queries:
  - Remember, this won't fail in the CLI. Since it prints sequentially, all of the state registers will be appropriately allocated at materialization. If you, however, create an iterator of the resulting bag, and iterate and print in random orders, you'll always see the most up-to-date register as the result of the ExprValue.
  ```sql
  SELECT (SELECT i FROM <<0, 1>>) FROM << {'a': 1}, {'a': 5} >> LET a * 2 AS i
  ```
  ```kt
// NOTE: This is for a different more complex query, but nonetheless, it shows what's wrong.
val iterator = actualExprValueResult.iterator()  // this is a bag

val item1 = iterator.next()
println(item1)
println(item1.ionValue)

val item2 = iterator.next()

println(item1)
println(item1.ionValue)

println(item2)
println(item2.ionValue)

---- RESULT :

{'projListSubQuery': <<1>>}
{projListSubQuery:$partiql_bag::[1]}
{'projListSubQuery': <<5>>}
{projListSubQuery:$partiql_bag::[1]}
{'projListSubQuery': <<5>>}
{projListSubQuery:$partiql_bag::[5]}
  ```

## Expected Behavior
- The resulting tuples of a SFW should be able to be accessed with correct values in whatever order.

## The Solution
- So, the solution was found after much debugging with @yliuuuu and @RCHowell . Essentially, at `compileBindingsToValues`, we need to mimic the clone of an ExprValue by taking the underlying IonValue and using it to create a new ExprValue using the ExprValueFactory. That way, it's not a reference to the global state's registers.
[...]

PR #927, opened 2022-12-09
Removed `ION_WITHOUT_ANNOTATION` test format

lziq:
## Relevant Issues
- [Closes/Related To] Issue #560 
- POC PR: https://github.com/partiql/partiql-lang-kotlin/pull/892 

## Description
- Removed `ION_WITHOUT_ANNOTATION` test format. Changed those test runners which used it to using `ION` test format. Later, we need to figure out what test cases need to get transformed from `ION` to `STRICT`. 
- The only adjustment to the affected test cases was adding `$BAG_ANNOTATION` and `$MISSING_ANNOTATION` to expected Ion results, since these annotations are no longer removed from the actual results by the test runners.
- Some test cases now have to specify `expectedPermissiveModeResult` argument of `runEvaluatorTestCase`, since the expected result in legacy mode (generally, `NULL`) differs from the expected result in permissive mode (generally, `MISSING`). This was a testing bug, previously masked away by the erasure of the annotations in the runners.

## Other Information
No backward incompatible changes.

### DIFF
--- a/lang/src/test/kotlin/org/partiql/lang/eval/EvaluatingCompilerExceptionsTest.kt
+++ b/lang/src/test/kotlin/org/partiql/lang/eval/EvaluatingCompilerExceptionsTest.kt
@@ -143,7 +143,7 @@ class EvaluatingCompilerExceptionsTest : EvaluatorTestBase() {
     fun missingAlias() =
         // Same query as previous test--but DO NOT throw exception this time because of UndefinedVariableBehavior.MISSING
         runEvaluatorTestCase(
-            sqlWithUndefinedVariable, expectedResult = "[null]",
+            sqlWithUndefinedVariable, expectedResult = "$BAG_ANNOTATION::[$MISSING_ANNOTATION::null]",
             target = EvaluatorTestTarget.COMPILER_PIPELINE,
             compileOptionsBuilderBlock = { undefinedVariable(UndefinedVariableBehavior.MISSING) }
         )
@@ -164,7 +164,7 @@ class EvaluatingCompilerExceptionsTest : EvaluatorTestBase() {
     fun missingQuotedAlias() =
         // Same query as previous test--but DO NOT throw exception this time because of UndefinedVariableBehavior.MISSING
         runEvaluatorTestCase(
-            sqlWithUndefinedQuotedVariable, expectedResult = "[null]",
+            sqlWithUndefinedQuotedVariable, expectedResult = "$BAG_ANNOTATION::[$MISSING_ANNOTATION::null]",
             target = EvaluatorTestTarget.COMPILER_PIPELINE,
             compileOptionsBuilderBlock = { undefinedVariable(UndefinedVariableBehavior.MISSING) }
         )
--- a/lang/src/test/kotlin/org/partiql/lang/eval/EvaluatingCompilerFromSourceByTests.kt
+++ b/lang/src/test/kotlin/org/partiql/lang/eval/EvaluatingCompilerFromSourceByTests.kt
@@ -49,26 +49,26 @@ class EvaluatingCompilerFromSourceByTests : EvaluatorTestBase() {
     fun rangeOverListWithBy() = runEvaluatorTestCase(
         "SELECT VALUE addr FROM someList BY addr",
         session,
-        "\$bag::[101,102,103]"
+        "$BAG_ANNOTATION::[101,102,103]"
     )
 
     @Test
     fun rangeOverBagWithBy() = runEvaluatorTestCase(
         "SELECT VALUE addr FROM someBag BY addr",
         session,
-        "\$bag::[111,112,113]"
+        "$BAG_ANNOTATION::[111,112,113]"
     )
     @Test
     fun rangeOverListWithAsAndAt() = runEvaluatorTestCase(
         "SELECT VALUE [i, v, z] FROM someList AS v AT i BY z",
         session,
-        "\$bag::[[0,1,101],[1,2,102],[2,3,103]]"
+        "$BAG_ANNOTATION::[[0,1,101],[1,2,102],[2,3,103]]"
     )
     @Test
     fun rangeOverBagWithAsAndAt() = runEvaluatorTestCase(
         "SELECT VALUE [i, v, z] FROM someBag AS v AT i BY z",
         session,
-        "\$bag::[[\$missing::null,11,111],[\$missing::null,12,112],[\$missing::null,13,113]]"
+        "$BAG_ANNOTATION::[[$MISSING_ANNOTATION::null,11,111],[$MISSING_ANNOTATION::null,12,112],[$MISSING_ANNOTATION::null,13,113]]"
     )
 
     @Test
@@ -77,7 +77,7 @@ class EvaluatingCompilerFromSourceByTests : EvaluatorTestBase() {
         // the result of the inner query is a bag, so i should always be MISSING
         // However, addr should still contain an address since the items of that bag are unchanged
         session,
-        "\$bag::[[\$missing::null,101,1],[\$missing::null,102,2],[\$missing::null,103,3]]"
+        "$BAG_ANNOTATION::[[$MISSING_ANNOTATION::null,101,1],[$MISSING_ANNOTATION::null,102,2],[$MISSING_ANNOTATION::null,103,3]]"
     )
 
     @Test
@@ -85,6 +85,6 @@ class EvaluatingCompilerFromSourceByTests : EvaluatorTestBase() {
         "SELECT VALUE [i, addr, v] FROM (SELECT VALUE v + 1000 FROM someList AS v) AS v AT i BY addr",
         // However, since we + 1000 to v in the inner query, we create a new value that does not have an address.
         session,
-        "\$bag::[[\$missing::null,\$missing::null,1001],[\$missing::null,\$missing::null,1002],[\$missing::null,\$missing::null,1003]]"
+        "$BAG_ANNOTATION::[[$MISSING_ANNOTATION::null,$MISSING_ANNOTATION::null,1001],[$MISSING_ANNOTATION::null,$MISSING_ANNOTATION::null,1002],[$MISSING_ANNOTATION::null,$MISSING_ANNOTATION::null,1003]]"
     )
 }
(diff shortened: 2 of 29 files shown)


lziq:
Note that in this PR, I adopted `ION` instead of `STRICT`, since the change from `ION_WITHOUT_ANNOTATION` to `ION` is definitely an improvement. However, whether we should change `ION` to `STRICT` test format remains questionable.  Probably, for some cases, we still need to use `ION` test format.

lang/src/test/kotlin/org/partiql/lang/eval/EvaluatingCompilerFromSourceByTests.kt
```diff
@@ -49,26 +49,26 @@ class EvaluatingCompilerFromSourceByTests : EvaluatorTestBase() {
     fun rangeOverListWithBy() = runEvaluatorTestCase(
         "SELECT VALUE addr FROM someList BY addr",
         session,
-        """[101, 102, 103]"""
+        "\$bag::[101,102,103]"
```
vgapeyev, line 52: I see some new tests use `$BAG_ANNOTATION` while others use `\$bag`.  Is there a reason for choosing one vs another in the specific cases?

Otherwise, since this is under your control within the PR, consistency would be better.

lang/src/test/kotlin/org/partiql/lang/eval/builtins/functions/CharacterLengthEvaluationTest.kt
```diff
@@ -17,7 +18,7 @@ class CharacterLengthEvaluationTest : EvaluatorTestBase() {
     @ParameterizedTest
     @ArgumentsSource(PassCases::class)
     fun characterLengthPassTests(testCase: ExprFunctionTestCase) =
-        runEvaluatorTestCase(testCase.source, expectedResult = testCase.expectedLegacyModeResult)
+        runEvaluatorTestCase(testCase.source, expectedResult = testCase.expectedLegacyModeResult, expectedPermissiveModeResult = testCase.expectedPermissiveModeResult)
```
vgapeyev, line 21: Why is this change?

lang/src/test/kotlin/org/partiql/lang/eval/EvaluatingCompilerExceptionsTest.kt
```diff
@@ -143,7 +143,7 @@ class EvaluatingCompilerExceptionsTest : EvaluatorTestBase() {
     fun missingAlias() =
         // Same query as previous test--but DO NOT throw exception this time because of UndefinedVariableBehavior.MISSING
         runEvaluatorTestCase(
-            sqlWithUndefinedVariable, expectedResult = "[null]",
+            sqlWithUndefinedVariable, expectedResult = "$BAG_ANNOTATION::[$MISSING_ANNOTATION::null]",
```
vgapeyev, line 146: Is it fair to say that all test adjustments amounted to adding one of these two annotations to expected ion values -- that is, there were absolutely no other reasons tests had to be adjusted as the result of changing `ION_WITHOUT_BAG_AND_MISSING_ANNOTATIONS` to `STRICT`?  
Whatever the answer, this is something to share in the PR description.

vgapeyev:
> I adopted `ION` instead of `STRICT`, since the change from `ION_WITHOUT_ANNOTATION` to `ION` is definitely an improvement. However, whether we should change `ION` to `STRICT` remains questionable. We might still want to use `ION` test format for some test cases.

Focusing on the specific tests in this PR, the ones started as `ION_WITHOUT_BAG_AND_MISSING_ANNOTATIONS`, could you share why you went with `ION` and not all the way to `STRICT`?  Is there a specific obstacle that makes switching to `STRICT` significantly more difficult in these tests? 

Since `STRICT` is the ultimate goal, you could be done and done with these tests, but now you'll have to plow through them again, now also commingled with all the other `ION` tests!

vgapeyev (commented):
My primary comment is the "why not `STRICT`" one.  Depending on how that turns out, the others might be mute, but they are examples of a general request about describing a PR.

lang/src/test/kotlin/org/partiql/lang/eval/builtins/functions/CharacterLengthEvaluationTest.kt
lziq: `char_legnth(missing)` outputs `null` in legacy typing mode, and outputs `missing` in permissive mode. Originally, since we were using `ION_WITHOUT_ANNOTATION`, `null` in ION can represent both `missing` & `null`. But now we need to change.

lang/src/test/kotlin/org/partiql/lang/eval/EvaluatingCompilerExceptionsTest.kt
vgapeyev: I can *imagine* reasons for `ION` to remain in some cases, but the question [in the review preamble, to which you seem to be responding here] is about the specific tests touched by this PR.  Can you *point to* a reason why these tests should be `ION` and not `STRICT`?

lziq: The bug of `ExprValue` iterator has not been addressed yet, https://github.com/partiql/partiql-lang-kotlin/issues/833

vgapeyev: I don't see this through yet.  Is there a specific test here for which #833 bug plays out?  Does it play out in *all* the tests if they use `STRICT` rather than `ION`?

lziq: This one, https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/src/test/kotlin/org/partiql/lang/eval/EvaluatingCompilerFromSourceByTests.kt#L75

lziq: Also, another question is that if we are querying against an Ion file or an `IonValue`, is the output supposed to be `IonValue` or `ExprValue`?

lziq: I think it should be `IonValue` in this case. Namely, if the binding data is in format of ION, the output should be in format of ION, since we are querying against ION.

lang/src/test/kotlin/org/partiql/lang/eval/builtins/functions/CharacterLengthEvaluationTest.kt
lziq: The `expectedPermissiveModeResult` is the same as `expectedResult`, if not specified.

lang/src/test/kotlin/org/partiql/lang/eval/EvaluatingCompilerExceptionsTest.kt
vgapeyev: If so, the `rangeOverListNested` test that you are pointing at can be at `ION`, while the majority still switches to `STRICT` -- the resting framework has a parameter on the `runEvaluatorTestCase` that can do that, from what I see.  
The reason why this test gets special treatment is worth documenting, with reference to #833. 

> Also, another question is that if we are querying against an Ion file or an IonValue, ....

I don't see how this reasoning is valid: in `PipelineEvaluatorTestAdapter`, even in the `ION` format, the actual value still surfaces as an ExprValue and then gets converted to Ion, just for the purpose of comparing to the expected (specified as Ion).  Isn't the whole purpose of `STRICT` format you introduced to avoid this round-about?

lziq: I mean one feature of PartiQL is to query against Ion files. In this case, users expect to get the result back in format of Ion. So if we are writing integration test to test this feature, I think the input & output should be Ion, instead of `ExprValue`, since we might not property transform `ExprValue` to `IonValue`.

lziq: If you are querying against a JSON file, you hope the query engine returns a JSON value. And if you are querying an ION file, you hope it returns an Ion value to you, right?

vgapeyev: Right now we are talking about the tests touched by this PR.  Do you say your arguments still apply?

lziq: I agree that we should use `STRICT` for all the test cases eventually. But my approach in this PR is that, we change `ION_WITHOUT_ANNOTATION` to `ION` first, and then let's see if we should switch to `STRICT` next. And probably, for some test cases, we are not able to do so. So we should figure this out in the next PRs. But in this PR, let's just remove `ION_WITHOUT_ANNOTATION`, and switch to `ION`, since this change for any test case always makes sense.

lziq: If you think we should look at each test case and try to switch it to `STRICT` just in this PR, I can do that.

lang/src/test/kotlin/org/partiql/lang/eval/builtins/functions/CharacterLengthEvaluationTest.kt
vgapeyev: I see now: this goes in conjunction of adding a permissive-result to the tests cases where it differs from the legacy-result, as on Line 33 below.  

This is a kind of thing worth mentioning in the PR description as a finding.

lang/src/test/kotlin/org/partiql/lang/eval/EvaluatingCompilerExceptionsTest.kt
vgapeyev: Yes.   

Here is my initial suggestion that started this conversation

> Since STRICT is the ultimate goal, you could be done and done with these tests, but now you'll have to plow through them again, now also commingled with all the other ION tests!

This pointed out that going straight to `STRICT` would've been a more efficient way of proceeding, since you (and I) would have looked at most of these tests once rather than twice.
If you really want to look at these tests again when eliminating `ION`, so be it.  

I still want to see an improved PR description, since that goes into the git commit.  Other comments in this PR have some of the suggestions on what to document.

vgapeyev (changes requested):

vgapeyev (approved):

PR #929, opened 2022-12-13
numeric built in functions

yliuuuu:
## Relevant Issues
- N/A

## Description
- This PR implemented the missing numeric function as stated in https://github.com/partiql/partiql-docs/blob/main/docs/builtins-partiql.adoc. 
- A big part of this PR focus on implementing BigDecimal operations that are not covered in Java's native BigDecimal class. 
  - An extra step was taken to normalize the result of BigDecimal operations: 
  - Consider the math context has precision p, then the result will be unscaleValue * 10 ^ (-scale) and unscaleValue has exactly p digits. 

- Currently in our code base, p is set to 38. Test cases are created to ensure the calculation is correct under p = 38. 

## Other Information
  - Yes.
  -  No.
  - No.
## License Information


lang/src/main/kotlin/org/partiql/lang/util/NumberExtensions.kt
```diff
@@ -256,3 +257,302 @@ val Number.isPosInf get() = when (this) {
...
+        val numerator = this.add(rMinusOne.multiply(xToR, operationMC), operationMC)
+
+        // (index*(x^(index-1))
+        val denominator = r.multiply(xToRMinusOne, operationMC)
+
+        // x = (n + (index-1)*(x^index)) / (index*(x^(index-1)))
+        resPrev = res
+
+        res = numerator.divide(denominator, operationMC)
+    } while (res.round(mathContext).subtract(resPrev.round(mathContext)).abs() > tolerance)
```
RCHowell, line 318: Do you know if this implementation guarantees convergence?

lang/src/main/kotlin/org/partiql/lang/util/NumberExtensions.kt
```diff
@@ -256,3 +257,302 @@ val Number.isPosInf get() = when (this) {
...
+    return res
+}
+
+/**
+ * Computes the square root of a given BigDecimal.
+ * See https://dl.acm.org/doi/pdf/10.1145/214408.214413
+ */
+fun BigDecimal.squareRoot(mathContext: MathContext = MATH_CONTEXT): BigDecimal {
+    if (this.signum() < 0) {
+        throw ArithmeticException("Cannot take root of a negative number")
```
RCHowell, line 328: What about when `this = 0`?

lang/src/main/kotlin/org/partiql/lang/util/NumberExtensions.kt
```diff
@@ -256,3 +257,302 @@ val Number.isPosInf get() = when (this) {
...
+    var approx = guess
+
+    val zeroPointFive = BigDecimal.ONE.divide(BigDecimal.valueOf(2))
+    do {
+        // plus 2 for precision buffering
+        val operatingPrecision = kotlin.math.max(
+            kotlin.math.max(guessPrecision, targetPrecision + 2),
+            normalizedPrecision
+        )
+        val tempMC = MathContext(operatingPrecision, RoundingMode.HALF_EVEN)
```
RCHowell, line 356: Do we have to create a MathContext every time? This creation is likely to slow things down. I see how the precision is changing though.

lang/src/main/kotlin/org/partiql/lang/util/NumberExtensions.kt
```diff
@@ -256,3 +257,302 @@ val Number.isPosInf get() = when (this) {
...
+}
+
+/**
+ * Computes the exponential value of a BigDecimal.
+ */
+private fun BigDecimal.expHelper(mathContext: MathContext): BigDecimal {
+    // For faster convergence, we break exponent into integer and fraction parts.
+    // e^x = e^(i+f) = (e^(1+f/i)) ^i
+    // 1 + f/i < 2
+    var intPart = this.divideToIntegralValue(BigDecimal.ONE).setScale(0)
```
RCHowell, line 397: Would this be simpler?
```suggestion
    var intPart = this.toBigInteger()
```

lang/src/main/kotlin/org/partiql/lang/util/NumberExtensions.kt
```diff
@@ -256,3 +257,302 @@ val Number.isPosInf get() = when (this) {
...
+
+    if (intPart.signum() == 0) {
+        return this.expTaylor(mathContext)
+    }
+
```
RCHowell, line 402: Did you look at https://github.com/eobermuhlner/big-math?

RCHowell: I've been looking at apache commons math4, but wow that's a lot!

lang/src/main/kotlin/org/partiql/lang/util/NumberExtensions.kt
```diff
@@ -256,3 +257,302 @@ val Number.isPosInf get() = when (this) {
...
+        .add(
+            fractionPart.divide(
+                intPart, mathContext
+            )
+        )
+
+    // e^(1+f/i)
+    val etoExpInner = expInner.expTaylor(mathContext)
+    // The build in power function can only handle int type, which max out at 999999999
+    val maxInt = BigDecimal.valueOf(999999999L)
```
RCHowell, line 415: Where is the max defined? a comment, a constant, etc.?

docs/documentation/Functions.md
```diff
@@ -982,6 +982,167 @@ FLOOR(`-inf`) = `-inf` -- Float
...
+
+pow(x, 0.0) is 1.0;
+
+pow(x, 1.0) == x;
+
+pow(x, NaN) is NaN;
+
+pow(NaN, x) is NaN for x != 0.0;
+
+pow(x, Inf) is NaN for abs(x) == 1.0
```
yliuuuu, line 1106: resurfacing the comment "what about `0^+inf`"

lang/src/main/kotlin/org/partiql/lang/eval/builtins/MathFunctions.kt
```diff
@@ -70,25 +103,136 @@ private fun floor(n: Number): Number = when (n) {
...
+}
+
+private fun pow(n: Number, p: Number): Number {
+    val (first, second) = if (n.isPosInf || n.isNegInf || n.isNaN) {
+        n to p.toDouble()
+    } else if (p.isPosInf || p.isNegInf || p.isNaN) {
+        n.toDouble() to p.toDouble()
+    } else {
+        coerceNumbers(n, p)
+    }
```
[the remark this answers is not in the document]
yliuuuu: This is to prevent the `coerceNumbers` trying to cast `+inf` etc to a bigDecimal, which leads to an error. Will add a comment later.

lang/src/main/kotlin/org/partiql/lang/util/NumberExtensions.kt
```diff
@@ -256,3 +257,302 @@ val Number.isPosInf get() = when (this) {
...
+ * where x needs to be positive.
+ */
+private fun BigDecimal.intRoot(
+    index: Int,
+    mathContext: MathContext
+): BigDecimal {
+    val operationMC = MathContext(
+        if (mathContext.precision + 2 < 0) Int.MAX_VALUE else mathContext.precision + 2,
+        mathContext.roundingMode
+    )
```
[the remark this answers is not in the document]
yliuuuu: This is generally for precision buffering, in this specific case, the result of the operation will be have a precision of `targetPrecision + 2`, which means we have two more significant digit, for the intermediate result. 
Consider: 
`1.1000226` + `1.1000225`  = `2.2000451`
rounding `2.2000451` => we get `2.20005`
If we round too early we have: 
1.10002 + 1.10002 = `2.20004`

Do we have the same MathContext for the entire engine? -> This is fixed at the moment, but we may consider making this configurable in the future in my opinion.

lang/src/main/kotlin/org/partiql/lang/util/NumberExtensions.kt
```diff
@@ -256,3 +257,302 @@ val Number.isPosInf get() = when (this) {
...
+
+    val tolerance: BigDecimal = BigDecimal.valueOf(5L).movePointLeft(mathContext.precision)
```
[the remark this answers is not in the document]
yliuuuu: Depending on if we want to make the mathContext configurable. Note that mathContext is a parameter required for the function call.

lang/src/main/kotlin/org/partiql/lang/util/NumberExtensions.kt
yliuuuu: currently trying to prove this...

lang/src/main/kotlin/org/partiql/lang/util/NumberExtensions.kt
yliuuuu: In this specific case, I believe the benefit outweighed the cost. Because bigDecimal operation's run time increase non-linearly w.r.t precision.

lang/src/main/kotlin/org/partiql/lang/util/NumberExtensions.kt
yliuuuu: Looking into this, the initial thought was, I was not confident on converting big decimal to big Integer then to big decimal will be faster.

lang/src/main/kotlin/org/partiql/lang/util/NumberExtensions.kt
yliuuuu: This is a java library built in constraint.  see https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#pow-int-

lang/src/main/kotlin/org/partiql/lang/util/NumberExtensions.kt
yliuuuu: Yes this in fact does. See https://math.stackexchange.com/questions/2079222/prob-18-chap-3-in-baby-rudin-behaviour-of-x-n1-fracp-1p-x-n-frac

lang/src/main/kotlin/org/partiql/lang/util/NumberExtensions.kt
yliuuuu: It can be handled via the code logic, but I will add a special case here.

RCHowell (dismissed):

RCHowell (approved):

Issue #908, opened 2022-11-29
Values with the same field name should be legal in one Ion struct 


lziq:
## Description
- We should allow access a value in a struct where there is another value with the same field name. However, we are not allowing so in ion structs, https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/src/main/kotlin/org/partiql/lang/eval/IonStructBindings.kt#L62. 

## To Reproduce
1. PartiQL structs:
```
SELECT a FROM [{'a':1, 'a':2}]
   | 
===' 
<<
  {
    'a': 1
  }
>>
--- 
OK!
```
2. Ion structs
```
SELECT a FROM `[{'a':1, 'a':2}]`
   | 
===' 
<<
org.partiql.lang.eval.EvaluationException: Multiple matches were found for the specified identifier
	Evaluator Error: at line 1, column 8: Binding name was 'a'

	at org.partiql.lang.eval.ExceptionsKt.err(Exceptions.kt:58)
	at org.partiql.lang.util.BindingHelpersKt.errAmbiguousBinding(BindingHelpers.kt:23)
	at org.partiql.lang.eval.IonStructBindings.handleMatches(IonStructBindings.kt:62)
	at org.partiql.lang.eval.IonStructBindings.caseInsensitiveLookup(IonStructBindings.kt:55)
	at org.partiql.lang.eval.IonStructBindings.get(IonStructBindings.kt:68)
```

## Expected Behavior
- Ion structs should behave the same as PartiQL structs

lziq:
Quotes from spec: 
```
When the tuple t has multiple attributes a, the tuple
path navigation t.a will return the first instance of a. Note that for tuples whose order is defined by schema, this is
well-defined, for unordered tuples, it is implementation defined which attribute is returned in permissive mode or an
error in type checking mode, which is described in Section 4.1.
```

lziq:
Since Ion structs are unordered, https://amzn.github.io/ion-docs/docs/spec.html, I think the current behavior is correct in strict type checking mode.

PR #940, opened 2022-12-16
Ion struct bug fix

lziq:
## Relevant Issues
- [Closes/Related To] Issue #908


lziq:
The correct behavior is correct

PR #942, opened 2022-12-16
Fixes build failures for the assertThrows invalid reference

johnedquinn:
## Relevant Issues
- None

## Description
- We've seen build failures for an invalid reference to assertThrows. This changes the method to use the Junit5 method, which can consistently be found.
- Marking this as a draft -- to be further discussed.

## Other Information
  - Updating a test file

## License Information

### DIFF
--- a/lang/src/test/kotlin/org/partiql/lang/syntax/PartiQLParserPrecedenceTest.kt
+++ b/lang/src/test/kotlin/org/partiql/lang/syntax/PartiQLParserPrecedenceTest.kt
@@ -1066,20 +1066,20 @@ class PartiQLParserPrecedenceTest : PartiQLParserTestBase() {
     @Test
     @Parameters
     @TestCaseName("{0}")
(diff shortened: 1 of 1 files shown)


vgapeyev:
Putting aside that the most of the delta in this PR is actually about stripping out (from `failureCases`) the portion of test data that has not been in use for a while, -- 
is there more context about the failures that are being resolved?  

My concern: so far, this file has been purely JUnit4+JUnitParams, while this PR introduces a dependency on JUnit5 for the first time.  At the same time, while JUnit4 `assertThrows` is being removed, another org.junit.Assert method, `assertEquals` remains the workhorse of this suite.  How is the status of these two methods so different?  Why is there any certainty this change fixes the problem (whatever it is) rather than merely being a happy coincidence? 

It is quite possible we'd end up accepting the coincidence as a win and move on, but it's worthwhile to see if we can at least document the problem before, possibly, just masking it away.

johnedquinn:
PR is no longer needed.

PR #946, opened 2022-12-22
Replaces CLI to use Pico CLI

johnedquinn:
## Relevant Issues
- None

## Description
- Replaces JOpt with Pico CLI because it's significantly cleaner and easier to read/write.
- Cleans up the main file substantially.
- Updates docs.
- Re-packages file to be contained within `org.partiql.cli`
- Adjusts the `shell.sh` script to pass in all arguments to the CLI
- Adds ability to pipe queries into the `partiql` command

## New Examples
As a preface, I updated my path to be:
```shell
PATH="$HOME/Development/partiql-lang-kotlin/partiql-app/partiql-cli/build/install/partiql-cli/bin:$PATH"
```
```shell
### Echo
echo "SELECT VALUE 'Hello World' FROM <<0>>" | partiql
# << 'Hello World' >>

### Cat
cat ~/Desktop/query.partiql | partiql
# [0, 1]

### File parameter
partiql ~/Desktop/query.partiql
# [0, 1]

### Shell
partiql
# Welcome to the PartiQL Shell!
# ...
```

## To Run
- Pull this branch.
- Run `./partiql-app/partiql-cli/shell.sh --help`
- See the available help message. I've also included a screenshot:
Screen Shot 2022-12-28 at 11 44 57 AM

## Other Information
  - Yes, but not for the library. CLI options have been modified.
- Any new external dependencies? **YES**
  - Added pico-cli. Removed jopt.

## License Information


partiql-app/partiql-cli/src/main/kotlin/org/partiql/pico/CommandPartiQL.kt
```diff
@@ -12,8 +12,22 @@
  *  language governing permissions and limitations under the License.
  */
 
-package org.partiql.cli
+package org.partiql.pico
```
johnedquinn, line 15: This is deceiving -- I deleted PartiQLCommand because we weren't using it. Then, I created a CommandPartiQL for PicoCLI.

docs/tutorials/Command Line Tutorial.md
```diff
@@ -173,15 +142,15 @@ The variables `animals` and `types` can both be bound to the execution environme
 To bind the environment file to the execution environment, start the REPL with the following command:
 
 ```shell
-$ ./gradlew :cli:run -q --console=plain --args='-e config.env'
+$ ./partiql-app/partiql-cli/shell.sh shell -e config.env
 ```
 
 **Note**: Shell expansions such as `~` do not work within the value of the `args` argument.
```
RCHowell, line 148:
```suggestion
```
we don't need this anymore because we aren't using the Gradle args options

partiql-app/partiql-cli/shell.sh
```diff
@@ -17,4 +17,4 @@ cli_path=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 
 cd "$cli_path"
 ../../gradlew :partiql-app:partiql-cli:install
-../partiql-cli/build/install/partiql-cli/bin/partiql
+../partiql-cli/build/install/partiql-cli/bin/partiql "$@"
```
RCHowell, line 20: I think it would make more sense to just have `shell` as the default

partiql-app/partiql-cli/src/main/kotlin/org/partiql/Main.kt
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Runs the PartiQL CLI.
+ */
+fun main(args: Array<String>) {
+    val ion = IonSystemBuilder.standard().build()
+    val valueFactory = ExprValueFactory.standard(ion)
+    val command = CommandLine(CommandPartiQL())
+        .addSubcommand(CommandQuery(valueFactory))
+        .addSubcommand(CommandShell(valueFactory))
```
RCHowell, line 36: Rather than subcommands, what do you think about running the shell if no query is provided? You can still provide the environment, but if no input file or query text is provided, then you get the REPL. This is typically how REPLs work such as python, ruby, lua, etc. I suppose ours is a _bit_ odd because the query is provided as a text parameter rather than a file path .. something we remedy by using stdin 

So the usage would be
```shell
# executes `select * from foo`
echo "select * from foo" | partiql

# executes the query within `foo.partiql`
partiql ./foo.partiql

# executes the repl
partiql
```

This is consistent with python, ruby, lua, and I presume many other interpreters with `(print (eval (read)))`..

partiql-app/partiql-cli/src/main/kotlin/org/partiql/Main.kt
```diff
@@ -0,0 +1,39 @@
...
+    val exitCode = command.execute(*args)
```
RCHowell, line 37: Rather than subcommands, let's go with the unofficial convention used by python, ruby, node, lua, etc.

```shell
# execute the query in `foo.partiql` — useful for tooling!
partiql foo.partiql

# execute the query `select * from foo` — today's usage
echo "select * from foo" | partiql

# execute the repl — no parameters, no input on stdin
partiql
```

partiql-app/partiql-cli/src/main/kotlin/org/partiql/functions/ReadFile.kt
```diff
@@ -12,7 +12,7 @@
  *  language governing permissions and limitations under the License.
  */
 
-package org.partiql.cli.functions
```
RCHowell, line 15: Everything in the `partiql-cli` gradle project should now be in the `org.partiql.cli` jvm package

partiql-app/partiql-cli/shell.sh
RCHowell, line 20: Depending on how well argument expansion/passthrough to the gradle generated script works here (I definitely did not thoroughly test this at the time) we could rename this script to `shell.sh` as it is now more than just a way to start the REPL

partiql-app/partiql-cli/src/main/kotlin/org/partiql/pico/CommandQuery.kt
```diff
@@ -0,0 +1,84 @@
...
+        val pipeline = options.pipeline
+        val input = when (inputFile) {
+            null -> EmptyInputStream()
+            else -> FileInputStream(inputFile!!)
+        }
+        val output = when (outputFile) {
+            null -> UnclosableOutputStream(System.out)
+            else -> FileOutputStream(outputFile!!)
+        }
+
```
RCHowell, line 65: Make this the root command and check if the input stream (what is now called `query`) is a file. If its not specified, check stdin. If nothing is there, then fire up the shell.

RCHowell (changes requested):

partiql-app/partiql-cli/src/main/kotlin/org/partiql/pico/CommandQuery.kt
```diff
@@ -0,0 +1,84 @@
...
+    @CommandLine.Option(names = ["-o", "--out"], description = ["The path to the output file"], paramLabel = "FILE")
+    var outputFile: File? = null
+
+    @CommandLine.Option(names = ["-u", "--out-format"], description = ["The output file format"], paramLabel = "FORMAT")
+    var outputFormat: OutputFormat = OutputFormat.PARTIQL
+
+    @CommandLine.Option(names = ["-w", "--wrap-ion"], description = ["Indicates that the input Ion file contains a sequence of Ion values rather than a single Ion collection"])
+    var wrapIon: Boolean = false
+
+    @CommandLine.Parameters(description = ["The PartiQL query to run"], paramLabel = "QUERY")
```
RCHowell, line 52: You'll want to add an `index` to this parameter. It doesn't really matter when there's only one argument, but it's good practice

partiql-app/partiql-cli/src/main/kotlin/org/partiql/Main.kt
johnedquinn: Ack. I like this idea.

I've updated the CLI to receive an optional file (arity "0..1") as a parameter. Performed the changes so we can read queries from standard input. Now, if the non-interactive configs haven't been modified, a file isn't passed as a parameter, and nothing is in standard input, we start the shell.

partiql-app/partiql-cli/shell.sh
johnedquinn: > ... rename this script to `shell.sh`

It's already `shell.sh`. Do you mean something like `partiql`? I'd be interested in creating a root-level directory called `bin` or `partiql-bin`.

Then, from the project root, we can run `./bin/partiql`. And, if users want, they can add the bin to their PATH, so they can just run `partiql`. On top of that, I haven't looked too much into it -- but I'd like the current script to not have to re-install and re-assemble on every invocation. I know we can set the Gradle task's inputs/outputs to avoid frequent assembly, but I haven't made any attempts yet.

partiql-app/partiql-cli/src/main/kotlin/org/partiql/cli/pico/PartiQLCommand.kt
```diff
@@ -0,0 +1,139 @@
...
+     */
+    override fun run() {
+        when {
+            executionOptions?.query != null -> runCli(executionOptions!!, executionOptions!!.query!!.inputStream())
+            executionOptions != null && System.`in`.available() > 0 -> runCli(executionOptions!!, System.`in`)
+            System.`in`.available() > 0 -> runCli(ExecutionOptions(), System.`in`)
+            shellOptions != null -> runShell(shellOptions!!)
+            else -> runShell()
+        }
+    }
```
RCHowell, line 67: nit to cut down on the assertions and density here

```suggestion
    override fun run() {
        val command = executionOptions ?: ExecutionOptions()
        val shell = shellOptions ?: ShellOptions()
        val stdin = System.`in`
        when {
            command.query != null -> runCli(command, command.query!!.inputStream())
            stdin.available() > 0 -> runCli(command, stdin)
            else -> runShell(shell)
        }
    }
```

partiql-app/partiql-cli/src/main/kotlin/org/partiql/cli/pico/PipelineOptions.kt
```diff
@@ -22,28 +23,27 @@ import org.partiql.lang.eval.ProjectionIterationBehavior
...
+    @CommandLine.Option(names = ["-T", "--typed-behavior"], description = ["Indicates how CAST should behave: [\${COMPLETION-CANDIDATES}]"], paramLabel = "OPT")
     var typedOpBehavior: TypedOpBehavior = TypedOpBehavior.HONOR_PARAMETERS
 
-    @CommandLine.Option(names = ["-r", "--projection-iter-behavior"], description = ["Controls the behavior of ExprValue.iterator in the projection result"], paramLabel = "BEHAVIOR")
+    @CommandLine.Option(names = ["-I", "--iter-behavior"], description = ["Controls the behavior of ExprValue.iterator in the projection result: [\${COMPLETION-CANDIDATES}]"], paramLabel = "OPT")
     var projIterBehavior: ProjectionIterationBehavior = ProjectionIterationBehavior.FILTER_MISSING
 
-    @CommandLine.Option(names = ["-v", "--undefined-variable-behavior"], description = ["Defines the behavior when a non-existent variable is referenced"], paramLabel = "BEHAVIOR")
+    @CommandLine.Option(names = ["-u", "--undef-behavior"], description = ["Defines the behavior when a non-existent variable is referenced: [\${COMPLETION-CANDIDATES}]"], paramLabel = "OPT")
```
RCHowell, line 46: Is there any particular reason to change the names/aliases from before? I like the consistency of using the single character as the first name alias. I would remove the `-I` altogether though as that's a loaded term/alias. I think you should keep `--undefined-variable-behavior` even though it's verbose because we have a good alias for it now. I'm inclined to remove the aliases from `--typing-mode` and `--typed-op-behavior`

RCHowell (changes requested):

partiql-app/partiql-cli/shell.sh
RCHowell: I definitely meant `partiql.sh` haha and just made a mistake

RCHowell: Robert from two days ago had too much egg nog

docs/tutorials/Command Line Tutorial.md
```diff
@@ -20,18 +20,38 @@ project root).
...
-To view all commands available, run the CLI with the `--help` option.
+To view all available options, run the CLI with the `--help` option.
 
-## Shell
+## Non-Interactive (Single Query Execution)
+
+To execute a single query, run:
+
+```shell
+./partiql-app/partiql-cli/shell.sh -q query.partiql
```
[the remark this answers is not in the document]
johnedquinn: I believe this was already updated.

partiql-app/partiql-cli/shell.sh
johnedquinn: Renamed!

partiql-app/partiql-cli/src/main/kotlin/org/partiql/cli/pico/PipelineOptions.kt
johnedquinn: ACK. I renamed to use the longer terms and removed ambiguous aliases.

RCHowell (approved):

PR #954, opened 2023-01-03
Adds support for shebang and fixes pipeline options

### LINKED ISSUE #952: Support PartiQL Shebang
## Relevant Issue/Bug
- Relevant to #946 

## Requested Solution/Feature
- Since we now support passing in queries via STDIN (see PR linked above), it would be nice to also support shebangs.
- Example file (`example.partiql`):
```
#!/usr/bin/env partiql

SELECT * FROM <<0, 1>>
```
- Then, we just need to run `./example.partiql`
- We'd just need to check for the existence of the shebang (as operating systems don't strip the first line) so we can remove it from the query definition.

## Describe Alternatives
- I've already implemented this locally (by removing the first line if it contains a shebang), but I'm seeing if there's a better recommended way. But nevertheless, it works. Also works with shebangs like `#!/usr/bin/env partiql --out-format ION_BINARY`

## Additional Context
- None

## DoD (Definition of Done)
- Shebangs working.

johnedquinn:
## Relevant Issues
- Closes #952 

## Description
- Adds support for shebang
- Fixes ability to use passed options for the pipeline.
- Adjusts how we determine whether STDIN is present by using `System.console()` -- this allows partiql to pipe its output to another partiql command.
- With the addition of standard input processing (from #946), we can seemingly add support for shebangs. Essentially, if a user specifies a shebang with the path to the `partiql` executable, this PR allows them to run PartiQL executable files.
- This can help some of our customers write queries in a fast-paced manner

## Usage
- Users of this feature will need to have the gradle-produced executable (`partiql`) on their PATH.
- Here's an example:
Screen Shot 2023-01-03 at 1 43 26 PM

- And here's an example of piping its output to itself:
Screen Shot 2023-01-05 at 5 34 08 PM

- And another example with pretty printing:
Screen Shot 2023-01-03 at 1 45 15 PM

- And here's an example with the experimental pipeline:
Screen Shot 2023-01-03 at 2 00 03 PM

## Open Questions
- Windows doesn't support shebangs -- I'm curious what we should do in this scenario.

## Other Information

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Adds ability to pipe queries to the CLI
+- Adds ability to run PartiQL files as executables by adding support for shebangs
 - Added numeric functions `abs`, `sqrt`, `exp`, `ln`, `pow`.
 
 ### Changed
--- a/docs/tutorials/Command Line Tutorial.md
+++ b/docs/tutorials/Command Line Tutorial.md
@@ -26,6 +26,8 @@ To view all available options, run the CLI with the `--help` option.
 
 ## Non-Interactive (Single Query Execution)
 
+### Using the Script
+
 To execute a single query, run:
 
 ```shell
@@ -33,6 +35,8 @@ To execute a single query, run:
 ```
 
 where `query.partiql` contains the PartiQL query to execute.
+
+### Using the `partiql` Command Directly
 
 Alternatively, you may pipe input into the native command:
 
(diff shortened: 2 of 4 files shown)


RCHowell:
Could you please document that you need either partiql.sh on the path, or you need to specify the gradle-produced executable script

johnedquinn:
> Could you please document that you need either partiql.sh on the path, or you need to specify the gradle-produced executable script

ACK. Recent revision rebases and updates PR description and documentation. Unfortunately, modifying the script to handle the shebang needs more research. See created feature request #962 .

RCHowell (approved):

PR #961, opened 2023-01-05
usage of experimental features in java

yliuuuu:
## Relevant Issues
- N/A

## Description
- An example usage of the experimental features annotation in java. 

## Other Information
  - No, example. 
  - No. 
  - No. 

## License Information

### DIFF
--- a/examples/src/main/java/org/partiql/examples/ExperimentalFeatureJavaExample.java
+++ b/examples/src/main/java/org/partiql/examples/ExperimentalFeatureJavaExample.java
@@ -1,0 +1,89 @@
+package org.partiql.examples;
+
+import com.amazon.ion.IonSystem;
+import com.amazon.ion.system.IonSystemBuilder;
+import kotlin.OptIn;
+import org.jetbrains.annotations.NotNull;
+import org.partiql.annotations.PartiQLExperimental;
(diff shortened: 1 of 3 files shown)


vgapeyev:
I'd question whether it is useful or even possible to have a meaningful and maintainable example of @OptIn here, in the examples sub-project?  What happens when the feature chosen for the example (PartiQLCompilerPipeline here?) graduates from the experimentation?  
This equally applies to the counterpart Kotlin example. 

More broadly, if we have already thought through the ceremony of using experimental annotations in this project and arrived at guidelines, please point me to them -- since this PR ought to be evaluated w.r.t. that.  If not, this looks like a thing worth figuring out how to do it right.

yliuuuu:
Upon examination, it seems like Java does not recognize the `Optin` annotation specified in Kotlin. Java users will be able to access the experimental APIs freely, and not be warned at all.

vgapeyev:
Ha-ha-ha!  How did we not see this right away?! :-)

So, sadly, these annotations will not be that useful for "enforced" communication, as our customers consume from Java, I presume.  Still, my inclination is to proceed as planned for now.  Even though they would be not much more than human-readable documentation, that's still useful.  Opinions?

yliuuuu:
Closing this.

Issue #966, opened 2023-01-10
PartiQL compiler pipeline doble evaluation


hpolloni:
## Description
The new PartiQL pipeline (planner, compiler, etc.) evaluates the plan twice: once when calling `statement.eval()` and another time when materializing the ExprValue. This is unexpected as execution of the plan should only happen once (during `statement.eval()`). 

## To Reproduce
1. Pull commit https://github.com/hpolloni/partiql-lang-kotlin/commit/030b355ecfc61b681aa188d5e0b8fbfde9d6ab26
2. Run the test.

## Expected Behavior
- Evaluation should happen once during `statement.eval()`

RCHowell:
Hey Herman, this is an interesting issue and we've had some previous troubles with materialization to IonValues. Just taking a quick look, and trying to understand how this is happening.

The values are never actually materialized when `eval()` is called as you would expect. Instead, you will get a BagExprValue which holds a yet-to-be-instantiated sequence — where the sequence is what `eval()` creates.

The bug appears in compiling bindings to values, [the sequence is peeked to determine if a BagExprValue or ListExprValue should be created](https://github.com/partiql/partiql-lang-kotlin/blob/main/lang/src/main/kotlin/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt#L289). This instantiates the sequence and is what's causing `exhausted` to be set to `true`.

This change prevents the sequence from being instantiated prematurely.
```kotlin
        return thunkFactory.thunkEnv(expr.metas) { env ->
            // var relationType: RelationType? = null
            // we create a snapshot for currentRegister to use during the evaluation
            // this is to avoid issue when iterator planner result
            val currentRegister = env.registers.clone()
            val elements = sequence {
                env.load(currentRegister)
                val relItr = bexprThunk(env)
                // relationType = relItr.relType
                while (relItr.nextRow()) {
                    yield(mapThunk(env))
                }
            }

            // Trick the compiler here to always initialize `relationType`
            ExprValue.newBag(elements)
            // when (elements.firstOrNull()) {
            //     null -> createOutputSequence(relationType, emptySequence())
            //     else -> createOutputSequence(relationType, elements)
            // }
        }
```

John originally used a meta to indicate whether or not to instantiate a bag or list which would prevent having to peek. We will make this change asap

hpolloni:
ah! thanks. That makes sense to me as well.

Issue #951, opened 2023-01-03
Indeterministic error during build process


yliuuuu:
## Description
- During local build, we sometimes receive an indeterministic error message with the following format: 
```
[Test Case]
Caused by:
 org.partiql.lang.eval.EvaluationException: Variable not in GROUP BY or aggregation function: [variable name]
```
The test case showing may change from time to time but the error message is largely consistent. The problem may get fixed by running build command again or clean Gradle cache. 
- I am not aware if the same error ever appear during Github workflow build. 

A full example: 
```
TrimEvaluationTest > runPassTests(ExprFunctionTestCase) > org.partiql.lang.eval.builtins.functions.TrimEvaluationTest.runPassTests(ExprFunctionTestCase)[27] FAILED
    org.partiql.lang.eval.evaluatortestframework.EvaluatorAssertionFailedError: Failure reason: FAILED_TO_EVALUATE_QUERY
    Note            : PartiQLCompilerPipeline (compile options unaltered)
    Group name      : null
    Query           : SELECT trim(both el from '   1ab1  ') AS trimmed
    FROM <<' 1'>> AS el
    Target pipeline : ALL_PIPELINES
    Expected result : [{trimmed:"ab"}]
    Actual result   : null
    Result format   : ION_WITHOUT_BAG_AND_MISSING_ANNOTATIONS
        at app//org.partiql.lang.eval.evaluatortestframework.AssertionsKt.assertDoesNotThrow(Assertions.kt:24)
        at app//org.partiql.lang.eval.evaluatortestframework.PipelineEvaluatorTestAdapter.privateRunEvaluatorTestCase(PipelineEvaluatorTestAdapter.kt:55)
        at app//org.partiql.lang.eval.evaluatortestframework.PipelineEvaluatorTestAdapter.runEvaluatorTestCase(PipelineEvaluatorTestAdapter.kt:28)
        at app//org.partiql.lang.eval.evaluatortestframework.MultipleTestAdapter.runEvaluatorTestCase(MultipleTestAdapter.kt:10)
        at app//org.partiql.lang.eval.EvaluatorTestBase.runEvaluatorTestCase(EvaluatorTestBase.kt:88)
        at app//org.partiql.lang.eval.EvaluatorTestBase.runEvaluatorTestCase$default(EvaluatorTestBase.kt:75)
        at app//org.partiql.lang.eval.builtins.functions.TrimEvaluationTest.runPassTests(TrimEvaluationTest.kt:22)
[...]

PR #969, opened 2023-01-13
Fixes list/bag ExprValue creation in plan evaluator

RCHowell:
## Relevant Issues

#966 
#951 

## Description

Description is in 966 and agg finder fix describe in 951. Peeking the sequence caused it to be instantiated prematurely, so we'll use John's original (and simpler) method of adding a meta to determine which ExprValue to create. The other fixes (it stopped at least? can't say definitively because ... nondeterministic) some nondeterministic behavior we were seeing

## Other Information
No

 No

No

## License Information

### DIFF
--- a/lang/src/test/kotlin/org/partiql/lang/compiler/memorydb/operators/GetByKeyProjectRelationalOperatorFactory.kt
+++ b/lang/src/test/kotlin/org/partiql/lang/compiler/memorydb/operators/GetByKeyProjectRelationalOperatorFactory.kt
@@ -57,10 +57,16 @@ class GetByKeyProjectRelationalOperatorFactory : ProjectRelationalOperatorFactor
         // Parse the tableId so we don't have to at evaluation-time
         val tableId = UUID.fromString(impl.staticArgs.single().textValue)
 
+        var exhausted = false
+
         // Finally, return a RelationExpression which evaluates the key value expression and returns a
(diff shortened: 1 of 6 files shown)


lang/src/test/kotlin/org/partiql/lang/compiler/memorydb/operators/GetByKeyProjectRelationalOperatorFactory.kt
```diff
@@ -57,11 +57,17 @@ class GetByKeyProjectRelationalOperatorFactory : ProjectRelationalOperatorFactor
         // Parse the tableId so we don't have to at evaluation-time
         val tableId = UUID.fromString(impl.staticArgs.single().textValue)
 
+        var exhausted = false
```
johnedquinn, line 60: Should the changes in the file be added in this PR?

johnedquinn (commented):
All looks good. Just one minor comment regarding the use of `exhausted`.

lang/src/test/kotlin/org/partiql/lang/compiler/memorydb/operators/GetByKeyProjectRelationalOperatorFactory.kt
RCHowell: It's just a test, and it would have warned us about the early sequence creation. I'm ok with keeping these changes

johnedquinn (approved):

PR #974, opened 2023-01-31
Fix Decimal Match Logic

yliuuuu:
## Relevant Issues
- [Closes/Related To] N/A

## Description
- Currently, PartiQL's decimal semantics remains vague. 
- Based on the SQL semantics: 
```
An exact numeric type has a precision P and a scale S. P is a positive integer that determines the number of significant digits in a particular radix R, where R is either 2 or 10. S is a non-negative integer. Every value of an exact numeric type of scale S is of the form n × 10^–S , where n is an integer such that –R^P ≤ n < R^P .

20) If a <precision> is omitted, then an implementation-defined <precision> is implicit.
21) The value of a <precision> shall be greater than 0 (zero). 
22) If a <scale> is omitted, then a <scale> of 0 (zero) is implicit.
23) If an <exact numeric type> contains <scale>, then the value of <scale> shall not be greater than the value of <precision> contained in that <exact numeric type>.
```

This PR fixes two things: 
1. Precision of 0 will now throw a Semantic error
```
1.00 is decimal(0,0)
   | 
org.partiql.lang.eval.EvaluationException: Precision 0 should be a positive integer
	Semantic Error: at line 1, column 9: Invalid precision or scale for decimal, <UNKNOWN> : <UNKNOWN>
```
2. Fix constrained decimal matching logic: 

Before: 
```
PartiQL> 1.0000 is decimal(5,3)
   | 
===' 
true
--- 
```
Now:
```
PartiQL>  1.0000 is decimal(5,3)
   | 
===' 
false
--- 
```

## Other Information
  - Yes. 
  - Yes. Bug fix for decimal matching, and forbid precision of 0. 
  - No. 

## License Information


RCHowell (dismissed):
Thank you for the fix. Could you please add these tests to the conformance test suite?

RCHowell:
Also, was there an associated SQL feature number associated with SQL exact numeric types?

yliuuuu:
Actually yes. This is a good call out. `E011` : Numeric data types. `E011-03` : DECIMAL and NUMERIC
data types.

RCHowell (approved):

PR #978, opened 2023-02-07
Fixes javadoc jar

RCHowell:
## Description

Dokka generated docs weren't getting added to the Javadoc JAR since I moved publishing to plugin. Now each module that uses the publishing plugin will appropriately create sources and javadoc jars as before.

**Workflow Changes**
## Workflow Changes

**Before - failing (cancelled, or fails after ~40mins)**

Before we had six workflows (three jobs, each run twice for java 11 and 17)

- Build and Test
  - tasks: build, test
- Randomized Tests
  - tasks: run randomized tests
- Report Generation
  - tasks: build, test, generate report, upload report

**After - success ~12mins**

- Randomized Tests
  - tasks: run randomized tests
- Report Generation
  - tasks: build, test, generate report, upload report

I removed the double build + test between jobs 1 and 2. Now we have four workflows (2 jobs, 2 java versions)

I also used the gradle action which tunes gradle (I don't know the exact details) to cache its setup between jobs and shouldn't use the gradle daemon. This fixed the workflow which was overwhelmed by building, testing, and generating docs four times.

**Skip Dokka**
You can disable running dokka to save several minutes on the build. Gradle performance is hard to measure because of caching, but it's generally faster to not do a task rather than doing said task. I have dokka skipped on CI
```
             `./gradlew clean build --no-build-cache` BUILD SUCCESSFUL in 8m 22s
 `DOKKA=false ./gradlew clean build --no-build-cache` BUILD SUCCESSFUL in 5m 14s
```

## Other Information
No

 No

No

## License Information

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,46 +22,17 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: recursive
+
       - name: Use Java ${{ matrix.java }}
         uses: actions/setup-java@v3
         with:
           distribution: 'corretto'
           java-version: ${{ matrix.java }}
-          cache: gradle
-      - name: Build and test
-        run: ./gradlew build
 
-  randomized-test:
-    runs-on: ubuntu-20.04
-    strategy:
-      matrix:
-        java: [ 11, 17 ]
-    steps:
-      - uses: actions/checkout@v2
-        with:
-          submodules: recursive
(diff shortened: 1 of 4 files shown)


vgapeyev (approved):
I confirmed that the doc jar now gets content.

I also tried to understand why and why it's the correct way, but comprehending enough Gradle for that will have to wait for another day.

PR #992, opened 2023-02-27
CREATE TABLE initial experimental syntax

vgapeyev:
An initial portion of experimental CREATE TABLE syntax, with basics of column declarations, with data types and NULL / NOT NULL constraints. 

## Other Information
  - Except that the previous plug-the-hole CREATE TABLE syntax would no longer work.

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Adds ability to pipe queries to the CLI.
 - Adds ability to run PartiQL files as executables by adding support for shebangs.
-- Adds initial syntax for CREATE TABLE.
+- Adds experimental syntax for CREATE TABLE, towards addressing 
+  [#36](https://github.com/partiql/partiql-docs/issues/36) of specifying PartiQL DDL.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,8 +49,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.9.2] - 2023-01-20
 
 ### Added
-- Adds ability to pipe queries to the CLI
-- Adds ability to run PartiQL files as executables by adding support for shebangs
+- Adds ability to pipe queries to the CLI.
+- Adds ability to run PartiQL files as executables by adding support for shebangs.
+- Adds initial syntax for CREATE TABLE.
 
 ### Changed
 
--- a/lang/src/main/antlr/PartiQL.g4
+++ b/lang/src/main/antlr/PartiQL.g4
@@ -95,16 +95,8 @@ tableDef
     ;
 
 tableDefPart
-    : columnDecl                        # TableDefColumn
-    | tableConstraint                   # TableDefConstr
-    ;
-
-columnDecl
-    : columnName columnDef
-    ;
-
-columnDef
-    : type columnConstraint*
+    : columnName type columnConstraint*                             # ColumnDeclaration
+    | ( CONSTRAINT tableConstraintName )?  tableConstraintDef       # TableConstraint
     ;
 
 columnConstraint
@@ -115,10 +107,6 @@ columnConstraintDef
     : NOT NULL                                  # ColConstrNotNull
     | NULL                                      # ColConstrNull
     | CHECK PAREN_LEFT expr PAREN_RIGHT         # ColConstrCheck
-    ;
-
-tableConstraint
-    : ( CONSTRAINT tableConstraintName )?  tableConstraintDef
     ;
 
 tableConstraintDef
--- a/lang/src/main/antlr/PartiQL.g4
+++ b/lang/src/main/antlr/PartiQL.g4
@@ -43,6 +43,11 @@ symbolPrimitive
     : ident=( IDENTIFIER | IDENTIFIER_QUOTED )
     ;
 
+tableName : symbolPrimitive;
+tableConstraintName : symbolPrimitive;
+columnName : symbolPrimitive;
+columnConstraintName : symbolPrimitive;
+
 /**
  *
  * DATA QUERY LANGUAGE (DQL)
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -49,8 +49,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.9.2] - 2023-01-20
 
 ### Added
-- Adds ability to pipe queries to the CLI
-- Adds ability to run PartiQL files as executables by adding support for shebangs
+- Adds ability to pipe queries to the CLI.
+- Adds ability to run PartiQL files as executables by adding support for shebangs.
+- Adds initial syntax for CREATE TABLE.
```
am357, line 54: How about adding a link to an issue in https://github.com/partiql/partiql-docs and labeling the implementation as experimental? something like [this](https://github.com/partiql/partiql-lang-kotlin/pull/831/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edL44-L46)

lang/src/main/antlr/PartiQL.g4
```diff
@@ -76,15 +81,38 @@ ddl
...
+    ;
+
+columnConstraint
+    : ( CONSTRAINT columnConstraintName )?  columnConstraintDef
+    ;
+
+columnConstraintDef
+    : NOT NULL                                  # ColConstrNotNull
+    | NULL                                      # ColConstrNull
+    | CHECK PAREN_LEFT expr PAREN_RIGHT         # ColConstrCheck
```
am357, line 109: Maybe it makes sense going only w/ `NOT NULL` and perhaps `NOT MISSING` for now until we flesh out the semantics of the rest in PartiQL (like `CHECK`).

lang/src/main/antlr/PartiQL.g4
```diff
@@ -76,15 +81,38 @@ ddl
...
+    ;
+
+tableConstraintDef
+    : CHECK PAREN_LEFT expr PAREN_RIGHT         # TblConstrCheck
+    ;
```
am357, line 114: Following my previous comment about column constraints, I think a more relevant constraint to support as experimental for now is the concept of open/close? meaning additional attributes can get added to the table or not.

lang/src/main/pig/partiql.ion
```diff
@@ -430,8 +430,8 @@ may then be further optimized by selecting better implementations of each operat
 
         // A data definition operation.
         (sum ddl_op
-            // `CREATE TABLE <symbol>`
-            (create_table table_name::symbol)
```
am357, line 434: Do we know why this implementation exists?

lang/src/test/kotlin/org/partiql/lang/syntax/PartiQLParserTest.kt
```diff
@@ -3543,14 +3543,58 @@ class PartiQLParserTest : PartiQLParserTestBase() {
...
-        "(ddl (create_table user))"
+        "CREATE TABLE \"user\" (\"lastname\" string)",
+        """
+            (ddl (create_table user (table_def
+                (column_declaration lastname (string_type)))))
+        """.trimIndent()
+    )
+
+    @Test
+    fun createTableWithConstraints() = assertExpression(
```
am357, line 3563: Can we also add couple of negative tests?

lang/src/test/kotlin/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransformTests.kt
```diff
@@ -877,20 +877,21 @@ class AstToLogicalVisitorTransformTests {
...
-    data class ProblemTestCase(val sql: String, val expectedProblem: Problem)
+    data class ProblemTestCase(val id: Int, val sql: String, val expectedProblem: Problem)
 
     @ParameterizedTest
     @ArgumentsSource(ArgumentsForProblemTests::class)
     fun `unimplemented features are blocked`(tc: ProblemTestCase) {
+        val id by lazy { "[in test case #${tc.id}]" }
         val problemHandler = ProblemCollector()
-        assertDoesNotThrow("Parsing TestCase.sql should not throw") {
+        assertDoesNotThrow("Parsing TestCase.sql should not throw $id") {
```
am357, line 887: Instead of adding _id_ couldn't we use `ProblemTestCase`'s _val_ using id in such a way seems odd IMO.

am357 (commented):
Thanks for putting this together, overall LGTM. Left some comments and some questions?

lang/src/test/kotlin/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransformTests.kt
vgapeyev: Did you mean using `ProblemTestCase`'s `sql` field? 

My purpose in introducing `id` was making it possible to point at the offending `ProblemTestCase` when the test fails.  The assert message as it was, was not informative. The `sql` field could work, but its contents are longer and there could be instances with only subtle differences, complicating the task of determining the correct one when fixing the tests.

vgapeyev (commented):
Responses to a few of the comments.  Will address the rest as suggested.

vgapeyev:
1st round of feedback addressed.

lang/src/main/antlr/PartiQL.g4
am357: Evaluation can also be evaluation of a `DML` statement like `INSERT`. In such cases `NOT MISSING` can mean that the attribute is non-optional and the incoming data must provide a value for such attribute. Having said that it makes sense to leave it out until we flesh out the `SchemaType` semantics.

lang/src/main/antlr/PartiQL.g4
am357: IMO we know that we need to support both open and close schemas; no objection on leaving it out though.

lang/src/test/kotlin/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransformTests.kt
am357: Makes sense, but wouldn't having a free flow `id` cause other problems; since we don't auto-generate or check for uniqueness?

vgapeyev: Yes, still not ideal. But we are in the context of comparing three viable alternatives that are there on hand for identifying the offending test: (1) leave it as it was, no identification; (2) identify with the full text of the query; (3) identify with a short integer.
This is internal testing; the driving principle for choosing among the 3 is efficiency of identifying the broken test that needs fixing. Meta-F on an integer beats the alternatives in my mind.

lang/src/main/antlr/PartiQL.g4
```diff
@@ -43,6 +43,11 @@ symbolPrimitive
     : ident=( IDENTIFIER | IDENTIFIER_QUOTED )
     ;
 
+tableName : symbolPrimitive;
```
am357, line 46: Can we also mention the same RFC and experimental notice in the comments here?

lang/src/test/kotlin/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransformTests.kt
am357: I guess using `id` is what made me think it's odd. It could be just a name for the test case; happy to park the discussion here though.

am357 (commented):
Thanks for making the changes—no major blocker just one request about adding the experimental notice w/ link to the RFC to the grammar too.

am357 (approved):

Issue #577, opened 2022-04-14
Migrate fully to JUnit 5


dlurton:
- There are many tests still using JUnit 4.
- There are now many test classes which mix JUnit 5's annotations and JUnit 4's assertions, which is messy.
- JUnit 4 tests don't seem to be running parallel under JUnit 5's backward compatibility runner.

One challenge:  we make somewhat heavy use of `junitparams`, which is tied to JUnit 4 and isn't needed any longer (since JUnit5 has such parameterization built in), so some non-trivial amount of mechanical refactoring will be required.  

This task will be completed when all references to JUnit 4 (and `junitparams`) are removed from the code in this repository.

Also, consider if #576 is required prior to starting this issue.  Should all tests should `import` their `assert*` functions instead of inheriting them from `TestBase`.

RCHowell:
Closing as duplicate https://github.com/partiql/partiql-lang-kotlin/issues/205

Issue #576, opened 2022-04-14
Remove `TestBase`


dlurton:
As mentioned in https://github.com/partiql/partiql-lang-kotlin/pull/573#discussion_r848788874, `TestBase` inherits from `org.junit.Assert`, which is how most of the inheriting test classes access the `assert*` methods (through Assert's protected methods.) This is disadvantages because these functions shadow anything imported, which creates some confusing situations sometimes in the IDE's code completion. This also means we have many JUnit5 tests that are using the JUnit4 assertion functions. This is not a huge problem, just a bit messy.

We should consider removing `TestBase` entirely

Issue #205, opened 2020-02-13
Migrate  to JUnit 5


therapon:
Not high priority but JUnit 5 has some quality of life improvements over JUnit4, for example: 
1. Parametrized tests support by default, you can even rerun a single parameter on IntelliJ 
1. Better support for asserting exceptions
1. `@DisplayName` lets you give more descriptive names on the test report
1. `@RepeatedTest` for repeating a test x amount of times, useful for tests that rely on random inputs 
1. `assertAll` support using lambdas to defer the assertions, we wouldn't need AssertJ anymore 

First one is something that would be particularly useful in IonSQL as it has a lot of parametrized tests

RCHowell:
More context in https://github.com/partiql/partiql-lang-kotlin/issues/577

RCHowell:
This work needs a rebase https://github.com/partiql/partiql-lang-kotlin/pull/994

PR #994, opened 2023-03-03
Try migrating from JUnit4 to JUnit5.

vgapeyev:
We currently have a mix of JUint4(+ JUnitParametrs) and JUnit5 tests (#577) as well as a questionable setup for importing JUnit4 assertions (#576). 

This PR is a peek into what would it take to fully transition to  JUnit5. There is nothing particularly new here: all changes are something that is already in use in other places. But it is worth to have a look at them, as a highlight of what will have to be done. 
My opinion is that the transition is doable, is worth it, -- but it would be somewhat of a mind-numbing  slog.

When there is agreement, I propose merging this PR and then following with the rest of the transition.   

A few specifics to note: 

* I decided to skip switching from `TestBase` to direct JUnit4 imports (as recommended in #576) and got straight to JUnit5 in the affected test suites, but take measures to make it possible to have a gradual transition, suite by suite.  

* In particular, this explains the `PartiQLParserTestBase_J5` duplicate of `PartiQLParserTestBase`, to deal with transitive inheritance from `TestBase`. (We can discuss whether we want to commit this like this in this PR and drop the `_J5` suffix later, or do it differently.)

* Introduction of `companion object`s (because @MethodSource needs a static method) is the most unpleasant thing, for its tab-shift and, sometimes, code relocation.  

See a couple more comments at specific spots. 

## Other Information

## License Information

### DIFF
--- a/lang/src/test/kotlin/org/partiql/lang/eval/NullIfEvaluationTest.kt
+++ b/lang/src/test/kotlin/org/partiql/lang/eval/NullIfEvaluationTest.kt
@@ -38,7 +38,6 @@ class NullIfEvaluationTest : EvaluatorTestBase() {
         fun testCase(expr1: String, expr2: String, expected: String) = NullIfTestCase(expr1, expr2, expected)
 
         @JvmStatic
-        @Suppress("unused")
         fun nullifEvaluationTests() = listOf(
             testCase("1", "1", "null"),
             testCase("1", "2", "1"),
--- a/lang/src/test/kotlin/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransformTests.kt
+++ b/lang/src/test/kotlin/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransformTests.kt
@@ -877,21 +877,20 @@ class AstToLogicalVisitorTransformTests {
         )
     }
 
-    data class ProblemTestCase(val id: Int, val sql: String, val expectedProblem: Problem)
+    data class ProblemTestCase(val sql: String, val expectedProblem: Problem)
 
     @ParameterizedTest
     @ArgumentsSource(ArgumentsForProblemTests::class)
     fun `unimplemented features are blocked`(tc: ProblemTestCase) {
-        val id by lazy { "[in test case #${tc.id}]" }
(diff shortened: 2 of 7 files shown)


lang/src/test/kotlin/org/partiql/lang/eval/NullIfEvaluationTest.kt
```diff
@@ -38,7 +38,6 @@ class NullIfEvaluationTest : EvaluatorTestBase() {
         fun testCase(expr1: String, expr2: String, expected: String) = NullIfTestCase(expr1, expr2, expected)
 
         @JvmStatic
-        @Suppress("unused")
```
vgapeyev, line 41: `@Suppress("unused")` on methods that create argument lists can now be removed in JUnit5 tests -- it looks like there are no longer warnings for this, if there ever were.

lang/src/test/kotlin/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransformTests.kt
```diff
@@ -877,21 +877,20 @@ class AstToLogicalVisitorTransformTests {
         )
     }
```
vgapeyev, line 878: This file's changeset restores the state of AstToLogicalVisitorTransformTests.kt to what it was before PR #992.  While figuring out JUnit5 here, I now saw another method of getting the identification of the failed test that I was after in PR992: re-run the suite with the failing test in IntelliJ -- it auto-numbers the cases of parameterized test.  It is not as good as the searchability of the id in PR992 (say, getting to the source of the offending parameters by clicking on the failure in the UI would be great), but is close enough and avoids the undeniable extra cruft of user-created IDs. 

Having said this, but looking at what happens in StatementRedactorTest in this PR (~80 test parameters with subtly different arguments), I am maybe getting by PR992 opinion back...

yliuuuu (approved):
I have no objection. My only comment is that we should finish migrate to Junit5 and clean up code base before 1.0

RCHowell:
Stale and needs rebase. Tagging this in issue #205

PR #1001, opened 2023-03-08
[sprout] Adds _ nested definition syntax

RCHowell:
## Relevant Issues
n/a

## Description

Adds `_::[ ... ]` and `_: { }` syntax to sum and product types to defined nested terms. This is like how Protobuf and other IDLs allows for nested definitions that aren't necessarily related to the "parent" definition. These behave the same as inline definitions and can be defined on either sum or product types. It's mostly a convenience to place definitions close to their usage

**Before**
```
sort::{
  specs: list::[sort_spec],
}

// .... some other lines — not always possible to place next to one another

sort_spec::{
  field: int,
  order: [ ASC, DESC ],
}
``` 

**After**
```
sort::{
  specs: list::[spec],
  _: [
    spec::{
      field: int,
      order: [ ASC, DESC ],
    }
  ]
}
```

**Generated Code** (no poems)
```kotlin
public data class Sort(public val specs: List<Spec>) : BaseNode() {

  public data class Spec(
    public val field: Int,
    public val order: Order
  ) : BaseNode() {

    public enum class Order {
      ASC,
      DESC,
    }
  }
}
```

## Other Information
No, sprout is an internal library

No

No

## License Information


yliuuuu:
It seems like the "old way" generates two separate classes like following: 
```
public data class Sort(
  public val specs: List<SortSpec>
) : ExampleNode()

public data class SortSpec(
  public val `field`: Int,
  public val order: Order
) : ExampleNode() {
  public enum class Order {
    ASC,
    DESC,
  }
}
```

No objection here, but it is worth to call out.

yliuuuu (approved):
One small thing to point out here is that in the description, you said support `_:: [....]` and `_: {...} `. I think the `_:{}` is a typo? 

So a couple learning notes here as well: 
1. Type: 
i. Product type: structure with fixed set of fields. Represented as struct. 
ii. Sum type: One of several defined structure. Represented as list. 
iii. Notice here the choice of list and struct has nothing to do with the order but the nature of type.
2. Annotation and field
i. Annotation is represented by `::`. In Sprout, product name and sum name are represented by annotation. I.e -> generated class name. Container type and optional tag also requires the double colon syntax. 
ii. Field is represented by `:` I.e., generated variable name.
3. Inline Definition: 
A product type can contain inline def: 
```
foo::{
  a: [...],               // inline sum foo.a
  b: v::[...],            // inline sum foo.v
  c: optional::[...],     // inline sum foo.c, optional field of foo
  d: optional::x::[...],  // inline sum foo.v, optional field of foo
  e: {...},               // inline product foo.e
  f: y::{...},            // inline product foo.y
  g: optional::{...},     // inline product foo.g, optional field of foo
  h: optional::z::{...},  // inline product foo.z, optional field of foo
 }
```

RCHowell:
> One small thing to point out here is that in the description, you said support _:: [....] and _: {...} . I think the _:{} is a typo?

No notice that `_: { ... }` syntax is used in the description. You are correct that struct and list were chosen because of their relationship to the algebraic types.

Annotations are just that .. annotations. Think of them as tags for a product or sum type definition. The special `_` syntax is a common sigil in programming languages for "*match* but ignore / don't bind".

The idea here is to *nest* definitions like Protobuf (or other IDLs that support nested defintions) without implying a relationship to the algebraic type.

For Product types, that is
```ion
foo::{
  a: int    // field a
  _: [       // field _ .. ignore, process as a list of nested definitions
    ...
  ]
}
```

For Sum types, it's
```
foo::[
  a::{ }  // variant a
  _::[    // variant _ .. ignore, process as a list of nested definitions
    ...
  ]
]
```

PR #1008, opened 2023-03-10
Moves partiql-lib to lib

RCHowell:
## Description

Now that we are splitting `lang/` into multiple artifacts, we want to be consistent with naming those artifacts `partiql-`. Libraries which not part of the partiql-lang-kotlin jar are kept in `lib/`

## License Information


vgapeyev:
Is `partiql-lang-kotlin/lib/sprout/src/main/kotlin/org/partiql/sprout/generator/target/dot/.gitkeep ` meant to be under source control?

vgapeyev (approved):
Looks fine to me, but out of context, re any plans/decisions that may have been made re this and other recent directory movements.  

If there is someone who is more in the loop on the matter, should maybe glance at this as well?

PR #1009, opened 2023-03-11
Fixes modeling of nested definitions and absolute import path

RCHowell:
## Description

Adding `_` syntax was hacky because there wasn't a nice way to express nested types in the model. This extends the model and fixes the generation (as well as fixes generated nested enum definitions).

## License Information


partiql-lib/partiql-sprout/src/main/kotlin/org/partiql/sprout/generator/target/kotlin/spec/KotlinNodeSpec.kt
```diff
@@ -62,33 +70,39 @@ sealed class KotlinNodeSpec(
...
         def = product,
         clazz = clazz,
         builder = TypeSpec.classBuilder(clazz),
         constructor = FunSpec.constructorBuilder(),
         companion = TypeSpec.companionObjectBuilder(),
-        types = types.toMutableList()
-    )
+        ext = ext,
+    ) {
+        override val children: List<KotlinNodeSpec> = nodes
```
yliuuuu, line 84: Do we not care about `Prop` here?

partiql-lib/partiql-sprout/src/main/kotlin/org/partiql/sprout/model/Model.kt
```diff
@@ -27,28 +27,34 @@ class Universe(
...
         override fun toString() = "sum::$ref"
     }
 
     /**
      * TypeDef.Product represents a structure of name/value pairs
      */
-    class Product(ref: TypeRef.Path, val props: List<TypeProp>, override val children: List<TypeDef>) : TypeDef(ref) {
+    class Product(ref: TypeRef.Path, val props: List<TypeProp>, override val types: List<TypeDef>) : TypeDef(ref) {
+
+        override val children: List<TypeDef> = props.filterIsInstance<TypeProp.Inline>().map { it.def } + types
```
yliuuuu, line 57: Confused here. is this a mismatch between the `children` in the core model and `children` in `KotlinNode`?

partiql-lib/partiql-sprout/src/main/kotlin/org/partiql/sprout/generator/target/kotlin/KotlinGenerator.kt
```diff
@@ -90,14 +90,9 @@ class KotlinGenerator(private val options: KotlinOptions) : Generator<KotlinResu
...
+        nodes = children.mapNotNull { it.generate(symbols) },
         clazz = symbols.clazz(ref),
-        children = children.mapNotNull { it.generate(symbols) },
-        types = props.filterIsInstance<TypeProp.Inline>().mapNotNull {
-            when (it.def) {
-                is TypeDef.Enum -> it.def.generate(symbols)
-                else -> null
-            }
-        }
+        ext = (props.enumProps(symbols) + types.enums(symbols)).toMutableList(),
```
yliuuuu, line 95: tracking enum. Where is this used?

partiql-lib/partiql-sprout/src/main/kotlin/org/partiql/sprout/generator/target/kotlin/spec/KotlinNodeSpec.kt
```diff
@@ -30,25 +30,33 @@ import org.partiql.sprout.model.TypeDef
...
+     */
+    abstract val children: List<KotlinNodeSpec>
 
     /**
      * Returns the built Pair<Base, Impl>
      */
     open fun build(): TypeSpec = with(builder) {
         primaryConstructor(constructor.build())
-        types.forEach { addType(it) }
+        ext.forEach { addType(it) }
```
yliuuuu, line 59: Tracking enums.

partiql-lib/partiql-sprout/src/main/kotlin/org/partiql/sprout/generator/target/kotlin/spec/KotlinNodeSpec.kt
RCHowell: Prop is separated at this point. Nested definitions (including inlines of prop) are assigend to no. Prop doesn't hold a typespec builder. Prop is simply a name-to-type holder.

partiql-lib/partiql-sprout/src/main/kotlin/org/partiql/sprout/model/Model.kt
RCHowell: You're right. There is a slight mismatch. This model keeps inline definitions within the TypeProp.Inline variant whereas the Kotlin generation separates them out. For the model, keeping inline definitions within TypeProp captures the fact that these definitions are inline nested definitions rather than normal nested definitions. For Kotlin generation, the distinction doesn't matter hence why it's dropped in the generator, but I figured the distinction was worth keeping in the model.

partiql-lib/partiql-sprout/src/main/kotlin/org/partiql/sprout/generator/target/kotlin/spec/KotlinNodeSpec.kt
RCHowell: ext is an extension point to add any KotlinPoet TypeSpec to a node. For the built in poems, this just means enums.

yliuuuu (commented):
The naming choice is a bit unfortunately, but no better opinion from my side. Want to make sure I understand this correctly. Thanks for being patient with me here.

yliuuuu (approved):
Thanks for the explanation.

yliuuuu (approved):

PR #1018, opened 2023-03-25
Fix conformance GH Actions workflow by using a fixed gradle version

alancai98:
Noticed that the GH Actions workflow for conformance started failing. Example failed run: https://github.com/partiql/partiql-lang-kotlin/actions/runs/4506051931/jobs/7932435216#step:4:38. I think this is due to the GH Actions workflow running with Gradle 8.0. This PR fixes the conformance gradle version to what's in the `gradle-wrapper`, `7.5.1`.

## License Information

### DIFF
--- a/.github/workflows/conformance-report.yml
+++ b/.github/workflows/conformance-report.yml
@@ -61,7 +61,11 @@ jobs:
         with:
           distribution: 'corretto'
           java-version: 17
-          cache: gradle
+      # Fix gradle version to what's used in gradle-wrapper
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@v2
+        with:
+          gradle-version: 7.5.1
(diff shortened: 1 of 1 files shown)


.github/workflows/conformance-report.yml
```diff
@@ -19,7 +19,11 @@ jobs:
         with:
           distribution: 'corretto'
           java-version: 17
-          cache: gradle
```
alancai98, line 22: Using the `actions/setup-java@v3` used to pull in Gradle 7.*. In some recent change to that action, Gradle 8.* was getting pulled in which caused build failures. Using Gradle's `gradle/gradle-build-action@v2` plugin allows us to fix the version to a specific version.

am357 (approved):

PR #1022, opened 2023-03-31
Adds EVERY, ANY, and SOME aggregate builtins

vgapeyev:
Adding a few built-in functions, some out of those being requested. 

* `TEXT_REPLACE(str, from, to)`, same as PostreSQL `REPLACE`, but using a different name because the latter is now a keyword in PartiQL DML. 

* `EVERY` -- a standard SQL aggregation on booleans.

Also, some touch-ups in `Functions.md`, `Aggregate Functions.md` -- for consistency and slightly better typesetting. 

## Other Information

## License Information


partiql-lang/src/main/kotlin/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -243,6 +253,10 @@ internal class EvaluatingCompiler(
                             ?: ExprValue.nullValue
                 }
             }
+            val everyAccFunc: (ExprValue?, ExprValue) -> ExprValue = { accumulated, nextItem ->
+                checkIsBooleanType("EVERY", nextItem)
+                accumulated?.let { ExprValue.newBoolean(it.booleanValue() && nextItem.booleanValue()) } ?: nextItem
+            }
```
RCHowell, line 259: I know performance isn't exactly a concern, but perhaps we use ExprAggregator so that we don't have to instantiate an ExprValue for every accumulated row just to be unpacked in the next accumulator.next call.

```kotlin
val everyAccFunc = { filter: (ExprValue) -> Boolean ->
    object : ExprAggregator {

        var everyIsTrue: Boolean = true

        override fun next(value: ExprValue) {
            if (value.isNotUnknown() && filter.invoke(value)) {
                checkIsBooleanType("EVERY", value)
                everyIsTrue = everyIsTrue && value.booleanValue()
            }
        }

        override fun compute() = ExprValue.newBoolean(everyIsTrue);
    }
}
```

partiql-lang/src/main/kotlin/org/partiql/lang/eval/builtins/BuiltinImplems.kt
```diff
@@ -0,0 +1,9 @@
+package org.partiql.lang.eval.builtins
+
+import org.partiql.lang.eval.ExprValue
+import org.partiql.lang.eval.stringValue
+
+internal fun characterLength(value: ExprValue): Int {
+    val str = value.stringValue()
+    return str.codePointCount(0, str.length)
+}
```
RCHowell, line 9: Rather than making a new file, perhaps we put this in ExprValue extensions as "stringValueCodepointLength"

```kotlin
private fun ExprValue.stringValuecodepointLength(): Int {
    val str = stringValue()
    return str.codePointCount(0, str.length)
}
```

Alternatively, since all builtin classes are internal, this can be a static function of `ExprFunctionCharLength`. I missed that on the initial factoring.

Any particular reason why we are adding ExprFunctionLength? Seems confusing because the input domain isn't included in the function name. There are also already two functions that do this. We also have two functions for the lengths of collections (cardinality (sql) and size (?) ). I don't think we need this one.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/builtins/ScalarBuiltinsExt.kt
```diff
@@ -497,3 +500,30 @@ internal object ExprFunctionToString : ExprFunction {
...
+        name = "text_replace",
+        requiredParameters = listOf(StaticType.TEXT, StaticType.TEXT, StaticType.TEXT),
+        returnType = StaticType.TEXT,
+    )
+
+    override fun callWithRequired(session: EvaluationSession, required: List<ExprValue>): ExprValue {
+        val string = required[0].stringValue()
+        val from = required[1].stringValue()
+        val to = required[2].stringValue()
+        return ExprValue.newString(string.replace(from, to))
```
RCHowell, line 527: I don't believe it matters now, but we should consider writing string manipulation functions with the byte arrays / codepoints rather than the Kotlin builtins. but honestly if the result is the same then it doesn't matter.

partiql-lang/src/test/kotlin/org/partiql/lang/eval/builtins/functions/TextReplaceExprFunctionTest.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    class TestCases : ArgumentsProviderBase() {
+        override fun getParameters(): List<Any> = listOf(
+            ExprFunctionTestCase("text_replace('abcdefabcdef', 'cd', 'XX')", "\"abXXefabXXef\""),
+            ExprFunctionTestCase("text_replace('abcdefabcdef', 'xyz', 'XX')", "\"abcdefabcdef\""),
+            ExprFunctionTestCase("text_replace('abcdefabcdef', 'defab', '')", "\"abccdef\""),
+            ExprFunctionTestCase("text_replace('abcabcabcdef', 'abcabc', 'XXX')", "\"XXXabcdef\""),
+            ExprFunctionTestCase("text_replace('abcabcabcdef', '', 'X')", "\"XaXbXcXaXbXcXaXbXcXdXeXfX\""),
+            ExprFunctionTestCase("text_replace('', 'abc', 'XX')", "\"\""),
+            ExprFunctionTestCase("text_replace('', '', 'XX')", "\"XX\""),
```
RCHowell, line 23: Let's add some multi-codepoint examples as well

RCHowell (changes requested):

vgapeyev (commented):
Responding to Robert's PR comments.  A few related commits are to follow.

jpschorr:
We should add these tests to the conformance test suite as well.

RCHowell (approved):

PR #1027, opened 2023-04-03
Introduces PartiQLSchemaInferencer and PlannerSession

johnedquinn:
## Relevant Issues
- None

## Description
- Introduces new public APIs: PartiQLSchemaInferencer and PlannerSession
- Internally, creates:
  - Metadata (consolidated ConnectorMetadata class)
  - ReferenceResolver (aids in resolving locals/globals)
  - PlanTyper (types the logical plan)
- Makes slight modifications to the PartiQL Plan
- Adjusts the Gradle build process to use the Sprout command to generate the Plan code and place it in the generatedSrc directory.
  - Therefore, I deleted all of the checked-in generated code
- Adjusts PlanRewriter to use a parameterized type instead of defaulting to Unit

## Other Information
  - New APIs

## License Information


partiql-lang/src/main/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencer.kt
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+
+    private const val DEFAULT_TABLE_NAME = "UNSPECIFIED"
+
+    private fun inferInternal(query: String, ctx: Context): ValueDescriptor {
+        val parser = PartiQLParserBuilder.standard().build()
+        val ast = parser.parseAstStatement(query)
+        val normalizedAst = ast.normalize()
+        return inferUsingLogicalPlan(normalizedAst, ctx)
```
yliuuuu, line 123: There is a function called `AstToPlan`. Any reason not to use it ?

partiql-lang/src/main/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencer.kt
```diff
@@ -0,0 +1,174 @@
...
+                metadata = ctx.metadata,
+                scopingOrder = PlanTyper.ScopingOrder.LEXICAL_THEN_GLOBALS,
+                customFunctionSignatures = emptyList(),
+                problemHandler = ctx.problemHandler
+            )
+        )
+        return convertSchema(typedPlan)
+    }
+
+    private fun convertSchema(rex: Rex): ValueDescriptor = when (rex) {
```
yliuuuu, line 149: Do you see the `valueDescriptor` getting replaced by a schema Type? If so, may be mark this in a comment?

yliuuuu (commented):
Great job. Left a couple comments here. I also have a couple high-level comments that is out of the scope for this PR, but will be nice to have as a follow up. 
1. May be we can mark the new Sprout generated Plan and this Plan Inferencer as experimental? 
2. May be we can take the opportunity to deprecated the old inferencer on AST. 
3. Is it possible to route the new plan inferencer to the existing inferencer tests?

partiql-lang/src/main/kotlin/org/partiql/lang/planner/transforms/plan/PlanTyper.kt
```diff
@@ -0,0 +1,1535 @@
...
+    }
+
+    override fun visitRexBinary(node: Rex.Binary, ctx: Context): Rex.Binary {
+        val lhs = visitRex(node.lhs, ctx).grabType() ?: handleMissingType(ctx)
+        val rhs = visitRex(node.rhs, ctx).grabType() ?: handleMissingType(ctx)
+        val args = listOf(lhs, rhs)
+        val type = when (node.op) {
+            Rex.Binary.Op.PLUS, Rex.Binary.Op.MINUS, Rex.Binary.Op.TIMES, Rex.Binary.Op.DIV, Rex.Binary.Op.MODULO -> when (hasValidOperandTypes(args, node.op.name, ctx) { it.isNumeric() }) {
+                true -> computeReturnTypeForNAry(args, PlanTyper::inferBinaryArithmeticOp)
+                false -> StaticType.NUMERIC // continuation type to prevent incompatible types and unknown errors from propagating
```
am357, line 300: I likely need to read the code in more depth but can you please elaborate on this?

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
```diff
@@ -0,0 +1,663 @@
...
+import org.partiql.types.BoolType
+import org.partiql.types.StaticType
+import org.partiql.types.StaticType.Companion.unionOf
+import org.partiql.types.StructType
+import java.net.URL
+import java.time.Instant
+import java.util.stream.Stream
+import kotlin.test.assertTrue
+
+class PartiQLSchemaInferencerTests {
```
am357, line 32: Does the code can handle `JOIN`s correctly today? If so, can we add the related tests (could be another PR). Otherwise, let's add an issue to track the `JOIN` support.

am357 (commented):
Thank you for the effort! Really great work in putting this together! this feature is a major leap towards making PartiQL functionalities modular which has been our vision for a while. 

No major comments but In addition to the individual comments, adding the following additional comments:
1. Let's ensure our APIs have proper documentation; currently if not mistaken, some don't and some are lean on the documentation side (can be in a separate PR); also thank you for addressing some already in [#1028](https://github.com/partiql/partiql-lang-kotlin/pull/1028).
2. If not there already, let's also create an issue for a creating Wiki section for our Plugin system.
3. I don't think we're overloading the term `Schema` even when considering one use in Plan nodes as an outcome of `PlanType::type` and one as the final output of the `PartiQLSchemaInferencer::infer`, but keen to understand your point of view on this. We can also discuss offline about this.

partiql-lang/src/main/kotlin/org/partiql/lang/planner/transforms/plan/PlanTyper.kt
johnedquinn: Essentially the same as [here](https://github.com/partiql/partiql-lang-kotlin/blob/965183a2b65d944acc541ad7e0f88929332680f6/partiql-lang/src/main/kotlin/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt#L447-L495).

For binary arithmetic operations, we make sure both operands are numeric, then we compute the return type using the inferBinaryArithmeticOp.

partiql-lang/src/main/kotlin/org/partiql/lang/planner/transforms/plan/PlanUtils.kt
```diff
@@ -0,0 +1,51 @@
...
+
+package org.partiql.lang.planner.transforms.plan
+
+import org.partiql.plan.Attribute
+import org.partiql.plan.Rel
+import org.partiql.plan.Rex
+import org.partiql.types.StaticType
+
+internal object PlanUtils {
+    internal fun getSchema(input: Rel): List<Attribute> = when (input) {
```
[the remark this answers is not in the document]
johnedquinn: With the above, where schema corresponds to a List<Attribute>, we can still support `<<[INT, DECIMAL]>>`.

`List<Attribute>` is essentially ordered Bindings<StaticType>. So, consider below:
```partiql
SELECT a
FROM << [1, 1.0] >> AS a
```

The SCAN operator's Schema (`List<Attribute>`:w
) would correspond to:
```
[
  Attribute('a', ListType(IntType(), DecimalType()))
]
```

And then, the PROJECT operator would maintain the same list of attributes. I'll add a test to be explicit about this.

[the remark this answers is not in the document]
am357: So when I said `<<[INT, DECIMAL]>>` as the schema I meant the output schema. W/ bindings we're still talking in terms of collection of tuples.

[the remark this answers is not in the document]
johnedquinn: Oh -- we do support it. I'm adding the following test cases which work without PR modification:
```kotlin
            SuccessTestCase(
                name = "Non-tuples in SELECT VALUE",
                query = "SELECT VALUE a FROM << [ 1, 1.0 ] >> AS a",
                expected = ValueDescriptor.TypeDescriptor(
                    BagType(ListType(unionOf(StaticType.INT, StaticType.DECIMAL)))
                )
            ),
            SuccessTestCase(
                name = "SELECT VALUE",
                query = "SELECT VALUE [1, 1.0] FROM <<>>",
                expected = ValueDescriptor.TypeDescriptor(
                    BagType(ListType(unionOf(StaticType.INT, StaticType.DECIMAL)))
                )
            ),
```

LMK if that's what you mean. As for the code snippet you referenced, that is a utility method to get the output schema of a single Rel.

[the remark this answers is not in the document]
am357: I see, that's nice!. In that case I'm confused about the `Schema` definition as a collection of `Attribute` vs collection of `StaticType`; let's discuss offline to flesh it out further.

johnedquinn:
Alright, I'm publishing a few changes to address all comments. Also responding here to some questions.

@yliuuuu:
1. I've marked the Inferencer as experimental.
2. As for deprecating the StaticTypeInferencer, this can be a follow-up, if you're aligned.
3. For the existing tests, it's extremely difficult since they attach assertions related to SourceLocationMetadata, which we don't have on the Plan nodes. They also don't differentiate between success vs error tests, so they leave everything up to the handlers. Makes test sharing extremely difficult.

@am357:
1. ACK. These all should have been addressed.
2. See #1031 
3. We'll discuss offline.

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
johnedquinn: Decided to add JOINs.

johnedquinn:
Ok, after porting the StaticTypeInferenceVisitorTransform tests locally, I made some minor changes to the PlanTyper. In a follow-up PR, I'll add the test porting from https://github.com/partiql/partiql-lang-kotlin/commit/d1e5e53d9e0e7add9d099a89679f4fbad635e8d0

partiql-lang/src/main/kotlin/org/partiql/lang/planner/transforms/plan/PlanUtils.kt
am357, line 23: Considering our most recent discussion, should the term `schema` in planner along w/relevant methods (E.g. `getSchema` get changed?

am357: Just saw https://github.com/partiql/partiql-lang-kotlin/issues/1032, thanks for creating the issue. If this is not going to be part of this PR just let me know so we punt it to a subsequent one.

PR #1049, opened 2023-04-19
Adds support for CURRENT_USER

### LINKED ISSUE #1048: Support SQL's CURRENT_USER
## Relevant Issue/Bug
- None

## Requested Solution/Feature
- The PartiQL AST, Plan, and Evaluator should support SQL's `CURRENT_USER`.

## Research
- According to SQL-92: "The data type of CURRENT_USER, SESSION_USER, and SYSTEM_USER is character string. Whether the character string is fixed-length or variable-length, and its length if it is fixed-length or maximum length if it is variable-length, are implementation-defined. The character set of the character string is SQL_TEXT."
- We can take a look at other implementations for potential guidance. For example, Trino has a dedicated node for `CURRENT_USER`, and when planning, the node gets translated into a hidden function call, which takes in a session and returns a string representing the current user of the session.
  - Using this model, we can still represent SQL's `CURRENT_USER` within the AST & Plan & during inference/evaluation.

## Additional Context
- None

## DoD (Definition of Done)
- [ ] CURRENT_USER AST Node added
- [ ] CUURENT_USER Function added
- [ ] CURRENT_USER added to EvaluationSession
- [ ] CURRENT_USER working within Evaluator
- [ ] CURRENT_USER tested with the PartiQLSchemaInferencer

johnedquinn:
## Relevant Issues
- Closes #1048 

## Description
- Adds syntactical support for CURRENT_USER by:
  - Adding an AST Node for `session_attribute`
  - Adding syntax support in ANTLR. Parsing keyword CURRENT_USER into a `session_attribute`
- Adds support for evaluation of the CURRENT_USER in Evaluating Compiler by:
  - Modifying compilation of SessionAttribute to look into the EvaluationSession.
- Adds support for evaluation of CURRENT_USER in the Planner Pipeline by:
  - Adding the hidden function `$__current_user`. Its signature specifies a return type of STRING/NULL. It loads the CURRENT_USER given the Session's Context.
  - Modifying AstToLogicalVT to transform the current_user AST node into a Logical node function call
- Adds support for schema inference of CURRENT_USER using experimental planner by:
  - Adding support within `RexConverter` to convert into a `Rex.Call`
- Adds support for pretty printing the Session Attribute within the Query Pretty Printer
- Updates CLI to make the current user the current user on the host machine using Java's system properties
- Adds tests
  - Related to parsing
  - Related to the AstToLogicial transform
  - Related to the evaluation of multiple scenarios of using CURRENT_USER
  - Related to schema inference in the PartiQLSchemaInferencer
  - Related to the pretty printer

## To Run

Install the PartiQL CLI and use CURRENT_USER anywhere in your queries! You can use both the STANDARD and EXPERIMENTAL pipeline flags.

Screenshot 2023-04-19 at 3 40 49 PM

## Open Questions
- [CLOSED] This PR adds a default user "UNKNOWN_USER" to the EvaluationSession, if the EvaluationSessionBuilder did not specify a user. However, this is NOT found in the SQL-92 Specification. The SQL-1992 standard does not have a concept of "default value" for CURRENT_USER, as it is a session-specific value that depends on the user who is currently connected to the database.
  - UPDATE: CURRENT_USER may be null. See research notes below. Updated the function and inferencer tests accordingly. Removed the default value.
- [CLOSED] I also named the function "$__current_user", because I couldn't find in the SQL-92 Specification how systems should represent the function internally (if specified). Multiple implementations allow for both `CURRENT_USER` and `CURRENT_USER()` however.
  - UPDATE: Researching into the spec doesn't say whether the identifier should also act as a function name. Postgres also takes the same approach. The approach of hiding the function name is not a 1-way door, as we can always expose the function in the future if there is a desire to do so.

## Research Notes
- According to SQL-1999: "When an SQL-session
is initiated, the SQL-session user identifier is determined in an implementation-defined manner,
unless the session is initiated using a <connect statement>. The value of the SQL-session user
identifier can never be the null value. The SQL-session user identifier can be determined by using
SESSION_USER."
- ". The current user identifier and the
current role name are called the current authorization identifiers.
At a given time, the value of the current user identifier or the current role name can be a null
value, but their values cannot both be null values at the same time."
- Subsection 4.34: "When an SQL-session is initiated other than through the use of an explicit <connect state-
ment>, there is an implementation-defined initial value of the SQL-session user identifier."

## Other Information

## License Information


partiql-lang/src/main/kotlin/org/partiql/lang/eval/EvaluationSession.kt
```diff
@@ -31,7 +31,8 @@ class EvaluationSession private constructor(
     val globals: Bindings<ExprValue>,
     val parameters: List<ExprValue>,
     val context: Map<String, Any>,
-    val now: Timestamp
+    val now: Timestamp,
+    val currentUser: String?
```
am357, line 35: Just bouncing back ideas here: considering addition of other variables like `CURRENT_ROLE` does it make sense to add a more general argument like `sysVariables`?

am357 (commented):
Overall LGTM—just one comment regarding the `curretUser` parameter in `EvaluationSession`.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/EvaluationSession.kt
vgapeyev: Note the `context` argument here already.  The doc comment for it appears to cover this purpose.

Said this, I, so far, continue to maintain my opinion that CURRENT_USER is best be a regular identifier, supplied through the global environment.  I do not see any functionality for it in the language, besides lookup -- which is the functionality of variables/identifiers. 

But I'll explore this PR some more before passing a firmer judgment :-)

johnedquinn: ACK -- the context parameter is a good point. I'll update to reflect.

As for the usage as a regular identifier, are you proposing that we modify `compileId` within `EvaluatingCompiler` to have a `when` statement to check for certain variables such as `current_user`?

I feel as if the most appropriate representation would be the function call -- as it takes in the EvaluationSession, and it wouldn't require modifying the `EvaluatingCompiler`.

vgapeyev (commented):
Upon looking through the PR's change set, I am now quite convinced that the approach with CURRENT_USER being a dedicated AST node merely creates a new code path for "passing the buck" until a certain value is looked up somewhere.  Such a path is already in place for identifiers / global variables. 
But, as the change set demonstrates, this new code path requires maintenance and testing, as well as replication for each new system variable in the future.

This may benefit from a 2nd or even 3rd opinion, but my recommendation remains: treat CURRENT_USER as an identifier that it to be provided by the global environment.  
Whether CURRENT_USER lookup from the global environment should be implemented right now is more of an open question. The user request we have at the moment is just to be able to parse a query that contains CURRENT_USER.  
If we do not like to live with the "unbound identifier error" (which I think would be just fine for some time) it should be possible to do something to augment the global environment in the CLI, similarly to what the PR does currently. This would be easy to change when we get to proper deliberate improvement project for global environment.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/EvaluationSession.kt
vgapeyev: To address the current need (to be able to parse a query that mentions CURRENT_USER) all that needs to be done, I think, is removing CURRENT_USER from the list of tokens, which should make it parse as an identifier. 

If we'd like a query with CURRENT_USER to evaluate in the CLI, the best and easiest at the moment would be to augment the global bindings that CLI uses with a binding for CURRENT_USER. 

Either way, I don't foresee a need to touch EvaluatingCompiler or anything else in the processing pipeline.

am357:
> Upon looking through the PR's change set, I am now quite convinced that the approach with CURRENT_USER being a dedicated AST node merely creates a new code path for "passing the buck" until a certain value is looked up somewhere. Such a path is already in place for identifiers / global variables. But, as the change set demonstrates, this new code path requires maintenance and testing, as well as replication for each new system variable in the future.
> 
> This may benefit from a 2nd or even 3rd opinion, but my recommendation remains: treat CURRENT_USER as an identifier that it to be provided by the global environment. Whether CURRENT_USER lookup from the global environment should be implemented right now is more of an open question. The user request we have at the moment is just to be able to parse a query that contains CURRENT_USER. If we do not like to live with the "unbound identifier error" (which I think would be just fine for some time) it should be possible to do something to augment the global environment in the CLI, similarly to what the PR does currently. This would be easy to change when we get to proper deliberate improvement project for global environment.

@vgapeyev I agree that having an AST node does not seem to be necessary but I see parsing and planning of the `CURRENT_USER` as an internal (system) function with specific signature makes more sense mainly b/c `CURRENT_USER` has an specific meaning rather than an arbitrary environment variable.

vgapeyev:
> ... `CURRENT_USER` has specific meaning rather than an arbitrary environment variable.

And what I am advocating for, is that the meaning (the STRING type of the variable and the id/name of the specific user) would be most naturally provided by the environment. 
Comparing to the alternative, look at it from this angle: what does the proposed "system built-in" `ExprFunctionCurrentUser` function do?  It looks up `session.currentUser`. The latter is added to the session at the same time as global binding are.  There is nothing more happening besides this.  Skip the middle man!

am357:
> Comparing to the alternative, look at it from this angle: what does the proposed "system built-in" `ExprFunctionCurrentUser` function do? It looks up `session.currentUser`. The latter is added to the session at the same time as global binding are. There is nothing more happening besides this. Skip the middle man!

The difference is when we have it modeled as a function in the logical plan, we've captured this semantic whereas having it as a free flow environment reference can mean anything.

vgapeyev:
> we've captured this semantic

From the point of view of the *language* what semantics is there to CURRENT_USER, besides it being of type STRING and yielding a string value when referenced? 
These are all that matters for CURRENT_USER in queries like 
```
SELECT * FROM Orders as x WHERE x.buyer = CURRENT_USER
```

To be fair, both a scalar value and a zero-arity function returning a scalar value would capture this semantics ("this is a string") equally well -- but provided the function is also made available through the environment.  What I am arguing against is putting a function into the standard library -- that does nothing but looks up from what amounts to the global environment.

am357:
> From the point of view of the language what semantics is there to CURRENT_USER, besides it being of type STRING and yielding a string value when referenced?

Isn't this relevant that it captures the SQL standard `CURRENT_USER` with whatever specification we have for it in SQL? I can see your point of view with regards to the same outcome but cannot see it appealing to leave it to the environment as yet another binding—this is all because it has a especial meaning and in my view needs especial treatment to ensure it reflects the intention of its use.

I can see similar approaches in [PostgresSQL](https://www.postgresql.org/docs/9.1/functions-info.html), [SQL Server](https://learn.microsoft.com/en-us/sql/t-sql/functions/current-user-transact-sql?view=sql-server-ver16), and [MySQL](https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_current-user) (haven't checked the rest).

johnedquinn:
A few questions:

1. This is a simple scenario, but what happens when you want to execute?
```sql
CREATE TABLE "CURRENT_USER" ( ... )
```
  - By expressing the CurrentUser as just a binding within the GlobalsEnvironment, we would rid users of the ability to use keywords as database variable names (with the use of quotes).
  - This is currently supported by Postgres, btw.
2. Currently, we model the global bindings environment to be an attribute of the EvaluationSession, but is this the right thing to do? In SQL, these ideas are separated. The bindings environment is effectively the collection of data being managed by the DBMS. The session is an active connection between the client and the DBMS, and it acts to allow clients to interact with the database. Our specification doesn't give hints that the data itself should be represented within the session. I bring this up because it might not be in our best interest to model session attributes as bindings within the global environment -- and I don't think we want to move in that direction.
  - This also brings me to SQL-1992 Section  21.2.7  "TABLES view", whose function is "Identify the tables defined in this catalog that are accessible to a given user." With the creation of the TABLES view, the statement, in PartiQL terms, adds a binding representing the data accessible to a particular user. When multiple users simultaneously connect to a database, the environment (separate from the session) contains all tables -- and the TABLES view shows which tables are allowed to be queried by any particular user. Similarly, when a user attempts to query any table, "The current \<authorization identifier\> determines the privileges for the execution of each SQL-statement. For direct SQL, the SQL-session \<authorization identifier\> is always the current \<authorization identifier\>." (Section 4.26)
  - If the environment were to directly hold the value of `CURRENT_USER`, how could the database hold multiple simultaneous user identifiers?
3. Thinking about the recent connector/catalog contributions -- when we call `ConnectorMetadata::getObjectType`, should catalogs be given the freedom to return any type besides `unionOf(STRING, NULL)` ? This is also somewhat related to question number 1. IMO, this API should be describing data managed by the DBMS -- not session details.
  - If CURRENT_USER (the keyword) is used as the argument for `ConnectorMetadata::getObjectType` -- we'd be giving external contributors the ablity to return ANY type. And, even if we make it explicit that it must be a string, they can return any string. Whereas, if it is maintained as part of the session, we wouldn't need to ask some connector for the current_user -- we'd have it already.

These 3 questions lead me to believe that we should be modeling `CURRENT_USER` as a function call that takes in the current session, rather than a binding within the bindings environment.

vgapeyev:
> cannot see it appealing to leave it to the environment as yet another binding

That's what I find actually appealing!  The language does not need to have any special treatment hard-wired to "know" that CURRENT_USER is different from any other identifier pointing to a string.  The environment can deal with the matters of whether this is a name of a known user, whether the user has authenticated, whether the user has rights to tables, whether the user has changed since a session has started, etc, etc, etc.  

Perhaps I should say that I am talking about a more sophisticated notion of "global environment" than we currently have in place.  I believe I said as much in one of the discussions, but did not put it in writing (that we should use this as an opportunity to schedule a project to properly design environments).  I do not suggest designing environments as part of this PR, but aspirationally it would be something like 
```
interface GlobalEnvironment {
  fun resolveToValue(id: BindingName): ExprValue
  fun resolveToFunction(id: BindingName): ExprFunction
  // ...
}
``` 
where it is an implementation of `resolveToValue` (or `resolveToFunction`, if we decide to go with `CURRENT_USER()` instead) that decides what to return for any given argument, be it `BindingName("CURRENT_USER")` or not.  
Of course, that is us who supply at least one implementation, and that's where all "CURRENT_USER is the current user" semantics would go. 

For now, just augmenting `EvaluationSession.globals`, somewhere before passing it down the pipeline, with a binding for CURRENT_USER would suffice.

am357:
> The environment can deal with the matters of whether this is a name of a known user, whether the user has authenticated, whether the user has rights to tables, whether the user has changed since a session has started, etc, etc, etc.

The point I'm trying to make is some functions like `CURRENT_USER` as they have especial meaning cannot be handed to an arbitrary environment declaration—do we want to give the ability to users to override `CURRENT_USER` to have a different signature? my point is, these are built-ins and should be treated as such.

vgapeyev:
> The point I'm trying to make is some functions like `CURRENT_USER` as they have especial meaning cannot be handed to an arbitrary environment declaration—do we want to give the ability to users to override `CURRENT_USER` to have a different signature? my point is, these are built-ins and should be treated as such.

And all I am trying to communicate is that the best place to localize these concerns is in defining and constraining what a "proper global environment" is, not in the language. 

Let me try to put my thinking as a criterion that a third-party arbitrage should be able to resolve: A solution to this matter will be acceptable if it does not contain special-casing on the `CURRENT_USER` token hardwired anywhere in the language, which includes tokenizer, parser, desugaring passes ("visitor transforms"), compiler, and interpreter/evaluator. 

Whether it is a value or a function is not important; it is important where is the special case that resolves `CURRENT_USER` token into something.  Ideally, there is exactly 1 place that does that, and it is outside the language.

vgapeyev:
> These 3 questions lead me to believe that we should be modeling `CURRENT_USER` as a function call that takes in the current session, rather than a binding within the bindings environment.

If you can implement this while satisfying the arbitration criterion I have formulated in the other comment, then I'd now have to accept it :-) 
I think this can be done for `CURRENT_USER()` (syntactically, a call to a zero-arity function), but not for `CURRENT_USER` (syntactically, an identifier lookup). 

The questions that lead the comment are all relevant questions to consider when designing the concept of the environment properly.  I believe they all can be addressed satisfactorily in the direction that I advocate for (the environment is not just a "dumb" list of bindings), but I hoped that, for this PR, we could agree on the contours of that, but implement something that is an approximation and perhaps a hack, but is close in spirit.

johnedquinn:
> Let me try to put my thinking as a criterion that a third-party arbitrage should be able to resolve: A solution to this matter will be acceptable if it does not contain special-casing on the CURRENT_USER token hardwired anywhere in the language, which includes tokenizer, parser, desugaring passes ("visitor transforms"), compiler, and interpreter/evaluator. -- @vgapeyev 

What does arbitrage mean in this scenario? Webster might be misleading me.

And can you elaborate on the special-casing of the `CURRENT_USER` token?

---

> I think this can be done for CURRENT_USER() (syntactically, a call to a zero-arity function), but not for CURRENT_USER (syntactically, an identifier lookup).

I don't think `CURRENT_USER` is syntactically an identifier lookup. According to the SQL-1992 BNF, it lies under the rule `<reserved word>` -- unrelated to `<identifier>`.

---

But, going back to the potential, future modification of the "GlobalEnvironment" -- I'm still unconvinced that we should be adding CURRENT_USER as a binding within the environment. SQL-1992 states:
> An SQL-session has an \<authorization identifier\> that is initially
         set to an implementation-defined value when the SQL-session is
         started, unless the SQL-session is started as a result of suc-
         cessful execution of a \<connect statement\>, in which case the
         \<authorization identifier\> of the SQL-session is set to the value
         of the implicit or explicit \<user name\> contained in the \<connect
         statement\>.

SQL-1992 also has a statement dedicated to setting the session's current user: `<set session authorization identifier statement>`

SQL is quite explicit about the current user's placement -- within the session itself. And since we have direct access to the session itself with ExprFunction, it still makes the most sense to me to leverage the use of built-in system functions.

vgapeyev:
Looking at the new iteration of the PR, does it maybe look more attractive now to implement SessionAttribute/CurrentUser AST node directly?  That is, by EvaluatingCompiler.compileCurrentUser method rather than by the translation to a built-in function call?   I believe a handful of fragments from the current implementation would just go away (SystemBuiltinsSql, transforms), which include some details that ought to be considered unsightly (error guarding against encountering the translated away node; obscure $__current_user function name; some references to CURRENT_USER_KEY and FUNCTION_NAME that really contribute to spaghetti dependencies).  

The concern that CURRENT_USER has the proper (string) type would be ensured with the direct approach as well as not better than with any other -- when it is a dedicated language form, one can impose on it what one wishes!

partiql-lang/src/main/pig/partiql.ion
```diff
@@ -76,6 +76,9 @@ may then be further optimized by selecting better implementations of each operat
 
         // The expressions that can result in values.
         (sum expr
+            // SQL Session Keywords (CURRENT_USER, CURRENT_ROLE, etc.)
```
am357, line 79: I think it makes sense to disallow `CURRENT_USER` as a keyword (E.g. when used as `SELECT "CURRENT_USER"` ...).

All the following are error in PostgresSQL 15:

```
CREATE TABLE MyTable (
  current_user VARCHAR(20)
);

CREATE TABLE MyTable (
  CURRENT_USER VARCHAR(20)
);

CREATE TABLE CURRENT_USER (
  id VARCHAR(20) 
);

```

Ref. https://www.db-fiddle.com/f/dsYWqRGdH4UDoPSHV5wKfC/0

As also discussed internally—echoing my view and also @vgapeyev's—we can just parse directly to a `Call` node.

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransformTests.kt
```diff
@@ -272,6 +273,41 @@ class AstToLogicalVisitorTransformTests {
...
+                            ExprFunctionCurrentUser.FUNCTION_NAME,
+                            emptyList()
+                        )
+                    )
+                }
+            ),
+            TestCase(
+                // Note:
+                // `SELECT * FROM bar AS b` is rewritten to `SELECT b.* FROM bar as b` by [SelectStarVisitorTransform].
+                // Therefore, there is no need to support `SELECT *` in `AstToLogicalVisitorTransform`.
```
vgapeyev, line 293: I doubt this comment is useful in these tests.

partiql-lang/src/test/kotlin/org/partiql/lang/syntax/PartiQLParserTest.kt
```diff
@@ -920,6 +920,36 @@ class PartiQLParserTest : PartiQLParserTestBase() {
...
+    fun currentUserUpperCase() = assertExpression(
+        "CURRENT_USER",
+        "(session_attribute CURRENT_USER)"
+    )
+
+    @Test
+    fun currentUserMixedCase() = assertExpression(
+        "CURRENT_user",
+        "(session_attribute CURRENT_user)"
+    )
```
vgapeyev, line 933: Not sure if I want to keep this can of worms open -- so only if you'd like...

Is there a good reason to preserve the case of the keyword as it was entered?  
This does not happen for other keywords (SELECT, WHERE).
I understand `session_attribute` as a generic implementation to use instead of several more concrete AST nodes for CURRENT_USER and friends. With the latter, case would not be preserved either.

vgapeyev (approved):
Approved, assuming some attention to a couple of comments.

vgapeyev (approved):

PR #1055, opened 2023-04-25
Adds asAlias property to INSERT AST node

### LINKED ISSUE #1043: INSERT with alias loses original table name which is replaced with the alias
## Description

Two issues:

- We don't appear to be modeling aliases correctly in the `dml_op` type of `pariql_ast` since there doesn't appear to be a place for the original name of the table *and* its alias.
- When `AS <alias>` is included following `INSERT INTO <table>`, the original table name is lost, having been replaced with the alias.

For example:

```
Welcome to the PartiQL shell!
Typing mode: LEGACY
Using version: 0.9.2-f3b92a82
PartiQL> insert into foo as f << { 'id': 1 } >>
   | !!
(dml 
    (operations
        (dml_op_list
        (insert
            (id f (case_insensitive) (unqualified))
            //  ^ wut?
            (bag 
                (struct
                    (expr_pair
                        (lit "id")
                        (lit 1)
                    )
                )
            )
        )
    )
  )
)

```
## Expected Behavior

- Both the original table name and alias are included in the AST at reasonable locations.

## Additional Context
- Java version: XXX
- PartiQL version: 0.9.2-f3b92a82
- Add any other context about the problem here.

johnedquinn:
## Relevant Issues
- Closes #1043 

## Description
- **Breaking**: Adds `asAlias` nullable property to the `insert` AST node.
- Modifies tests corresponding to the change
- This PR does NOT add the alias to the DML Conflict Action -- as the current planner does NOT support any action besides EXCLUDED -- which does not reference the alias. This will need to be addressed once we extend support to the conflict action value to include SET/VALUE according to the [RFC](https://github.com/partiql/partiql-docs/blob/main/RFCs/0011-partiql-insert.md).

## Other Information
  - Yes, the public AST node `insert` now contains another property called `asAlias`

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,9 +27,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Breaking**: Adds a new property `asAlias` to the `insert` AST node.
+
 ### Deprecated
 
 ### Fixed
+
+- Parsing INSERT statements with aliases no longer loses the original table name. Closes #1043.
 
 ### Removed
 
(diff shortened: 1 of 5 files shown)


johnedquinn:
Closing, as #1061 also resolves this issue.

Issue #825, opened 2022-10-04
Optimize Evaluation Performance When Subqueries Has Aggregate Functions In `SELECT` Clause 


lziq:
Assume that a table `data` has 2 integer columns, which are `id` & `num`. `n` is the number of rows it has. In the following query, 
```
SELECT t.id, (SELECT max(num) FROM data) AS x FROM data AS t
```
for the subquery `SELECT max(num) FROM data`, currently, PartiQL repeatedly evaluate it as many times as `n`. So totally, the inner `SELECT` clause is evaluated for `O(n^2)` time, which is a huge time cost. The point is that we all know that the subquery `SELECT max(num) FROM data` returns a constant and it does not change as `t` varies. Thus, we are able to design some mechanism to optimize the performance of this query, probably by using a cache or transform the subquery into a constant. 

This issue can cause serious problems to PartiQL users since it is very common for users to write aggregate functions in subqueries, e.g. if they want to select the `id` where `num` is maximum, they can write the following query:
```
SELECT t.id 
FROM data AS t 
WHERE t.num = (SELECT max(num) FROM data) 
```
However, the `O(n^2)` time cost will make it probably impossible to use PartiQL to run this query at users' side.

vgapeyev:
While true for these specific queries, the situation is more subtle in general. 
Consider slight variations of the above queries, 
```
SELECT t.id, (SELECT max(num) FROM t) AS x 
FROM data AS t

SELECT t.id 
FROM data AS t 
WHERE t.num = (SELECT max(num) FROM t.s)
```
where the aggregation subquery is dependent on each tuple (t here) in the data source (data here) rather than on the data source as a whole.  In that case, the proposed optimization would not be valid. 

So, in addition to how implement the memoization, a prior question is how to determine eligibility for it and limit the optimization to eligible situations only.
Further, thee is nothing special about SELECT subqueries as the subject of seeking an optimization like this: it can be applicable to other complex expressions.  Also, re the "host" query, there is nothing special about SQL-style SELECT either: the concern equally applies to SELECT VALUE and PIVOT.   

Properly, this problem should be addressed by algebraic optimization, where an unnecessarily-repeated computation would be moved outside a loop on whose iteration variables it does not depend.  

Having said the above, after analyzing the current code, it does appear possible to hack-in the memoization optimization, in a very limited scope, essentially only for the example in the ticket.

vgapeyev:
[...]

PR #1059, opened 2023-04-28
A couple of touch-ups in SELECT compilation; free variables 

vgapeyev:
## Relevant Issues
- Related To Issue #825

## Description
A couple of touch-ups in SELECT compilation; free variables 

Changes in this PR were made in course of working on the "headstrong" optimization in SELECT (PR xxx), but could be seen useful independently. 
Each change is in a separate commit. 
`freeVariables()` is unused as here (except for tests), but is used in the following PR xxx.  It should find use in more places, hence it is in this PR.

## Other Information
  - Internal changes.

## License Information


partiql-lang/src/main/kotlin/org/partiql/lang/eval/PartiqlAstExtensions.kt
```diff
@@ -65,3 +65,114 @@ internal fun PartiqlAst.Expr.getStartingSourceLocationMeta(): SourceLocationMeta
...
+                    val binders = setOfNotNull(n.asAlias?.text, n.atAlias?.text, n.byAlias?.text)
+                    (accum - binders) + walkExpr(n.expr, emptySet())
+                }
+            }
+
+        // Similarly to FROM, accounts for, e.g.,  LET expr as x, 2*x+1 as z   -- 2nd x is not free.
+        private fun addFreeOfLets(ns: List<PartiqlAst.LetBinding>, accum: Set<String>): Set<String> =
+            ns.foldRight(accum) { n, acc ->
+                (acc - n.name.text) + walkExpr(n.expr, emptySet())
+            }
```
johnedquinn, line 174: Can you help me clarify the expected behavior when an identifier is used within the original expression and its textual format is the alias?

Consider:
```sql
SELECT a
FROM t AS t
LET t AS t
```

In the visitor, we would:
1. Evaluate PROJECT. We'd accumulate `set(a)`.
2. Evaluate LET: We'd accumulate: `(acc - n.name.text) + walkExpr(n.expr, emptySet())` == `(set(a) - t) + set(t)` == `set(a) + set(t)` == `set(a, t)`. Unsure if this is expected.
3. Evalute SCAN: `(accum - binders) + walkExpr(n.expr, emptySet())` == `(set(a, t) - set(t)) + set(t)` == `set(a) + set(t)` == `set(a, t)`. Also unsure if this is expected.

partiql-lang/src/test/kotlin/org/partiql/lang/eval/PartiqlAstExtensionsTests.kt
```diff
@@ -43,4 +43,76 @@ class PartiqlAstExtensionsTests : EvaluatorTestBase() {
...
+
+            VarsTestCase("SELECT t.a as x, t.b as y, s as z  FROM Tbl as t at i", setOf("Tbl", "s")),
+            VarsTestCase("SELECT t, x FROM Tbl t, t as x", setOf("Tbl")),
+
+            VarsTestCase("SELECT s, t, x*x, y as z FROM Tbl as t LET t.a + t.b as x, t.a-z as y", setOf("Tbl", "s", "z")),
+            VarsTestCase("SELECT 5                 FROM Tbl as t LET t.a + t.b as x, t.a-z as y", setOf("Tbl", "z")),
+            VarsTestCase("SELECT s, t, x*x, y as z FROM Tbl as t LET t.a + t.b as x, 2*x   as y", setOf("Tbl", "s")),
+
+            VarsTestCase("        SELECT t, x, y, z FROM t as x, Foo as y", setOf("t", "Foo", "z")),
+            VarsTestCase("SELECT (SELECT t, x, y, z FROM t as x, Foo as y) as s FROM Tbl as t", setOf("Tbl", "Foo", "z")),
```
johnedquinn, line 115: This is an interesting scenario -- could you also include a subquery within the FROM? Maybe:
```
SELECT x
FROM (
    SELECT t, x, y, z FROM t as x, Foo as y
) AS t
```

This is similar to my other question. With the PR as it is, is `t` expected to be labeled as a free variable?

partiql-lang/src/main/kotlin/org/partiql/lang/eval/PartiqlAstExtensions.kt
```diff
@@ -65,3 +65,114 @@ internal fun PartiqlAst.Expr.getStartingSourceLocationMeta(): SourceLocationMeta
...
+
+        // Accounts for situations when an earlier binder is referenced in a later FROM item, such as t in 2nd item in
+        //   FROM Tbl as t, t as x
+        // Such a reference should not be counted as free.
+        private fun addFreeOfFrom(n: PartiqlAst.FromSource, accum: Set<String>): Set<String> =
+            when (n) {
+                is PartiqlAst.FromSource.Join -> {
+                    val acc = addFreeOfFrom(n.right, accum)
+                    addFreeOfFrom(n.left, acc)
+                }
```
johnedquinn, line 159: Should we search the JOIN condition for free variables?

johnedquinn (commented):
Leaving a couple (potentially confused) questions to better understand free variables. Appreciate any information you can give.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/PartiqlAstExtensions.kt
vgapeyev: Yes, this is as expected.  I have added a couple tests inspired by this example.

Intuitively, "free variables" of an expression are "all references to variables whose definitions are external/global w.r.t. this expression". 

For the sake of discussion, let's mark each occurrence of `t` in the example (but still think of it as the same variable):

```sql
SELECT a
FROM t1 AS t2
LET t3 AS t4
```

Then, what we want is for `freeVars` is to reflect the fact that the "externals" are `a` and `t1` -- which it does.  For the remaining, t2, t3, t4 do not enter the final free variables set: `t3` is a reference to the binder `t2`, and `t4` is a binder itself (available to be referenced in SELECT, but isn't being used there).

partiql-lang/src/main/kotlin/org/partiql/lang/eval/PartiqlAstExtensions.kt
vgapeyev: Yes, this was an oversight; good catch!

vgapeyev (commented):
Addressed the feedback so far.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/PartiqlAstExtensions.kt
```diff
@@ -65,3 +65,115 @@ internal fun PartiqlAst.Expr.getStartingSourceLocationMeta(): SourceLocationMeta
...
+            // Note: in `SELECT e AS x ...`, free variables of `e` get into the result,
+            // but `x` *does not*, due to `x` being a 'symbol', not an 'id' in the AST.
+            // (For reference, 'x' *should not* count as either free or bound because it is not a variable,
+            // it is a spec to create an attribute in a struct.)
+            current += walkProjection(node.project, emptySet())
+
+            node.limit?.let { current += walkExpr(it, emptySet()) }
+            node.offset?.let { current += walkExpr(it, emptySet()) }
+
+            node.order?.let { current += walkOrderBy(it, emptySet()) }
```
yliuuuu, line 128: Not sure if the intention is to call all the free variable function on the normalized AST only or not. For Order by, if you run on a unnormalized ast, the following two are false. On a normalized ast, the following are true. Would be great if you can clarify your intension in the function comments (and the test cases since they are not running on normalized ast). 
```
VarsTestCase("SELECT t.x + 1 as y FROM Tbl t ORDER BY y", setOf("Tbl"))
VarsTestCase("SELECT x + 1 as y FROM Tbl t ORDER BY y",setOf("Tbl", "x"))
```

yliuuuu (commented):
Has a small comments regarding the usage of the `freeVariables()` extension function, otherwise looks good to me.

vgapeyev (commented):
Responding to Yingtao's feedback.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -3159,21 +3137,19 @@ private sealed class ProjectionElement
...
- *
- * For `SELECT * FROM foo, bar, bat`, `exprThunks` would contain a compiled expression for each of `foo`, `bar` and
- * `bat`.
+ * Represents a wildcard projection into the final result.
+ * For `SELECT x.* FROM foo as x`, [thunk] contains the compiled expression for 'x'.
+ * Note that this does not cover `SELECT * FROM ...`, because it is assumed to have been translated
+ * into `SELECT x.*, y.*, z.* FROM ... `.
  */
-private class MultipleProjectionElement(val thunks: List<ThunkEnv>) : ProjectionElement()
+private class MultipleProjectionElement(val thunk: ThunkEnv) : ProjectionElement()
```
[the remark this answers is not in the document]
yliuuuu: Oh sorry this is just my own refreshing note.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/PartiqlAstExtensions.kt
```diff
@@ -65,3 +65,115 @@ internal fun PartiqlAst.Expr.getStartingSourceLocationMeta(): SourceLocationMeta
...
+
+            node.having?.let { current += walkExpr(it, emptySet()) }
+
+            node.group?.let {
+                val freeInGroup = walkGroupBy(it, emptySet())
+                val boundByGroup =
+                    (it.groupAsAlias?.let { setOf(it.text) } ?: emptySet()) +
+                        it.keyList.keys.flatMap { it.asAlias?.let { setOf(it.text) } ?: emptySet() }
+                current = (current - boundByGroup) + freeInGroup
+            }
```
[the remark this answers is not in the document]
vgapeyev: Done now.  

Frankly, I hoped no one would notice -- the difference seems to matter only when one uses delimited identifiers, while the current implementation of case sensitivity does not seem to have a sane explanation.  Fixing the latter is better be out of scope for this PR.

yliuuuu (approved):
LGTM.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/PartiqlAstExtensions.kt
```diff
@@ -65,3 +66,137 @@ internal fun PartiqlAst.Expr.getStartingSourceLocationMeta(): SourceLocationMeta
...
+            // SELECT clauses are processed in the reverse of evaluation order, since that's how variable scoping extends.
+
+            // Note: in `SELECT e AS x ...`, free variables of `e` get into the result,
+            // but `x` *does not*, due to `x` being a 'symbol', not an 'id' in the AST.
+            // (For reference, 'x' *should not* count as either free or bound because it is not a variable,
+            // it is a spec to create an attribute in a struct.)
+            current += walkProjection(node.project, emptySet())
+
+            node.limit?.let { current += walkExpr(it, emptySet()) }
+            node.offset?.let { current += walkExpr(it, emptySet()) }
```
vgapeyev, line 149: TODO: `LIMIT` and `OFFSET` may need special treatment. 
Say, in a `LIMIT x+5`, the variable `x` makes sense only if it is bound outside the current SELECT query (i.e., `x` is bound globally or in an outer, containing query). 
So, this occurrence of `x` should be considered free for this SELECT, and therefore be counted in its feeeVariables() set. 

Of course, this is confusing if the id `x` is also bound in FROM.  So may want to introduce a warning about that?

PR #1060, opened 2023-04-28
Do not recompute iteration-stable items in a SELECT list

vgapeyev:
## Relevant Issues
- Closes Issue #825 

## Description

This introduces a fairly generalized and corrected version of the optimization suggested in #825, now formulated as: 

> When it can be determined that an item in a SELECT list does not depend on any of the iteration variables active in its context, it is sufficient to compute the item upon encountering it first during iterations and reuse the computed value in each subsequent encounter.

See in-code comments on how the optimization works.  

Making the optimization possible required:

* Computing the set of free variables of an expression.  See PR #1059.

* Providing information about globally-defined variables during compilation.  
Done by adding a GlobalsCheck class parameter to EvaluatingCompiler -- an object that can tell whether a given variable is defined in the global environment.  For now, it is implemented as a wrapper around an EvaluationSession.  
To make this addition work throughout the code base, the primary needed adjustment was in the execution order of creating a pipeline + calling compilation vs session creation -- a session now has to be created first.  
Overall, this addition can be seen as a precursor to properly introducing static environment into compilation.  

## Other Information
- Any backward-incompatible changes? YES
  - Adds a new argument to `EvaluatingCompiler`

## License Information


examples/src/main/java/org/partiql/examples/S3JavaExample.java
```diff
@@ -28,11 +28,7 @@
 import org.jetbrains.annotations.NotNull;
 import org.partiql.examples.util.Example;
 import org.partiql.lang.CompilerPipeline;
-import org.partiql.lang.eval.Bindings;
-import org.partiql.lang.eval.EvaluationSession;
-import org.partiql.lang.eval.ExprValue;
-import org.partiql.lang.eval.ExprValueExtensionsKt;
-import org.partiql.lang.eval.Expression;
+import org.partiql.lang.eval.*;
```
johnedquinn, line 31: Avoid the wildcard import.

examples/src/main/java/org/partiql/examples/EvaluationJavaExample.java
```diff
@@ -18,10 +18,7 @@
 import org.jetbrains.annotations.NotNull;
 import org.partiql.examples.util.Example;
 import org.partiql.lang.CompilerPipeline;
-import org.partiql.lang.eval.Bindings;
-import org.partiql.lang.eval.EvaluationSession;
-import org.partiql.lang.eval.ExprValue;
-import org.partiql.lang.eval.Expression;
+import org.partiql.lang.eval.*;
```
johnedquinn, line 21: Wildcard import.

EDIT -- Some of the example files got converted. Can you fix that?

partiql-lang/src/main/kotlin/org/partiql/lang/eval/GlobalsCheck.kt
```diff
@@ -0,0 +1,21 @@
+package org.partiql.lang.eval
+
+interface GlobalsCheck {
```
johnedquinn, line 3: From what I can see, `GlobalsCheck` does not need to be made public. Its only use outside of this library is within the CLI, and all we do is initialize a `GlobalsCheck` using an `EvaluationSession` and pass it to `CompilerPipeline.Builder`.

So, an alternative to this, is to make this class `internal` and have the `CompilerPipeline.Builder` also build the `internal GlobalsCheck` using the passed `EvaluationSession`.

Even going beyond visibility -- it seems like all we need access to is an `EvaluationSession's` globals. Why not just use the session's globals directly? The `CompilerPipeline` also exposes `globalTypeBindings` as well.

If we remove public visibility, half the files touched don't actually need to be in this PR. If we remove it entirely, we may only need to make slight modifications.

johnedquinn (changes requested):
I haven't gone through all of the code, as it seems like many files can be excluded from this PR given the recommendation to reduce visibility or remove the class entirely.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/GlobalsCheck.kt
```diff
@@ -0,0 +1,21 @@
...
+        fun of(session: EvaluationSession): GlobalsCheck = SessionGlobalsCheck(session)
+
+        val empty: GlobalsCheck = EmptyGlobalsCheck()
+    }
+
+    private class SessionGlobalsCheck(val session: EvaluationSession) : GlobalsCheck {
+        override fun isDefined(name: BindingName): Boolean {
+            return session.globals[name] != null
+        }
+    }
```
yliuuuu, line 16: Something about global binding name lookup during compilation time is troubling me, I feel like this should be done in the evaluation time.

vgapeyev: There is nothing inherently wrong with that.  Such information is commonly provided to compilation by what is commonly known as "static environment", for which `GlobalsCheck` is a stand-in -- until we decide to do that properly, outside of this PR. 

Note that `GlobalsCheck` is an interface and it exposes only static information to the compiler. While an instance of this interface can be created from a run-time object (as it is here), it does not have to be.  

Part of you unease could merely be because the instance-creation code here is in the same file as the interface.  This is certain sloppiness that I might address while working on ideas from John.

partiql-lang/src/test/kotlin/org/partiql/lang/eval/EvaluatingCompilerSelectHeadstrongTest.kt
```diff
@@ -0,0 +1,78 @@
+package org.partiql.lang.eval
+
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ArgumentsSource
+import org.partiql.lang.eval.evaluatortestframework.EvaluatorTestCase
+import org.partiql.lang.util.ArgumentsProviderBase
+
+/** Queries that demonstrate the "headstrong" optimization of SELECT items,
+ *  implemented in [EvaluatingCompiler.compileProjectItemSource]. */
+class EvaluatingCompilerSelectHeadstrongTest : EvaluatorTestBase() {
```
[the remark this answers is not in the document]
vgapeyev: I agree the tests do not demonstrate the effect of the optimization, but it happens during their execution.  If the optimization code is merged and runs in production, it's better to have some kind of test for it running in the suite rather than let examples get stale in a doc somewhere.

PR #1061, opened 2023-05-01
Adds support for DML INSERT's WHERE/ALIAS in Parser, AST, and Plans

### LINKED ISSUE #1038: Parser support for the WHERE clause with INSERT INTO ... ON CONFLICT REPLACE EXCLUDED
## Relevant Issue/Bug

Last year [we asked to support `INSERT INTO ... ON CONFLICT REPLACE EXCLUDED`](https://github.com/partiql/partiql-lang-kotlin/issues/704).  We would like to extend that request to include the optional `WHERE` clause mentioned in [RFC #11](https://github.com/partiql/partiql-docs/blob/main/RFCs/0011-partiql-insert.md).

## Requested Solution/Feature

The parser should be extended to support the optional `WHERE` clause, such that `INSERT` statements like this can be parsed:

```sql
INSERT INTO my_fine_table AS s 
    << {
        'id': $id,
        'some_field': $some_field,
        'last_modified': $last_modified
    } >>
WHERE s.last_modified <= $last_modified 
-- ^ new syntax is here.
```

This query has "upsert" semantics and includes a form of opportunistic locking.

## DoD (Definition of Done)

- The parser can parse this statement.
- The relevant nodes in the AST, logical and physical plans include `WHERE` clause.

### LINKED ISSUE #1043: INSERT with alias loses original table name which is replaced with the alias

### LINKED ISSUE #1052: Allow PartiQL applications to rewrite DML nodes into custom `project` operators
## Requested Solution/Feature

My team requires the PartiQL-Kotlin’s query planner API to be extended to allow us customization of how the resolved logical plan is transformed into a physical plan.  This allows us to rewrite `INSERT` statements into custom `bexpr.project` operators.

This is not easy to accomplish today because:

* The `bexpr.project` operator does not exist in any domain except the physical plan domain, but by the time a query is rewritten into a physical plan, the original `dml` nodes have already been replaced with a SFW style query.
* While it is possible to rewrite these SFW queries into the needed custom `bexpr.project` operator, this rewrite would be overly complicated, be very painful to create (excessive code) and have high maintenance overhead, since it now depends on the exact expressions used to in the SFW query's projection, which really should be considered an implementation detail.

This is the preferred solution as it has the least amount of work and is the most flexible:

* Make `LogicalResolvedToDefaultPhysicalVisitorTransform`  `public` and `open`.  This allows PartiQL applications to inherit from this to customize the transformation from logical to physical plan.
* Enhance query planner API to allow the application supply its own instance of `PartiqlLogicalResolvedToPartiqlPhysicalVisitorTransform`.

## Describe Alternatives

* Option B: (2nd choice, due to it being less flexible):
    * Enhance the query planner API to allow the application to supply a callback which can opt to rewrite certain resloved-logical plan relational operators into physical.
* Option C:
    * Move `bexpr.project` and `impl` to the resolved-logical domain.
    * Enhance the query planner API to allow custom passes over the resolved-logical plan.
      * Applications can then handle this translation in such a pass.

## DoD (Definition of Done)

- PartiQL applications can rewrite DML nodes into custom `project` operators without depending on PartiQL's implementation details.

johnedquinn:
## Relevant Issues
- Closes #1038
- Closes #1043 
- Closes #1052 
  - By adding support for the alias and condition in the physical plan, there is no longer a need to rewrite the DML statement into a projection operator -- as requested in 1052.

## Description
- Adds support for DML INSERT's WHERE/ALIAS in Parser, AST, and Plans

## Changes
- Adds support for ALIAS syntax
  - Adds a new property `as_alias` to the `insert` AST node.
  - Adds `target_alias` property to the `dml_insert`, `dml_replace`, and `dml_update` nodes within the
  Logical and Logical Resolved plans
  - Adds new physical plan field representing the register allocated to the index.
- Adds support for CONDITION in `ON CONFLICT DO [REPLACE|UPDATE] EXCLUDED WHERE <condition>`
  - Adds new property `condition` to the AST nodes of `do_replace` and `do_update`
  - Adds `condition` property to the `dml_replace` and `dml_update` nodes within the
  Logical and Logical Resolved plans
  - Adds new physical plan field representing the conflict condition.
- Removes `dml_query` from the physical plan and adds the `statement.dml` and `dml_operation` to the physical plan.
  - This allows users to rewrite the physical plan without using the "loose" contract that used to be `dml_query` -- which was effectively just a struct with key-value pairs.

**Note**: The pipeline currently doesn't support the physical plan for `ON CONFLICT DO UPDATE EXCLUDED`. This PR still provides the changes for this clause, but it does not complete the required physical plan changes to support full planning of this statement.

## Testing
- Added tests for the PartiQLParser, the AstToLogicalVisitorTransform, the LogicalToLogicalResolvedVisitorTransform, and the LogicalResolvedToDefaultPhysicalVisitorTransform
- Compilation of DML has changed from extracting the key-value pairs from the evaluated physical plan to immediately evaluating the `table_unique_id` and `rows` expressions. These cases are covered by the integration tests under `IntegrationTests.kt`.

## To Check
For the PR reviewer, you can use these features by starting up the experimental planner in the CLI and adding an empty table to the global environment:

Screenshot 2023-05-01 at 2 14 08 PM

Then, explain the `ast`, `logical`, `logical_resolved`, and `physical` plans for example queries. For example, here is the `logical_resolved`:

Screenshot 2023-05-01 at 2 15 47 PM

## Other Information
  - `partiql.ion` has modifications. See the CHANGELOG.

## License Information

### DIFF
--- a/partiql-lang/src/main/antlr/PartiQL.g4
+++ b/partiql-lang/src/main/antlr/PartiQL.g4
@@ -197,7 +197,7 @@ conflictAction
    [ WHERE <condition> ]
 */
 doReplace
-    : EXCLUDED;
+    : EXCLUDED ( WHERE condition=expr )?;
     // :TODO add the rest of the grammar
 
 /*
@@ -207,7 +207,7 @@ doReplace
    [ WHERE <condition> ]
 */
 doUpdate
-    : EXCLUDED;
+    : EXCLUDED ( WHERE condition=expr )?;
     // :TODO add the rest of the grammar
 
 updateClause
--- a/partiql-lang/src/main/antlr/PartiQL.g4
+++ b/partiql-lang/src/main/antlr/PartiQL.g4
@@ -161,6 +161,8 @@ removeCommand
 insertCommandReturning
     : INSERT INTO pathSimple VALUE value=expr ( AT pos=expr )? onConflictClause? returningClause?;
 
+// TODO: Update grammar to disallow the mixing and matching of `insert`, `insertLegacy`, `onConflict`, and `onConflictLegacy`
+//  See https://github.com/partiql/partiql-lang-kotlin/issues/1063 for more details.
 insertCommand
     : INSERT INTO pathSimple VALUE value=expr ( AT pos=expr )? onConflictClause?  # InsertLegacy
     // See the Grammar at https://github.com/partiql/partiql-docs/blob/main/RFCs/0011-partiql-insert.md#2-proposed-grammar-and-semantics
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,12 +28,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Adds support for SQL's CURRENT_USER in the AST, EvaluatingCompiler, experimental planner implementation, and Schema Inferencer.
   - Adds the AST node `session_attribute`.
   - Adds the function `EvaluationSession.Builder::user()` to add the CURRENT_USER to the EvaluationSession
+- Adds support for parsing and planning of `INSERT INTO .. AS <alias> ... ON CONFLICT DO [UPDATE|REPLACE] EXCLUDED WHERE <expr>`
 
 ### Changed
+
+- **Breaking**: Adds a new property `as_alias` to the `insert` AST node.
+- **Breaking**: Adds new property `condition` to the AST nodes of `do_replace` and `do_update`
+- **Breaking**: Adds `target_alias` property to the `dml_insert`, `dml_replace`, and `dml_update` nodes within the
+  Logical and Logical Resolved plans
+- **Breaking**: Adds `condition` property to the `dml_replace` and `dml_update` nodes within the
+  Logical and Logical Resolved plans
 
 ### Deprecated
 
 ### Fixed
+
+- Parsing INSERT statements with aliases no longer loses the original table name. Closes #1043.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Adds the AST node `session_attribute`.
   - Adds the function `EvaluationSession.Builder::user()` to add the CURRENT_USER to the EvaluationSession
 - Adds support for parsing and planning of `INSERT INTO .. AS <alias> ... ON CONFLICT DO [UPDATE|REPLACE] EXCLUDED WHERE <expr>`
+- Adds the `statement.dml` and `dml_operation` node to the experimental PartiQL Physical Plan.
 
 ### Changed
 
@@ -46,6 +47,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Parsing INSERT statements with aliases no longer loses the original table name. Closes #1043.
 
 ### Removed
+
+- **Breaking**: Removes node `statement.dml_query` from the experimental PartiQL Physical Plan. Please see the added
+  `statement.dml` and `dml_operation` nodes.
 
 ### Security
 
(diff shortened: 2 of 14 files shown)


dlurton:
I also think that `PartiQLResult.Insert`, and `PartiQLResult.Delete` (and related code) may need to be removed as well.  I see that `PartiQLResult.Explain` was added not to long ago, so probably it and `PartiQLResult.Value` can be removed as well.

johnedquinn:
I completely agree with retaining the structure of the DML within the physical plan and have consulted @am357 about it as well. I've just added recent changes to reflect our position. If you'll notice, I didn't remove the DML `PartiQLResults` -- the reason for this is that we can still retain its intended use by modifying how we compile the plan. See the changes in `PartiQLCompilerDefault` and `PhysicalPlanCompilerImpl`. Re-marking as Ready For Review.

dlurton (commented):
I did not review every detail--but I did bring in my DML rewrite WIP into this branch on my local machine to verify that this would be sufficient to meet my use case, and it looks like it is.  So, overall, this LGTM.  

However, I did accidentally discover that inclusion of the `VALUE <expr>` clause (which I'm pretty sure was a mistake from the past since it's redundant with `VALUES << <expr> >>`), causes an internal error caused by a `ClassCastException` within the ANTLR AST -> PartiQL AST visitor.   (We have to keep around the `VALUE` clause for backward compatibility...) 

i.e. this query will reproduce the issue: `INSERT INTO foo VALUE { 'id': 42 } ON CONFLICT DO REPLACE EXCLUDED WHERE TRUE`

johnedquinn:
@dlurton -- regarding the query for reproducing the issue. It seems that this actually has to do with the mixing-and-matching of the "legacy" INSERT command with the "current" ON CONFLICT clause. The grammar says:

```antlr
insertCommand
    : INSERT INTO pathSimple VALUE value=expr ( AT pos=expr )? onConflictClause?  # InsertLegacy
    | INSERT INTO symbolPrimitive asIdent? value=expr onConflictClause?           # Insert
    ;

onConflictClause
    : ON CONFLICT WHERE expr DO NOTHING                                           # OnConflictLegacy
    | ON CONFLICT conflictTarget? conflictAction                                  # OnConflict
    ;
```
While the grammar says we can mix-and-match the "legacy" and "current" commands/clauses, we actually don't allow it in the visitor. Ideally, this should be expressed more explicitly within our grammar. I spoke to @am357 , and it seems that we are maintaining the "legacy" insert and "legacy" on-conflict clauses for backward-compatibility.

Anyways, this issue is present in `main`, and I'll leave it out of the scope of this PR.

partiql-lang/src/main/antlr/PartiQL.g4
```diff
@@ -197,7 +197,7 @@ conflictAction
    [ WHERE <condition> ]
 */
 doReplace
-    : EXCLUDED;
+    : EXCLUDED ( WHERE condition=expr )?;
```
am357, line 200: Unrelated to this line: should we add a note in the Grammar (or anywhere else applicable) that we can't mix and match the `legacy` with the other?

am357 (approved):

Issue #998, opened 2023-03-07
Implement evaluation of a representative fragment of graph query


vgapeyev:
Start on implementing graph query evaluation, as defined in [RFC-0033](https://github.com/partiql/partiql-docs/blob/main/RFCs/0033-graph-query.md). 

Partial work for this issue is in PR #981, whose completion depends on Issues #996 and #997.

vgapeyev:
Closed with PR #1074

Issue #997, opened 2023-03-07
A mechanism to bring externally-defined graphs into the evaluation environment


vgapeyev:
Make it possible to bring into the evaluation environment bindings for externally-defined graphs, such as through the format indicated in #996.

In the specification, PartiQL evaluation is defined w.r.t. a *database environment*, while in the implementation there is not yet a flexible way for a user to configure such an environment. (We have an EvaluationSession API, which requires programming, and `global_env` facility in the CLI, which only supports external data originating as Ion.)

The specific need for this issue is limited to graphs, but this work can be useful as a vehicle for understanding how to bring external data in general.

vgapeyev:
Closed with PR #1064 and PR #1073.

Issue #996, opened 2023-03-07
A format to define graph data externally


vgapeyev:
Come up with a file/data format to describe a graph value and a reader for making the graph available in PartiQL for graph query evaluation, as an instance of the data model specified in [RFC-0025](https://github.com/partiql/partiql-docs/blob/main/RFCs/0025-graph-data-model.md) and partially implemented in PR #881.

vgapeyev:
Closed with PR#1062.

PR #1062, opened 2023-05-04
File format for external graphs

vgapeyev:
Introduces an ISL-based file format for defining graphs externally. 

## Relevant Issues
- This PR addresses #996 from the inter-related trio of #996, #997, #998.

## Description
The goal of this format is to offer *a* way to define graphs to be brought into PartiQL runtime for querying with GPML.  It specifically is not a goal to specify *the* embedding of graphs into Ion. 

At the moment, the format is not used anywhere in the PartiQL evaluation pipelines, but there are unit tests with graph examples conforming to the format. 

## Other Information

## License Information


partiql-lang/src/main/kotlin/org/partiql/lang/graph/ExternalGraphReader.kt
```diff
@@ -0,0 +1,37 @@
...
+class GraphValidationException(message: String) : RuntimeException(message)
+object ExternalGraphReader {
+
+    private val ion: IonSystem = IonSystemBuilder.standard().build()
+    private val iss = IonSchemaSystemBuilder.standard()
+        .addAuthority(getResourceAuthority(ion))
+        .withIonSystem(ion)
+        .build()
+    private val graphSchema: Schema = iss.loadSchema("graph.isl")
+    private val graphType = graphSchema.getType("graph")!!
```
am357, line 20: `graph.isl` and `graph` to constants?

partiql-lang/src/main/resources/org/partiql/schemas/graph.isl
```diff
@@ -0,0 +1,71 @@
...
+type::{
+    name: DirectedPair,
+    type: list,
+    ordered_elements: [ NodeId, NodeId ],
+    annotations: closed::[dir]
+}
+
+type::{
+    name: UndirectedPair,
+    type: sexp,
```
RCHowell, line 62: Because annotations have no semantic meaning, does that mean the distinction between directed and undirected pairs is simply the bracketing?
```ion
[ a, b ]  // directed
( a, b )  // undirected
```

This is a bit subtle and since both ion lists and s-expressions are ordered there isn't necessarily a relationship between directed and undirected to the orderedness of the structure representing the pair. Unfortunately Ion doesn't have an unordered collection, but I think using s-expression operators you could improve the visual clues

```ion
( a -> b )  // directed, also explicitly shows direction
( a -- b )  // undirected
```

This is valid and Ion and is consistent with [Graphviz DOT](https://graphviz.org/doc/info/lang.html)

> An edgeop is -> in directed graphs and -- in undirected graphs.

partiql-lang/src/test/resources/graphs/rfc0025-example.ion
```diff
@@ -0,0 +1,16 @@
...
+  nodes: [
+    { id: n1, labels: ["a"] },
+    { id: n2, labels: ["a"] },
+    { id: n3, labels: ["b"] }
+  ],
+
+  edges: [
+    { id: e1, labels: ["x"], ends: [n2, n1] },
+    { id: e2, labels: ["y"], ends: dir::[n1, n3] },
+    { id: e3, labels: ["y"], ends: [n2, n3] }
```
RCHowell, line 14: Example of previous comment
```
    { id: e1, _: (n2 -> n1), labels: ["x"] },
    { id: e2, _: (n1 -> n3), labels: ["y"] },
    { id: e3, _: (n2 -> n3), labels: ["y"] }

// vs

    { id: e1, labels: ["x"], ends: [n2, n1] },
    { id: e2, labels: ["y"], ends: dir::[n1, n3] },
    { id: e3, labels: ["y"], ends: [n2, n3] }
```

partiql-lang/src/main/resources/org/partiql/schemas/graph.isl
```diff
@@ -0,0 +1,71 @@
...
+    name: Edge,
+    annotations: closed::[edge],
+    type: struct,
+    fields: closed::{
+        id: { type: EdgeId, occurs: required },
+        labels: { type: list, element: string},
+        ends: { one_of: [DirectedPair, UndirectedPair], occurs: required },
+        payload: { type: EdgePayload, occurs: optional }
+    }
+}
```
RCHowell, line 51: This pattern of a product wrapping a single union type field has come up pretty often in protobuf, isl, pig, sprout, and I'm sure many more models. You have variants with some shared fields, so you create an outer product time called "common" and the variants are in something called "body". In this case we have `Edge` and `Ends variants DirectedPair, UndirectedPair`.

In a perfect world, we would define an Edge as a union type with common properties (id, labels, ends) amongst variants but we don't have the modeling power so we have to do this product-wrapped-union thing.

Purely aesthetics, but I've found the name for the "body" union type to not carry much meaning and can be dropped without suffering. In Ion, we can use `_` as an "empty" symbol, so your `ends` can be removed. Also moving required fields first. Then we get

```ion
type::{
    name: Edge,
    annotations: closed::[edge],
    type: struct,
    fields: closed::{
        id: { type: EdgeId, occurs: required },
        _: { one_of: [DirectedPair, UndirectedPair], occurs: required }
        labels: { type: list, element: string},
        payload: { type: EdgePayload, occurs: optional }
    }
}
```

And we can have edges like
```
edge::{ id: a, _: ( x -> y ) }
```

partiql-lang/src/main/resources/org/partiql/schemas/graph.isl
```diff
@@ -0,0 +1,71 @@
...
+    are general Ion.  That is, it is not specified how a fragment of a payload can be itself recognized as a graph.
+
+    A file contains a graph if it contains a single Ion struct of ISL type Graph, as defined here.
+
+    Annotations specified in this schema (graph, node, edge, dir, undir) are for mnemonic purposes only --
+    graph authors are not required to use them and processors should not attribute semantic meaning to their
+    presence or absence.
+*/
+
+type::{
```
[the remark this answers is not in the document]
vgapeyev: I am not sure what you mean here...  
Referring to the fact that nodes and edges are abstract entities in the data model and labels, payloads (as well as ends) are given as functions from node/edge sets to some other sets?  

I guess, some other designs for the format are possible, perhaps more closely reflecting the data model formulation (whatever would be the merits of that). 
This is a one I came up with.  It does not give too many opportunities for writing something wrong, if I may brag so :-)

partiql-lang/src/main/resources/org/partiql/schemas/graph.isl
```diff
@@ -0,0 +1,71 @@
...
+    name: graph,  // ExternalGraphValidationTests breaks when this is named Graph ... no idea
```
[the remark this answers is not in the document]
vgapeyev: Thank you!

partiql-lang/src/main/resources/org/partiql/schemas/graph.isl
vgapeyev: I like this idea!  

The `[ ]` vs `( )` was indeed too clever.  All I liked it for was avoiding a required heavy-weight discriminator.

partiql-lang/src/main/resources/org/partiql/schemas/graph.isl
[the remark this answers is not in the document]
am357: > Referring to the fact that nodes and edges are abstract entities in the data model and labels, payloads (as well as ends) are given as functions from node/edge sets to some other sets?
Yes, and not too opinionated about this since it should not represent our actual data model; I think the only merit would be for reader to get a better sense of our model via this representation.

partiql-lang/src/main/resources/org/partiql/schemas/graph.isl
vgapeyev: I feel like passing on the points here at the moment...

Overall, the schema is simple enough that factoring out the commonality between nodes and edges could only increase understanding burden for someone who needs to write a graph in a hurry, I feel.  

As for `ends` vs `_` -- I am inclined to be explicit, a.k.a "self-describing".

partiql-lang/src/main/resources/org/partiql/schemas/graph.isl
vgapeyev: Done.  

I went for an extra syntactic frill of allowing differing length: `<-` and `<--`, etc.

vgapeyev (commented):
I believe I addressed all the actionable comments in the coming 2 commits.

partiql-lang/src/main/resources/org/partiql/schemas/graph.isl
```diff
@@ -0,0 +1,81 @@
...
+type::{
+    name: DirectedPair,
+    type: sexp,
+    ordered_elements: [ NodeId, DirectedMarker, NodeId ]
+}
+
+type::{
+    name: DirectedMarker,
+    type: symbol,
+    valid_values: [ '->', '-->', '<-', '<--' ]
```
RCHowell, line 62: Why the three character `---` and `-->` rather than the two character ones? Do they mean something different? I'm inclined to keep it simple unless the three character variants have some particular meaning.

RCHowell (changes requested):

partiql-lang/src/main/resources/org/partiql/schemas/graph.isl
vgapeyev: Just meant to cater to different syntactic tastes.  Not important. Will stick with length-2.

am357 (approved):

RCHowell (approved):

PR #1064, opened 2023-05-05
A reader for external graphs

vgapeyev:
## Relevant Issues
- This is another PR towards the trio of issues #996, #997, #998.

## Description
Reads into memory graphs represented in Ion according to the format introduced in PR #1062.

The in-memory representation `SimpleGraph` is what was developed for the POC #981, with parts not needed here removed.   

## Other Information

## License Information


partiql-lang/src/main/kotlin/org/partiql/lang/graph/ExternalGraphReader.kt
```diff
@@ -34,4 +46,98 @@ object ExternalGraphReader {
...
+    }
+
+    /* Returns a map from node IDs in the source graph to newly-built in-memory nodes.
+    *  This map is used later to recognize node IDs used in edge definitions.
+    */
+    internal fun readNodes(nodesList: IonList): Map<String, SimpleGraph.Node> {
+        val pairs = nodesList.toList().map { readNode(it as IonStruct) }
+        val duplicates = pairs.map { it.first }.groupBy { it }.filterValues { it.size > 1 }.keys
+        if (duplicates.isNotEmpty()) throw GraphReadException(
+            "Identifiers used for more than one node: ${duplicates.joinToString()}."
```
RCHowell, line 82: Perhaps include the duplicate identifier names in the error message

partiql-lang/src/main/kotlin/org/partiql/lang/graph/ExternalGraphReader.kt
```diff
@@ -1,14 +1,25 @@
 package org.partiql.lang.graph
```
RCHowell, line 1: I believe all of the internal methods in this class could actually be private.

partiql-lang/src/main/kotlin/org/partiql/lang/graph/ExternalGraphReader.kt
```diff
@@ -34,4 +46,98 @@ object ExternalGraphReader {
...
+                else -> throw GraphReadException(
+                    "BUG: At edge $id, the entry at 'ends' field should have been validated as either IonList or IonSexp, but got: $ends"
+                )
+            }
+        }
+
+        internal fun endNode(seq: IonSequence, idx: Int): SimpleGraph.Node {
+            val id = (seq.get(idx) as IonSymbol).symbolValue().assumeText()
+            return allNodes.getOrElse(id) {
+                throw GraphReadException("Node id $id is used in an edge, but it was not introduced at a node")
```
RCHowell, line 139:
```suggestion
                throw GraphReadException("Node id $id is used in an edge, but it was not defined as a node")
```

RCHowell (changes requested):

partiql-lang/src/main/kotlin/org/partiql/lang/graph/Graph.kt
```diff
@@ -0,0 +1,32 @@
...
+    interface Elem {
+        val labels: Set<String>
+        val payload: ExprValue
+    }
+
+    interface Node : Elem
+    interface Edge : Elem
+    interface EdgeDirected : Edge
+    interface EdgeUndir : Edge
+}
```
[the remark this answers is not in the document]
vgapeyev: Sigh.  Rather than `Graph`, it's better to look at `SimpleGraph`, which exposes the edges and nodes.  

`Graph` is an artifact of connecting back to the computational engine in POC #981, so it has some "access methods", but different ones.  So I removed them out of this PR to avoid questions why they are not being used anywhere.  I was reluctant removing `Graph` completely, because that would require tweaking override dependencies between it and `SimpleGraph`, inviting more merge conflicts in the future.

partiql-lang/src/main/kotlin/org/partiql/lang/graph/ExternalGraphReader.kt
vgapeyev: `${duplicates.joinToString()}` does just that?

partiql-lang/src/main/kotlin/org/partiql/lang/graph/ExternalGraphReader.kt
vgapeyev: Good point -- except for the `read()` methods. 

The fact that those are `internal` is because of `SimpleGraph` in their return type, which I had designated internal so far.  I am not sure yet where this will end up eventually.  These read methods ought to be public, but I did not yet figure out how, to make that possible, to change their return type to a public type (`Graph` would work) while still being able to do tests in `ExternalGraphReaderTests` (which need methods from `SimpleGraph` that `Graph` does not have). I hope there is some king of a test/package wizardry that can reconcile this tension.

partiql-lang/src/main/kotlin/org/partiql/lang/graph/ExternalGraphReader.kt
```diff
@@ -1,14 +1,25 @@
 package org.partiql.lang.graph
 
+import com.amazon.ion.IonList
+import com.amazon.ion.IonSequence
+import com.amazon.ion.IonSexp
+import com.amazon.ion.IonString
+import com.amazon.ion.IonStruct
+import com.amazon.ion.IonSymbol
```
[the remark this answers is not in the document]
vgapeyev: Ok, we settled, in a separate exchange, that IonElement is the way of the future -- and it is in the future for the work here :-)

partiql-lang/src/main/kotlin/org/partiql/lang/graph/ExternalGraphReader.kt
vgapeyev: Changed on my side.

partiql-lang/src/main/kotlin/org/partiql/lang/graph/SimpleGraph.kt
```diff
@@ -0,0 +1,31 @@
...
+typealias Labels = Set<String>
+
+/** A straightforward implementation of in-memory graphs. */
+internal class SimpleGraph(
+    internal val nodes: List<Node>,
+    internal val directed: List<Triple<Node, EdgeDirected, Node>>,
+    internal val undir: List<Triple<Node, EdgeUndir, Node>> // order of Nodes in the Triple "doesn't matter"
+) : Graph {
+
+    // Intentionally, not a data class -- want to use pointer equality
```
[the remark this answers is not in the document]
vgapeyev: Hmmm...  I think the automatically-generated `equals` for data classes is an important part of what they promise, maybe even the most important part -- the assurance that if you have constructed two instances of a data class from the same pieces, the instances are equal -- for the purposes of putting them in sets, etc, even if the instances are referentially (pointer) distinct. 

So, even though it is possible to override `equals` in a data class and Kotlin manual implies that's ok, I think it is a bad idea as a matter of practice.  I frankly, do not see why this is not prohibited!

vgapeyev (commented):
I believe this responds to all the comments so far. A commit with suggested tweaks is on the way.  

There are conflicts with the upstream PR#1062 -- I'll resolve after that one itself settles.

partiql-lang/src/main/kotlin/org/partiql/lang/graph/SimpleGraph.kt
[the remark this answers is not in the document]
RCHowell: Nothing actionable here then, just different perspectives on what a data class is.

partiql-lang/src/main/kotlin/org/partiql/lang/graph/ExternalGraphReader.kt
RCHowell: As you've said, let's do private for everything except that which is exposed for tests being internal. And correct, public methods cannot return internal/private types.

partiql-lang/src/main/kotlin/org/partiql/lang/graph/ExternalGraphReader.kt
RCHowell: Scheduling an optometrist appointment now... 😆

partiql-lang/src/main/kotlin/org/partiql/lang/graph/Graph.kt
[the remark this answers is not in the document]
RCHowell: At the time of review, I couldn't unpack 

> The in-memory representation SimpleGraph is what was developed for the POC https://github.com/partiql/partiql-lang-kotlin/pull/981, with parts not needed here removed.

But I see what you're doing now. Then I will review the update to #981 on these matters. This is good here.

vgapeyev:
I am sorry for the mess with commits while bringing in changes from #1062 and additional local work.  

Instead of doing my usual straightforward merge, I attempted a rebase, trying my luck this time with a rebase on top of a local merge...  This ended up with a merge anyway, even a messier one. 
For the future, reserving rebases to straight-line commit sequences only.

partiql-lang/src/main/kotlin/org/partiql/lang/graph/ExternalGraphReader.kt
vgapeyev: Done.

vgapeyev:
Ok, after some mess with updating for upstream work in #1062, at least the diff now looks as intended and the last pending comment addressed.  

But I understand if reviewers would prefer to wait until #1062 is merged and this one is re-targeted to main.

partiql-lang/src/main/kotlin/org/partiql/lang/graph/ExternalGraphReader.kt
```diff
@@ -1,16 +1,26 @@
...
 import com.amazon.ionschema.Schema
+import org.partiql.lang.eval.ExprValue
 import org.partiql.lang.partiqlisl.getResourceAuthority
 import java.io.File
 
 class GraphValidationException(message: String) : RuntimeException(message)
+class GraphReadException(message: String) : RuntimeException(message)
 
-/** A validator for external graphs represented in Ion in accordance with the graph.isl schema.
+internal typealias EdgeTriple<E> = Triple<SimpleGraph.Node, E, SimpleGraph.Node>
```
RCHowell, line 21: Nit, internal so doesn't really matter but an upper bound on `E` would be nice.

I can't tell if `E` is for Edge or Element. Either way
```suggestion
internal typealias EdgeTriple<E : Edge> = Triple<SimpleGraph.Node, E, SimpleGraph.Node>
```

RCHowell (approved):
One nit

partiql-lang/src/main/kotlin/org/partiql/lang/graph/ExternalGraphReader.kt
vgapeyev: We think similarly! -- I tried this before.  Alas, "Bounds are not allowed on type alias parameters".

PR #1083, opened 2023-05-17
Fixes JOIN Predicate and Allows Duplicates in StructType

johnedquinn:
## Relevant Issues
- None

## Description
- Fixes JOIN predicate by adding the correct bindings to the input to the predicate for typing
- Modifies StructType attribute `fields` to be a List<Field>. This allows for ordering of the Struct as well as duplicate keys.
- Adds attribute, `isOrdered`, to StructType. As the PartiQL Spec allows for ordered tuples in the presence of schema, I opted to add it directly to StructType.
  - For referencing attributes of an ORDERED StructType, the behavior is well-defined in grabbing the first attribute:
    - > When the tuple t has multiple attributes a, the tuple path navigation t.a will return the first instance of a. Note that for tuples whose order is defined by schema, this is well-defined. For unordered tuples, it is implementation defined which attribute is returned in permissive mode or an error in type checking mode, which is described in Section 4.1.
    --- PartiQL Spec Page 13
  - For referencing attributes of an UNORDERED StructType, it is implementation-defined.
- Removed the ambiguity with referencing `a` in the following:
  ```
  SELECT * FROM << { 'a': 0 } >> AS t JOIN << { 'a': true } >> ON a = false
  ```

## Testing

- Made sure all tests pass in StaticTypeTests which tests `isInstance` for Structs. Added tests relevant to ordering.
- Added tests for the additions in the PartiQLSchemaInferencer
- Modified the LocalPlugin to return all structs to be ORDERED. Reflected those changes in the tests.

## Other Information
  - See CHANGELOG.

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     externally-defined graphs into the evaluation environment. 
   - Evaluation of straight-path patterns with simple label matching and 
     all directed/undirected edge patterns.
-- Adds new attribute, `isOrdered`, to StructType. See the KDoc for more information.
+- Adds new `TupleConstraint` variant, `IsOrdered`, to represent ordering in `StructType`. See the KDoc for more information.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     externally-defined graphs into the evaluation environment. 
   - Evaluation of straight-path patterns with simple label matching and 
     all directed/undirected edge patterns.
+- Adds new attribute, `isOrdered`, to StructType. See the KDoc for more information.
 
 ### Changed
 
@@ -46,7 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fixes the ability for JOIN predicates to access the FROM source aliases and columns.
+- Fixes the ability for JOIN predicates to access the FROM source aliases and corresponding attributes.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     externally-defined graphs into the evaluation environment. 
   - Evaluation of straight-path patterns with simple label matching and 
     all directed/undirected edge patterns.
-- Adds new `TupleConstraint` variant, `IsOrdered`, to represent ordering in `StructType`. See the KDoc for more information.
+- Adds new `TupleConstraint` variant, `Ordered`, to represent ordering in `StructType`. See the KDoc for more information.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,9 +38,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Breaking**: The `fields` attribute of `org.partiql.types.StructType` is no longer a `Map<String, StaticType>`. It is
+  now a `List<org.partiql.types.StructType.Field>`, where `Field` contains a `key (String)` and `value (StaticType)`. This
+  is to allow duplicates within the `StructType`.
+
 ### Deprecated
 
 ### Fixed
+
+- Fixes the ability for JOIN predicates to access the FROM source aliases and columns.
 
 ### Removed
 
--- a/partiql-lang/src/main/kotlin/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
+++ b/partiql-lang/src/main/kotlin/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
@@ -1288,7 +1288,7 @@ internal class StaticTypeInferenceVisitorTransform(
 
         override fun transformExprStruct(node: PartiqlAst.Expr.Struct): PartiqlAst.Expr {
             val struct = super.transformExprStruct(node) as PartiqlAst.Expr.Struct
-            val structFields = mutableListOf<Pair<String, StaticType>>()
+            val structFields = mutableListOf<StructType.Field>()
             var closedContent = true
             struct.fields.forEach { expr ->
                 val nameExpr = expr.first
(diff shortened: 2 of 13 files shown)


partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
```diff
@@ -714,9 +715,98 @@ class PartiQLSchemaInferencerTests {
...
+                expected = BagType(
+                    StructType(
+                        fields = listOf(
+                            StructType.Field("a", unionOf(INT, STRING))
+                        ),
+                        contentClosed = true,
+                        constraints = setOf(TupleConstraint.Open(false), TupleConstraint.UniqueAttrs(true))
+                    )
+                )
+            ),
```
yliuuuu, line 797: Spec Page 4: 
> Unlike SQL, PartiQL allows the possibility of duplicate attribute names, in the interest of compatibility with nonstrict JSON/Ion datasets. However PartiQL does not encourage duplicate attribute names; navigation into tuples with the conventional dot notation (Section 4) is tuned to the assumption that the attribute names are unique.


Page 13:
> Tuple navigation in tuples with duplicate attributes When the tuple t has multiple attributes a, the tuple path navigation t.a will return the first instance of a. Note that for tuples whose order is defined by schema, this is well-defined, for unordered tuples, it is implementation defined which attribute is returned in permissive mode or an error in type checking mode, which is described in Section 4.1.

yliuuuu: If the intention here is to say that the plan will delegate the decision as which "a" to output(based on one of the comments above), then I personally think it is OK. But maybe worth to document this decision more formally.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
```diff
@@ -1310,15 +1310,15 @@ internal class StaticTypeInferenceVisitorTransform(
...
 
             val hasDuplicateKeys = structFields
-                .groupingBy { it.first }
+                .groupingBy { it.key }
                 .eachCount()
                 .any { it.value > 1 }
 
             if (hasDuplicateKeys) {
                 TODO("Duplicate keys in struct is not yet handled")
             }
```
yliuuuu, line 1319: Are we planning on fixing the staticTypeInferencer in the future?

partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
```diff
@@ -479,17 +479,42 @@ data class StructType(
...
+    public constructor(
+        fields: Map<String, StaticType>,
+        contentClosed: Boolean = false,
+        primaryKeyFields: List<String> = listOf(),
+        constraints: Set<TupleConstraint> = setOf(),
+        metas: Map<String, Any> = mapOf(),
+    ) : this(
+        fields.map { Field(it.key, it.value) },
+        contentClosed,
+        primaryKeyFields,
```
yliuuuu, line 492: Why is declaring `primaryKeyFields` and `contentClosed` limited to field being a Map?

CHANGELOG.md
```diff
@@ -38,10 +38,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+- **Breaking**: The `fields` attribute of `org.partiql.types.StructType` is no longer a `Map<String, StaticType>`. It is
+  now a `List<org.partiql.types.StructType.Field>`, where `Field` contains a `key (String)` and `value (StaticType)`. This
+  is to allow duplicates within the `StructType`.
+
 ### Deprecated
 
 ### Fixed
 
+- Fixes the ability for JOIN predicates to access the FROM source aliases and columns.
```
am357, line 49: `columns` -> `attr`?

partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
am357: I think it's a good decision to contain `primaryKeyFields` and `contentClosed` to the previous public interface mainly b/c I'm of the opinion that they don't belong in `StructType` at all and this is a technical debt.

partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
```diff
@@ -479,17 +479,42 @@ data class StructType(
...
+        constraints,
+        metas
+    )
+
+    /**
+     * The key-value pair of a StructType, where the key represents the name of the field and the value represents
+     * its [StaticType]. Note: multiple [Field]s within a [StructType] may contain the same [key]. Therefore, when referencing
+     * a field of a [StructType], it is important to gather all fields of the same [key]. To determine the [StaticType]
```
am357, line 500: > Note: multiple [Field]s within a [StructType] may contain the same [key]. Therefore, when referencing
     * a field of a [StructType], it is important to gather all fields of the same [key]. To determine the [StaticType]

Maybe here we just have to mention that this allows having multiple identical keys different types in the struct as specified by the spec, etc.

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
am357: I think it makes sense to have the union of the types as part of logical plan. Basically there can be two possibilities:
1. Either we don't have a schema for the struct value which in that case there are no fields in the struct type.
2. We have a schema with duplicate keys and in that case, it's at least helpful to know what are the possible types for evaluation.

Considering this, @yliuuuu I can still see your concern with defining this more formally as a spec. (how we type our AST and represents the types in the plan) — maybe we can just open an issue in `partiql-spec` repo to track this?

partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
```diff
@@ -475,6 +485,7 @@ data class StructType(
     // TODO remove `contentClosed` and `primaryKeyFields` if after finalizing our type specification we're
     // still going with `StructType`.
     val contentClosed: Boolean = false,
+    val isOrdered: Boolean = false,
```
am357, line 488: Why this hasn't been modeled as a constraint?

partiql-lang/src/main/kotlin/org/partiql/lang/eval/visitors/StaticTypeInferenceVisitorTransform.kt
johnedquinn: I opted to defer as I don't know the future of the StaticTypeInferencer.

am357 (approved):

PR #1086, opened 2023-05-19
Moves PIG AST & Plans

johnedquinn:
## Relevant Issues
- None

## Description
- Moves `partiql.ion` into a new project, `partiql-ast`
- Updates all references
- Note: Now that the classes are outside of `lang` there were many instances of "smart casts" and non-null assertions that needed to be resolved. Hence, those small, functionally-equivalent changes.

## Other Information
  - See CHANGELOG

## License Information


partiql-ast/build.gradle.kts
```diff
@@ -0,0 +1,46 @@
...
+    id(Plugins.pig)
+}
+
+dependencies {
+    api(Deps.pigRuntime)
+    api(Deps.ionElement)
+}
+
+pig {
+    namespace = "org.partiql.ast.domains"
```
RCHowell, line 28: I think we should keep the package as `org.partiql.lang.domains` that way this is not a breaking change. The goal is to cleanly split artifacts of partiql-lang-kotlin for independent consumption. Ideally doing so without repackaging (ie breaking things) where possible.

partiql-ast/src/main/kotlin/org/partiql/ast/domains/AsAnyElement.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import com.amazon.ionelement.api.IonElement
+
+// FIXME: Workaround for a bug in PIG that is fixed in its next release:
+//  https://github.com/partiql/partiql-ir-generator/issues/41
+// FIXME: This does NOT conform to PartiQL's standards of forbidding top-level functions/variables. However,
+//  given the issue above, this is required. To reduce some of the grossness of top-level functions, we have specified
+//  the file's JvmName further above.
+internal fun List<IonElement>.asAnyElement() =
+    this.map { it.asAnyElement() }
```
RCHowell, line 13: I can do a PIG release today so we don't have to keep this around.

RCHowell (changes requested):

partiql-ast/src/main/kotlin/org/partiql/ast/domains/AsAnyElement.kt
johnedquinn: Works for me! LMK when you do and I'll remove.

RCHowell: And PIG plugin release since we are actually a consumer of the plugin .. which is a consumer of PIG.

RCHowell: I believe we can make these changes now because moving these domains without repackaging is not a breaking change.

johnedquinn: Alright, with the plugin release, I've now set the target to `main` and have removed the utility function.

RCHowell:
Can you confirm that the partiql.ion file makes it in the `partiql-ast.jar`? As we have customers which extract this

johnedquinn:
> Can you confirm that the partiql.ion file makes it in the `partiql-ast.jar`? As we have customers who extract this

Screenshot 2023-05-23 at 3 24 51 PM

RCHowell (approved):

PR #1096, opened 2023-05-23
Update partiql-tests to the newest; related doc tweaks.

vgapeyev:
This checks out the git submodule for the most recent version of partiql-tests conformance suite.  
According to a manually-run comparison report, all new failures (and successes) are due to the the new tests that got added to the conformance suite since last refresh.

Also, some documentation tweaks -- see specific comments. 

In addition to the needed improvement with enabling/disabling the conformance suite runs, it would be helpful to get a mechanism for running smaller subsets of the suite (at least being able to limit to a directory), especially when running with IntelliJ UI.

A yet another issue is `DefaultConformanceTestRunner` that runs conformance tests during regular builds, w.r.t. a list of known failures.  After partiql-tests refresh this list became significantly out-of-date. 
This PR disables DefaultConformanceTestRunner for the time being.  The issue might be more significant than deferred maintenance in updating the fail list -- we need to figure out more comfortable / useful workflow with conformance tests overall. 

## Other Information
  - internal

## License Information

### DIFF
--- a/test/partiql-tests-runner/README.md
+++ b/test/partiql-tests-runner/README.md
@@ -7,9 +7,6 @@ This package enables:
 1. Verifying conformance tests are defined correctly (e.g. verify evaluation environment is not missing a table)
 2. Identifying areas in the Kotlin implementation that diverge from the [PartiQL Specification](https://partiql.org/assets/PartiQL-Specification.pdf)
 
-Eventually, the `partiql-test-runner` module will replace the `partiql-pts` and `partiql-testscript` modules along with some other 
-tests in `lang` that were ported to `partiql-tests` 
-(see [partiql-lang-kotlin#789](https://github.com/partiql/partiql-lang-kotlin/issues/789)).
 
 ## Run Conformance Tests Locally
 
@@ -21,3 +18,18 @@ tests in `lang` that were ported to `partiql-tests`
 PARTIQL_TESTS_DATA=/path/to/partiql-tests/data \
 ./gradlew :test:partiql-tests-runner:test --tests "*ConformanceTestsReportRunner" -PconformanceReport
 ```
+The report is written into file `test/partiql-tests-runner/conformance_test_results.ion`.
+
+## Run Conformance Tests in IntelliJ
+
+The above project property `-PconformanceReport` is checked in `test/partiql-tests-runner/build.gradle.kts`,
+to prevent the conformance test suite from executing during a normal project-build test run. 
+Unfortunately, this also disables running `ConformanceTestsReportRunner` via IntelliJ UI for unit tests. 
+To make that possible locally, temporarily comment out the check in `test/partiql-tests-runner/build.gradle.kts`.
+
+## Compare Conformance Reports locally
+
+```shell
+./gradlew :test:partiql-tests-runner:run --args="pathToFirstConformanceTestResults pathToSecondConformanceTestResults firstCommitId secondCommitId pathToComparisonReport"
+```
+The last argument, `pathToComparisonReport` is for the `.md` file into which to write the comparison report.
--- a/test/partiql-tests-runner/README.md
+++ b/test/partiql-tests-runner/README.md
@@ -20,11 +20,11 @@ PARTIQL_TESTS_DATA=/path/to/partiql-tests/data \
 ```
 The report is written into file `test/partiql-tests-runner/conformance_test_results.ion`.
 
-## Run Conformance Tests in IntelliJ
+## Run Conformance Tests in UI
 
 The above project property `-PconformanceReport` is checked in `test/partiql-tests-runner/build.gradle.kts`,
-to prevent the conformance test suite from executing during a normal project-build test run. 
-Unfortunately, this also disables running `ConformanceTestsReportRunner` via IntelliJ UI for unit tests. 
+to exclude the conformance test suite from executing during a normal project-build test run. 
+Unfortunately, this also disables running `ConformanceTestsReportRunner` in a UI runner. 
 To make that possible locally, temporarily comment out the check in `test/partiql-tests-runner/build.gradle.kts`.
 
 ## Compare Conformance Reports locally
--- a/test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/TestRunner.kt
+++ b/test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/TestRunner.kt
@@ -456,7 +456,8 @@ class TestRunner {
      * succeed with the expected result. Ensures that tests included in the failing list fail.
      *
      * These tests are included in the normal test/building.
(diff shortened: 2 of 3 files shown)


test/partiql-tests-runner/README.md
```diff
@@ -7,9 +7,6 @@ This package enables:
 1. Verifying conformance tests are defined correctly (e.g. verify evaluation environment is not missing a table)
 2. Identifying areas in the Kotlin implementation that diverge from the [PartiQL Specification](https://partiql.org/assets/PartiQL-Specification.pdf)
 
-Eventually, the `partiql-test-runner` module will replace the `partiql-pts` and `partiql-testscript` modules along with some other 
-tests in `lang` that were ported to `partiql-tests` 
-(see [partiql-lang-kotlin#789](https://github.com/partiql/partiql-lang-kotlin/issues/789)).
```
vgapeyev, line 12: I believe this remark was outdated.

test/partiql-tests-runner/README.md
```diff
@@ -21,3 +18,18 @@ tests in `lang` that were ported to `partiql-tests`
...
 ./gradlew :test:partiql-tests-runner:test --tests "*ConformanceTestsReportRunner" -PconformanceReport
 ```
+The report is written into file `test/partiql-tests-runner/conformance_test_results.ion`.
+
+## Run Conformance Tests in IntelliJ
+
+The above project property `-PconformanceReport` is checked in `test/partiql-tests-runner/build.gradle.kts`,
+to prevent the conformance test suite from executing during a normal project-build test run. 
+Unfortunately, this also disables running `ConformanceTestsReportRunner` via IntelliJ UI for unit tests. 
+To make that possible locally, temporarily comment out the check in `test/partiql-tests-runner/build.gradle.kts`.
```
vgapeyev, line 28: If there is a better way to run the conformance suite as unit tests in IntelliJ, let's write it up here. 

Otherwise, something better is desirable.  It might be possible to derive something suitable from https://docs.gradle.org/current/userguide/java_testing.html#sec:configuring_java_integration_tests

alancai98: The way I was running the conformance tests in `DefaultConformanceTestRunner` was running the specific parameterized test function (e.g. `validatePartiQLEvalTestData` and `validatePartiQLEvalEquivTestData`). This didn't require commenting out the following line. And let me run the tests with IntelliJ. Perhaps we could add a comment about this workaround.

alancai98: Running the entire class of tests though required me to comment out those lines.

alancai98 (approved):
Document changes look reasonable to me. Left a comment about another way to run the conformance tests locally and doc nits. Based off the PR description, we should probably cut a couple more issues:

---

> ... we need to figure out more comfortable / useful workflow with conformance tests overall.

For this point, could we create a GitHub issue? May be a helpful place to keep both our thoughts on this topic.

---

> In addition to the needed improvement with enabling/disabling the conformance suite runs, it would be helpful to get a mechanism for running smaller subsets of the suite (at least being able to limit to a directory), especially when running with IntelliJ UI.

This could warrant a separate GitHub issue. I guess adding some way to filter based off the namespace/directory of the test could be a reasonable starting place.

test/partiql-tests-runner/README.md
```diff
@@ -21,3 +18,18 @@ tests in `lang` that were ported to `partiql-tests`
 PARTIQL_TESTS_DATA=/path/to/partiql-tests/data \
 ./gradlew :test:partiql-tests-runner:test --tests "*ConformanceTestsReportRunner" -PconformanceReport
 ```
+The report is written into file `test/partiql-tests-runner/conformance_test_results.ion`.
+
+## Run Conformance Tests in IntelliJ
+
+The above project property `-PconformanceReport` is checked in `test/partiql-tests-runner/build.gradle.kts`,
+to prevent the conformance test suite from executing during a normal project-build test run. 
+Unfortunately, this also disables running `ConformanceTestsReportRunner` via IntelliJ UI for unit tests.
```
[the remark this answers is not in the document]
vgapeyev: Changed to a "UI runner".

test/partiql-tests-runner/README.md
vgapeyev: I tried to run the functions directly, but the exclude from `build.gradle.kts`, even though it's on the `ConformanceTestsReportRunner` *class*, applies to the member functions as well.  
Perhaps your experience was different because it was indeed for `DefaultConformanceTestRunner`, which was not excluded (till this PR, hopefully temporarily...)

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/TestRunner.kt
```diff
@@ -456,6 +456,7 @@ class TestRunner {
      * succeed with the expected result. Ensures that tests included in the failing list fail.
      *
      * These tests are included in the normal test/building.
+     * Update May 2023: Now excluded from the normal build, because fail lists are out of date.
```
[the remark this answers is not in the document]
vgapeyev: I think we are on the same page here: it's important to run conformance tests regularly, probably within the regular pre-commit test run, and some sort of fail list / exclusion list is a must in that case.  But it should not be too burdensome to maintain it! I'd like to figure out something automated or tool-assisted.

Luckily, our conformance tests are still well-duplicated in unit tests (if I am not mistaken), so let's hope excluding them from builds won't create too many loopholes.

vgapeyev (commented):
Alan, thanks for the review!  I adjusted docs in a few places, so probably will need another approval click. 

Re tickets for the TODO items, I aim to post some in a few days, after I gain some more experience working with the conformance suite.

test/partiql-tests-runner/README.md
alancai98: Ah, I see now. Ignore my followup comment then. I was running the functions without this PR's changes locally so my workaround doesn't actually work. For now, the only way to run the tests locally would be to comment out that `exclude` line in `build.gradle.kts`.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/TestRunner.kt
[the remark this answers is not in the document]
alancai98: Yeah, I agree. Finding some good way to maintain the "fail list / exclusion list" is important, which can also be useful for the Rust implementation.

> Luckily, our conformance tests are still well-duplicated in unit tests (if I am not mistaken), so let's hope excluding them from builds won't create too many loopholes.

This should be the case other than for the `pts` tests (which were removed from the `partiql-lang-kotlin` codebase). Regardless, the PR's GitHub Actions CI script will help identify if there are any regressions there.

alancai98 (approved):

PR #1098, opened 2023-05-23
Initializes partiql-parser subproject with ANTLR sources

RCHowell:
## Description

No external change. Moving `Parser` implementation to an impl package to follow Explicit API structure.

## Other Information
- Updated Unreleased Section in CHANGELOG: **[YES/NO]**
Yes

- Any backward-incompatible changes? **[YES/NO]**
No

- Any new external dependencies? **[YES/NO]**
No

## License Information

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1100, opened 2023-05-24
Removes deprecated and unused syntax classes/methods

johnedquinn:
## Relevant Issues
- None

## Description
- Removes deprecated and unused syntax classes/methods
- See CHANGELOG for all breaking changes.
- Also moved/removed some internal classes/methods.

## Other Information

## License Information


partiql-lang/src/test/kotlin/org/partiql/lang/util/testdsl/IonResultTestCase.kt
```diff
@@ -15,7 +14,7 @@ import org.partiql.lang.eval.evaluatortestframework.ExpectedResultFormat
 import org.partiql.lang.eval.evaluatortestframework.PartiQLCompilerPipelineFactory
 import org.partiql.lang.eval.evaluatortestframework.PipelineEvaluatorTestAdapter
 import org.partiql.lang.mockdb.MockDb
-import org.partiql.lang.syntax.PartiQLParser
+import org.partiql.lang.syntax.impl.PartiQLParser
```
RCHowell, line 17: I'm thinking I should've updated these in https://github.com/partiql/partiql-lang-kotlin/pull/1098

I will actually do that before you rebase

RCHowell (changes requested):

partiql-lang/src/main/kotlin/org/partiql/lang/syntax/impl/PartiQLPigVisitor.kt
```diff
@@ -1102,8 +1104,8 @@ internal class PartiQLPigVisitor(val customTypes: List<CustomType> = listOf(), p
     }
 
     override fun visitDateFunction(ctx: PartiQLParser.DateFunctionContext) = PartiqlAst.build {
-        if (!DATE_TIME_PART_KEYWORDS.contains(ctx.dt.text.toLowerCase())) {
-            throw ctx.dt.err("Expected one of: $DATE_TIME_PART_KEYWORDS", ErrorCode.PARSE_EXPECTED_DATE_TIME_PART)
```
RCHowell, line 1106: It doesn't matter that much, but because we have an enum I don't think we need the `KEYWORDS` part. Instead of a set `contains`, you use can use `safeValueOf(...)`. Additionally, you could use the enum DateTimePart.values instead of building the keyword list.

RCHowell (approved):

PR #1101, opened 2023-05-24
Removes deprecated ExprValueFactory along with other deprecated classes

johnedquinn:
## Relevant Issues
- None

## Description
- Removes deprecated ExprValueFactory along with other deprecated classes. See CHANGELOG for all details.

## Other Information

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,17 +41,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 - **Breaking**: Removes deprecated `org.partiql.annotations.PartiQLExperimental`
-- **Breaking**: Removes deprecated constructor for `org.partiql.lang.CompilerPipeline`
-- **Breaking**: Removes deprecated `valueFactory`, `build`, `builder`, `standard` in `org.partiql.lang.CompilerPipeline`
-- **Breaking**: Removes deprecated constructor in `org.partiql.lang.CompilerPipeline`
-- **Breaking**: Removes deprecated `ofIonStruct()` from `org.partiql.lang.eval.Bindings`
 - **Breaking**: Removes deprecated/unused `blacklist()` and `denyList()` from `org.partiql.lang.eval`
 - **Breaking**: Removes deprecated enum `LEGACY` in `org.partiql.lang.eval.CompileOptions`
-- **Breaking**: Removes deprecated `org.partiql.lang.eval.ExprValueFactory`
-- **Breaking**: Removes deprecated `isNull()` in `org.partiql.lang.eval.ExprValueType`
-- **Breaking**: Removes deprecated `convert()` in `org.partiql.lang.eval.io.DelimitedValues.ConversionMode`
-- **Breaking**: Removes deprecated `exprValue()` from `org.partiql.lang.eval.io.DelimitedValues`
-- **Breaking**: Removes deprecated constructor for `org.partiql.lang.eval.physical.EvaluatorState`
+- **Breaking**: Removes deprecated `org.partiql.lang.eval.ExprValueFactory`, as well as all methods that had its instance
+  among arguments. The counterparts of these methods without an ExprValueFactory are still available. The affected methods
+  include: `ofIonStruct()` in `org.partiql.lang.eval.Bindings`, a constructor of `org.partiql.lang.CompilerPipeline`,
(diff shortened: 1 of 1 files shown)


RCHowell:
Thanks for cleaning this up. Will approve after rebase.

CHANGELOG.md
```diff
@@ -40,6 +40,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- **Breaking**: Removes deprecated constructor in `org.partiql.lang.CompilerPipeline`
+- **Breaking**: Removes deprecated `ofIonStruct()` from `org.partiql.lang.eval.Bindings`
+- **Breaking**: Removes deprecated/unused `blacklist()` and `denyList()` from `org.partiql.lang.eval`
+- **Breaking**: Removes deprecated enum `LEGACY` in `org.partiql.lang.eval.CompileOptions`
+- **Breaking**: Removes deprecated `org.partiql.lang.eval.ExprValueFactory`
+- **Breaking**: Removes deprecated `isNull()` in `org.partiql.lang.eval.ExprValueType`
+- **Breaking**: Removes deprecated `convert()` in `org.partiql.lang.eval.io.DelimitedValues.ConversionMode`
+- **Breaking**: Removes deprecated `exprValue()` from `org.partiql.lang.eval.io.DelimitedValues`
+- **Breaking**: Removes deprecated constructor for `org.partiql.lang.eval.physical.EvaluatorState`
+- **Breaking**: Removes deprecated `org.partiql.lang.eval.visitors.GroupKeyReferencesVisitorTransform`
```
vgapeyev, line 55: This list of **Breaking** removals may sound more alarming than necessary.  In many cases, a method that has an `ExprValueFactory` argument is being removed, while its counterpart without such an argument is still there. One example is `Bindings.ofIonStruct`, but  there are more. 

My preference would be to spell this out in CHANGELOG to preclude any panic, while reducing the repetitiveness as well.  Something like: 

> **Breaking**: Removes deprecated `org.partiql.lang.eval.ExprValueFactory`, as well as all methods (previously deprecated as well) that had its instance among arguments, namely: `ofIonStruct()` in `org.partiql.lang.eval.Bindings`, ...[TODO]....  The counterparts of these methods without an `ExprValueFactory` are still available.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/CompileOptions.kt
```diff
@@ -79,10 +79,6 @@ enum class TypingMode {
  * Indicates how CAST should behave.
  */
 enum class TypedOpBehavior {
-    @Deprecated(message = "TypedOpBehavior.LEGACY is an old compile option that will be removed in an upcoming release", replaceWith = ReplaceWith("TypedOpBehavior.HONOR_PARAMETERS"))
-    /** The old behavior that ignores type arguments in CAST and IS. */
-    LEGACY,
-
```
vgapeyev, line 85: With `HONOR_PARAMETERS` as the only value remaining in the enum, did we overlook the opportunity to deprecate all of `TypedOpBehavior`, as opposed to just its `LEGACY` value?

Could we be a little daring and remove `TypedOpBehavior` in this version cycle as well, skipping the deprecation ceremony? (This would be a separate PR.) 
From what I see, the primary concern would be various configuration methods that take a `TypedOpBehavior` argument.  But most (all?) of those default the argument to `HONOR_PARAMETERS` already -- that is, the argument can be skipped, and chances are is being skipped by users.  While the other value for the argument is already impossible. In other words, anyone who would bump into an upgrade hiccup with disappeared `TypedOpBehavior` would have bumped into a similar hiccup with disappeared `LEGACY`.

vgapeyev (changes requested):
This is pretty much an approval, just asking for a CHANGELOG edit. 

Also, let's bring up the matter of removing TypedOpBehavior while skipping the deprecation ceremony.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/CompileOptions.kt
johnedquinn: Let's absolutely do that. I can look into it after this PR.

vgapeyev (approved):
Good to go as far as I am concerned.  
(Not sure what concern Robert had when mentioning "rebase".)

RCHowell (approved):

PR #1116, opened 2023-06-07
Adds syntactic and planning support for EXCLUDED

### LINKED ISSUE #1111: Please support `excluded` table in `INSERT ... ON CONFLICT ... WHERE` predicate
## Relevant Issue/Bug

I list this as a feature, not a bug, because it seems clear that this is work as yet undone, instead of incorrect behavior of completed code.  https://github.com/partiql/partiql-lang-kotlin/pull/1061 implemented support for the `WHERE` clause, but did not include support for the `excluded` table.

## Requested Solution/Feature

Two problems are preventing the `excluded` table from working today:

### Parse Error

I would expect that this query parse correctly:

```sql
INSERT INTO foo AS f << {'id': 1, 'version': 42} >> 
WHERE foo.version < excluded.version
```

... but I receive the following exception:

```
org.partiql.lang.syntax.ParserException: mismatched input 'WHERE' expecting {<EOF>, ';'}
	Parser Error: at line 1, column 69: unexpected token found, WHERE : WHERE
```

On experimentation, I found that this error happens any time the keyword `excluded` is included in the `WHERE` clause.  (Parsing succeeds if `excluded.version` is replaced with `whatever.bar`).  On inspection of `PartiQL.g4` it appears that `EXCLUDED` is not treated as an expression.  

### Variable Resolution

The `excluded` variable must be added to the local scope for variable resolution within the `WHERE` clause.  Without this, any reference to the `excluded` table will be reported as an undefined variable.

## Describe Alternatives

AFAIK there are no alternatives.

## Additional Context

My team intends to use the `excluded` table in the very near future.  The workaround is to use parameters and reference them instead of the `excluded` table, but this will only work when inserting a single record.  We would like to support multi-record inserts.

```
INSERT INTO foo AS f << {'id': $id, version: $version } >> 
WHERE foo.version < $version
```

## DoD (Definition of Done)

- The `excluded` table can be referenced in the `WHERE` clause of `INSERT INTO ... ON CONFLICT ... WHERE...` without causing any errors.

johnedquinn:
## Relevant Issues
- Closes #1111 

## Description
- Adds support for `EXCLUDED` by:
  - Modifying `varRefExpr` in ANTLR to allow for non-reserved keywords. Marks `EXCLUDED` as non-reserved. It is still reserved immediately following `DO REPLACE` and `DO UPDATE`.
  - Modifying PIG Logical/LogicalResolved `dml_operation.dml_update` and `dml_operation.dml_replace` to have an optional `row_alias` (string).
    - I do this because `dml_update` is used for both `UPSERT` and `INSERT ... ON CONFLICT DO UPDATE ...`. `UPSERT` doesn't have the ability to use the `excluded` table.
    - Same goes for `dml_replace` with `REPLACE` and `INSERT ... ON CONFLICT DO REPLACE ...`.
  - So, the `row_alias` becomes a variable declaration and therefore gets a dedicated variable in the local registers.
  - For identifier resolution, we allow the ON CONFLICT ACTION `<condition>` have access to the new variable by adding it to the scope.

## Testing
- Added parsing tests for non-reserved keywords
- Added parsing tests for using `EXCLUDED` in the `<condition>`
- Added `AstToiLogical` and `LogicalToLogicalResolved` tests. Updated `LogicalResolvedToPhysical` tests to make sure the `rowAlias` was being passed.

## Other Information
  - See CHANGELOG.

## License Information

### DIFF
--- a/partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransformTests.kt
+++ b/partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransformTests.kt
@@ -836,6 +836,41 @@ class AstToLogicalVisitorTransformTests {
                     }
                 }
             ),
+
+            // A collection of values
+            TestCase(
+                "INSERT INTO foo AS f <<{'id': 1, 'name':'bob'}, {'id': 2, 'name':'alice'}>> ON CONFLICT DO REPLACE EXCLUDED WHERE f.id > 2",
+                PartiqlLogical.build {
+                    PartiqlLogical.build {
+                        dml(
+                            identifier("foo", caseInsensitive()),
+                            dmlReplace(
+                                varDecl("f"),
+                                condition = gt(
+                                    listOf(
+                                        path(
+                                            id("f", caseInsensitive(), unqualified()),
+                                            listOf(pathExpr(lit(ionString("id")), caseInsensitive()))
+                                        ),
+                                        lit(ionInt(2))
+                                    )
+                                ),
+                                rowAlias = varDecl(AstToLogicalVisitorTransform.EXCLUDED)
+                            ),
+                            bag(
+                                struct(
+                                    structField(lit(ionString("id")), lit(ionInt(1))),
+                                    structField(lit(ionString("name")), lit(ionString("bob")))
+                                ),
+                                struct(
+                                    structField(lit(ionString("id")), lit(ionInt(2))),
+                                    structField(lit(ionString("name")), lit(ionString("alice")))
+                                )
+                            )
+                        )
+                    }
+                }
+            ),
             // Testing using excluded non-reserved keyword in condition
             TestCase(
                 "INSERT INTO foo AS f <<{'id': 1, 'name':'bob'}>> ON CONFLICT DO REPLACE EXCLUDED WHERE excluded.id > 2",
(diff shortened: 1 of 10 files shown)


partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransformTests.kt
```diff
@@ -822,7 +823,38 @@ class AstToLogicalVisitorTransformTests {
...
+                                    structField(lit(ionString("id")), lit(ionInt(1))),
+                                    structField(lit(ionString("name")), lit(ionString("bob")))
                                 )
+                            )
+                        )
+                    }
+                }
+            ),
+            // Testing using excluded non-reserved keyword in condition
+            TestCase(
```
am357, line 840: Can we also add a test with a collection of values, e.g.:

```SQL
INSERT INTO foo AS f <<{'id': 1, 'name':'bob'}, {'id': 2, 'name':'Alice'}>> ON CONFLICT DO REPLACE EXCLUDED WHERE f.id > 2
```

I can tell that it passes but would be good to have it as part of the tests.
```

am357 (commented):
Overall very clean—just a comment about adding an additional test.

dlurton (approved):
LGTM, thanks.

am357 (approved):

Issue #1082, opened 2023-05-17
Parsing of a larger fragment of label patterns in graph query


vgapeyev:
The currently-parsed grammar of graph patterns, in its part for checking labels on nodes and edges, only supports single-label patterns, as in `-[x:feeds]->`. 

The GPML paper defines a richer language of label patterns, with disjunction, conjunction, negation, grouping, and wildcards.  

This ticket if for completing the label patterns in the parser, **with priority for adding disjunction/alternation** `|`(as in `-[:feeds|contains]->`) **and negation** `!` (as in `-[:!feeds]->`).

vgapeyev:
Completed with PR #1139.

Issue #1122, opened 2023-06-13
Revisit cases sensitivity of identifiers


vgapeyev:
In short: What is currently implemented for identifiers in partiql-lang-kotlin: 
 - Is not what is specified in SQL; 
 - Not specified or described anywhere; 
 - The implementation is hard to follow and understand, as it is not encapsulated, but rather spread around and duplicated at use sites places.

Ideally, we implement what SQL prescribes, while taking care of properly abstracting it. 
This will surely be backward-incompatible, but everyone's sanity can be worth it. 

## SQL

In SQL, there are two syntactic variants of identifiers: regular identifiers like `foo` and delimited/quoted identifiers like `"fOO"`. 
- A regular identifier is considered case-insensitive and has the fully-capitalized variant as the canonical form. 
- A delimited identifier is case-sensitive. 
- The regular vs delimited distinction is only lexical; after parsing and canonicalization, all identifiers are essentially strings. This is to say that `foo`, `Foo`, `fOo`, `FOO` and, notably, `"FOO"` are all the same identifier, canonically being `FOO`. This identifier is distinct from `"foo"`, `"Foo"`, `"fOo"`, all of which are also distinct from each other. 
- The equivalence of identifiers is the equality of these canonicalized strings, in case-sensitive way. 
- Either lexical kind of an identifier (regular or delimited) can be used either for binding a variable (as in FROM or GROUP BY) or as a reference variable to something bound elsewhere.  

## partiql-lang-kotlin

In contrast, in the current implementation: 
- While there are lexical regular and delimited identifiers of the same lexical appearance as in SQL, there are three kinds of identifiers internally. The first distinction is based on whether it is an identifier used as binder vs used as a reference. 
- For a binder, the case of characters in the lexical appearance is preserved semantically, while the fact whether the identifier was regular or delimited is discarded. The binder identifiers are used as keys for things like environments.
- For a reference identifier, a delimited identifier becomes "an identifier with case-sensitive lookup", while a regular identifier becomes "an identifier with case-insensitive lookup".  For the latter, the canonical representation is lower case. 
-  The lookup of a reference identifier in an environment depends on whether it is case-sensitive or case-insensitive.
- The case-sensitive lookup is straightforward, using case-sensitive string equality. 
[...]

PR #1139, opened 2023-06-30
Parsing of GPML label patterns.

vgapeyev:
## Relevant Issues
- Closes Issue #1082 

## Description
- Now parsing the full assortment of node/edge label patterns: named label, disjunction, conjunction, negation, grouping, and wildcard. 

## Other Information

  - Limited to the experimental (graphs) fragment of the language: the grammar and the AST have changed slightly and will cause compilation errors.  

## License Information


CHANGELOG.md
johnedquinn, line 1: While these changes occur in an experimental area of our language (not strictly enforced via experimental annotations), should we still highlight these changes to the AST as breaking? As we currently have users of this experimental language, we could make sure that the next release is a "major" version release with information about what has changed. This would largely serve as a mechanism to ensure that we release this as a new "major" version versus accidentally breaking someone else's pipeline via a minor version bump.

partiql-parser/src/main/antlr/PartiQL.g4
```diff
@@ -408,10 +408,28 @@ edgeWSpec
...
+    | labelFactor                        # LabelTermFactor
+    ;
+
+labelFactor
+    : BANG labelPrimary                  # LabelFactorNot
+    | labelPrimary                       # LabelFactorPrimary
+    ;
+
+labelPrimary
+    : symbolPrimitive                    # LabelPrimaryName
```
johnedquinn, line 429: From the GPML paper, I believe this would only be `IDENTIFIER`.

However, this would severely limit labels that share the name of reserved keywords. Just adding this comment here to make sure this was intentional. That being said -- it seems like label names are always case-sensitive whether they are quoted or not. Am I right in this assumption?

partiql-lang/src/main/kotlin/org/partiql/lang/graph/Graph.kt
```diff
@@ -66,11 +66,11 @@ interface Graph {
...
 sealed class LabelSpec {
-    /** A graph element matches when one of its labels is [name].*/
-    data class OneOf(val name: String) : LabelSpec()
+    /** A graph element (node or edge) matches when its label set contains [name]. */
+    data class Name(val name: String) : LabelSpec()
 
-    /** A graph element always matches (even when it is not labeled at all).*/
-    object Whatever : LabelSpec()
+    /** A graph element matches as long as it has a label. */
+    object Wildcard : LabelSpec()
```
johnedquinn, line 73: While this is *somewhat* considered experimental (under graph, but not annotated with an experimental annotation), if anyone has used this API -- these changes would be considered breaking. IMO, not worthy of deprecation, however, a notice in the CHANGELOG could be a good idea (especially so that we internally remember to release this as a major version bump).

johnedquinn (commented):
Overall this looks good to me! Coincides with the GPML paper. I do have some questions regarding case-sensitivity and avoiding breaking others' code, but once resolved, I can approve.

partiql-parser/src/main/antlr/PartiQL.g4
vgapeyev: Are you concerned that a label can only be a non-delimited/regular identifier? I did not go back to sift through the paper, but I would not expect it to bother about this issue. My expectation is that a primary label would be a general SQL identifier, i.e. both non-delimited and delimited being allowed. 

Here, I did not dwell on the matter much: it was a `symbolPrimitive` before [old Line 414 above], so it remains that. 
This should be resolved while dealing with #1122, which I have great hope of getting to soon.

partiql-lang/src/main/kotlin/org/partiql/lang/graph/Graph.kt
vgapeyev: Ok, I'll add better communication about this.

partiql-parser/src/main/antlr/PartiQL.g4
johnedquinn: I guess it's more about case-sensitivity -- and retaining whether the name was delimited or not. The reason is that in `SimpleGraph` -- when matching a Graph against a LabelSpec, we check in a case-sensitive manner whether the graph's labels contain a LabelSpec. As this was the functionality prior to this PR, I can ignore this as out-of-scope until #1122 is addressed. Thanks!

johnedquinn (dismissed):
LGTM!

RCHowell (approved):

PR #1091, opened 2023-05-22
Initializes PartiQL Value classes

RCHowell:
## Description

This PR introduces an iteration of ExprValue (split from evaluator) informed by IonElement. Notably supporting annotations #1093 and defining Reader/Writer interfaces

## Other Information
- Updated Unreleased Section in CHANGELOG: **[YES/NO]**
NO

- Any backward-incompatible changes? **[YES/NO]**
NO

- Any new external dependencies? **[YES/NO]**
  - `org.jetbrains.kotlinx:kotlinx-collections-immutable`

## License Information

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1142, opened 2023-07-03
Initialize partiql-parser package with partiql-ast IR

RCHowell:
## Relevant Issues

Rebase of https://github.com/partiql/partiql-lang-kotlin/pull/1110

- [x] Rebase #1054 
- [x] Update AST definitions from #1054 
- [x] Merge #1109
- [x] Define types within AST
- [x] Introduce all parser rule ambiguities to AST definition
- [x] Re-implement `toLegacyAst()` without synthetic locations
- [x] Implement `org.partiql.lang.syntax.PartiQLParser` interface via shim
- [x] Validate via all existing tests
- [x] Merge #1091 
- [x] Use PartiQL Value for literals

## Description

This PR introduces the new PartiQL AST and PartiQL Parser built on the latest IR classes. Both the partiql-ast and partiql-parser packages can be consumed as artifacts independently from partiql-lang-kotlin.

**About the AST**

The AST package is simply the type definitions file with a default factory. The easiest way to review is by looking at the definitions. https://github.com/partiql/partiql-lang-kotlin/blob/ast-rebase/partiql-ast/src/main/resources/partiql_ast.ion

Notably this AST is able to use PIG 1.0 features such as sets, maps, enums, imported types, and additional scalar types. We are able to better model the AST without concern for V0 AST compatibility as there is a layer of indirection between the AST's data structures and the serialized s-expression format. It is possible to convert to the legacy AST via the ToLegacyAst.kt helper. We also get to drop the dependency on pig-runtime and its workarounds.

Where possible, node definitions are nested (see DML and GPML as an example) and redundancy is eliminated (see unary and binary expressions as an example). Because we can now model sums of sum types, we can fix the inheritance of certain nodes, details in #1110.

**About the Parser**

The parser package is effectively the existing PartiQLPigParser implementation (credit @johnedquinn) updated to produce the new AST rather than the PIG AST. One notable difference is that metadata (ie source locations) is not tagged onto nodes, rather it is kept in a metadata map exactly like the PartiQL Rust implementation. A benefit of this approach is that we decouple ANTLR parse tree nodes from AST nodes because we can build a source location from the start-end of a grammar rule rather than from a single token as done in org.partiql.lang.syntax parsers.

Note also that the partiql-parser package and the existing AST use the same grammar found in partiql-parser/src/main/antlr/PartiQL.g4. This means the ANTLR-based implementations of both the old and new interfaces use the same grammar.

## Other Information
TODO

No. This change includes additions with the ability to produce the 

No

Yes

## License Information


partiql-parser/build.gradle.kts
```diff
@@ -15,12 +15,13 @@
 plugins {
     id(Plugins.antlr)
     id(Plugins.conventions)
-    // id(Plugins.publish)
+    id(Plugins.publish)
 }
 
 dependencies {
     antlr(Deps.antlr)
     implementation(project(":partiql-ast"))
```
johnedquinn, line 23: Do you think `partiql-ast` should be `api`?

The reason for that is if someone consumes the `Parser` interface, the `parseAstStatement` method returns an API defined in `partiql-ast`.

partiql-ast/build.gradle.kts
```diff
@@ -22,6 +22,13 @@ plugins {
 dependencies {
     api(Deps.pigRuntime)
     api(Deps.ionElement)
+    implementation(project(":partiql-types"))
```
johnedquinn, line 25: Do you think this should be `api`?

johnedquinn (changes requested):
This is some great work. I have one additional comment (beyond the small comments I left): we need to update `partiql-lang/build.gradle.kts` to use `api` instead of `libs(project())` for the parser.

Beyond my small comments, I think this PR is ready to be shipped. Thanks for all the effort!

partiql-parser/src/main/kotlin/org/partiql/parser/PartiQLParserBuilder.kt
```diff
@@ -0,0 +1,35 @@
...
+        @JvmStatic
+        public fun standard(): PartiQLParserBuilder {
+            return PartiQLParserBuilder()
+        }
+    }
+
+    public fun build(): PartiQLParser {
+        return PartiQLParserDefault()
+    }
+}
```
yliuuuu, line 35: Should it takes in a list of custom types as well?

partiql-lang/src/main/kotlin/org/partiql/lang/syntax/PartiQLParserBuilder.kt
```diff
@@ -29,12 +30,25 @@ import org.partiql.lang.types.CustomType
  */
 class PartiQLParserBuilder {
```
yliuuuu, line 31: Should we plan on moving this to the parser package or should we just go directly to deprecate and remove pattern here?

yliuuuu (commented):
Overall looks good to me, only have some small nit.
One high level comment is, the new Parser that parse to the new AST package does not have its own test suit. I understand that the ShimParser is working as expected, but do we planning on port all those parser tests to test directly on the new AST. 
Given some of the key distinguishes between the new ast and old one, (using partiQL value, different source location, etc), may worth to have those test cases some points in the future. (huge work for sure, not in scope of this PR for sure, just want to understand the road map here). 

Another high level comment is: would be nice to have some of the PR description ported to package level README.

partiql-ast/build.gradle.kts
RCHowell: Yes

partiql-parser/build.gradle.kts
RCHowell: Yes. Good callout.

partiql-lang/src/main/kotlin/org/partiql/lang/syntax/PartiQLParserBuilder.kt
RCHowell: It exists in the new package. This one will be deprecated then removed.

partiql-ast/src/main/kotlin/org/partiql/lang/ast/IsTransformedOrderByAliasMeta.kt
```diff
@@ -30,6 +28,7 @@ import org.partiql.lang.eval.visitors.OrderBySortSpecVisitorTransform
  */
 class IsTransformedOrderByAliasMeta private constructor() : Meta {
```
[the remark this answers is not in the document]
RCHowell: I simply moved the metas to not have a dependency cycle. There are a handful of ways we might clean these up, but are you ok to consider that out of scope for this PR?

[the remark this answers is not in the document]
yliuuuu: Yes. Should have stated that.

johnedquinn (changes requested):
Need to update `partiql-lang/build.gradle.kts` to use `api` instead of the `libs(project())`

partiql-ast/src/main/kotlin/org/partiql/ast/Ast.kt
```diff
@@ -0,0 +1,45 @@
...
+
+public abstract class MyFactory : AstBaseFactory() {
+
+    override fun identifierSymbol(symbol: String, caseSensitivity: Identifier.CaseSensitivity): Identifier.Symbol {
+        return ComparableIdentifier(_id(), symbol, caseSensitivity)
+    }
+}
+
+//
+class ComparableIdentifier(
```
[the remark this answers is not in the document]
RCHowell: Yes, that snuck in by accident. Thanks

johnedquinn (dismissed):
Great work!

vgapeyev (commented):
I continue studying this work, but thought of posting a few disjoint comments that have accumulated so far.

partiql-ast/README.adoc
```diff
@@ -0,0 +1,356 @@
...
+The interfaces are generated from `resources/partiql_ast.ion` (details in `lib/sprout/README`)
+
+=== Node
+
+[source,kotlin]
+----
+public interface AstNode {
+
+    // Every node gets an _id for associating any metadata
+    public val _id: String
```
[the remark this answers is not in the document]
RCHowell: > I'd be concerned that Random.nextInt() is not a reliable method to get uniqueness.

Correct, you can override if you require uniqueness. I considered this good enough, but I can make it truly unique if necessary. Related, I wanted the factory to be stateless

partiql-ast/README.adoc
```diff
@@ -0,0 +1,356 @@
...
+    override fun equals(other: Any?): Boolean {
+        if (other == null || other !is Identifier.Symbol) return false // different type
+        if (other === this) return true // same object
+        return when (caseSensitivity) {
+            Identifier.CaseSensitivity.SENSITIVE -> this.symbol == other.symbol
+            Identifier.CaseSensitivity.INSENSITIVE -> this.symbol.lowercase() == other.symbol.lowercase()
+        }
+    }
+}
+----
```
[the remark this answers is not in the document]
RCHowell: I used this as an example because we had discussed this exact limitation last fall. This example shows how the current design is able to address situations in which generated sources do not handle edge cases.

I believe it is useful to provide customers with the facility to produce their own concrete implementations of AST nodes (if they want, hence a default). Here we are indeed eating our own dogfood while overcoming basic limitations that are trivial when hand-writing ASTs. The default factory is _our_ default factory, but not _the_ default factory.

> Use of [the abstract factory] pattern enables interchangeable concrete implementations without changing the code that uses them, even at run. [wikipedia](https://en.wikipedia.org/wiki/Abstract_factory_pattern)


Is your primary concern that this is a bad example?

partiql-ast/src/main/kotlin/org/partiql/ast/Ast.kt
```diff
@@ -0,0 +1,45 @@
...
+    // internal default overrides here
+}
+
+public abstract class MyFactory : AstBaseFactory() {
+
+    override fun identifierSymbol(symbol: String, caseSensitivity: Identifier.CaseSensitivity): Identifier.Symbol {
+        return ComparableIdentifier(_id(), symbol, caseSensitivity)
+    }
+}
+
```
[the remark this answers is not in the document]
RCHowell: Removed in latest commit

partiql-parser/src/main/kotlin/org/partiql/parser/impl/PartiQLParserDefault.kt
```diff
@@ -0,0 +1,2004 @@
...
+         * Each visit attaches source locations from the given parse tree node; constructs nodes via the factory.
+         */
+        private inline fun <T : AstNode> translate(ctx: ParserRuleContext, block: AstFactory.() -> T): T {
+            val node = factory.block()
+            if (ctx.start != null) {
+                locations[node._id] = SourceLocation(
+                    line = ctx.start.line,
+                    offset = ctx.start.charPositionInLine + 1,
+                    length = (ctx.stop?.stopIndex ?: ctx.start.stopIndex) - ctx.start.startIndex + 1,
+                    lengthLegacy = ctx.start.text.length, // LEGACY LENGTH
```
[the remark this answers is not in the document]
vgapeyev: Yes, this is where `locations` is located, pardon the pun. But where does one read out from `locations` the data that is written into it here?  In other words, how is `locations` used?

I guess it's in the future, and I am just showing being disoriented by the the scope or phasing of this large project.  Intuitively, I thought this PR would have included this reading, so I'd like to understand why it's ok for it not to.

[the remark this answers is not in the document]
RCHowell: Locations is provided in the parser result. This is also how PartiQL Rust implements it.

```kotlin
public interface PartiQLParser {

    @Throws(PartiQLSyntaxException::class, InterruptedException::class)
    public fun parse(source: String): Result

    public data class Result(
        val source: String,
        val root: AstNode,
        val locations: SourceLocations,
    )
}
```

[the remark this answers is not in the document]
RCHowell: The locations you initially referred to are a mutable map used internally by the parse tree visitor as we produce the AST. This mutable map is "frozen" and returned in the parser result which is (the second thing you're asking about) how you access them.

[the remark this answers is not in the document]
vgapeyev: Thanks!  That's what I missed indeed. 
Expecting a mutable in/out parameter (which `location` in the internal `Visitor` is) is a rusty tool in my forensics kit.

partiql-ast/README.adoc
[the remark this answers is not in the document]
vgapeyev: My foremost concern is understanding what ready-to-roll uses we have for the significant extra complexity that Abstract Factory pattern introduces.  Not nice-to-haves that a future customer might find useful, but 2-3 usecases (that is, more than 1!) that we intend to roll out ourselves -- otherwise we won't prove the dog food is edible.

The prototypical usecase for Abstract Factory is making it possible to develop a GUI application, such as a game, while abstracting away from a specific windowing toolkit, so that the application can run with with the look and feel of any toolkit for which someone would provide a concrete factory. 

PartiQL AST (the interfaces in partiql-ast/org.partiql.ast.Types) is the analog of a "windowing toolkit abstraction", while PartiQL parser and PartiQL compiler are analogues of a "GUI game".  What would be useful analogues of the concrete windowing toolkits that we are going to roll out as alternatives to the one concrete AST in partiql-ast/org.partiql.ast.**impl**.Types ?  

I can vaguely imagine something on a scale of an AST variant that contains something like instrumentation to count uses of nodes during compilation or optimization transformations -- but I don't have a defensible case for this and don't recall us planning for anything of a kind.  The case-sensitivity example is not convincing because it's something we should just provide (or not!) in the default AST implementation -- case sensitivity ought to be part of the PartiQL AST contract (or not! -- taken care of outside the AST, as currently). 

To put it differently, what do we give up if org.partiql.ast.Types and org.partiql.ast.**impl**.Types are one and the same thing (just concrete classes, no interfaces)?

This probably requires a discussion more sprawling than can fit in PR comments...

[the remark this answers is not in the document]
RCHowell: I'm aware of the pattern background, and I see your concern now. The motivation for this was to have the flexibility of hand-written AST nodes, but the convenience of generated ones. This pattern enables us to get everything generated, but hand-write when we desire. The default ~GUI~ AST toolkit is what we provide, but customers can provide their own ~look and feel~ nodes when desired/required. 

As you'll see in the PIG AST and the Physical Plan. Various mechanisms have been added for customer extension such as the metadata `<string, object>` map as well as the `impl` string on the `partiql_physical` domain (which is a point of customer extension, but is also related to the "physical"-ness of the domain). 

My effort here was to provide a more powerful mechanism for both ourselves and customers to extend AST nodes. That is, they can use the generated identifiers to store metadata in a richer-typed structure than the object map (see SourceLocations) for an example. And they can handwrite implementations of our AST node interfaces.

My opinion is that, if I could go back in time, the AST would be internal and hand-written. It is the driving cause of major-version churn due to its constant evolving and surface area, as we as its inflexibility due to the generator limitations. I don't this PR solves that issue, but it gives us some better defaults and flexibility. If the AST were internal, then the separation of interfaces and indirection of instantiation would be superfluous. But I found this pattern to be a reasonable compromise given our situation.

johnedquinn (approved):

vgapeyev (dismissed):

partiql-types/src/main/kotlin/org/partiql/value/datetime/impl/SqlDate.kt
```diff
@@ -33,8 +33,8 @@ internal data class SqlDate private constructor(
         @JvmStatic
         @Throws(DateTimeException::class)
         fun of(year: Int, month: Int, day: Int): Date {
-            if (year < 1 || year > 9999)
-                throw DateTimeException("Expect Year Field to be between 1 to 9999, but received $year")
+            if (year < 0 || year > 9999)
+                throw DateTimeException("Expect Year Field to be between 0 to 9999, but received $year")
```
johnedquinn, line 37: Is this change expected? If so:

See [Ion Spec](https://amazon-ion.github.io/ion-docs/docs/spec.html#timestamp)
Also, SQL-1999 Section 6.1 Table 11

partiql-ast/README.adoc
```diff
@@ -0,0 +1,364 @@
...
+
+For example, let's implement a `toString(case: Case)` function on some basic nodes.
+
+[source,kotlin]
+----
+//
+// Usage:
+//  node.accept(ToString, Case.UPPER) // ~ node.toString(Case.UPPER)
```
vgapeyev, line 273: I am puzzled by this comment, esp. in the context of the comment in the next snippet.  
Does `~` mean "not" or "something like"?

partiql-ast/README.adoc
```diff
@@ -0,0 +1,364 @@
...
+[source,kotlin]
+----
+// Traverse the tree collecting all node names
+object AstClassNameCollector  {
+
+    // Public static entry for Java style consumption
+    @JvmStatic
+    fun collect(node: AstNode): List<String> {
+        val acc = mutableListOf<String>()
+        node.accept(ToString, acc) // recall, we have node.toString()
```
vgapeyev, line 306: I am puzzled by this comment, esp. in the context of the comment in the previous snippet.
Is this meant to refer to `accept` in the previous snippet (per the comment there) or to AstNode.toString?  In either case, why?: `ToString` on this line is the class defined just below in this snippet.

partiql-ast/src/main/kotlin/org/partiql/ast/helpers/ToIon.kt
```diff
@@ -95,8 +96,21 @@ internal object ToIon : PartiQLValueBaseVisitor<IonElement, Unit>() {
...
 
     override fun visitTimestamp(v: TimestampValue, ctx: Unit) = v.toIon {
-        val sql = java.sql.Timestamp.valueOf(value)
-        ionTimestamp(Timestamp.forSqlTimestampZ(sql))
+        val offset = when (val z = v.value.timeZone) {
+            TimeZone.UnknownTimeZone -> 0
+            is TimeZone.UtcOffset -> z.totalOffsetMinutes
+            null -> 0
+        }
+        val timestamp = Timestamp.forSecond(
```
johnedquinn, line 104: See this Javadoc note regarding unknown timezone: https://javadoc.io/doc/software.amazon.ion/ion-java/1.0.0/software/amazon/ion/Timestamp.html#forSecond-int-int-int-int-int-java.math.BigDecimal-java.lang.Integer-

vgapeyev (commented):
Could you remove these comments or make it more clear what is the point?

partiql-ast/README.adoc
RCHowell: "something like" poor choice on my part. I can update this

partiql-ast/README.adoc
RCHowell: It's meant to re-iterate how the dispatch works, but I see these comments have caused more harm than help.

partiql-ast/README.adoc
```diff
@@ -0,0 +1,364 @@
...
+//
+object ToString :  AstBaseVisitor<String, Case>() {
+
+    override fun defaultVisit(node: AstNode, ctx: Case) = defaultReturn(node, ctx)
+
+    override fun defaultReturn(node: AstNode, ctx: Case): String = when (ctx) {
+        Case.UPPER -> node::class.simpleName.uppercase()
+        Case.LOWER -> node::class.simpleName.lowercase()
+        Case.PASCAL -> node::class.simpleName
+        Case.SNAKE -> snakeCaseHelper(node::class.simpleName)
```
vgapeyev, line 283: OTOH (re previous 2 comments) maybe rename `ToString` completely, since it is far from what one can expect from `toString`?  Maybe `AstClassName`?

vgapeyev (dismissed):

partiql-ast/README.adoc
RCHowell: It's not the ast class name, it's the Java simpleName

johnedquinn (approved):
Great work!

PR #1144, opened 2023-07-03
Discussion: streamline visiting in PartiQLPigVisitor

vgapeyev:
This PR is to discuss a suggestion to remove `visit` and `visitOrNull` convenience methods in PartiQLPigVisitor.  So far, this PR just points at the issues and shows examples of proposed changes.  If there is agreement, that would need to be repeated for the remaining uses of the method. 
But in case I am overlooking something, let's discuss here. 

Specifics are in PR comments at proposed changes, but here is the overall reasoning, in logical order. 

- (1) Let's avoid reflection-based cast, when the regular `as` cast can be done. [See at `visitFromClause`, and also `visitTableCrossJoin`, `visitExpr`]

- (2) Sometimes, a generic `visit()` call can be replaced with a call to a specific visitor. [See `visitExprTermWrappedQuery`]

- (3) Similar considerations apply to `visitOrNull`: 
  - `visitOrNull` hides several opportunities for direct `visitXxx` calls [see `visitDeleteCommand,`  `visitCaseExpr`]
  - if there are cases where further dispatch may be necessary, the argument about avoiding reflection cast would still apply. 

After addressing all the similar instances, there should not be need left for `visit()` and `visitOrNull()`.

Separately but relatedly, as I ended up repeating reverse-engineering ANTLR parsing details several times already (whenever I needed to modify it), I have also attempted to put in a pragmatic summary and guidelines in `PartiQLPigVisitor`. 

## Other Information
  - Internal matter.

## License Information


partiql-lang/src/main/kotlin/org/partiql/lang/syntax/impl/PartiQLPigVisitor.kt
```diff
@@ -813,7 +852,8 @@ internal class PartiQLPigVisitor(val customTypes: List<CustomType> = listOf(), p
      *
      */
 
-    override fun visitFromClause(ctx: PartiQLParser.FromClauseContext) = visit(ctx.tableReference(), PartiqlAst.FromSource::class)
+    override fun visitFromClause(ctx: PartiQLParser.FromClauseContext) =
+        visit(ctx.tableReference()) as PartiqlAst.FromSource
```
vgapeyev, line 856: The `PartiQLPigVisitor.visit()` call in the original code uses reflection to perform a cast in the result, compared to the regular cast on the result of `AbstractParseTreeVisitor.visit()` in the replacement code.  Reflection is expensive and better be avoided. 
     
The verbosity of the two approaches is comparable, while the internals of `visit()` do not encapsulate much action.

partiql-lang/src/main/kotlin/org/partiql/lang/syntax/impl/PartiQLPigVisitor.kt
```diff
@@ -963,7 +1003,8 @@ internal class PartiQLPigVisitor(val customTypes: List<CustomType> = listOf(), p
      *
      */
 
-    override fun visitExprTermWrappedQuery(ctx: PartiQLParser.ExprTermWrappedQueryContext) = visit(ctx.expr(), PartiqlAst.Expr::class)
+    override fun visitExprTermWrappedQuery(ctx: PartiQLParser.ExprTermWrappedQueryContext) =
+        visitExpr(ctx.expr())
```
vgapeyev, line 1007: Here we can avoid a cast (at this call site, at least), by calling a more specialized method (like `visitExpr` here), which already returns a result of the desired type. 
In many cases, this approach is already in place -- see calls to `visitSymbolPrimitive`, `visitPathSimple`, so this is nothing new.

partiql-lang/src/main/kotlin/org/partiql/lang/syntax/impl/PartiQLPigVisitor.kt
```diff
@@ -278,8 +317,8 @@ internal class PartiQLPigVisitor(val customTypes: List<CustomType> = listOf(), p
 
     override fun visitDeleteCommand(ctx: PartiQLParser.DeleteCommandContext) = PartiqlAst.build {
         val from = visit(ctx.fromClauseSimple(), PartiqlAst.FromSource::class)
-        val where = visitOrNull(ctx.whereClause(), PartiqlAst.Expr::class)
-        val returning = visitOrNull(ctx.returningClause(), PartiqlAst.ReturningExpr::class)
+        val where = ctx.whereClause()?.let { visitWhereClause(it) }
+        val returning = ctx.returningClause()?.let { visitReturningClause(it) }
```
vgapeyev, line 321: The same reasoning as for `visit()` applies to `visitOrNull` here: 
- Remove the abstraction and deal with the null case in-line. 
- Avoid reflection for doing the cast. 
- Use an opportunity for a more specific method (`visitWhereClause`,   
  `visitReturningClause`).

Again, the overall verbosity is comparable if not better.

vgapeyev (commented):
At-code remarks promised in the PR intro.

partiql-lang/src/main/kotlin/org/partiql/lang/syntax/impl/PartiQLPigVisitor.kt
```diff
@@ -1479,11 +1520,13 @@ internal class PartiQLPigVisitor(val customTypes: List<CustomType> = listOf(), p
         else -> ctx.map { visitOrNull(it, clazz) }
     }
 
+    // TODO: Inspect `visitOrNull` call sites and replace with direct `visitXxx` calls or 'as' casts, as appropriate.
     private fun <T : PartiqlAst.PartiqlAstNode> visitOrNull(ctx: ParserRuleContext?, clazz: KClass<T>): T? = when (ctx) {
         null -> null
         else -> clazz.cast(visit(ctx))
     }
```
RCHowell, line 1527: Nice PR. Thank you for the cleaning effort! This has some background. The visitOrNull with the reflection based cast came from a similar Java implementation.

We can improve the readability (and use the `as` cast) with a type reified inline. [Here's what I did for the new implementation.
](https://github.com/partiql/partiql-lang-kotlin/blob/d353514d147f60b2c4ff207c55f69b65acfd73c0/partiql-parser/src/main/kotlin/org/partiql/parser/impl/PartiQLParserDefault.kt#L1839)

These inline functions are like macros

```suggestion
private inline fun <reified T : AstNode> visitOrNull(ctx: ParserRuleContext?): T? = ctx?.let { visit(it) as T }
```


Then we can replace the current `visitOrNull` to achieve the same affect. Additionally, it would be nice to pass a method reference for which visit to call. But at that point, we should just do the explicit visit method.

RCHowell (changes requested):

partiql-lang/src/main/kotlin/org/partiql/lang/syntax/impl/PartiQLPigVisitor.kt
vgapeyev: Thanks for pointing out this inline+reified language construct!  With it, my major concern (about reflection) goes away.  Then, going with reified visit/visitOrNull or not comes down to a matter of preference, I think.  

In my opinion, the bodies of these functions are so trivial that I do not feel the abstraction is worth it.  The call sites do not give much mental savings either: 
```
   visit(ctx.exprBagOp()) as PartiqlAst.Expr
```  
vs
```
   visit<PartiqlAst.Expr>(ctx.exprBagOp())
```

Regardless the above, do you I read it right you support doing 
```
   ctx.whereClause()?.let { visitWhereClause(it) }
```
when a specific visitor is available?

RCHowell:
I suggest we also remove all over the overrides which simply invoke super as it's unnecessary indirection.

partiql-lang/src/main/kotlin/org/partiql/lang/syntax/impl/PartiQLPigVisitor.kt
RCHowell: Yes, `ctx.whereClause()?.let { visitWhereClause(it) }` when we know the type would be preferable. I recall only a few places where we didn't now the type .. meaning the visitOrNull is becoming less useful as time goes on

vgapeyev:
> I suggest we also remove all over the overrides which simply invoke super as it's unnecessary indirection.

@johnedquinn , what do you think about this suggestion from Robert?  
I presume he refers to these methods: https://github.com/partiql/partiql-lang-kotlin/blob/b2df81246961eee7348ffbd67452119ddb754f6a/partiql-lang/src/main/kotlin/org/partiql/lang/syntax/impl/PartiQLPigVisitor.kt#L1571C1-L1605C1
I have to admit I couldn't figure out the comment that precedes them. (Even if I maybe benefited from the IDE behavior without knowing they were responsible!) 

This matter should be independent from the rest of this PR.

johnedquinn:
> > I suggest we also remove all over the overrides which simply invoke super as it's unnecessary indirection.
> 
> @johnedquinn , what do you think about this suggestion from Robert? I presume he refers to these methods: https://github.com/partiql/partiql-lang-kotlin/blob/b2df81246961eee7348ffbd67452119ddb754f6a/partiql-lang/src/main/kotlin/org/partiql/lang/syntax/impl/PartiQLPigVisitor.kt#L1571C1-L1605C1 I have to admit I couldn't figure out the comment that precedes them. (Even if I maybe benefited from the IDE behavior without knowing they were responsible!)
> 
> This matter should be independent from the rest of this PR.

Yes, it really just has to do with IDE behavior -- would be fine to be stripped out. When you write `visit`, it brings up all of the methods that have not yet been overridden. Just helped me with the initial ANTLR work where I had to write hundreds of visit methods.

vgapeyev:
> > > I suggest we also remove all over the overrides which simply invoke super as it's unnecessary indirection.
> > 
> > 
> > @johnedquinn , what do you think about this suggestion from Robert? I presume he refers to these methods: https://github.com/partiql/partiql-lang-kotlin/blob/b2df81246961eee7348ffbd67452119ddb754f6a/partiql-lang/src/main/kotlin/org/partiql/lang/syntax/impl/PartiQLPigVisitor.kt#L1571C1-L1605C1 I have to admit I couldn't figure out the comment that precedes them. (Even if I maybe benefited from the IDE behavior without knowing they were responsible!)
> > This matter should be independent from the rest of this PR.
> 
> Yes, it really just has to do with IDE behavior -- would be fine to be stripped out. When you write `visit`, it brings up all of the methods that have not yet been overridden. Just helped me with the initial ANTLR work where I had to write hundreds of visit methods.

Oh see: this list would effectively **exclude** its methods from the IDE drop down that shows up when one types "override fun visit...".  I can see how that was helpful then, but not as much anymore.

partiql-lang/src/main/kotlin/org/partiql/lang/syntax/impl/PartiQLPigVisitor.kt
RCHowell: > deal with the null case in-line

Agreed. The visitOrNull is coming from Java where the sugar isn't there

RCHowell (approved):

vgapeyev:
@RCHowell thank you for the approval!  
Would you also approve #1161?  
I swear #1161 has nothing besides carrying out the changes from this #1144 all the way through, in detail, complete boring detail, and nothing than the boring detail!

vgapeyev:
Superseded with #1161, which contains all the commits from here.

vgapeyev:
Superseded with https://github.com/partiql/partiql-lang-kotlin/pull/1161, which contains all the commits from here.

PR #1150, opened 2023-07-12
Service Loader Directory

yuxtang-amazon:
## Relevant Issues
- [Closes/Related To] Issue #1149 

## Description
- Utilized URLClassLoader to enable Java Service Loader to load custom functions from `~/.partiql/plugins` directory.

## Other Information
YES

NO

NO

YES

## License Information


johnedquinn (changes requested):
Really neat implementation!

Some things to consider now are:
1. How can we show to the PR reviewers that this works?
2. How can we test that this works? Especially if someone in the far future accidentally makes a breaking change.

partiql-cli/build.gradle.kts
```diff
@@ -62,6 +61,26 @@ tasks.register<GradleBuild>("install") {
     tasks = listOf("assembleDist", "distZip", "installDist")
 }
 
+tasks.named("build").configure {
+    dependsOn("generateMockDBJAR")
+}
```
johnedquinn, line 66: Also, seems like this causes a bug.  If you run `./gradlew clean`, then run `./gradlew :partiql-cli:test`, there are failures.

This is likely caused by the order of task creation.

partiql-cli/src/main/kotlin/org/partiql/cli/pico/PipelineOptions.kt
```diff
@@ -70,14 +70,22 @@ internal class PipelineOptions {
     )
     var undefinedVarBehavior: UndefinedVariableBehavior = UndefinedVariableBehavior.ERROR
 
+    @CommandLine.Option(
+        names = ["--plugins"],
+        description = ["Directory containing plugin jar files"],
```
johnedquinn, line 75:
```suggestion
        description = ["Directory containing plugins"],
```

partiql-cli/src/main/kotlin/org/partiql/cli/pico/PipelineOptions.kt
```diff
@@ -70,14 +70,22 @@ internal class PipelineOptions {
...
+        paramLabel = "DIR"
+    )
+    var pluginDir: String = System.getProperty("user.home") + "/.partiql/plugins"
```
johnedquinn, line 78: Look at `PipelineOptions::environmentFile`. It could be useful to use a File instead. Then, when you create the default, you can use `File::resolve(".partiql").resolve("plugins")` -- which is system agnostic.

partiql-cli/src/main/kotlin/org/partiql/cli/utils/ServiceLoaderUtil.kt
```diff
@@ -92,20 +92,35 @@ import org.partiql.value.stringValue
...
+            val pluginFolders = pluginsDir.listFiles { file -> file.isDirectory }.orEmpty()
+            val files = pluginFolders.flatMap { folder ->
+                folder.listFiles { file -> file.isFile && file.extension == "jar" }.orEmpty().toList()
+            }
+            val plugins = if (files.isNotEmpty()) {
+                val classLoader = URLClassLoader.newInstance(files.map { it.toURI().toURL() }.toTypedArray())
+                ServiceLoader.load(Plugin::class.java, classLoader)
+            } else {
+                ServiceLoader.load(Plugin::class.java)
+            }
```
johnedquinn, line 123: For now, we can remove the `else`.

partiql-cli/build.gradle.kts
```diff
@@ -62,6 +61,26 @@ tasks.register<GradleBuild>("install") {
...
+
+val mockDbJarPath = "$buildDir/tmp/plugins/mockdb"
+val mockDbJarFile = "$buildDir/tmp/plugins"
+
+tasks.register<Jar>("generateMockDBJAR") {
+    dependsOn(":plugins:partiql-mockdb:build")
+    archiveBaseName.set("mockdb")
+    archiveVersion.set("0.1.0")
```
johnedquinn, line 74: Have you considered using [Gradle copy](https://docs.gradle.org/current/userguide/working_with_files.html#sec:copying_single_file_example) instead?

johnedquinn (changes requested):
Overall -- this is really great work. It's exciting to see how you've hacked the build to remove `testImplementation` and replace it with ServiceLoader.

partiql-cli/build.gradle.kts
```diff
@@ -62,6 +61,26 @@ tasks.register<GradleBuild>("install") {
     tasks = listOf("assembleDist", "distZip", "installDist")
 }
 
+tasks.named("build").configure {
+    dependsOn("generateMockDBJAR")
+}
+
+val mockDbJarPath = "$buildDir/tmp/plugins/mockdb"
+val mockDbJarFile = "$buildDir/tmp/plugins"
```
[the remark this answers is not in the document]
johnedquinn: Just regarding accuracy of naming ^

partiql-cli/src/main/kotlin/org/partiql/cli/utils/ServiceLoaderUtil.kt
yuxtang-amazon: I don't think we can remove the `else`. Because the default file path is "~/.partiql/plugins" and so the files might be empty.

johnedquinn: Since the CLI is an application, people don't use this utility directly.

Therefore, since we are the only users of `ServiceLoaderUtil` and since we don't provide any implementations of `Plugin`, we will never make use of the statement within the `else` block.

The only time we load `Plugins` is when external customers provide their implementations of `Plugin` -- which is through the filesystem.

johnedquinn (changes requested):

johnedquinn (approved):
Great work!

PR #1160, opened 2023-07-24
Removes project-all from the AST

RCHowell:
## Description

A project-item in the projection-list has been modeled as either an expression with an alias, or a project-all. The project-all is really a path expression with an unpivot at the end ie `foo.bar.*` but have had two ways of expressing this. This PR removes the duplication.

This is also matches our grammar

## Other Information
No, these changes are not released

No

No

Yes

## License Information


RCHowell:
Closing this as we don't want the syntax `t.* as x` to be allowed. This should be captured in the grammar, then we'll keep projectItemAll in the AST as this is a special spread projection

PR #1168, opened 2023-07-31
Add FunctionManager and Pluggable Functions Implementations

yuxtang-amazon:
## Relevant Issues
- [Closes/Related To] Issue #1149 

## Description
- Develops a FunctionManager for function overloading and redefines internal Function structures
- Defines PartiQLFunction and Plugin interfaces, converts PartiQLFunction to ExprFunctions and adds ServiceLoader to 
   PartiQL CLI to load pluggable functions
- Adds --plugin command line operation and completes unit tests & integration tests of Java Service Loader
- Moves PartiQLFunction parameters and signatures to partiql-types
- Adds @deprecated notice and give a replacement API to use in place of the deprecated methods

## Other Information
YES

YES
  - Deprecates Map<String, ExprFunction> representation of functions in the CompilerPipeline
  and experimental PartiQLCompilerPipeline. Please use List to represent functions instead.
  - Deprecates Arguments class, callWithOptional() and callWithVariadic() methods in the ExprFunction
  with a Deprecation Level of ERROR. Please invoke callWithRequired() instead.
  - Deprecates optionalParameter and variadicParameter in the FunctionSignature with a Deprecation
  Level of ERROR. Please use multiple implementations of ExprFunction and use the LIST ExprValue to
  represent variadic parameters instead.

NO

YES

## License Information


partiql-ast/src/main/kotlin/org/partiql/ast/helpers/ToLegacyAst.kt
johnedquinn, line 1: I believe these changes were accidentally included in this PR.

partiql-ast/src/main/pig/partiql.ion
johnedquinn, line 1: Remove.

partiql-ast/src/main/resources/partiql_ast.ion
johnedquinn, line 1: Remove

partiql-lang/src/main/kotlin/org/partiql/lang/graph/GpmlTranslator.kt
johnedquinn, line 1: Remove

partiql-lang/src/main/kotlin/org/partiql/lang/graph/Graph.kt
johnedquinn, line 1: Remove

partiql-lang/src/main/kotlin/org/partiql/lang/graph/SimpleGraph.kt
johnedquinn, line 1: Remove

partiql-lang/src/main/kotlin/org/partiql/lang/syntax/impl/PartiQLPigVisitor.kt
johnedquinn, line 1: Remove

partiql-lang/src/test/kotlin/org/partiql/lang/eval/TypedNullTests.kt
johnedquinn, line 1: Remove

partiql-parser/src/main/antlr/PartiQL.g4
johnedquinn, line 1: Remove

partiql-parser/src/main/antlr/PartiQLTokens.g4
johnedquinn, line 1: Remove

partiql-parser/src/main/kotlin/org/partiql/parser/impl/PartiQLParserDefault.kt
johnedquinn, line 1: Remove

partiql-lang/src/main/kotlin/org/partiql/lang/CompilerPipeline.kt
```diff
@@ -227,12 +223,28 @@ internal class CompilerPipelineImpl(
...
+
+    @Deprecated(
+        "Functions should be a list.",
+        level = DeprecationLevel.WARNING
+    )
+    constructor(
+        ion: IonSystem,
+        parser: Parser,
+        compileOptions: CompileOptions,
+        functionMap: Map<String, ExprFunction>,
```
johnedquinn, line 241:
```suggestion
        functions: Map<String, ExprFunction>,
```

The signature needs to match the original.

partiql-lang/src/main/kotlin/org/partiql/lang/CompilerPipeline.kt
```diff
@@ -47,7 +47,7 @@ data class StepContext(
      * Includes built-in functions as well as custom functions added while the [CompilerPipeline]
      * was being built.
      */
-    val functions: @JvmSuppressWildcards Map<String, ExprFunction>,
```
johnedquinn, line 50: We need to deprecate this constructor and offer another constructor that allows for a list.

partiql-lang/src/main/kotlin/org/partiql/lang/CompilerPipeline.kt
```diff
@@ -79,7 +79,7 @@ interface CompilerPipeline {
      * Includes built-in functions as well as custom functions added while the [CompilerPipeline]
      * was being built.
      */
-    val functions: @JvmSuppressWildcards Map<String, ExprFunction>
```
johnedquinn, line 82: We'll need to provide an alternative to the `functions` property (maybe `functionList`) and deprecate the old property.

partiql-lang/src/main/kotlin/org/partiql/lang/compiler/PartiQLCompilerBuilder.kt
```diff
@@ -62,6 +62,15 @@ class PartiQLCompilerBuilder private constructor() {
...
 
+    @Deprecated(
+        message = "customFunctions should be a list. Use PartiQLCompilerBuilder with List<ExprFunction> instead.",
+        level = DeprecationLevel.WARNING
+    )
+    constructor(customFunctions: Map<String, ExprFunction>) :
+        this() {
+            this.customFunctions = customFunctions.values.toList()
+        }
+
```
johnedquinn, line 73: It seems like we can delete this addition.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/ExprFunction.kt
```diff
@@ -17,10 +17,8 @@ package org.partiql.lang.eval
 import org.partiql.lang.errors.ErrorCode
 import org.partiql.lang.types.FunctionSignature
 
+@Deprecated("As function overloading is now supported, this class is deprecated.", level = DeprecationLevel.ERROR)
 sealed class Arguments
-data class RequiredArgs(val required: List<ExprValue>) : Arguments()
-data class RequiredWithOptional(val required: List<ExprValue>, val opt: ExprValue) : Arguments()
-data class RequiredWithVariadic(val required: List<ExprValue>, val variadic: List<ExprValue>) : Arguments()
```
johnedquinn, line 23: We'll need to retain these -- however, let's deprecate.

partiql-lang/src/test/kotlin/org/partiql/lang/syntax/PartiQLParserMatchTest.kt
johnedquinn, line 1: Remove

johnedquinn (changes requested):
Great work here! Seems like Vladimir's GPML PR was stripped with this PR, so you'll need to manually revert those files. Still have a couple files to review, but left some comments in the meantime.

partiql-lang/src/main/kotlin/org/partiql/lang/CompilerPipeline.kt
yuxtang-amazon: I don't think we need to deprecate this because the `data class StepContext` is only invoked inside `CompilerPipelineImpl` where we already add deprecate notice to functionMap. In this way,  `CompilerPipelineImpl` always convert functionMap to a list so StepContext will just use list.

partiql-lang/src/main/kotlin/org/partiql/lang/CompilerPipeline.kt
yuxtang-amazon: I don't think we need to deprecate this because the data class StepContext is only invoked inside CompilerPipelineImpl where we already add deprecate notice to the old functionMap. In this way, CompilerPipelineImpl always convert functionMap to a list so StepContext will just use list.

partiql-lang/src/main/kotlin/org/partiql/lang/CompilerPipeline.kt
yuxtang-amazon: I don't think we need to deprecate this because in `CompilerPipelineImpl`, I already add deprecate notice to the old functionMap. In this way, `CompilerPipelineImpl` always convert functionMap to a list so `CompilerPipeline` interface don't need to have the map representation.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/ExprFunction.kt
yuxtang-amazon: Since the `sealed class Arguments` has been deprecated, if I retain the 3 lines, we will get errors:
Using 'Arguments' is an error. As function overloading is now supported, this class is deprecated.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/ExprFunction.kt
```diff
@@ -98,9 +73,4 @@ interface ExprFunction {
  * @param session The current [EvaluationSession].
  * @param args The argument list.
  */
-fun ExprFunction.call(session: EvaluationSession, args: Arguments): ExprValue =
```
johnedquinn, line 101: Since this is a public function, we'll also need to deprecate this and add the alternative.

johnedquinn:
All looks great to me! Great job! Let's wait for another team member to review/approve before merging.

partiql-lang/src/main/kotlin/org/partiql/lang/CompilerPipeline.kt
```diff
@@ -227,14 +236,31 @@ internal class CompilerPipelineImpl(
     private val ion: IonSystem,
     private val parser: Parser,
     override val compileOptions: CompileOptions,
+    override val functionList: List<ExprFunction>,
     override val functions: Map<String, ExprFunction>,
```
RCHowell, line 240: We really don't need to change the field type. It's unfortunate that it is a map, but there could be some subtle behavior issues. Do we guarantee that a function's key in the map is the same as its name in the signature. Sure we do when we add via the builder, but that's not the case if someone provides the function themselves.

Could you please confirm the behavior of a function whose signature name differs than the one given in the map?

partiql-lang/src/main/kotlin/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -129,16 +127,28 @@ private typealias ThunkEnvValue<T> = ThunkValue<Environment, T>
...
+    @Deprecated(
+        "Functions should be a list. Use EvaluatingCompiler with List<ExprFunction> instead.",
+        level = DeprecationLevel.WARNING
+    )
+    constructor(
+        functions: Map<String, ExprFunction>,
+        customTypedOpParameters: Map<String, TypedOpParameter>,
+        procedures: Map<String, StoredProcedure>,
+        compileOptions: CompileOptions = CompileOptions.standard()
+    ) : this(functions.values.toList(), customTypedOpParameters, procedures, compileOptions)
```
RCHowell, line 144: Really if we want a list of functions internally, we could just use the `map.values` iterable and not have to add all of these deprecations and duplications. It's not ideal, but it does simplify things.

But again, using the list is different because we'll be using function names from the signature rather than the names in the given map (which may differ!)

partiql-lang/src/main/kotlin/org/partiql/lang/eval/ExprFunction.kt
```diff
@@ -48,39 +58,16 @@ interface ExprFunction {
...
-     * It is assumed that the [ExprFunction]s will always be called after validating
-     * the arguments for the `arity` and the types of arguments.
-     * [EvaluatingCompiler] validates the arguments before calling [ExprFunction]s.
-     * Hence the implementations are not required to deal with it.
-     *
-     * @param session The current [EvaluationSession].
-     * @param required The required arguments.
-     * @param opt The optional arguments.
-     */
+    @Deprecated("As function overloading is now supported, please provide multiple implementations of ExprFunction and invoke callWithRequired() instead.", level = DeprecationLevel.ERROR)
```
RCHowell, line 64:
```suggestion
    @Deprecated("Please define overloaded functions by providing each alternative as its own ExprFunction; each function is invoked by callWithRequired() rather than callWithOptional().", level = DeprecationLevel.ERROR)
```

partiql-lang/src/main/kotlin/org/partiql/lang/eval/impl/FunctionManager.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.partiql.types.StaticType
+
+/**
+ * Replaces the map of functions to achieve function overloading.
+ * Supports getting all functions and getting the function map by name
+ */
+internal class FunctionManager(
+    val functions: List<ExprFunction>
+) {
+    val functionMap: Map<String, List<ExprFunction>> = functions.groupBy { it.signature.name }
```
RCHowell, line 18: make private

partiql-lang/src/test/kotlin/org/partiql/lang/eval/evaluatortestframework/PartiQLCompilerPipelineFactory.kt
```diff
@@ -81,7 +81,7 @@ internal class PartiQLCompilerPipelineFactory() : PipelineFactory {
             compiler = PartiQLCompilerBuilder.standard()
                 .options(evaluatorOptions)
                 .customTypes(legacyPipeline.customDataTypes)
-                .customFunctions(legacyPipeline.functions.values.toList())
```
RCHowell, line 84: Here's where we get a hint that the function name within the key is lost when compiling the plan

partiql-types/src/main/kotlin/org/partiql/types/function/FunctionParameter.kt
```diff
@@ -0,0 +1,15 @@
...
+ * This currently only contains [ValueParameter], however, this is designed in this
+ * manner to allow for other types of parameters in the future. TypeParameter may be
+ * of use.
+ */
+public sealed class FunctionParameter {
+    public class ValueParameter(
+        public val name: String, // A human-readable name to help clarify use
+        public val type: PartiQLValueType // The parameter's type
+    ) : FunctionParameter()
+}
```
RCHowell, line 15: These are public and published so I think we should use javadoc style comments

RCHowell (changes requested):

docs/wiki/tutorials/Pluggable Functions Tutorial.md
```diff
@@ -0,0 +1,136 @@
...
+
+1. In your project's main resources directory (usually `src/main/resources`), create a new directory named `META-INF/services`.
+
+2. Within this directory, create a new file named after the full interface name as `org.partiql.spi.Plugin`.
+
+3. Inside this file, write the fully qualified name of your Plugin implementation class. If you have multiple implementations, each should be listed on a new line.
+
+Here's an example if your `Plugin` implementation class is `org.partiql.plugins.mockdb.LocalPlugin`:
+```Kotlin
+org.partiql.plugins.mockdb.LocalPlugin
```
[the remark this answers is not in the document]
yuxtang-amazon: Thanks for the comment. I want to clarify if the Java documentation is https://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html#jls-13.1

RCHowell (approved):

johnedquinn (approved):

PR #1169, opened 2023-07-31
Errors in types

RCHowell:
## Relevant Issues

Moves Errors out of lang and drops the ExprValue dependency. This is a prerequisite to further modularization

## Description

See CHANGELOG entry.

Moves `org.partiql.lang.errors` in partiql-lang to `org.partiql.errors` in partiql-types — as we move towards unifying the error handling logic across the various stages of compilation.

## Other Information
YES

YES

NO

YES

## License Information


vgapeyev (commented):
I am ok with this move in principle -- short of some sort of a holistic re-design/simplification of error handling, we have to do something like this to keep moving with more interesting matters of modularization. 

However, since many previously-internal things in the moved files now became public, this raises the perennial concern about increasing the versionable surface. Any thoughts about mitigating that?  Communication/indication to users that these are not really public/versionable is probably the only thing feasible, but through what mechanism?

partiql-lang/src/test/kotlin/org/partiql/lang/errors/ParserErrorsTest.kt
```diff
@@ -12,7 +12,7 @@
  *  language governing permissions and limitations under the License.
  */
 
-package org.partiql.lang.errors
+package org.partiql.errors
```
johnedquinn, line 15: For all the `partiql-lang/src/test/kotlin/org/partiql/lang/errors` tests, was this change intentional?

partiql-types/src/main/kotlin/org/partiql/errors/ErrorCode.kt
```diff
@@ -12,37 +12,34 @@
...
 private val LOC_TOKEN = LOCATION + (TOKEN_INFO)
 private val LOC_TOKEN_STR = LOCATION + (setOf(Property.TOKEN_STRING))
 
-enum class ErrorBehaviorInPermissiveMode {
+public enum class ErrorBehaviorInPermissiveMode {
     THROW_EXCEPTION, RETURN_MISSING
 }
 
-internal const val UNBOUND_QUOTED_IDENTIFIER_HINT =
+public const val UNBOUND_QUOTED_IDENTIFIER_HINT: String =
```
johnedquinn, line 27: Does this need to be `public`?

partiql-types/src/main/kotlin/org/partiql/errors/ErrorCode.kt
```diff
@@ -365,6 +362,12 @@ enum class ErrorCode(
...
 
+    /**
+     * NOTE: This is unused and may be removed!
+     *
+     * To remove the dep on ExprValue, we introduce the EXPECTED_ARGUMENT_TYPES argument
+     *  Property.EXPECTED_ARGUMENT_TYPES -> ExprValueType.values().filter { it.isText }
+     */
     EVALUATOR_CONCAT_FAILED_DUE_TO_INCOMPATIBLE_TYPE(
         ErrorCategory.EVALUATOR,
         LOCATION + setOf(Property.ACTUAL_ARGUMENT_TYPES),
```
johnedquinn, line 373:
```suggestion
        LOCATION + setOf(Property.ACTUAL_ARGUMENT_TYPES, Property.ACTUAL_ARGUMENT_TYPES,),
```

I believe you need to add this.

johnedquinn (changes requested):

partiql-lang/src/test/kotlin/org/partiql/lang/errors/ParserErrorsTest.kt
RCHowell: No it was the find-replace. It shouldn't matter where the tests live, but I think we should keep them in the existing path to minimize the diff

partiql-types/src/main/kotlin/org/partiql/errors/ErrorCode.kt
RCHowell: It is used in two places, but honestly need not exist because it's just a debug tring https://github.com/search?q=repo%3Apartiql%2Fpartiql-lang-kotlin%20UNBOUND_QUOTED_IDENTIFIER_HINT&type=code

We can remove this (which also relates to Vladimir's concern)

partiql-types/src/main/kotlin/org/partiql/errors/ErrorCode.kt
RCHowell: Do you mind clarifying why? Did you mean to put the same thing twice?

johnedquinn: Oh, sorry: I meant to add `EXPECTED_ARGUMENT_TYPES` instead.

And it loosely tells the end user what properties are available. So, since your expect `EXPECTED_ARGUMENT_TYPES` to be in the property value map, it can be accessed by the end-user.

johnedquinn (approved):

PR #1180, opened 2023-08-16
Adds syntactical support for CREATE FUNCTION

johnedquinn:
# DRAFT NOTES
- Marking as draft as there are a couple things needed:
  1. CHARACTER without an argument assumes the length to be 1. However, in order to parse `CHARACTER(<CML>)`, where `<CML>` is the implementation-defined max-length, we don't have a way to say unlimited length. 
  2. Similar scenarios for FLOAT, DECIMAL, NUMERIC, etc. Instead of parsing the arguments -- I removed the arguments as we don't have implementation-defined maximums.
  3. We need data classes for AST so we can assert equivalence to expected ASTs
  4. Also leaving it open-ended for testing failures. Should we assert error locations?
  5. Left the SQL:1999 relevant grants in a README in `resources`.

## Relevant Issues
- None

## Description
- Adds syntactical support for CREATE FUNCTION to the new parser.
- Adds the CREATE FUNCTION statement to Sprout AST definition
- Adds SQL:1999's Built-In Functions to resources. Converted several statements to comply with PartiQL (PartiQL doesn't have implementation-defined maximums for string length, precision, etc. Also, PartiQL doesn't support types such as INTERVAL, BIT, and BIT VARYING -- left those out).
  - I did not add support for relevant `GRANT` statements for the built-in functions.

## Other Information
  - Not yet

## License Information


RCHowell:
Do you know of, or have an example of vararg and type parameters?

johnedquinn:
> Do you know of, or have an example of vararg and type parameters?

No -- this PR includes what is specified in SQL:1999 for `<function specification>`, which doesn't support variadic parameters or type parameters. Even in SQL:2023, I don't see support for variadic or type parameters, but I'd need to double-check.

RCHowell:
Related to https://github.com/partiql/partiql-docs/blob/function-resolution/docs/functions.adoc

johnedquinn:
As this is not a priority, I'm closing this and saving the state in fork: https://github.com/johnedquinn/partiql-lang-jvm/releases/tag/v0.12.1-create-function. I'll pick this back up when necessary.

PR #1181, opened 2023-08-17
Initializes PartiQL Planner sub-project

johnedquinn:
## Relevant Issues
- None

## Description
- Initializes PartiQL Planner sub-project
- Migrates existing planner to new sub-project
- See the CHANGELOG for all relevant information

## Notes
- I had to make some things public, but, when possible, I annotated them with experimental
- Testing is meh due to equals function not being generated. Had to work around this. See the notes in the new sub-projects build.gradle.kts and the test files.

## Other Information
  - Yes, see CHANGELOG.

## License Information


partiql-lang/src/main/kotlin/org/partiql/lang/compiler/PartiQLCompilerPipeline.kt
```diff
@@ -56,7 +56,7 @@ class PartiQLCompilerPipeline(
         @JvmStatic
         fun standard() = PartiQLCompilerPipeline(
             parser = PartiQLParserBuilder.standard().build(),
-            planner = PartiQLPlannerBuilder.standard().build(),
+            planner = PartiQLPhysicalPlannerBuilder.standard().build(),
```
RCHowell, line 59: I'm curious why this is all being renamed to PartiQLPhysicalPlannerBuilder rather than keeping as is. Recall that the only differences between the logical and physical domains is the "operation" key which means the "physical" planning responsibilities are almost nothing. I had envisioned the planner is purely logical, and the compiler is responsible for adding the operators. 

We would need to define how one might decorate a logical plan with some name, but the physical plan plays no substantial role to warrant a big split.

partiql-lang/src/main/kotlin/org/partiql/lang/planner/PartiQLPhysicalPass.kt
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+package org.partiql.lang.planner
+
+import org.partiql.lang.domains.PartiqlPhysical
+
+/**
+ * Defines a [PartiQLPlannerPass] of a [PartiqlPhysical.Plan]
+ */
+fun interface PartiQLPhysicalPass : PartiQLPlannerPass<PartiqlPhysical.Plan>
```
RCHowell, line 22: I know this is similar to today, but I believe we be focusing on making the logical domain the public interface for planning and rewriting as the logical plan is what we transport. The physical plan stuff is really Kotlin internal only.

Similar comment from substrait
> There is no true distinction between logical and physical operations in Substrait. By convention, certain operations are classified as physical, but all operations can be potentially used in any kind of plan. A particular set of transformations or target operators may (by convention) be considered the “physical plan” but this is a characteristic of the system consuming substrait as opposed to a definition within Substrait.

That is, the physical part is system specific, but what is defined by the PartiQL Canonical Plan (ie public interface) is not physical

RCHowell: But because the "impl" field is so minimal, I'm beginning to think each we should allow annotating logical plan nodes

RCHowell (changes requested):

partiql-lang/src/main/kotlin/org/partiql/lang/compiler/PartiQLCompilerPipeline.kt
johnedquinn: Given that we're only trying to move out the planner (returning a logical plan) to a new subproject (with `PartiQLPlanner` being its name), I decided to adjust `:partiql-lang`'s planner (returning a physical plan) to `PartiQLPhysicalPlanner`. My thinking was that, eventually, we'll want to decide on where the physical planning/compiling components should be moved, but for now, we really just need to strip out the logical plan components and make it *slightly* easier to tell which component returns a logical plan vs a physical plan.

This way, consumers of the experimental `PartiQLCompilerPipeline` can still use the API almost as-is, and we can begin modifying the `PartiQLPlanner` to return the Sprout-generated logical plan.

partiql-lang/src/main/kotlin/org/partiql/lang/planner/PartiQLPhysicalPass.kt
johnedquinn: I'd personally recommend implementing this in a separate PR. This code linked doesn't change any API. I just moved this top-level interface from PartiQLPlannerPass.kt to conform with one top-level interface/class per file.

RCHowell:
Closing as this was rebased on partiql-planner, with the moved code landing in paritql-eval https://github.com/partiql/partiql-lang-kotlin/commit/86b8ddb03506d786dc3f78e0b46483a9ff4a3c54

PR #1182, opened 2023-08-21
Adds support for Sprout-generated equality/hashCode

johnedquinn:
## Relevant Issues
- None

## Description
- Converts generation of Sprout product types to be abstract classes
- Adds default  equality and hashCode methods to each product
- Previous generation resulted in interfaces
- Added a `:test/sprout-tests` subproject to run integration tests on generated code

## Other Information
  - Yes, see CHANGELOG
  and [Code Style Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CODE_STYLE.md)? **NO**

## License Information

### DIFF
--- a/lib/sprout/src/main/kotlin/org/partiql/sprout/generator/target/kotlin/KotlinGenerator.kt
+++ b/lib/sprout/src/main/kotlin/org/partiql/sprout/generator/target/kotlin/KotlinGenerator.kt
@@ -2,11 +2,13 @@ package org.partiql.sprout.generator.target.kotlin
 
 import com.squareup.kotlinpoet.AnnotationSpec
 import com.squareup.kotlinpoet.ClassName
+import com.squareup.kotlinpoet.CodeBlock
 import com.squareup.kotlinpoet.FunSpec
 import com.squareup.kotlinpoet.KModifier
 import com.squareup.kotlinpoet.ParameterSpec
 import com.squareup.kotlinpoet.PropertySpec
 import com.squareup.kotlinpoet.TypeSpec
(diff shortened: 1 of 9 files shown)


lib/sprout/src/main/kotlin/org/partiql/sprout/generator/target/kotlin/KotlinGenerator.kt
```diff
@@ -192,4 +188,51 @@ class KotlinGenerator(private val options: KotlinOptions) : Generator<KotlinResu
...
+    /**
+     * Adds `equals` method to the core abstract class
+     */
+    private fun KotlinNodeSpec.Product.addEqualsMethod() {
+        val equalsFunctionBodyBuilder = CodeBlock.builder().let { body ->
+            body.addStatement("if (this === other) return true")
+            body.addStatement("if (other !is %T) return false", this.clazz)
+            this.props.forEach { prop ->
+                body.addStatement("if (%N != other.%N) return false", prop.name, prop.name)
+            }
```
RCHowell, line 201: I think we should deeply compare collections as well

RCHowell (changes requested):

lib/sprout/src/main/kotlin/org/partiql/sprout/generator/target/kotlin/KotlinGenerator.kt
johnedquinn: This should be deeply comparing as-is. I just pushed tests to show some examples. Thanks!

RCHowell (approved):

PR #1184, opened 2023-08-25
Changes PartiQLValue types to be nullable

RCHowell:
## Description

Makes all PartiQLValues nullable in PartiQLValue as this matches IonJava as well as SQL semantics. 

This PR includes:

- Addition of PartiQLValueType.ANY
- All runtime types `T` are the static type `T|NULL`
- General cleanup of reading/writing Ion values
- Defines equality for collection types
- Includes lookup optimized version of the `StructValue`
- Extract `toIon` behavior and remove that from `org.partiql.ast`. Now the Ion writer and `toLegacyAst` helper use the same logic.
- Adds PartiQLValueTextWriterBuilder
- Adds detailed README

**RENDERED** — https://github.com/partiql/partiql-lang-kotlin/blob/goodbye-nulls/partiql-types/README.adoc

## Other Information
YES

YES

NO

YES

## License Information


partiql-types/src/main/kotlin/org/partiql/value/PartiQLValue.kt
```diff
@@ -405,398 +446,115 @@ public abstract class SexpValue<T : PartiQLValue> : CollectionValue<T> {
...
+     * @param other
+     * @return
+     */
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other !is StructValue<*>) return false
+        if (other.annotations != this.annotations) return false
+
+        // both null.struct
+        if (other.isNull && this.isNull) return true
```
johnedquinn, line 506:
```suggestion
        if (other.isNull && this.isNull) return true
        if (other.isNull || this.isNull) return false
```
Same goes for bag/list/sexp.

johnedquinn (commented):
Overall looks great! Minor comments.

partiql-types/src/main/kotlin/org/partiql/value/impl/BagValueImpl.kt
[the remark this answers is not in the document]
RCHowell: ack

partiql-types/src/main/kotlin/org/partiql/value/PartiQLValue.kt
RCHowell: Good catch. I think this would do the trick.

- true iff both null
- false iff only one is null
- continues if both non-null

```suggestion
if (this.isNull || other.isNull) return this.isNull == other.isNull
```

johnedquinn (approved):
Thanks for this! LGTM.

PR #1186, opened 2023-08-28
Updates FunctionSignature with method specifications

RCHowell:
## Relevant Issues
https://github.com/partiql/partiql-docs/pull/48

## Description

Adds `isNullCall` and `isNullable` to function properties, the former being part of SQL data definition.

IsNullable is added to tell if the return type is nullable (default) or not, as there are certain operators which can statically know to never return null and we wish to model this for the static type inferencer.

## Other Information
YES

YES

NO

YES 

## License Information


partiql-types/src/main/kotlin/org/partiql/types/function/FunctionSignature.kt
```diff
@@ -8,11 +8,83 @@ import org.partiql.types.PartiQLValueType
...
+     */
+    private val specific = buildString {
+        append(name.uppercase())
+        append("__")
+        append(parameters.joinToString("_") { it.type.name })
+        append("__")
+        append(returns.name)
+    }
+
+    override fun toString(): String = buildString {
```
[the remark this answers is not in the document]
RCHowell: It does. I will add these.

partiql-types/src/main/kotlin/org/partiql/types/function/FunctionSignature.kt
```diff
@@ -8,11 +8,103 @@ import org.partiql.types.PartiQLValueType
...
+            if (p1.type != p2.type) return false
+        }
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = name.hashCode()
+        result = 31 * result + returns.hashCode()
+        result = 31 * result + parameters.hashCode()
+        result = 31 * result + isDeterministic.hashCode()
```
johnedquinn, line 106:
```suggestion
        result = 31 * result + isDeterministic.hashCode()
        result = 31 * result + isNullCall.hashCode()
        result = 31 * result + isNullable.hashCode()
```

partiql-types/src/main/kotlin/org/partiql/types/function/FunctionSignature.kt
```diff
@@ -8,11 +8,103 @@ import org.partiql.types.PartiQLValueType
...
+            other.parameters.size != parameters.size
+        ) {
+            return false
+        }
+        // all other parts equal, compare parameters (ignore names)
+        for (i in parameters.indices) {
+            val p1 = parameters[i]
+            val p2 = other.parameters[i]
+            if (p1.type != p2.type) return false
+        }
```
johnedquinn, line 98: Is this just to avoid checking equality for parameter names?

Also, if we do this, we'll just need to make sure `other.parameters.size == this.parameters.size` so we don't get an `OutOfBoundsException`

johnedquinn (changes requested):

partiql-types/src/main/kotlin/org/partiql/types/function/FunctionSignature.kt
```diff
@@ -8,11 +8,103 @@ import org.partiql.types.PartiQLValueType
...
+            for (i in parameters.indices) {
+                val p = parameters[i]
+                val ws = (extent - p.name.length) + 1
+                appendLine()
+                append(indent).append(p.name.uppercase()).append(" ".repeat(ws)).append(p.type.name)
+                if (i != parameters.size - 1) append(",")
+            }
+        }
+        appendLine(" )")
+        append(indent).appendLine("RETURNS $returns")
```
am357, line 74: Is `returns` syntax going to cater for `isNullable`?

RCHowell: Unsure at the moment. I've only included what is currently valid SQL syntax

partiql-types/src/main/kotlin/org/partiql/types/function/FunctionSignature.kt
RCHowell: That check is already there

johnedquinn (approved):

partiql-types/src/main/kotlin/org/partiql/types/function/FunctionSignature.kt
am357: Can we have a TODO? or even having it as part of SQL's comment (`--`) in the Syntax?

johnedquinn (approved):

PR #1190, opened 2023-09-01
Fixes bug in FilterScanToKeyLookupPass

dlurton:
## Relevant Issues
N/A.

## Description

`FilterScanToKeyLookupPass` will now properly ignore primary key equality expressions where both sides reference the candidate row are rewritten.  i.e.:

    SELECT * FROM foo AS f WHERE f.id = f.bar

... was previously being rewritten to:

    (bindings_to_values
        (local_id 0)
        (project
            (impl <user-impl>)
            (var_decl 0)
            (path (local_id 0) (lit bar) (case_insensitive)))
        )
    )

Which is an invalid plan because when the dynamic argument to the project operator is evaluated the local variable 0 hasn't been initialized yet.

## Other Information
 See the diff.

No.

No

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,9 @@ classes in `:partiql-ast` and `:partiql-plan`.
 
 ### Fixed
 
+- Fix a bug in `FilterScanToKeyLookup` pass wherein it was rewriting primary key equality expressions with references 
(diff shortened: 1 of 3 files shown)


yliuuuu (approved):
LGTM.

PR #1197, opened 2023-09-06
Timestamp staticType

yliuuuu:
## Relevant Issues
- [Closes/Related To] N/A

## Description
- Add Timestamp StaticType. 
- The idea is to introduce minimal backward incompatible change. 
- Customer should not need to opt-in for timestamp class, or modify their existing codebase if they wish to continue use the ion timestamp. 
- However, if customer want to specify added feature such as precision, with/without timezone, then an explicit opt-in is required. 

## Other Information
  - Yes. 

 - Should not be. But would appreciate a second option on this. 
 

  - No. 

Yes. 

## License Information

### DIFF
--- a/partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
+++ b/partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
@@ -373,11 +373,13 @@ public class TimestampType internal constructor(internal val timestampType: Part
     @PartiQLTimestampExperimental
     public val withTimeZone: Boolean = timestampType.withTimeZone
 
-    // Preserve the original semantics. An arbitrary timestamp type without time zone
+    // Preserve the original semantics (ion timestamp). An arbitrary timestamp type with time zone.
     public constructor(metas: Map<String, Any> = mapOf()) : this(PartiQLTimestampType(null, true, metas))
 
     /**
-     * @param precision specifies the number of digits in the fractional seconds. If omitted, the default value is 6.
+     * @param precision specifies the number of digits in the fractional seconds.
+     *  If omitted, the default value is 6 as specified in the SQL spec.
+     *  If null, then the timestamp can have arbitrary constraint.
      * @param withTimeZone If true, then the underlying data must be associated with either unknown timezone(-00:00) or an UTC offset.
      * @param metas Metadata associated with the timestamp type.
      */
--- a/partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
+++ b/partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
@@ -372,7 +372,15 @@ public class TimestampType internal constructor(internal val timestampType: Part
     public val precision: Int? = timestampType.precision
     @PartiQLTimestampExperimental
     public val withTimeZone: Boolean = timestampType.withTimeZone
-    public constructor(metas: Map<String, Any> = mapOf()) : this(PartiQLTimestampType(metas = metas))
+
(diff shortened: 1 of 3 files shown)


partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
```diff
@@ -360,13 +364,70 @@ public data class TimeType(
...
+
+    override val metas: Map<String, Any> = timestampType.metas
+    override val allTypes: List<StaticType> = listOf(this)
+    @PartiQLTimestampExperimental
+    public val precision: Int? = timestampType.precision
+    @PartiQLTimestampExperimental
+    public val withTimeZone: Boolean = timestampType.withTimeZone
+
+    // Preserve the original semantics. An arbitrary timestamp type without time zone
+    public constructor(metas: Map<String, Any> = mapOf()) : this(PartiQLTimestampType(null, true, metas))
```
johnedquinn, line 377:
```suggestion
    public constructor(metas: Map<String, Any> = mapOf()) : this(PartiQLTimestampType(null, false, metas))
```

partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
```diff
@@ -360,13 +364,70 @@ public data class TimeType(
...
+
+    /**
+     * @param precision specifies the number of digits in the fractional seconds. If omitted, the default value is 6.
```
johnedquinn, line 380: Might be good to mention why 6 is default. Non-blocking.

partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
yliuuuu: Actually. The comment is wrong. Ion Timestamp always has time zone ( unknown or UTC-offset).

partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
```diff
@@ -360,13 +364,70 @@ public data class TimeType(
...
 
-public data class TimestampType(override val metas: Map<String, Any> = mapOf()) : SingleType() {
-    override val allTypes: List<StaticType>
-        get() = listOf(this)
+public class TimestampType internal constructor(internal val timestampType: PartiQLTimestampType = PartiQLTimestampType()) : SingleType() {
+
+    override val metas: Map<String, Any> = timestampType.metas
+    override val allTypes: List<StaticType> = listOf(this)
+    @PartiQLTimestampExperimental
+    public val precision: Int? = timestampType.precision
```
johnedquinn, line 372: Null precision means arbitrary precision as defined in Ion I'm assuming?

partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
yliuuuu: I will update the comment.

partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
yliuuuu: Yes. Will add to kdoc.

partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
yliuuuu: Will add to kdoc.

johnedquinn (approved):

PR #1211, opened 2023-09-15
Adds support for thread interruption in compilation and execution

johnedquinn:
## Relevant Issues
- None

## Description
- Adds support for thread interruption in compilation and execution
- Adds support for CLI users to use CTRL-C to cancel long-running compilation or execution of queries.
- Adds an `isInterruptible` flag to `CompileOptions` to allow users to opt-in.

## Impact Analysis
- Measured the compilation and execution of long-running queries (7 cross joins, 10 cross joins with aggregations, 10 cross joins with aggregation functions and grouping). See the JMH changes in this PR. We saw an average speedup of:
  - Compilation: -2.36%
  - Compilation and Evaluation: -1.35%
  - Compilation, Evaluation, and Materialization: -0.71%
  - Total: -1.47%
- While the impact to speed is low, I still opted to added a flag (`isInterruptible`) to the `CompileOptions` to allow users to opt-in to the new changes.

## Other Information

## License Information

### DIFF
--- a/partiql-lang/src/jmh/kotlin/org/partiql/jmh/benchmarks/CompilerInterruptionBenchmark.kt
+++ b/partiql-lang/src/jmh/kotlin/org/partiql/jmh/benchmarks/CompilerInterruptionBenchmark.kt
@@ -1,0 +1,307 @@
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at:
+ *
+ *      http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ */
+
+package org.partiql.jmh.benchmarks
+
+import org.openjdk.jmh.annotations.Benchmark
+import org.openjdk.jmh.annotations.BenchmarkMode
+import org.openjdk.jmh.annotations.Fork
+import org.openjdk.jmh.annotations.Measurement
+import org.openjdk.jmh.annotations.Mode
+import org.openjdk.jmh.annotations.OutputTimeUnit
+import org.openjdk.jmh.annotations.Scope
+import org.openjdk.jmh.annotations.State
+import org.openjdk.jmh.annotations.Warmup
+import org.openjdk.jmh.infra.Blackhole
+import org.partiql.jmh.utils.FORK_VALUE_RECOMMENDED
+import org.partiql.jmh.utils.MEASUREMENT_ITERATION_VALUE_RECOMMENDED
+import org.partiql.jmh.utils.MEASUREMENT_TIME_VALUE_RECOMMENDED
+import org.partiql.jmh.utils.WARMUP_ITERATION_VALUE_RECOMMENDED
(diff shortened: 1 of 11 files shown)


partiql-lang/src/jmh/kotlin/org/partiql/jmh/benchmarks/CompilerInterruptionBenchmark.kt
```diff
@@ -0,0 +1,307 @@
...
+    @Fork(value = FORK_VALUE)
+    @Measurement(iterations = MEASUREMENT_ITERATION_VALUE, time = MEASUREMENT_TIME_VALUE)
+    @Warmup(iterations = WARMUP_ITERATION_VALUE, time = WARMUP_TIME_VALUE)
+    fun evalCrossJoinWithoutInterruptible(state: MyState, blackhole: Blackhole) {
+        val result = state.pipelineWithoutInterruption.compile(state.crossJoinsAst).evaluate(state.session) as PartiQLResult.Value
+        val value = result.value
+        blackhole.consume(value)
+    }
+
+//    TODO: See the note in the KDoc for the class.
```
yliuuuu, line 198: I understand that the eval test including aggregation is slow because we materialize all the data during aggregation. But I am not fully convinced that we need to comment those out through.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -150,6 +150,15 @@ internal open class EvaluatingCompiler(
             "compilationContextStack was empty.", ErrorCode.EVALUATOR_UNEXPECTED_VALUE, internal = true
         )
 
+    private val interruptionCheck: () -> Unit = when (compileOptions.interruptible) {
```
yliuuuu, line 153: Could you please add a comment here stating out where we currently conduct the interruptionCheck?

partiql-lang/src/main/kotlin/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -150,6 +150,15 @@ internal open class EvaluatingCompiler(
...
+        true -> { ->
+            if (Thread.interrupted()) {
+                throw InterruptedException()
+            }
+        }
+        false -> { -> }
```
yliuuuu, line 159: Nit: I found the false -> { -> } a little weird. Maybe: 
```suggestion
        true -> { _ -> if (Thread.interrupted()) throw InterruptedException() }
        false -> { _ -> Unit}
```

yliuuuu (changes requested):
As discussed offline, the choice of where we place the interruptionCheck makes sense to me. It would be nice if you could leave a comment stating where we place the interruptionCheck, otherwise it looks good to me.

partiql-lang/src/jmh/kotlin/org/partiql/jmh/benchmarks/CompilerInterruptionBenchmark.kt
johnedquinn: They will run on every push to `main` according to our GH workflows.

yliuuuu: Point being the fact that they are slow does not interfere with the ability of them being a valid baseline for aggregation performance isn't it?

yliuuuu: Hmm wait. The measurement would be incorrect because those test includes materialization while others does not. If that is the case, then I would think commenting them out should be good. 

Maybe update the kdoc to reflect this?

yliuuuu (approved):
LGTM

PR #1218, opened 2023-09-20
Adds preliminary support for JOINs.

johnedquinn:
## Relevant Issues
- None

## Description
- Adds preliminary support for JOINs.
- We need to add support for `SELECT *` so that we can perform more comprehensive checks. Will add to follow-up PR.

## Other Information
  - Not `main`.

  - All changes to `plan_0_1` haven't been released yet.

## License Information

### DIFF
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/transforms/RelConverter.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/transforms/RelConverter.kt
@@ -17,6 +17,7 @@ import org.partiql.plan.Rex
 import org.partiql.planner.Env
 import org.partiql.types.StaticType
 import org.partiql.value.PartiQLValueExperimental
+import org.partiql.value.boolValue
 import org.partiql.value.stringValue
 
 /**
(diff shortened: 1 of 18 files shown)


partiql-planner/src/main/kotlin/org/partiql/planner/transforms/RelConverter.kt
```diff
@@ -244,14 +245,24 @@ internal object RelConverter {
...
             val props = emptySet<Rel.Prop>()
-            val joinType = convertJoinType(node)
+            val condition = node.condition?.let { RexConverter.apply(it, env) } ?: rex(StaticType.BOOL, rexOpLit(boolValue(true)))
+            val joinType = when (node.type) {
+                From.Join.Type.LEFT_OUTER, From.Join.Type.LEFT -> Rel.Op.Join.Type.LEFT
+                From.Join.Type.RIGHT_OUTER, From.Join.Type.RIGHT -> Rel.Op.Join.Type.RIGHT
+                From.Join.Type.FULL_OUTER, From.Join.Type.FULL -> Rel.Op.Join.Type.FULL
+                From.Join.Type.COMMA,
+                From.Join.Type.INNER,
+                From.Join.Type.CROSS -> Rel.Op.Join.Type.INNER // Cross Joins are just INNER JOIN ON TRUE
```
RCHowell, line 261: From the spec

```
The FROM clause expressions:

l , r ⇔ l CROSS JOIN r ⇔ l JOIN r ON TRUE
```

johnedquinn: Yes, exactly.

So, from the SQL:1999 (Section 7.7, Syntax Rules #4):
```
If a <qualified join> or <natural join> is specified and a <join type> is not specified, then INNER
is implicit.
```

So, even more specifically:
```
l , r <==> l CROSS JOIN r <==> l JOIN r ON TRUE <==> l INNER JOIN r ON TRUE
```

RCHowell: Just putting the clarification here. The representation is much nicer than what was there before. I don't see why we wouldn't merge these changes now along with the CASE typing and transpilation

PR #1226, opened 2023-09-26
Add `EXCLUDE` to the parser, ast, plan, and plan schema inferencer

alancai98:
## Relevant Issues
- Spec issue https://github.com/partiql/partiql-spec/issues/39

## Description
Experimental addition for the `EXCLUDE` operator to the
- parser
- ast
- partiql-plan
- partiql-plan schema inferencer

This operator serves to remove nested attributes and values from binding tuples before they're projected using `SELECT`/`PIVOT`.

## Other Information
- Updated Unreleased Section in CHANGELOG: Yes

- Any backward-incompatible changes? No.

  and [Code Style Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CODE_STYLE.md)? Yes

## License Information


partiql-lang/src/main/kotlin/org/partiql/lang/planner/transforms/plan/PlanTyper.kt
```diff
@@ -227,7 +338,8 @@ internal object PlanTyper : PlanRewriter<PlanTyper.Context>() {
         return node.copy(
             input = input,
             common = node.common.copy(
-                typeEnv = typeEnv
+                typeEnv = typeEnv,
+                properties = input.getProperties()
```
alancai98, line 342: This change was for fixing the `ORDER BY` schema inference behavior.

Tests with corrected behavior: https://github.com/partiql/partiql-lang-kotlin/blob/e1582745db1fc79929d11ecd3d924e131d4bcf9c/partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt#L658-L678.

Previously these tests would output a bag rather than a list.

CHANGELOG.md
```diff
@@ -29,13 +29,15 @@ Thank you to all who have contributed!
 ## [Unreleased]
 
 ### Added
+- **EXPERIMENTAL** Adds `EXCLUDE` to parser, ast, plan, and plan schema inferencer
```
am357, line 32: Could we also be explicit and make it clear that we are not supporting the evaluation at this stage.

am357 (commented):
Thank you for taking working on this feature!

I didn't pay too much attention to the algorithm as it's implemented in the `PlanTyper`. I focused more on the schema inferencer tests and left couple of comment. I also didn't bother with Kotlin style recommendation and alike.

CHANGELOG.md
alancai98: Sure, added a note to the CHANGELOG

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
```diff
@@ -128,6 +132,91 @@ class PartiQLSchemaInferencerTests {
         }
     }
 
+    @ParameterizedTest
```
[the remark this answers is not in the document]
alancai98: Thanks for bringing those cases up. 

> Does it make sense to also include a negative test to show that the exclude doesn't work for the following case:
> 
> ```sql
> SELECT * EXCLUDE t1 FROM Table1 AS t1, Table2 AS t2
> ```

This was previously syntactically correct. In the latest commit (https://github.com/partiql/partiql-lang-kotlin/pull/1226/commits/801437e9d5b6e72190aed6687eca1e71fbd62577), I changed the grammar to require one or more `excludeExprSteps`. So `EXCLUDE t1` is no longer syntactically correct. Added a couple parse error tests to `PartiQLParserTest.kt`.

---

> Also do we support select list(E.g., `SELECT t.a EXCLUDE`)? if so, can we include the tests for that?

There was an existing test with a select list: https://github.com/partiql/partiql-lang-kotlin/blob/e7bc9ec80efbec1c105239745d7b82292d13c000/partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt#L1355-L1382, though the name was incorrect. I also added a new test w/ `SELECT VALUE`.

partiql-parser/src/main/antlr/PartiQL.g4
```diff
@@ -334,6 +334,20 @@ windowSortSpecList
 havingClause
     : HAVING arg=exprSelect;
 
+excludeClause
+    : EXCLUDE excludeExpr (COMMA excludeExpr)*;
+
+excludeExpr
+    : symbolPrimitive excludeExprSteps+;  // Require 1 more `excludeExprSteps`. Disallow `EXCLUDE a`
```
johnedquinn, line 341: Can you explain why a step is required? I imagined that the items within the `EXCLUDE` clause would follow the same resolution rules as the projection items. If ambiguities are encountered, errors are thrown.

partiql-parser/src/main/antlr/PartiQL.g4
```diff
@@ -334,6 +334,20 @@ windowSortSpecList
...
+
+excludeExprSteps
+    : PERIOD symbolPrimitive                            # ExcludeExprTupleAttr
+    | BRACKET_LEFT attr=LITERAL_STRING BRACKET_RIGHT    # ExcludeExprCollectionAttr
+    | BRACKET_LEFT index=LITERAL_INTEGER BRACKET_RIGHT  # ExcludeExprCollectionIndex
+    | BRACKET_LEFT ASTERISK BRACKET_RIGHT               # ExcludeExprCollectionWildcard
+    | PERIOD ASTERISK                                   # ExcludeExprTupleWildcard
+    ;
```
johnedquinn, line 349: Can you give me a breakdown of what these semantically mean for your PR?

If I'm guessing:
- `EXCLUDE t.*` (when `t` is a struct) will make `t` an empty struct?
- `EXCLUDE t[*]` (when `t` is a list) will make `t` an empty list?
- `EXCLUDE t[2]` (when `t` is a list) will remove the item at index 2
- `EXCLUDE t[*].a` (when `t` is a list of structs) will remove attribute a from all items in `t`.

Follow up questions:
- Is `t.*.*[1]` allowed?

partiql-ast/src/main/resources/partiql_ast.ion
```diff
@@ -560,6 +561,22 @@ select::[
...
+    exclude_expr::{
+      root: '.identifier.symbol',
+      steps: list::[step],
+    },
+    step::[
+      exclude_tuple_attr::{ symbol: '.identifier.symbol' },
+      exclude_collection_index::{ index: int },
+      exclude_tuple_wildcard::{},
+      exclude_collection_wildcard::{},
+    ]
```
johnedquinn, line 576: Why not use the available `'.expr.path'`? The steps are almost the same (with the exception that you require literals). You could just make it a parser or semantic check to verify that the root is a symbol and your steps are literals? LMK what you think.

Also, I do like how you use the spec's naming of `tuple_wildcard` and `collection_wildcard`.

partiql-ast/src/main/pig/partiql.ion
```diff
@@ -219,6 +220,21 @@ may then be further optimized by selecting better implementations of each operat
             // For `<expr> [AS <id>]`
             (project_expr expr::expr as_alias::(? symbol)))
 
+        // EXCLUDE clause
+        (product exclude_op exprs::(* exclude_expr 1))
+
+        (product exclude_expr root::identifier steps::(* exclude_step 0))
```
johnedquinn, line 226: From your `.g4` you write that at least one step is required.

```suggestion
        (product exclude_expr root::identifier steps::(* exclude_step 1))
```

partiql-plan/src/main/resources/partiql_plan.ion
```diff
@@ -78,6 +78,25 @@ arg::[
   }
 ]
 
+// Exclude expr and step
+exclude_expr::{
+  root: string,
+  root_case: case,
+  steps: list::[exclude_step],
+}
```
johnedquinn, line 86: Same question as `partiql_ast`. Why not just path?

johnedquinn (commented):
Still need to finish reviewing some last files, but sending some comments for immediate feedback.

partiql-parser/src/main/antlr/PartiQL.g4
alancai98: Also worth mentioning `EXCLUDE` is currently modeled as a binding tuple operator (evaluated before the `SELECT` projection). We are trying to disallow excluding a top-level binding explicitly (e.g. disallow `SELECT ... EXCLUDE t FROM s, t`) since we don't have a clear use case for this.

partiql-parser/src/main/antlr/PartiQL.g4
alancai98: Yes to all.

partiql-ast/src/main/resources/partiql_ast.ion
alancai98: > Why not use the available `'.expr.path'`? The steps are almost the same (with the exception that you require literals). You could just make it a parser or semantic check to verify that the root is a symbol and your steps are literals? LMK what you think.

I had initially experimented with modeling these exclude expressions using `'.expr.path'` but this ended up making the schema inference on the plan a bit messier with the additional casts. I chose to model it as a separate node here to follow what the DML/DDL `path` (or simple path) had done as a separate node.

partiql-plan/src/main/resources/partiql_plan.ion
alancai98: (same as above comment in `partiql_ast`)

partiql-ast/src/main/resources/partiql_ast.ion
alancai98: I was under the notion that `expr`s return a PartiQL value. An individual exclude expr doesn't return a PartiQL value

partiql-lang/src/main/kotlin/org/partiql/lang/planner/transforms/plan/PlanTyper.kt
```diff
@@ -155,6 +158,114 @@ internal object PlanTyper : PlanRewriter<PlanTyper.Context>() {
...
+            return s.copy(fields = outputFields)
+        }
+
+        fun excludeExprStepsCollection(c: CollectionType, steps: List<ExcludeStep>, lastStepAsOptional: Boolean): StaticType {
+            var elementType = c.elementType
+            when (steps.first()) {
+                is ExcludeStep.CollectionIndex -> {
+                    if (steps.size > 1) {
+                        elementType = excludeExprSteps(elementType, steps.drop(1), lastStepAsOptional = true, ctx)
+                    }
```
yliuuuu, line 246: Sanity check here. 
So if we have `{'a' : [0, 1, 2]}` and exclude `a[0]`, the result will be `{'a' : [1,2]}`, hence optional, i.e., `unionOf(missing, int)`.

yliuuuu (commented):
Nice work. Is there any RFC in draft or document that comes along with this feature at the moment?

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
```diff
@@ -1070,6 +1155,895 @@ class PartiQLSchemaInferencerTests {
...
+                    FROM <<
+                        {
+                            'a': {
+                                'b': {
+                                    'c': 0,
+                                    'd': 'foo'
+                                }
+                            }
+                        }
+                    >> AS t""",
```
johnedquinn, line 1981: What happens if you added a similar, but different, test?
```
            ErrorTestCase(
                name = "invalid exclude collection wildcard",
                query = """SELECT * EXCLUDE t.a[*]
                    FROM <<
                        {
                            'a': {
                                'b': {
                                    'c': 0,
                                    'd': 'foo'
                                }
                            }
                        },
                        {
                            'a': [
                                {
                                    'c': 0,
                                    'd': 'foo'
                                }
                            ]
                        }
                    >> AS t""",
```

Since `t.a` is a union of `struct` and `list`?

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
```diff
@@ -1070,6 +1155,895 @@ class PartiQLSchemaInferencerTests {
...
+                            ]
+                        }
+                    >> AS t""",
+                problemHandler = assertProblemExists {
+                    Problem(
+                        UNKNOWN_PROBLEM_LOCATION,
+                        PlanningProblemDetails.InvalidExcludeExpr
+                    )
+                }
+            ),
```
johnedquinn, line 2046: Could you add a test regarding attributes that do not exist in a tuple? I'm curious about the behavior.

johnedquinn, line 2046: I'm curious about the behavior when an EXCLUDE attribute is removed and referenced later on in the exclude list. Something like:
```
SELECT * EXCLUDE t.a, t.a.b FROM << { 'a': { 'b': 1 } } >> AS t 
```

partiql-lang/src/main/kotlin/org/partiql/lang/planner/transforms/plan/PlanTyper.kt
johnedquinn: Can you elaborate on this?

I would assume that the static type of `a` would be a `ListType(INT)`. IMO, even if we had:
```
SELECT * EXCLUDE t.a[0]
FROM << { 'a': [0] } >> AS t
```

I would assume the static type of `t.a` would still be `ListType(INT)` because we can't know how long the list actually is (especially if `t` is part of the catalog). If `t.a` started without any elements, I also think it would still be `ListType(INT)`. That being said @alancai98 , do we expect this to be a runtime error when we attempt to exclude an index that doesn't exist (essentially `OutOfBoundsException`)? How about when we try to exclude an attribute in a tuple that doesn't exist? This one we can sometimes know statically, but what happens at runtime if it's a union type?

alancai98:
> Is there any RFC in draft or document that comes along with this feature at the moment?

I have parts of the RFC in a draft. Still working on defining the operational semantics in terms of existing clauses (e.g. PIVOT, UNPIVOT, CASE WHEN) before I'm ready to publish the draft.

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
johnedquinn: I think we might need to add support for it outright, since `null` is modeled as a distinct StaticType. AKA, very similar query that represents many flows that other SQL implementations might experience:
```sql
SELECT * EXCLUDE t.a[*]
FROM <<
  null,
  {
    'a': [
      {
        'c': 0,
        'd': 'foo'
      }
    ]
  }
>> AS t
```

Especially since everything is nullable by default in SQL. Similar experience:
```sql
-- I know this isn't real syntax, but it's meant for demonstration.
CREATE TABLE t1 (
  a STRUCT (
    c INT,
    d STRING
  )
);

-- A similar syntax that Redshift supports would be
CREATE TABLE t1 (
  a SUPER
);
```

`a` in both queries would be nullable, and therefore the following query would be wrongly typed:
```sql
SELECT * EXCLUDE t.a.*
FROM t1 AS t
```

If I'm not mistaken, the type currently returned would be something like:
```
BagType(
  StructType(c: IntType(), d: StringType())
)
```

Since `t` would be the `unionOf(NULL, StructType( ... ))`. But, I imagine the expected would be:
```
BagType(
  StructType()
)
```

alancai98: I agree w/ the point that `EXCLUDE` over type unions are important after some offline discussion. I still need some clarification on what the desired behavior should be w.r.t. typed unions.

For example, consider 

```
SELECT t
EXCLUDE t.a.b
FROM <<
    {
        'a': 1 {
            'b': 2,
            'c': 3
        }
    },
    {
        'a': NULL
    }
>> AS t
```

Currently the schema is something like:

```
Bag(
    Struct(
        a: unionOf(null, struct(b: int, c: int))
    )
)
```
^ basically ignores the union type for `a`.

Would the schema you propose then be:
```
Bag(
    Struct(
        a: unionOf(null, struct(c: int)) -- excluding `b` from the struct in the union
    )
)
```
?

alancai98: Thanks for confirming. Latest commit https://github.com/partiql/partiql-lang-kotlin/pull/1226/commits/d60d08cd1ab4f80bdce344e51c4983a0760ced7f should cover the union case as well as some other edge cases discussed offline.

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
```diff
@@ -1072,6 +1072,1179 @@ class PartiQLSchemaInferencerTests {
...
+                                    contentClosed = true,
+                                    constraints = setOf(TupleConstraint.Open(false), TupleConstraint.UniqueAttrs(true))
+                                )
+                            )
+                        ),
+                        contentClosed = true,
+                        constraints = setOf(TupleConstraint.Open(false), TupleConstraint.UniqueAttrs(true), TupleConstraint.Ordered)
+                    )
+                )
+            ),
```
yliuuuu, line 2247: A couple more test cases that can be beneficial: 
```
-- demonstrate expression in exclude don't not lead to typing error.  
SELECT * EXCLUDE t.b
                    FROM <<
                        {
                            'a': <<
                                { 'b': 0 },
                                { 'b': 1 },
                                { 'b': 2 }
                            >>
                        }
                    >> AS t
                    
-- demonstrate expression in exclude don't not lead to typing error.  
SELECT * EXCLUDE nonsense.a
                    FROM <<
                        {
                            'a': <<
                                { 'b': 0 },
                                { 'b': 1 },
                                { 'b': 2 }
                            >>
                        }
                    >> AS t 
                            
-- I am assuming
-- bag ( struct ( list ( unionOf( struct(b: int, c: unionOf(missing, int) ), struct(b: int, c: unionOf(missing, NULL, MISSSING) ), struct(b: int, c: unionOf(missing, decimal) ))) ) ) 
SELECT * EXCLUDE t.a[0].c
                    FROM <<
                        {
                            'a': [
                                { 'b': 0,
                                  'c': 0
                                },
                                { 'b': 1,
                                  'c': NULL
                                },
                                { 'b': 2,
                                  'c': 0.1
                                 }
                            ]
                     
                        } 
                    >> AS t                               
```

yliuuuu (commented):
Had some suggestion on putting in a couple additional test cases.

Also, it would be nice to explicitly list out what the subset of behavior that is considered stable.

am357 (commented):
If not mistaken, all the tests in this PR are working based off of literal data. It is great that we are examining the schema inferencing using various literals but perhaps for most of the use-cases we are doing the schema inference based on some base tables (in a catalog)—which will automatically remove most of the edge cases we see in the tests as a serious concern. This is b/c in my view, literal semi-structured data implies partial or no-schema which doesn't require static validation as such.

@alancai98 apologies to not to mention this earlier, does it make sense adding tests using a schema in the catalog? We don't need to go as deep as the current tests, but just to show the four exclude cases.

partiql-lang/src/main/kotlin/org/partiql/lang/planner/transforms/plan/PlanTyper.kt
```diff
@@ -155,6 +158,150 @@ internal object PlanTyper : PlanRewriter<PlanTyper.Context>() {
...
+                properties = input.getProperties()
+            )
+        )
+    }
+
+    private fun excludeExpr(attrs: List<Attribute>, expr: ExcludeExpr, ctx: Context): List<Attribute> {
+        val resultAttrs = mutableListOf<Attribute>()
+        attrs.forEach { attr ->
+            val rootId = expr.root
+            if (attr.name == rootId || (expr.rootCase == Case.INSENSITIVE && attr.name.equals(expr.root, ignoreCase = true))) {
```
johnedquinn, line 203: I'm not sure if this would be required as part of the RFC, but If we wanted to make sure we don't come into a situation such as:
```sql
SELECT * EXCLUDE nonsense.a
FROM t
```
we could:
```kotlin
val attrsExist = attrs.find { attr -> attr.name == .. } != null
if (attrsExist.not()) { handleUnresolvedDescriptor(..) }
attrs.forEach { ->
  if (attr.name == ..) {
    ..
```

It should make sure that we don't accidentally break future queries. It's IMO better to error now even though we aren't sure what the RFC will say. Then, if we allow nonexistent bindings after the RFC, we can always remove the error? Then you can add the second test that @yliuuuu mentioned above.
> @yliuuuu A couple more test cases that can be beneficial:

LMK what you think.

johnedquinn (commented):
This is some pretty solid work. Thanks for addressing my comments thus far. I left a single comment regarding nonexistent bindings. One you resolve that comment plus the others from the other maintainers, I think we can ship this.

johnedquinn:
Responding to  @am357:
> This is b/c in my view, literal semi-structured data implies partial or no-schema which doesn't require static validation as such.
>
> ...
> 
> @alancai98 apologies to not to mention this earlier, does it make sense adding tests using a schema in the catalog? We don't need to go as deep as the current tests, but just to show the four exclude cases.

So you know, @am357, we actually do perform static analysis on literals. That being said, I agree about adding just the base 4 exclude cases to use the catalog too, as our users will likely be using this feature the most.

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
alancai98: Yes to your assumption. I will add those test cases.

partiql-lang/src/main/kotlin/org/partiql/lang/planner/transforms/plan/PlanTyper.kt
johnedquinn: Also, it would surely help users who write:
```
SELECT a
EXCLUDE a.b
FROM <<
  { 'a': { 'b': 1, 'c': 2 } }
>> AS t
```

Since we are specifically looking in the bindings coming from the SCAN, when we look for the `EXCLUDE a.b` (and specifically looking for `a`), we are silently ignoring the fact that `a` isn't in the EXCLUDE's input bindings. So, even though the projection list can use `a`, the `EXCLUDE` is silently letting this data to pass through. The current implementation, if I'm not mistaken, will return:
```
<<
  { 'a': { 'b': 1, 'c': 2 } }
>>
```

johnedquinn: Since resolution of the `EXCLUDE` expressions is not defined, I think it's perfectly fine to mandate that users write `... EXCLUDE t.a.b` explicitly. Though, if we silently let them write `a.b` without notifying them that `a` isn't a binding (and therefore they think that their query will strip `b`), we'll be leaking data.

alancai98: That's fair. I had originally wanted to punt on the erroring behavior to form a consistent mental model for when to give an error in the RFC. The more I think about it though, if the exclude expr root is not in the binding tuples, then we should give some error. I'll adjust to at least give an error in this case (i.e. when the root doesn't exist)

am357 (approved):

johnedquinn (dismissed):
Thanks for this great PR! LGTM.

yliuuuu (approved):
LGTM. Thanks for the great work!

am357 (approved):

johnedquinn (approved):

PR #1229, opened 2023-09-27
Update mockdb to use partiql value schema

RCHowell:
## Description

This work is to extend the local plugin to use a richer schema representation which we are using for planner and transpiler testing. We shouldn't have two local plugins, so this PR converges them.

I have updated the local plugin README with how to use this plugin.

## Other Information
No

The schema language has changed, but this is our local testing plugin.

No

Yes

## License Information


plugins/partiql-local/src/main/kotlin/org/partiql/plugins/local/LocalCatalog.kt
```diff
@@ -0,0 +1,70 @@
...
+            when (curr) {
+                is FsTree.T -> break
+                is FsTree.D -> {
+                    curr = curr.children.firstOrNull { step.isEquivalentTo(it.name) }
+                    if (curr != null) match.add(curr.name)
+                }
+            }
+        }
+        // All steps matched and we're at a leaf
+        if (curr is FsTree.T) {
```
RCHowell, line 40: There is a potential bug here, but it's actually undefined behavior. Is it possible for a schema and object to have the same name? ie should this file structure be possible

```ion
catalog/
    ├── example
    │   └── other.ion
    └── example.ion
```

partiql-lang/src/test/resources/catalogs/aws/ddb/pets.ion
johnedquinn, line 1: So, the original MockDB had tables with attributes, bags with structs, top-level ints, etc. If I'm reading this correctly, is `aws.ddb.pets` a collection of structs? Or just a struct? The example in the README makes me think it is just a standalone struct.

plugins/partiql-local/README.md
```diff
@@ -1,90 +1,97 @@
...
 ## Catalog 
 
-Your Catalog is specified as the `localdb_root` directory from your configuration file above. If not specified, it defaults to
-`${HOME}/.partiql/localdb`.
+Your Catalog is specified as the `root` directory from your configuration file above. If not specified, it defaults to
+`${HOME}/.partiql/local`.
 
 Each Catalog holds Directories. Here's an example filesystem using the Configuration File from further above:
 ```text
 fs (Connector: localdb) (Root: /Users)
```
johnedquinn, line 28:
```suggestion
fs (Connector: local) (Root: /Users)
```

partiql-lang/src/test/resources/catalogs/aws/ddb/pets.ion
RCHowell: It should definitely be a bag of structs. This PR is wrong — but all the tests passed because (I believe) the coercion to singleton bag in the FROM source ...

I will need to update the table schemas on this PR.

plugins/partiql-local/src/main/kotlin/org/partiql/plugins/local/LocalSchema.kt
```diff
@@ -0,0 +1,212 @@
...
+    "type" to ionString("list"),
+    "items" to elementType.toIon()
+)
+
+private fun SexpType.toIon(): IonElement = ionStructOf(
+    "type" to ionString("sexp"),
+    "items" to elementType.toIon()
+)
+
+private fun StructType.toIon(): IonElement {
```
johnedquinn, line 201: We're missing constraints here

johnedquinn (changes requested):

johnedquinn (approved):

Issue #1221, opened 2023-09-20
Implement PartiQL bitwise operators


am357:
## Relevant Issue/Bug
- N/A

## Requested Solution/Feature
- Add support for _bitwise operators_ in PartiQL:

Operator | Description | Example | Result
-- | -- | -- | --
\|\| | concatenation | B'10001' \|\| B'011' | 10001011
& | bitwise AND | B'10001' & B'01101' | 00001
\| | bitwise OR | B'10001' \| B'01101' | 11101
\# | bitwise XOR | B'10001' \# B'01101' | 11100
\~ | bitwise NOT | \~ B'10001' | 01110
<< | bitwise shift left | B'10001' << 3 | 01000
\>> | bitwise shift right | B'10001' \>> 2 | 00100

```SQL
-- PostgresSQL: https://www.db-fiddle.com/f/g952RU7t652W32aiuWvsk6/8

SELECT B'110' | B'101' AS bit_or;
-- output: 111

SELECT 30 & 199 & 7 AS bit_and;
-- output: 6
```

## Describe Alternatives
- We could use a custom function for this with the following sample signatures for the immediate narrow use-case:
```
INT -> INT -> INT // E.g., bit_and(2, 3)
INT -> INT -> INT -> INT // E.g., bit_and(2, 3, 4)
```

## Additional Context
Prior Art:

1. PostgresSQL
   - https://www.postgresql.org/docs/9.4/functions-bitstring.html
   - https://www.db-fiddle.com/f/g952RU7t652W32aiuWvsk6/3
1. SQLite
   - https://www.db-fiddle.com/f/g952RU7t652W32aiuWvsk6/5
1. MySQL
   - https://www.db-fiddle.com/f/g952RU7t652W32aiuWvsk6/4
1. SQLServer
   - https://learn.microsoft.com/en-us/sql/t-sql/language-elements/bitwise-operators-transact-sql?view=sql-server-ver16

## DoD (Definition of Done)
- [ ] The required types are implemented
- [ ] The operator precedence is determined and documented in `partiql-doc` or `partiql-spec`
- [ ] The implementation is merged to `main`
- [ ] The conformance tests are created
- [ ] Mini RFC?

PR #1232, opened 2023-09-28
Bitwise and plan

yliuuuu:
## Relevant Issues
- [Closes/Related To] Issue #1221 

## Description
- Type Inferencer for Bitwise And. 
- Wiki Page explaining semantics for bitwise and operator. 

## Other Information
  - Yes. 

  - No. 

  - No. 

No. 

## License Information


docs/wiki/documentation/Numeric Data Type & Arithmetic Operations.md
```diff
@@ -206,4 +206,29 @@ Example
...
+```
+
+Return Type:
+: If one operand is of `INT` type, returns `INT` type. 
+: Else if one operand is of INT8 type, returns `INT8` type. 
+: Else if one operand is of INT4 type, returns `INT4` type. 
+: Else return `INT2` type. 
+
+Note: 
+: Type precedence of the bitwise operator is lower than the plus/minus operator and higher than the predicates.
```
alancai98, line 229: Could link the partiql-docs issue since this could change: https://github.com/partiql/partiql-docs/issues/50.

CHANGELOG.md
```diff
@@ -30,9 +30,9 @@ Thank you to all who have contributed!
 ## [Unreleased]
 
 ### Added
-- Support parsing and evaluation of Bitwise AND operator (&).
-  - The Bitwise And Operator only works for integer operands. 
-  - The operator precedence may change based on the pending operator precedence [RFC](https://github.com/partiql/partiql-docs/issues/50). 
+- Support parsing, planning and evaluation of Bitwise AND operator (&).
```
alancai98, line 33: nit: missing comma

```suggestion
- Support parsing, planning, and evaluation of Bitwise AND operator (&).
```

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
```diff
@@ -1070,6 +1071,58 @@ class PartiQLSchemaInferencerTests {
...
+                name = "BITWISE_AND_9",
+                query = "CAST(1 AS INT4) & 2",
+                expected = StaticType.unionOf(StaticType.INT, MISSING)
+            ),
+            SuccessTestCase(
+                name = "BITWISE_AND_10",
+                query = "CAST(1 AS INT8) & 2",
+                expected = StaticType.unionOf(StaticType.INT, MISSING)
+            ),
+
```
alancai98, line 1125: nice set of tests! could also add
- test w/ null propagation (1 & NULL) -> NULL
- test w/ missing propagation (1 & MISSING) -> MISSING and (NULL & MISSING) -> MISSING
- data type mismatch (1 & 'non-int') -> error and output type of union of int types

alancai98 (dismissed):
Looks good! Minor comments and some additional tests you could add

alancai98 (approved):

RCHowell:
I'm a bit late to the party, but `1 & NULL` isn't an error. It's an odd query and would be constant folded to NULL, but it shouldn't be an error as the tests suggest. In fact, the error "value is always null or missing" _probably_ doesn't make sense in most contexts.

yliuuuu:
> I'm a bit late to the party, but `1 & NULL` isn't an error. It's an odd query and would be constant folded to NULL, but it shouldn't be an error as the tests suggest. In fact, the error "value is always null or missing" _probably_ doesn't make sense in most contexts.

+1 on that. Loop in @johnedquinn for more context on the error `ExpressionAlwaysReturnsNullOrMissing`

johnedquinn:
Yeah, it's an artifact from the behavior of the `StaticTypeInferencer`. I do think it's a useful warning, though it's misclassified as a `ProblemSeverity` of `ERROR`. Should probably be `WARNING`.

PR #1237, opened 2023-10-03
Adds support for planning/transpilation of SELECT *

johnedquinn:
## Relevant Issues
- None

## Description
- Adds support for planning/transpilation of SELECT *
- Adds support for planning/transpilation of SELECT a.*
- Adds support for planning/transpilation of SELECT a.b.*
- Adds support for planning/transpilation of SELECT a.*, a.b

## Other Information
  - Not being added to `main` yet.

  - I'm planning on updating the definition of `Rex.TupleUnion.Arg`, but it's not in this PR yet.

## License Information


partiql-planner/src/testFixtures/kotlin/org/partiql/planner/test/Parsing.kt
```diff
@@ -127,7 +128,7 @@ public fun StructElement.toStructType(): StaticType {
         val type = it.getAngry<IonElement>("type").toStaticType()
         StructType.Field(name, type)
     }
-    return StructType(fields)
+    return StructType(fields, contentClosed = true, constraints = setOf(TupleConstraint.Open(false)))
```
RCHowell, line 131: Latest in `main` has handling of constraints.

https://github.com/partiql/partiql-lang-kotlin/blob/main/plugins/partiql-local/src/main/kotlin/org/partiql/plugins/local/LocalSchema.kt#L125-L139

lib/partiql-transpiler/src/test/resources/targets/redshift/redshift_results_00.ion
```diff
@@ -56,19 +56,46 @@ target::{
...
+        returns.sr_addr_sk AS sr_addr_sk, returns.sr_store_sk AS sr_store_sk, returns.sr_reason_sk AS sr_reason_sk,
+        returns.sr_ticket_number AS sr_ticket_number, returns.sr_return_quantity AS sr_return_quantity,
+        returns.sr_return_amt AS sr_return_amt, returns.sr_return_tax AS sr_return_tax,
+        returns.sr_return_amt_inc_tax AS sr_return_amt_inc_tax, returns.sr_fee AS sr_fee,
+        returns.sr_return_ship_cost AS sr_return_ship_cost, returns.sr_refunded_cash AS sr_refunded_cash,
+        returns.sr_reversed_charge AS sr_reversed_charge, returns.sr_store_credit AS sr_store_credit,
+        returns.sr_net_loss AS sr_net_loss
+        FROM tpc_ds."store" AS store
+        LEFT OUTER JOIN tpc_ds."store_returns" AS returns
+        ON store.s_store_sk = returns.sr_store_sk
```
RCHowell, line 80: I kind of think we should use a different suite for `SELECT *` with tables that have few columns just so this isn't so large

partiql-planner/src/main/kotlin/org/partiql/planner/transforms/RelConverter.kt
```diff
@@ -106,13 +108,22 @@ internal object RelConverter {
...
         }
         val op = rexOpStruct(fields)
-        rex(StaticType.STRUCT, op)
+        val type = StructType(
+            fields = fields.mapIndexed { i, it -> StructType.Field(schema[i].name, schema[i].type) },
+            contentClosed = true,
+            constraints = setOf(
+                TupleConstraint.Open(false)
+            )
+        )
```
RCHowell, line 117: I see why this was added, but my logic for not doing this was that PlanTyper will rewrite it anyways with the appropriate field types. There's no harm in having it here too

partiql-planner/src/main/kotlin/org/partiql/planner/typer/PlanTyper.kt
```diff
@@ -272,7 +272,8 @@ internal class PlanTyper(
          */
         override fun visitRexOpPath(node: Rex.Op.Path, ctx: StaticType?): Rex = rewrite {
             // 1. Resolve path prefix
-            val (root, steps) = when (val rootOp = node.root.op) {
+            val visitedRoot = visitRex(node.root, ctx)
```
RCHowell, line 275: Nice catch, but I'm concerned this would cause problems because we might resolve the root accidentally when we should have considered the whole path like line 279.

By visiting the root first we might accidentally resolve to some global `x` rather than something like `x.y.z`. I don't have an example at the moment, but I think we need to consider the whole path for resolution in 280

RCHowell (changes requested):

partiql-planner/src/main/kotlin/org/partiql/planner/transforms/RelConverter.kt
johnedquinn: I've since removed the default constructor from the RelConverter and have moved it to the `NormalizeSelect`.

PR #1242, opened 2023-10-06
Jvm field ir

RCHowell:
## Description

This PR improves the efficiency and simplicity of generated IR data structures. As we've been debugging larger and larger trees, we have noticed some hidden bloat that comes with using Kotlin over Java — ie each Kotlin field has a private java field along with public getter/setter. The solution is to mark each field as @JvmField, but this is not quite as simple as it sounds in our case.

For us, the IR is defined by an interface with internal implementations. A Kotlin interface can have fields (because they're really Java methods) which an implementation overrides/implements. These are non-final fields because they're actually method implementations; but the syntax is deceiving. You cannot mark non-final fields as @JvmField which means we can't actually use interfaces here.

But these brings light to Vladimir's suggestion; why bother with all the machinery? It's indeed simpler and more efficient to use data classes with `@JvmField` on all parameters. As others have suggested, we could have avoid the shenanigans with Java and Lombok 😉

There are three changes here that are minimal in practice but have nice access efficiency implications.

1. Each field becomes a final @JvmField (+1 access efficiency)
2. Using `data` class to get Kotlin's official hashcode/equals (+1 maintainability); we also get destructuring
3. Remove factory/abstract factory in favor of top-level constructor functions (+1 simplicity/ergonomics)

While the factory is gone, we keep the Java style builders and Kotlin type-safe DSL.

```kotlin
// Before

object Ast : AstFactoryBase() {}

abstract class Foo : Node {
	abstract val x: String
	abstract val y: Bool
}

open class FooImpl internal constructor(
	override val x: String,
	override val y: Bool,
) : Foo() {

   // our own copy/hashcode/equals/toString
}

Ast.foo(...)            // 1. construct with factory
Ast.create { foo(...) } // 2. factory in scope


// After, reduction

data class Foo(
	@JvmField val x: String,
	@JvmField val y: Bool
) : Node()
// inherit copy/hashcode/equals/toString/componentN

// Ast.kt
Foo("x", true)   // 1. construct directly, no indirection
foo("y", false)  // 2. construct with "factory" method
```

This second way of doing things reduces the complexity of the Java API while maintaining the fluent builders.

## Breaking Change (Kotlin only)

This is a very minimal breaking change. It's as simple as removing the factory in node construction and importing the top-level function. The interfaces are the exact same; only node construction has changed. 

```kotlin
// before

import org.partiql.ast.Ast

val a = Ast.foo("x", false)

// after

import org.partiql.ast.foo

val a = foo("x", false)
```

The symbols from the Java API don't change, but instead of factory methods called on a singleton object, they are static methods.

## Other Information
Yes

Yes, the factory class in no longer necessary. 

No

Yes

## License Information


CHANGELOG.md
```diff
@@ -36,12 +39,14 @@ Thank you to all who have contributed!
 ### Fixed
 
 ### Removed
+- Removed IR factory in favor of static top-level functions. Change `Ast.foo()`
+  to `foo()`
```
johnedquinn, line 43: Should we mark this as breaking?

lib/sprout/src/main/kotlin/org/partiql/sprout/generator/target/kotlin/poems/KotlinBuilderPoem.kt
```diff
@@ -197,10 +105,10 @@ class KotlinBuilderPoem(symbols: KotlinSymbols) : KotlinPoem(symbols) {
 
         // DSL Function
         val funcDsl = FunSpec.builder(builderName).returns(clazz)
-        funcDsl.addStatement("val builder = %T(${ props.joinToString { it.name }})", builderType)
+        funcDsl.addStatement("val builder = %T(${props.joinToString { it.name }})", builderType)
 
         // Java builder `build(factory: Factory = DEFAULT): T`
```
johnedquinn, line 110: Nit - Comment update

lib/sprout/src/main/kotlin/org/partiql/sprout/generator/target/kotlin/poems/KotlinUtilsPoem.kt
```diff
@@ -114,7 +102,7 @@ class KotlinUtilsPoem(symbols: KotlinSymbols) : KotlinPoem(symbols) {
 
     private fun KotlinNodeSpec.Product.rewriter(): FunSpec {
         val visit = product.ref.visitMethodName()
-        val constructor = symbols.camel(product.ref)
+        // val constructor = symbols.camel(product.ref)
```
johnedquinn, line 105: Comment left behind

johnedquinn (commented):
Very clear PR. Great work. There were some straggling comments and a file potentially left behind. I can approve once cleaned.

partiql-lang/src/jmh/kotlin/org/partiql/jmh/benchmarks/PartiQLParserBenchmark.kt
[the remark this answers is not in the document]
RCHowell: Yes it includes the `org.partiql.parser` implementation.

johnedquinn (approved):

PR #1244, opened 2023-10-10
Adds planning/typing support of SELECT *

johnedquinn:
## Relevant Issues
- None

## Description
- Goes ahead with the PartiQL Specification transformation for `SELECT -> SELECT VALUE`. Specifically uses `TupleUnion` for `SELECT *`. Removes NormalizeSelectList and NormalizeSelectStar. Adds NormalizeSelect.
- Adds an internal `RexReplacer`

## Other Information
  - No. Not main.

## License Information


partiql-ast/src/main/kotlin/org/partiql/ast/normalize/NormalizeSelect.kt
```diff
@@ -0,0 +1,252 @@
...
+            is From.Join -> lhs.aliases() + rhs.aliases()
+            is From.Value -> {
+                val asAlias = asAlias?.symbol ?: error("AST not normalized, missing asAlias on FROM source.")
+                val atAlias = atAlias?.symbol
+                val byAlias = byAlias?.symbol
+                listOf(Triple(asAlias, atAlias, byAlias))
+            }
+        }
+
+        private val col = { index: Int -> "_${index + 1}" }
```
RCHowell, line 234: nit: Do you mind moving this to the top with comment?

partiql-ast/src/main/kotlin/org/partiql/ast/normalize/NormalizeSelect.kt
```diff
@@ -0,0 +1,252 @@
...
+        @OptIn(PartiQLValueExperimental::class)
+        private fun buildCaseWhenStruct(expr: Expr, index: Int): Expr.Case {
+            return ast {
+                exprCase {
+                    branches.add(
+                        exprCaseBranch(
+                            condition = exprIsType(expr, typeStruct()),
+                            expr = expr
+                        )
+                    )
```
RCHowell, line 218: Sometimes the `+=` syntax on mutable lists can reduce nesting
```suggestion
                    branches += exprCaseBranch(
                        condition = exprIsType(expr, typeStruct()),
                        expr = expr
                    )
```

partiql-ast/src/main/kotlin/org/partiql/ast/normalize/NormalizeSelect.kt
```diff
@@ -0,0 +1,252 @@
...
+            val alias = expr.toBinder(i)
+            selectProjectItemExpression(expr, alias)
+        }
+
+        // t -> t AS t
+        private fun String.simple() = ast {
+            val expr = exprVar(id(this@simple), Expr.Var.Scope.DEFAULT)
+            val alias = id(this@simple)
+            selectProjectItemExpression(expr, alias)
+        }
```
RCHowell, line 268: removing the `ast { ... }` block should clarify `this` without have to use @

partiql-planner/src/main/kotlin/org/partiql/planner/PartiQLPlannerDefault.kt
```diff
@@ -34,17 +35,28 @@ internal class PartiQLPlannerDefault(
...
             plan = pass.apply(plan, onProblem)
         }
 
         return PartiQLPlanner.Result(plan, emptyList())
     }
+
+    private fun fold(stmt: org.partiql.plan.Statement, env: Env, onProblem: ProblemCallback): org.partiql.plan.Statement {
+        val foldedPlan = ConstantFolder.fold(stmt)
+        val typer = PlanTyper(env, onProblem)
+        return typer.resolve(foldedPlan)
```
RCHowell, line 60: This might be expensive, but we aren't at the point of optimizing yet. Why do we need another typing pass?

partiql-planner/src/main/kotlin/org/partiql/planner/transforms/RelConverter.kt
```diff
@@ -96,7 +90,11 @@ internal object RelConverter {
             }
             // SELECT VALUE ... FROM
             is Select.Value -> {
-                val constructor = projection.constructor.toRex(env)
+                val projectionOp = rel.op as? Rel.Op.Project ?: error("SELECT VALUE should have a PROJECT underneath")
+                assert(projectionOp.projections.size == 1) {
+                    "Expected SELECT VALUE projection to have a single binding. However, it looked like ${projectionOp.projections}"
+                }
+                val constructor = rex(StaticType.ANY, rexOpVarResolved(0))
```
RCHowell, line 97: I don't think it necessarily needs a PROJECT rel.op, but the input rel.op (regardless of what it is) must output exactly one binding.

partiql-planner/src/main/kotlin/org/partiql/planner/typer/PlanTyper.kt
```diff
@@ -531,24 +538,52 @@ internal class PlanTyper(
...
 
+        /**
+         * We need special handling for:
+         * ```
+         * CASE
+         *   WHEN a IS STRUCT THEN a
+         *   ELSE { 'a': a }
+         * END
+         * ```
+         * When we type the above, we can't just assume
```
RCHowell, line 555: comment is cut off

partiql-planner/src/main/kotlin/org/partiql/planner/typer/PlanTyper.kt
```diff
@@ -895,7 +970,10 @@ internal class PlanTyper(
             }
             rexOpPathStepSymbol(identifierSymbol(it.name, case))
         }
-        return rexOpPath(root, steps)
+        return when (steps.isEmpty()) {
+            true -> root.op
+            false -> rexOpPath(root, steps)
+        }
```
RCHowell, line 976: nice

partiql-planner/src/main/kotlin/org/partiql/planner/typer/RexReplacer.kt
```diff
@@ -0,0 +1,42 @@
...
+        return RexReplacerImpl.visitRex(rex, params)
+    }
+
+    private class ReplaceParams(val replace: Rex, val with: Rex)
+
+    private object RexReplacerImpl : PlanRewriter<ReplaceParams>() {
+
+        override fun visitRex(node: Rex, ctx: ReplaceParams): Rex {
+            if (node == ctx.replace) { return ctx.with }
+            return visitRexOp(node.op, ctx) as Rex
```
RCHowell, line 39:
```suggestion
            val op = visitRexOp(node.op, ctx) as Rex.Op
            return if (op !== node.op) rex(node.type, op) else node
```

RCHowell (changes requested):
I'm not done with the review, but have left some initial comments.

partiql-planner/src/main/kotlin/org/partiql/planner/PartiQLPlannerDefault.kt
johnedquinn: Unfortunately, we don't have the ability to fold an expression and tell all of it's dependents. Take the `SELECT *` transformation for example:
```
SELECT VALUE TUPLEUNION(
  CASE
    WHEN t IS STRUCT THEN t
    ELSE { '_1': t }
  END
)
FROM <<
  { 'a': 1}
>> AS t
```

When we originally type this, the StaticType of `Rex.Op.Select` will be `BAG(UNION(STRUCT(a), STRUCT(_1)))`. However, once it passes through the constant folder, we drop the second branch since the first branch's condition will always be true. Since the type of `Rex.Op.Select` is dependent upon the type of the CASE expression, we need to retype the plan with the second branch removed.

johnedquinn: That being said, this also reinforces our curiosity in typing during the conversion from AST to Plan. If a user explicitly writes a `CASE` with multiple potential output types, then I'd assume that, similar to Kotlin, the `CASE`'s output would also be multiple output types (even if we folded out branches that are never to execute). However, if we normalize a `SELECT *` to internally use a `STRUCT` or `TUPLEUNION`, then we can sidestep saying it could be multiple output types. We can definitively say it will be a single type.

RCHowell: > When we originally type this, the StaticType of Rex.Op.Select will be BAG(UNION(STRUCT(a), STRUCT(_1))). However, once it passes through the constant folder, we drop the second branch since the first branch's condition will always be true.

Why do we not know that `t` is a struct when we initially type? I think it's naive to type to

 `<< ( closed::{'a': int}, closed::{_1: any} >>`

What I think you should do is _type each branch_, prune branches, then union the remaining branch types. 

I'm writing this on github so I'm making some assumptions here

```kotlin
fun visitRexOpCase(op: Rex.Op.Case, ctx: StaticType): Rex {
   val branches: List<Rex.Op.Case.Branch> = op.branches.map { // type the branch cond+rex }.prune()
   return when (branches.size) {
      0 -> { // err always missing, no valid branch }
      1  -> branches.first.rex // drop the case-when
     else -> {
         // rewrite the case-when with new list of branches
         val type = unionOf(branches.map { it.rex.type })
         val op = rexOpCase(branches)
         rex(type, op)
     }
   }
}

fun List<Rex.Op.Case.Branch>.prune(): List<Rex.Op.Case.Branch> {
    val newBranches = mutableList<Rex.Op.Case.Branch>()
    // apply logic very similar to your constant folding
    // ie cond == true, then short circuit returning a singleton list of this branch
    // cond == false, drop this branch!
}
```

partiql-planner/src/main/kotlin/org/partiql/planner/typer/PlanTyper.kt
```diff
@@ -531,24 +542,94 @@ internal class PlanTyper(
...
+            }
+
+            // Calculate final expression (short-circuit to first branch if the condition is always TRUE).
             val resultTypes = visitedBranches.map { it.rex }.map { it.type }
-            return rex(AnyOfType(resultTypes.toSet()).flatten(), node.copy(branches = visitedBranches))
+            val firstBranch = newBranches.firstOrNull() ?: error("CASE_WHEN has NO branches.")
+            return@plan when (isLiteralTrue(firstBranch.condition)) {
+                true -> firstBranch.rex
+                false -> rex(type = StaticType.unionOf(resultTypes.toSet()).flatten(), node.copy(branches = newBranches))
+            }
```
RCHowell, line 565: This is entirely stylistic, but I'm leaving this here for some info / discussion. I've found loop control flow when using map / fold / mapNotNull to be more difficult to reason about than just using mutable lists and simple for loops; when I believe the stdlib code implements these functions with mutable lists. In this case, we don't actually map the elements so a simple `filter` would suffice. Consider the snippet

```kotlin
val newBranches = visitedBranches.mapNotNull { branch ->
    val conditionFoldedOp = branch.condition.op as? Rex.Op.Lit ?: return@mapNotNull branch
    val conditionBooleanLiteral = conditionFoldedOp.value as? BoolValue ?: return@mapNotNull branch
    when (conditionBooleanLiteral.value) {
        false -> null
        else -> branch
     }
}

// compared to
val newBranches = visitedBranches.filterNot { isLiteralBool(it.condition, false) }
```

partiql-planner/src/main/kotlin/org/partiql/planner/typer/PlanTyper.kt
```diff
@@ -531,24 +542,94 @@ internal class PlanTyper(
...
         }
 
+        @OptIn(PartiQLValueExperimental::class)
+        private fun isLiteralTrue(rex: Rex): Boolean {
+            val op = rex.op as? Rex.Op.Lit ?: return false
+            val value = op.value as? BoolValue ?: return false
+            return value.value ?: false
+        }
```
RCHowell, line 573:
```suggestion
        @OptIn(PartiQLValueExperimental::class)
        private fun isLiteralBool(rex: Rex, bool: Boolean): Boolean {
            val op = rex.op as? Rex.Op.Lit ?: return false
            val value = op.value as? BoolValue ?: return false
            return value.value == bool
        }
```

partiql-planner/src/main/kotlin/org/partiql/planner/typer/PlanTyper.kt
RCHowell: Now that I'm looking at it you can reduce a lot. You also don't need the plan block.

```kotlin
val branches = node.branches
                   .map { visitRexOpCaseBranch(it, it.rex.type) }
                   .filterNot { isLiteralBool(it.condition, false) }
return when (branches.size) {
    0 -> error("CASE_WHEN has NO branches.")
    1 -> branches.first().rex
    else -> {
         val types = branches.map { it.rex.type }.toSet()
         val type = AnyOfType(types).flatten()
         val op = rexOpCase(branches)
         rex(type, op)
    }
}
```

partiql-planner/src/main/kotlin/org/partiql/planner/typer/RexReplacer.kt
```diff
@@ -0,0 +1,45 @@
...
+ *  or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+ *  language governing permissions and limitations under the License.
+ */
+
+package org.partiql.planner.typer
+
+import org.partiql.plan.Rex
+import org.partiql.plan.rex
+// import org.partiql.plan.rex
```
RCHowell, line 19: comment left in

RCHowell (changes requested):
Overall looks good but could use some small clarity things. Could you please add a detailed comment about the typing of tupleunion? I understand the typing of case-when, but the typing of tupleunion is more complicated it seems and is deserving of a comment

partiql-planner/src/main/kotlin/org/partiql/planner/typer/PlanTyper.kt
johnedquinn: Agreed. The logic is a bit different, but yeah I'll update this. Good call.

partiql-planner/src/main/kotlin/org/partiql/planner/transforms/RelConverter.kt
```diff
@@ -80,11 +71,14 @@ internal object RelConverter {
     // IGNORE — so we don't have to non-null assert on operator inputs
     private val nil = rel(relType(emptyList(), emptySet()), relOpErr("nil"))
 
+    private var uniqueNumber: Long = 0L
+    private fun getNextUniqueName(): String = "\$__${uniqueNumber++}"
```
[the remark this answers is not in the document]
johnedquinn: Responding below.

partiql-planner/src/main/kotlin/org/partiql/planner/transforms/RelConverter.kt
```diff
@@ -256,7 +169,20 @@ internal object RelConverter {
...
         }
 
-        override fun visitFromValue(node: From.Value, nil: Rel): Rel {
+        /**
+         * Note: The name of the [Rel.Binding] doesn't actually matter. We make it unique to avoid any potential conflicts
+         * in the future.
+         */
+        override fun visitSelectValue(node: Select.Value, input: Rel): Rel = plan {
+            val rex = RexConverter.apply(node.constructor, env)
+            val schema = listOf(relBinding(name = getNextUniqueName(), rex.type))
```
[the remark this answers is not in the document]
johnedquinn: Discussed offline, I'm using `toBinder`.

RCHowell (approved):
Nice work!

PR #1250, opened 2023-10-20
Hidden functions

yliuuuu:
## Relevant Issues

## Description
- During parsing, say the user enter `trim_leading(val)`, the function will be parsed; the AST would be `Call(id(trim_leading, case_sensitive), val)`
- Because in the header we generated a function signature `fun trim_leading(a)`, the function resolver will recognized the function. 
- But the trim leading function actually represent a special form of trim, `TRIM( LEADING FROM a)`. 
- In this PR, we stop this scenario by adding a Unicode private use char to the key in Function Map for those functions that should be hidden. 

## Other Information
  - No. Change to unreleased feature. 

  - No. Change to unreleased feature. 
  
  - No. 

Yes. 

## License Information

### DIFF
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/Header.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/Header.kt
@@ -38,18 +38,12 @@ import org.partiql.value.PartiQLValueType.SYMBOL
 import org.partiql.value.PartiQLValueType.TIME
 import org.partiql.value.PartiQLValueType.TIMESTAMP
 
-/**
- * A structure for scalar function lookup.
- */
 private typealias FnMap<T> = Map<String, List<T>>
 
 /**
- * Map session attributes to underlying function name.
+ * Unicode non-character to be used for name sanitization
  */
-internal val ATTRIBUTES: Map<String, String> = mapOf(
-    "CURRENT_USER" to "\$__current_user",
-    "CURRENT_DATE" to "\$__current_date"
-)
+private val HIDDEN_FLAG: Char = Char(0xFDD0)
 
 /**
  * A place for type and function definitions. Eventually these will be read from Ion files.
@@ -73,7 +67,9 @@ internal class Header(
      */
     public fun lookup(ref: Fn.Unresolved): List<FunctionSignature.Scalar> {
         val name = getFnName(ref.identifier)
-        return functions.getOrDefault(name, emptyList())
+        return if (ref.isHidden)
+            functions.getOrDefault("${HIDDEN_FLAG}_$name", emptyList())
+        else functions.getOrDefault(name, emptyList())
     }
 
     /**
@@ -92,7 +88,7 @@ internal class Header(
             return null
         }
         val name = castName(targetType)
-        val casts = functions.getOrDefault(name, emptyList())
+        val casts = functions.getOrDefault("${HIDDEN_FLAG}_$name", emptyList())
         for (cast in casts) {
             if (cast.parameters.isEmpty()) {
                 break // should be unreachable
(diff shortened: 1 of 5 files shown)


partiql-planner/src/main/kotlin/org/partiql/planner/Header.kt
```diff
@@ -45,15 +45,7 @@ private typealias FunctionMap = Map<String, List<FunctionSignature>>
...
  */
-// private val OP: Char = Char(0xFDEF)
-
-/**
- * Map session attributes to underlying function name.
- */
-internal val ATTRIBUTES: Map<String, String> = mapOf(
-    "CURRENT_USER" to "\$__current_user",
-    "CURRENT_DATE" to "\$__current_date"
-)
```
RCHowell, line 56: Nice. Glad you were able to get rid of these.

partiql-planner/src/main/kotlin/org/partiql/planner/transforms/RelConverter.kt
```diff
@@ -419,7 +419,7 @@ internal object RelConverter {
                 schema.add(binding)
                 val args = agg.args.map { arg -> arg.toRex(env) }
                 val id = AstToPlan.convert(agg.function)
-                val fn = fnUnresolved(id)
+                val fn = fnUnresolved(id, false)
```
RCHowell, line 422: perfect

partiql-planner/src/main/kotlin/org/partiql/planner/transforms/RexConverter.kt
```diff
@@ -478,20 +480,15 @@ internal object RexConverter {
             val call = when (node.field) {
                 DatetimeField.TIMEZONE_HOUR -> error("Invalid call DATE_DIFF(TIMEZONE_HOUR, ...)")
                 DatetimeField.TIMEZONE_MINUTE -> error("Invalid call DATE_DIFF(TIMEZONE_MINUTE, ...)")
-                else -> call("date_diff_${node.field.name.lowercase()}", arg0, arg1)
+                else -> call("date_diff_${node.field.name.lowercase()}", true, arg0, arg1)
```
RCHowell, line 483: Since all of these are `call("..", true, ...)` I'm wondering if you should make `call("..", args...)` the default with true. Then you make another function `userCall("..", args...)` or something like that with a better name where the arg is `false`. This would cut down on the diff and we won't have a hanging boolean arg

partiql-planner/src/main/kotlin/org/partiql/planner/Header.kt
```diff
@@ -165,6 +162,10 @@ internal class Header(
             return functions.flatMap { it.sortedWith(functionPrecedence) }.groupBy { it.name }
         }
 
+        public fun combineInternal(vararg functions: List<FunctionSignature>): FunctionMap {
+            return functions.flatMap { it.sortedWith(functionPrecedence) }.groupBy { "${HIDDEN_FLAG}_${it.name}" }
```
RCHowell, line 166: Perfect. The lookup key is what we hide

RCHowell (changes requested):

partiql-planner/src/main/kotlin/org/partiql/planner/transforms/RexConverter.kt
yliuuuu: I thought, and internally debated about this. The benefit I see using this approach is that it forces the implementer to explicitly write out the whether or not the function is hidden. 

That said, two functions with different names would work.

partiql-planner/src/main/kotlin/org/partiql/planner/Header.kt
```diff
@@ -536,19 +540,22 @@ internal class Header(
...
 
+        // NULLIF(x, y)
         private fun nullIf(): List<FunctionSignature> = nullableTypes.map { t ->
             FunctionSignature(
                 name = "null_if",
                 returns = t,
                 parameters = listOf(
                     FunctionParameter("value", t),
-                    FunctionParameter("nullifier", BOOL),
+                    FunctionParameter("nullifier", BOOL), // TODO: why is this BOOL?
```
RCHowell, line 550: You're right, it should be `lhs: t, rhs: t` — we need not fix that here though.

RCHowell (approved):

RCHowell (approved):

PR #1254, opened 2023-10-30
Use int32 as the default integer type.

RCHowell:
## Relevant Issues
- N/A

## Description

This PR types integer literals as an `int32` -> `int64` -> `int`. Typing as the unbounded integer type for literal values make compatibility with other SQL databases cumbersome .. requiring either explicit CASTs (which also require special handling during typing) or re-mapping INT types during planning.

Generally, a CAST from INT to INT4 or INT8 is _possibly_ lossy; there may be an INT that does not fit in an INT4. So `cast_int(<int32>)` and `cast_int(<int64>)` both have return type `missing|int`. We could examine the argument to see if its both a literal and fits; but this is precisely what bounding in the parser accomplishes.

## Other Information
No

No, the planner is not released.

No

Yes

## License Information


partiql-parser/src/main/kotlin/org/partiql/parser/impl/PartiQLParserDefault.kt
```diff
@@ -1930,9 +1932,31 @@ internal class PartiQLParserDefault : PartiQLParser {
         }
 
         override fun visitLiteralInteger(ctx: GeneratedParser.LiteralIntegerContext) = translate(ctx) {
-            val n = ctx.LITERAL_INTEGER().text.toInt()
-            val v = intValue(n.toBigInteger())
-            exprLit(v)
+            val n = ctx.LITERAL_INTEGER().text
+
+            // 1st, try parse as int
```
johnedquinn, line 1937: Leaving this as an optional change. If we aren't going to do anything with the exceptions:

```kotlin
val text = ctx.LITERAL_INTEGER().text
val integer = text.toIntOrNull()?.let { int32Value(it) }
    ?: text.toLongOrNull()?.let { int64Value(it) }
    ?: text.toBigIntegerOrNull()?.let { intValue(it) }
    ?: throw error(ctx, "Unable to parse integer: $text.")
exprLit(integer)
```

PR #1259, opened 2023-11-02
Adds quoting of subqueries to pretty-printer

RCHowell:
## Description

This adds quoting of subqueries via the pretty-printer. It also exposes a place to control expression quoting behavior.

Yes

## License Information

### DIFF
--- a/partiql-ast/src/main/kotlin/org/partiql/ast/sql/SqlDialect.kt
+++ b/partiql-ast/src/main/kotlin/org/partiql/ast/sql/SqlDialect.kt
@@ -48,6 +48,23 @@ public abstract class SqlDialect : AstBaseVisitor<SqlBlock, SqlBlock>() {
 
     // IDENTIFIERS & PATHS
 
+    /**
+     * Default behavior is quote all SFW queries.
+     *
+     * @param node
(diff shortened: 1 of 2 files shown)


partiql-ast/src/main/kotlin/org/partiql/ast/sql/SqlDialect.kt
```diff
@@ -48,6 +48,23 @@ public abstract class SqlDialect : AstBaseVisitor<SqlBlock, SqlBlock>() {
 
     // IDENTIFIERS & PATHS
 
+    /**
+     * Default behavior is quote all SFW queries.
+     *
+     * @param node
+     * @param head
+     */
+    public open fun visitExprQuoted(node: Expr, head: SqlBlock) = when (node) {
```
johnedquinn, line 57:
```suggestion
    /**
     * Default behavior is to wrap all SFW queries with parentheses.
     *
     * @param node
     * @param head
     */
    public open fun visitExprWrapped(node: Expr, head: SqlBlock) = when (node) {
```

johnedquinn (changes requested):

PR #1262, opened 2023-11-02
In-memory Connector

yliuuuu:
## Relevant Issues

## Description
- Adds in memory connector for testing purpose. 
- Fixes the CI builds. 

Notes on CI builds: 
1. Github actions used packaged code (JAR) for build. 
      - We can not turn folder/files in Jar to a Java File object easily, especially with the current set up in which the local plugin takes a path and we retrieve the path and turn it to `File` object later. 
      - Solution:
          -   During build process, we list out the path to each individual test resource, and store the path into a single file. 
          -   We retrieve the content of resource using path, and load the content to a in-memory connector. 
2. Github checkout action is not running against the code in the RP. Instead, it runs against some merged version that is based on top of the target branch. 
     Fix : https://github.com/actions/checkout/blob/v3.0.2/README.md#user-content-checkout-pull-request-head-commit-instead-of-merge-commit
     
Commits: 
-  [a4719e4](https://github.com/partiql/partiql-lang-kotlin/pull/1262/commits/a4719e44cd6cde889a4f3509d49776c381b4c8e3) : Implemented and tested the in-memory connector
- [130e2ab](https://github.com/partiql/partiql-lang-kotlin/pull/1262/commits/130e2abed3382650a2ff2cd2d087f3bac6fced62) : Switch the schema inferencer to use in-memory plugin. This includes the logic to generate a `resource_path.txt` file during build process, and use the file to populate the in-memory connector. Also, currently we have a couple test cases that are failing due to the `ANY` function resolution in plan. I added a `IgnoredTestCase` to workaround this issue. 
- [681c593](https://github.com/partiql/partiql-lang-kotlin/pull/1262/commits/681c59368844874c7717745b0fd2483deabfe907): This commits changes the loading logic of test case provider. 
- [a358d71](https://github.com/partiql/partiql-lang-kotlin/pull/1262/commits/a358d719ee4ef63628659d2416f06d631ee06cd3) : This commits change the checkout action to run on pull request head commit. 
     

## Other Information
  - No. 

  -  No. 
 
  - No. 
  
Yes. 

## License Information

### DIFF
--- a/partiql-lang/src/test/kotlin/org/partiql/lang/planner/SchemaLoader.kt
+++ b/partiql-lang/src/test/kotlin/org/partiql/lang/planner/SchemaLoader.kt
@@ -1,0 +1,229 @@
+package org.partiql.lang.planner
+
+import com.amazon.ionelement.api.IonElement
+import com.amazon.ionelement.api.ListElement
+import com.amazon.ionelement.api.StringElement
+import com.amazon.ionelement.api.StructElement
+import com.amazon.ionelement.api.SymbolElement
+import com.amazon.ionelement.api.ionListOf
+import com.amazon.ionelement.api.ionString
+import com.amazon.ionelement.api.ionStructOf
+import com.amazon.ionelement.api.ionSymbol
+import org.partiql.types.AnyOfType
+import org.partiql.types.AnyType
+import org.partiql.types.BagType
+import org.partiql.types.BlobType
+import org.partiql.types.BoolType
+import org.partiql.types.ClobType
+import org.partiql.types.DateType
+import org.partiql.types.DecimalType
+import org.partiql.types.FloatType
+import org.partiql.types.GraphType
+import org.partiql.types.IntType
+import org.partiql.types.ListType
+import org.partiql.types.MissingType
+import org.partiql.types.NullType
+import org.partiql.types.SexpType
+import org.partiql.types.StaticType
+import org.partiql.types.StringType
+import org.partiql.types.StructType
+import org.partiql.types.SymbolType
+import org.partiql.types.TimeType
+import org.partiql.types.TimestampType
+import org.partiql.types.TupleConstraint
+
+// TODO: This code is ported from plugins/partiql-local/src/main/kotlin/org/partiql/plugins/local/LocalSchema.kt
+//  In my opinion, the in-memory connector should be independent of schema file format,
+//  hence making it inappropriate to leave the code in plugins/partiql-memory
+//  We need to figure out where to put the code.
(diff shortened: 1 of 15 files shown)


yliuuuu:
Pause this work. 

The issue is related to local plugin. 

The local plugin attempts retrieve catalog/schema via `xxx.class.getResource(...)`. It treats directory as catalog. 

In local build the files system approach works fine as we can directly access the file via URI. 

For some reasons, in GitHub work flow, the build runs jar instead of a local file system. We could not convert a folder in Jar to Java File Object. 

TODO: 
1. Figure out why GitHub work flow runs the jar instead of extracting from file system. 
2. If necessary, we may need a in-memory connector.

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
```diff
@@ -136,33 +137,54 @@ class PartiQLSchemaInferencerTests {
...
+                        map[catalogName]!!.add(subPath to staticType)
+                    } else {
+                        map[catalogName] = mutableListOf(subPath to staticType)
+                    }
+                }
+            }
+            map.forEach { (k: String, v: MutableList<Pair<String, StaticType>>) ->
+                it[k] = MemoryCatalog.of(*v.toTypedArray())
+            }
+        }
```
RCHowell, line 167: I see, rather than porting the catalogs from the .ion files you create the mem catalog from them. This works, but would be nice to clean up the resources in a later PR. Again, these are testing concerns so nothing pressing

partiql-lang/src/test/kotlin/org/partiql/lang/planner/SchemaLoader.kt
```diff
@@ -0,0 +1,229 @@
...
+import org.partiql.types.StructType
+import org.partiql.types.SymbolType
+import org.partiql.types.TimeType
+import org.partiql.types.TimestampType
+import org.partiql.types.TupleConstraint
+
+// TODO: This code is ported from plugins/partiql-local/src/main/kotlin/org/partiql/plugins/local/LocalSchema.kt
+//  In my opinion, the in-memory connector should be independent of schema file format,
+//  hence making it inappropriate to leave the code in plugins/partiql-memory
+//  We need to figure out where to put the code.
```
RCHowell, line 38: I think moving the catalogs from resources to in-memory is a better long term solution and don't need to duplicate this.

RCHowell (approved):

PR #1266, opened 2023-11-14
Adds additional PlanTyper test cases

yliuuuu:
## Relevant Issues

## Description
- Rebased of PR #1267 . 
- PartiQL Value Type Change: 
    - Add in `Fixed Precision Decimal` to temporary unblock type coercion issues. 
    - FLOAT + FIXED PRECISION DECIMAL -> FLOAT
    - FLOAT + DECIMAL -> DECIMAL
- Type Precedence Change: 
    - Bool should not be coercible to numeric type. 
    - Symbol is coercible  to a String, but String is not coercible to a Char. 
- Rewrite: 
   - Not equal operator get re-write to `Negate(Eq)`
- Function Resolver: 
   -  If the matched functions requires all the parameters to be casted to another Typer, then we do not return such matched function as the resolved function.
- Typer:
   -  A lookup table to resolve logical functions and boolean connectives. 
- Header Function
  - Equal Function should be nullCall. 
  - Boolean Connectives `AND`, `OR`, are nullable. 
  - Predicate `Is NULL`, `IS MISSING` are not nullcall, not nullable and is not missingCall. 
  - IS Type Assertion , i.e., IS INT2,  are not nullcall (should they be?), not nullable, but is missingCall. 

## Other Information
  - < If NO, why? >

  - < If YES, why? >

## License Information


RCHowell (approved):

partiql-planner/src/main/kotlin/org/partiql/planner/typer/FnResolver.kt
```diff
@@ -426,9 +433,10 @@ internal class FnResolver(private val headers: List<Header>) {
             INT32,
             INT64,
             INT,
-            DECIMAL,
+            FIX_PRECISION_DECIMAL,
```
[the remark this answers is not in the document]
yliuuuu: The type precedence. 

Currently FLOAT + DECIMAL(arbitrary precision)  -> DECIMAL(arbitrary precision)
FLOAT + FIXED_PRECISION_DECIMAL -> FLOAT

partiql-planner/src/main/kotlin/org/partiql/planner/typer/PlanTyper.kt
```diff
@@ -659,26 +687,49 @@ internal class PlanTyper(
...
+                        val type = AnyOfType(ref.type.allTypes.filterIsInstance<StructType>().toSet())
+                        val replacementVal = ref.copy(type = type)
+                        when (ref.op is Rex.Op.Var.Resolved) {
+                            true -> RexReplacer.replace(result, ref, replacementVal)
+                            false -> result
+                        }
+                    }
+                    val type = rex.toUnionType().flatten()
+
+                    return rexOpCaseBranch(condition, result.copy(type))
```
yliuuuu, line 709: Similar hack should be used when the call is dynamic. We should consider if there is a way to better modeling this.

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
```diff
@@ -3190,12 +3190,13 @@ class PartiQLSchemaInferencerTests {
...
                 expected = TYPE_BOOL
             ),
+            // MISSING = 1
             ErrorTestCase(
                 name = "Case Sensitive failure",
                 catalog = CATALOG_DB,
                 catalogPath = DB_SCHEMA_MARKETS,
                 query = "order_info.\"CUSTOMER_ID\" = 1",
-                expected = TYPE_BOOL
+                expected = NULL
```
yliuuuu, line 3199: MISSING = 1 returns NULL Type.

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
```diff
@@ -3209,14 +3210,14 @@ class PartiQLSchemaInferencerTests {
                 catalog = CATALOG_DB,
                 catalogPath = DB_SCHEMA_MARKETS,
                 query = "(order_info.customer_id = 1) AND (order_info.marketplace_id = 2)",
-                expected = TYPE_BOOL
+                expected = StaticType.unionOf(BOOL, NULL)
```
yliuuuu, line 3213: Because our function header says the Logical functions AND/OR are `nullable`, the inferencer returns `unionOf(BOOL, NULL)` even for input like `BOOL AND BOOL` OR `NULL AND NULL`. 

The workaround at this stage would be to have a look up table in the inferencer.

RCHowell:
Happy to review after rebase. Thank you.

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
RCHowell: We could also model special operator functions differently in the unresolved plan since that is internalized. Food for thought, no changes requested at the moment.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/FnResolver.kt
```diff
@@ -429,6 +436,7 @@ internal class FnResolver(private val headers: List<Header>) {
             DECIMAL,
             FLOAT32,
             FLOAT64,
+            DECIMAL_ARBITRARY, // Arbitrary precision decimal has a higher precedence than FLOAT
```
RCHowell, line 439: Nice, this is very clear now.

RCHowell (approved):

PR #1267, opened 2023-11-14
Adds support for dynamic dispatch of functions

johnedquinn:
## Relevant Issues
- None

## Description
- Adds support for dynamic dispatch of functions.
- I've opted to model the possible "candidates" directly within the plan under a dynamic function call. We re-use the same logic of static calls (inserting the coercions where necessary) and maintaining the order of potential functions to dynamically invoke.
  - By modeling it this way, we can remove all the `any -> any` functions from the PartiQL header.
  - We also don't need to generate a massive `CASE WHEN` -- we can let the engine handle this specifically.
  - We handle MISSING appropriately.
- In order to do this effectively, I compute the permutations of the potential types of the arguments. Consider the following invocation: `BINARY_PLUS(INT4|INT8, STRING|INT8|INT2)`. The permutations would result in the following "possible" function calls:
  - `BINARY_PLUS(INT4, STRING) -> MISSING` (doesn't exist)
  - `BINARY_PLUS(INT4, INT8)` -> `BINARY_PLUS(CAST(<INT4> AS INT8), INT8) -> INT8` (exists, but needs to coerce)
  - `BINARY_PLUS(INT4, INT2)` -> `BINARY_PLUS(INT4, CAST(<INT2> AS INT4)) -> INT4` (exists, but needs to coerce)
  - `BINARY_PLUS(INT8, STRING) -> MISSING` (doesn't exist)
  - `BINARY_PLUS(INT8, INT8) -> INT8` (exists and no coercions necessary)
  - `BINARY_PLUS(INT8, INT2)` -> `BINARY_PLUS(INT8, CAST(<INT2> AS INT8)) -> INT8` (exists, but needs to coerce)
  - So, in this scenario, we have computed that the result would be `INT8 | INT4 | MISSING`. `MISSING` when a function cannot be dynamically invoked due to incompatible types.
- NOTE: This does not deal with aggregation functions yet. That is for a subsequent PR.

## Other Information
- Updated Unreleased Section in CHANGELOG: **NO** -- planner branch
- Any backward-incompatible changes? **YES** -- no, only to planner branch

## License Information

### DIFF
--- a/partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
+++ b/partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
@@ -480,7 +480,7 @@ class PartiQLSchemaInferencerTests {
                         )
                     )
                 }
-            ).toIgnored("Plus op will be resolved to PLUS__ANY_ANY__ANY"),
+            ),
         )
 
         @JvmStatic
@@ -498,7 +498,7 @@ class PartiQLSchemaInferencerTests {
             SuccessTestCase(
                 name = "BITWISE_AND_3",
                 query = "1 & 2",
-                expected = StaticType.INT4
+                expected = INT4
             ),
             SuccessTestCase(
                 name = "BITWISE_AND_4",
@@ -561,7 +561,7 @@ class PartiQLSchemaInferencerTests {
                         PlanningProblemDetails.UnknownFunction("bitwise_and", listOf(INT4, STRING))
                     )
                 }
-            ).toIgnored("Bitwise And opearator will be resolved to BITWISE_AND__ANY_ANY__ANY"),
+            ),
         )
 
         @JvmStatic
@@ -2964,7 +2964,7 @@ class PartiQLSchemaInferencerTests {
                         )
                     )
                 }
-            ).toIgnored("Between will be resolved to BETWEEN__ANY_ANY_ANY__BOOL"),
+            ),
             SuccessTestCase(
                 name = "LIKE",
                 catalog = CATALOG_DB,
@@ -2987,7 +2987,7 @@ class PartiQLSchemaInferencerTests {
                         )
                     )
                 }
-            ).toIgnored("Like Op will be resolved to LIKE__ANY_ANY__BOOL"),
+            ),
             SuccessTestCase(
                 name = "Case Insensitive success",
                 catalog = CATALOG_DB,
@@ -3058,7 +3058,7 @@ class PartiQLSchemaInferencerTests {
                         )
                     )
                 }
-            ).toIgnored("And Op will be resolved to AND__ANY_ANY__BOOL"),
+            ),
             ErrorTestCase(
                 name = "Bad comparison",
                 catalog = CATALOG_DB,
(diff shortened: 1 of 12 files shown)


yliuuuu (commented):
We can remove the `IsMissable` in function signature for now.

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
```diff
@@ -2515,6 +2524,172 @@ class PartiQLSchemaInferencerTests {
...
+                    FROM <<
+                        { 'a': CAST(1 AS INT8), 'b': CAST(1.0 AS DECIMAL) },
+                        { 'a': CAST(1 AS INT4), 'b': TRUE },
+                        { 'a': 'hello world!!', 'b': DATE '2023-01-01' }
+                    >> AS t
+                """.trimIndent(),
+                expected = BagType(
+                    StructType(
+                        fields = mapOf(
+                            "c" to unionOf(INT4, INT8, MISSING, DECIMAL),
```
yliuuuu, line 2613: Why is this returns a union of number types? 
We have `a` whose type is unionOf(INT8, INT4, String), `b` whose type is unionOf(Decimal, Bool, Date), 
Should the plus operator returns unionOf(Decimal, Missing)?

johnedquinn: Bool can coerce to INT4 and INT8

yliuuuu: Got it. But I don't think we should allow this.

johnedquinn:
For reviewers, commit https://github.com/partiql/partiql-lang-kotlin/pull/1267/commits/0af05147b58d682ee7b88787de67778963a8e928 is a larger diff than necessary. I refactored the signature of FunctionSignature.Scalar using IntelliJ to remove `isMissable`, and it auto-formatted all instance creations of Scalar.

partiql-plan/src/main/resources/partiql_plan_0_1.ion
```diff
@@ -115,10 +118,34 @@ rex::{
...
+        _: [
+
+          // Represents a viable candidate.
+          //
+          // fn: The resolved function
+          // coercions: Represents the coercion that should take place -- if not null. The first item is to be wrapped
+          //  over the first argument of dynamic.args and so on.
+          candidate::{
+            fn: '.fn.resolved',
+            coercions: list::[optional::fn]
```
RCHowell, line 144: Why not `optional::'.fn.resolved'`?

johnedquinn: Each candidate function is statically known, so this can't be optional. This is the equivalent of "specific".

RCHowell: I meant for the coercions

johnedquinn: Ah, I see what you're getting at. Totally agree -- let me address, and I'll resubmit. Thanks!

RCHowell (approved):

PR #1271, opened 2023-11-29
Modifies how we model paths in the plan

johnedquinn:
## Relevant Issues
- None

## Description
- Modifies how we model paths in the plan
- As discussed offline, when dealing with the semantics of pathing, there are 3 core operations:
  1. Indexing into an array. Ex: `a[0], a[1 + 1]`
  2. Sensitive lookup in a tuple. Ex: `a."b", a['b'], a[CAST('a' || 'b' AS STRING)]`
  3. Insensitive lookup in a tuple. Ex: `a.b`
- The above will be part of the PUBLIC API. However, in order to appropriately handle the subtle differences between `a.b`, `a."b"`, and `a['b']` when we haven't typed yet-- the internal APIs contain:
  1. Indexing into an array. Ex: `a[0], a[1 + 1]`
  2. Sensitive lookup on a tuple. Ex: `a['b'], a[CAST('a' || 'b' AS STRING)]`
  3. Dot-notation lookup on a tuple/schema. Ex: `a.b` and `a."b"`. This has a subtle difference to the public API.
- We convert the dot-notation lookup to either a sensitive or insensitive lookup after typing/resolving. In my opinion, this should happen at the AST transformation, however, that is not currently supported.
- See the added `PartiQLSchemaInferencerTests` for more information.

## Other Information
  - Yes, but for the unreleased planner.

## License Information

### DIFF
--- a/partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
+++ b/partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/PartiQLSchemaInferencerTests.kt
@@ -35,6 +35,7 @@ import org.partiql.types.BagType
 import org.partiql.types.ListType
 import org.partiql.types.SexpType
 import org.partiql.types.StaticType
+import org.partiql.types.StaticType.Companion.ANY
 import org.partiql.types.StaticType.Companion.BAG
 import org.partiql.types.StaticType.Companion.BOOL
 import org.partiql.types.StaticType.Companion.DATE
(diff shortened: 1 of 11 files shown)


RCHowell:
Will review after rebase. Thank you.

partiql-plan/src/main/resources/partiql_plan_0_1.ion
```diff
@@ -110,9 +110,21 @@ rex::{
...
+          // Case-insensitive lookup. The key MUST be a literal string. Ex: a.b
+          symbol::{ key: string },
+
+          // For arrays. Ex: a[*]
+          // TODO: Do we need this? According to specification: [1,2,3][*] ⇔ SELECT VALUE v FROM [1, 2, 3] AS v
           wildcard::{},
+
+          // For tuples. Ex: a.*
+          // TODO: Do we need this? According to specification: {'a':1, 'b':2}.* ⇔ SELECT VALUE v FROM UNPIVOT {'a':1, 'b':2} AS v
           unpivot::{},
```
RCHowell, line 128: Perfect!

partiql-planner/src/main/kotlin/org/partiql/planner/transforms/RexConverter.kt
```diff
@@ -145,11 +148,28 @@ internal object RexConverter {
...
                     is Expr.Path.Step.Symbol -> {
-                        val identifier = AstToPlan.convert(it.symbol)
-                        rexOpPathStepSymbol(identifier)
+                        when (it.symbol.caseSensitivity) {
+                            org.partiql.ast.Identifier.CaseSensitivity.SENSITIVE -> rexOpPathStepKey(
+                                rex(
+                                    type = StaticType.STRING,
+                                    op = rexOpLit(stringValue(it.symbol.symbol))
+                                )
+                            )
```
RCHowell, line 170: We want this to be a `rexOpPathSymbol` right? For global resolution.

```
a."x" != a["x"]  -- !! only for global variable names, the prior is treated as a qualified identifier
```

RCHowell (changes requested):

partiql-planner/src/main/kotlin/org/partiql/planner/transforms/RexConverter.kt
johnedquinn: With your recent PR for internalizing the unresolved IRs, I added the retention of this subtle difference to the internal IR. Then, converting from the internal to the public, I transform it accordingly. See the updated PR description.

RCHowell (approved):

PR #1274, opened 2023-11-30
Fixes CLI hanging on invalid queries

### LINKED ISSUE #1230: v0.13.1 CLI hangs on invalid queries
## Description
Running an invalid query in the v0.13.1 CLI causes the CLI to hang. Previously, this would result in a PartiQL error.

## To Reproduce
1. Pull down v0.13.1 CLI or build `main` branch CLI
2. Run an invalid query (syntax or evaluation)
3. Query hangs unless CTRL+C is used
```
Welcome to the PartiQL shell!
Typing mode: LEGACY
Using version: 0.13.1-8a8fdd09
PartiQL> SELECT;
^C==='
'Compilation cancelled.'
---
OK!
```

## Expected Behavior
Expect the previous behavior of giving a PartiQL error without hanging.

## Additional Context
- Java version: 11
- PartiQL version: 0.13.1

johnedquinn:
## Relevant Issues
- Closes #1230 

## Description
- There was a bug introduced in our implementation of the CLI's RunnableWriter and RunnablePipeline.
  - To break it down:
    1. The main thread was designed to send interrupt signals to the relevant `RunnableWriter` and `RunnablePipeline` which would result in the tear-down of the child thread. We would then re-create the child thread and instantiate a new runnable in its place to continue to allow user inputs.
    2. There was a bug when the writer or pipeline were to fail for reasons beyond thread interruption. Essentially, on invalid queries, the child thread would die (due to unhandled exceptions), and the main thread would not know of its passing. So, the CLI would "hang" -- since we wouldn't be able to pass data between the runnables via the blocking queues.
- This PR fixes this bug by handling invalid queries.

## Reviewer Testing
- I personally recommend running the following queries:
```
-- Regular query
SELECT a, b
FROM << { 'a': 1, 'b': 2 } >> AS t;

-- Regular query that should fail during eval
SELECT a, b, c, d
FROM << { 'a': 1, 'b': 2 } >> AS t;

-- Regular query that should fail during parse
SELECT a;

-- This adds to the global bindings
!add_to_global_env { 'a': 1 }

-- This should work
a;

-- This should hang on compilation. Use CTRL-C to escape.
SELECT COUNT(*)
FROM
  ([1, 2, 3, 4]) as x1,
  ([1, 2, 3, 4]) as x2,
  ([1, 2, 3, 4]) as x3,
  ([1, 2, 3, 4]) as x4,
  ([1, 2, 3, 4]) as x5,
  ([1, 2, 3, 4]) as x6,
  ([1, 2, 3, 4]) as x7,
  ([1, 2, 3, 4]) as x8,
  ([1, 2, 3, 4]) as x9,
  ([1, 2, 3, 4]) as x10,
  ([1, 2, 3, 4]) as x11
GROUP BY x1._1;

-- This should essentially run for a very long time during evaluation. Use CTRL-C to cancel.
SELECT *
FROM
  ([1, 2, 3, 4]) as x1,
  ([1, 2, 3, 4]) as x2,
  ([1, 2, 3, 4]) as x3,
  ([1, 2, 3, 4]) as x4,
  ([1, 2, 3, 4]) as x5,
  ([1, 2, 3, 4]) as x6,
  ([1, 2, 3, 4]) as x7,
  ([1, 2, 3, 4]) as x8,
  ([1, 2, 3, 4]) as x9,
  ([1, 2, 3, 4]) as x10,
  ([1, 2, 3, 4]) as x11;

-- If you start up the shell with --pipeline EXPERIMENTAL, you can run this:
EXPLAIN SELECT 1 FROM <<0, 1, 2>>
```

## Other Information

## License Information

### DIFF
--- a/partiql-cli/src/main/kotlin/org/partiql/cli/shell/RunnableWriter.kt
+++ b/partiql-cli/src/main/kotlin/org/partiql/cli/shell/RunnableWriter.kt
@@ -51,7 +51,6 @@ internal class RunnableWriter(
                     } catch (t: Throwable) { printThrowable(t) }
                     is RunnablePipeline.Output.Error -> printThrowable(value.throwable)
                 }
-                out.println()
                 out.flush()
                 donePrinting.set(true)
             }
@@ -66,6 +65,7 @@ internal class RunnableWriter(
                 formatter.formatTo(result.value, out)
                 out.println()
                 out.info(BAR_2)
+                out.success("OK!")
             }
             is PartiQLResult.Explain.Domain -> {
                 val explain = ExplainFormatter.format(result)


yliuuuu (approved):
I can verify the fix works. 

I feel like the state management is a bit tricky to understand. 

Maybe we can think about if there is a better way to model this later.

PR #1281, opened 2023-12-04
add support of text and bool type to comparison

yliuuuu:
## Relevant Issues
- [Closes/Related To] N/A

## Description
- Add support of Boolean Type and Text Type (Symbol, String, Clob) to comparison predicate (greater than, greater than or equal, less than, less than or equal.)
- Add support for Text type in Between predicate. 
- SQL spec permits comparison between: 
    - Number
    - Character String
    - Binary String
    - datetime
    - Interval
    - Boolean 

## Other Information
  - No. 

  -  No. 

  -  No. 

- Yes. 

## License Information

### DIFF
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/PartiQLHeader.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/PartiQLHeader.kt
@@ -254,19 +254,19 @@ public object PartiQLHeader : Header() {
         ),
     )
 
-    private fun lt(): List<FunctionSignature.Scalar> = types.numeric.map { t ->
+    private fun lt(): List<FunctionSignature.Scalar> = (types.numeric + types.text + BOOL).map { t ->
         binary("lt", BOOL, t, t)
     }
 
-    private fun lte(): List<FunctionSignature.Scalar> = types.numeric.map { t ->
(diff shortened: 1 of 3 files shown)


partiql-planner/src/main/kotlin/org/partiql/planner/PartiQLHeader.kt
```diff
@@ -347,7 +347,7 @@ public object PartiQLHeader : Header() {
         )
     }
 
-    private fun between(): List<FunctionSignature.Scalar> = types.numeric.map { t ->
+    private fun between(): List<FunctionSignature.Scalar> = (types.numeric + types.text).map { t ->
```
johnedquinn, line 350:
```suggestion
    private fun between(): List<FunctionSignature.Scalar> = (types.numeric + types.text + BOOL).map { t ->
```

Given the other changes, should we add this?

yliuuuu: I choose to not add this in this PR because I hope once we have finalized missing prorogation rule, we can just perform a rewrite on Between. 

`Bool BETWEEN Bool AND Bool` on itself does not make much sense to me, and at the moment this is no use case require this function.

johnedquinn (approved):

PR #1288, opened 2023-12-08
fix typer/transfomrer

yliuuuu:
## Relevant Issues

## Description
- We need explicitly override the `visitRexOpPathStep` in the typer as in some case we need the resolved node as the step. i.e., the path step is a call node
- The Typer should throw an problem code upon encounter an unresolved function. Hence in `PlanTransform` we can just error out. 
    - I am not sure what is the purpose of the error node in the resolved plan. If we don't need this, I can have a follow up PR to remove it. 

## Other Information
  - No.

  -  No.
  
  - No.
  
No. 

## License Information


RCHowell (changes requested):
I didn't want to error out because we want to collect all plan errors. I'm not sure the best way go here. Ideally we don't have an error node, but I don't think we can throw?

I'm approving because I think throwing is better than having a hidden error node in the **resolved** plan

RCHowell (approved):

PR #1291, opened 2023-12-08
Adds int32 and int64 variants of date_add

RCHowell:
## Relevant Issues
N/A

## Description

Adds int32 and int64 operators to the date_add function. Before we only had the arbitrary-sized integer which we don't want as default.

## Other Information
No

No

No

Yes

## License Information

### DIFF
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/PartiQLHeader.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/PartiQLHeader.kt
@@ -590,23 +590,26 @@ public object PartiQLHeader : Header() {
     private fun extract(): List<FunctionSignature.Scalar> = emptyList()
 
     private fun dateAdd(): List<FunctionSignature.Scalar> {
+        val intervals = listOf(INT32, INT64, INT)
         val operators = mutableListOf<FunctionSignature.Scalar>()
         for (field in DatetimeField.values()) {
             for (type in types.datetime) {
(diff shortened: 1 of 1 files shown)


partiql-planner/src/main/kotlin/org/partiql/planner/PartiQLHeader.kt
```diff
@@ -590,23 +590,26 @@ public object PartiQLHeader : Header() {
     private fun extract(): List<FunctionSignature.Scalar> = emptyList()
 
     private fun dateAdd(): List<FunctionSignature.Scalar> {
+        val intervals = listOf(INT32, INT64, INT)
```
yliuuuu, line 593:
```suggestion
        val intervals = listOf(INT8, INT16, INT32, INT64, INT)
```

yliuuuu: What if the input is not a literal? Would you still prefer an implicit cast?

yliuuuu: DateAdd(Day, a_int16_type, b_timestamp_type)

RCHowell: Yeah that's fine. We mostly need int32 for our target systems

RCHowell: The default was INT which should never be the default. The default integer should always be INT32. Then INT64 is between them, hence why I did those three

yliuuuu (approved):

PR #1294, opened 2023-12-11
Adds INNER JOIN, LEFT JOIN, Path Step Key & Fixes ExprStruct

johnedquinn:
## Relevant Issues
- None

## Description
- Adds INNER JOIN, LEFT JOIN, Path Step Key & Fixes ExprStruct
- Going to add the other JOINS after this. Wanted to get the ExprStruct and PathStepKey out.

## Other Information

## License Information

### DIFF
--- a/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelJoin.kt
+++ b/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelJoin.kt
@@ -2,11 +2,15 @@ package org.partiql.eval.internal.operator.rel
 
 import org.partiql.eval.internal.Record
 import org.partiql.eval.internal.operator.Operator
+import org.partiql.value.BoolValue
+import org.partiql.value.PartiQLValue
+import org.partiql.value.PartiQLValueExperimental
 
 internal abstract class RelJoin : Operator.Relation {
 
     abstract val lhs: Operator.Relation
     abstract val rhs: Operator.Relation
+    abstract val condition: Operator.Expr
 
     private val lhsStored = mutableListOf<Record>()
     private val rhsStored = mutableListOf<Record>()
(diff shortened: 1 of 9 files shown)


partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelJoin.kt
```diff
@@ -0,0 +1,65 @@
+package org.partiql.eval.internal.operator.rel
+
+import org.partiql.eval.internal.Record
+import org.partiql.eval.internal.operator.Operator
+import org.partiql.value.BoolValue
+import org.partiql.value.PartiQLValue
+import org.partiql.value.PartiQLValueExperimental
+
+internal abstract class RelJoin : Operator.Relation {
```
RCHowell, line 9:
```suggestion
internal abstract class RelJoinNestedLoop : Operator.Relation {
```

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelJoin.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    @OptIn(PartiQLValueExperimental::class)
+    override fun next(): Record? {
+        lhsIterator.forEach { lhsRecord ->
+            rhsIterator.forEach { rhsRecord ->
+                val input = lhsRecord + rhsRecord
+                val result = condition.eval(input)
+                getOutputRecord(result.isTrue(), lhsRecord, rhsRecord)?.let { return it }
+            }
+        }
```
RCHowell, line 50: This is simple enough for now, but we don't have to store both relations for this. You keep a pointer to the left-hand-side record and right-hand-side record. Advance the right-hand-side until condition holds, then yield. Once right-hand-side is exhausted, increment the left hand side.

We have to really test our PartiQLValue equality and hash now and we can use different join algorithms

RCHowell (approved):

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelJoin.kt
johnedquinn: Agreed. I'm going to merge this now with the intention of modifying this in the next PR. There's only one way to go and that's up.

PR #1295, opened 2023-12-12
Adds support for FULL/RIGHT OUTER JOIN and improves performance of JOINs

johnedquinn:
## Relevant Issues
- A follow-up to #1294 

## Description
- Adds support for FULL/RIGHT OUTER JOIN and improves performance of JOINs
- I needed to update the equality of `BagValues`. Before, `BagValues` weren't taking into account being unordered, so I grouped and counted the elements for equality. I also needed to update `StructValues` to allow for `hashCode`. I essentially delegated the fields of a struct to some private lazily instantiated `_fields` variable.

## Other Information

## License Information


partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelJoinOuterFull.kt
```diff
@@ -0,0 +1,50 @@
...
+    private var previousRhs: Record? = null
+
+    override fun next(): Record? {
+        if (previousLhs != null && previousRhs != null) {
+            previousRhs!!.padNull()
+            val newRecord = previousLhs!! + previousRhs!!
+            previousLhs = null
+            previousRhs = null
+            return newRecord
+        }
```
RCHowell, line 38: What about checking the join condition?

partiql-types/src/main/kotlin/org/partiql/value/PartiQLValue.kt
```diff
@@ -471,10 +471,13 @@ public abstract class StructValue<T : PartiQLValue> : PartiQLValue, Sequence<Pai
 
     public abstract val fields: Sequence<Pair<String, T>>?
```
RCHowell, line 472: I kind of think this should just be a list or array. Array makes more sense honestly because of fixed size rather than an immutable List, but either way not a sequence

RCHowell (changes requested):

partiql-types/src/main/kotlin/org/partiql/value/PartiQLValue.kt
johnedquinn: I believe you're tackling this in your PR for modifying PartiQL Value.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelJoinOuterFull.kt
johnedquinn: Great catch! Updated and I added a test to use TRUE as the condition.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelJoinNestedLoop.kt
```diff
@@ -0,0 +1,79 @@
...
+    private var rhsRecord: Record? = null
+
+    override fun open() {
+        lhs.open()
+        rhs.open()
+        rhsRecord = rhs.next()
+    }
+
+    @OptIn(PartiQLValueExperimental::class)
+    abstract fun getOutputRecord(result: Boolean, lhs: Record, rhs: Record): Record?
```
RCHowell, line 27: `join`

https://www.youtube.com/watch?v=PEgk2v6KntY

RCHowell (approved):

PR #1303, opened 2023-12-13
Adds getBindings to PartiQL Plugin

RCHowell:
This PR introduces a `getBindings()` method to PartiQL Plugins. A plugin will return a PartiQLBindings object which returns a PartiQLValue given a BindingPath.

I have implemented this for the MemoryPlugin which is now used to load globals from the conformance test suite. This PR also includes an implementation of a TestExecutor for the ongoing evaluator work.

## License Information


johnedquinn (approved):
This looks good to me. In the future, we'll want to determine whether we only want `getBindings` to return a PartiQLValue, but this unblocks us today.

RCHowell:
Replaced by https://github.com/partiql/partiql-lang-kotlin/pull/1317

PR #1305, opened 2023-12-13
Use Iterable for PartiQLValue collections rather than Sequence

RCHowell:
## Relevant Issues
N/A

## Description
A sequence gets exhausted after consumed once which does not make it a good option for the **default** backing field. This PR replaces the collection data structure with the Iterable interface. This allows requesting a new iterator which gives freedom in the implementation of the backing structure. You can use sequences or lazily-generated iterators, but importantly this is not the default.

## Other Information
Yes

No. PartiQLValue is considered experimental.

No

Yes

## License Information

### DIFF
--- a/partiql-types/src/main/kotlin/org/partiql/value/PartiQLValue.kt
+++ b/partiql-types/src/main/kotlin/org/partiql/value/PartiQLValue.kt
@@ -67,10 +67,7 @@ public sealed interface ScalarValue<T> : PartiQLValue {
 @PartiQLValueExperimental
 public sealed interface CollectionValue<T : PartiQLValue> : PartiQLValue, Iterable<T> {
 
-    public val elements: Collection<T>?
-
     override val isNull: Boolean
-        get() = elements == null
 
     override fun iterator(): Iterator<T>
 
@@ -388,8 +385,8 @@ public abstract class BagValue<T : PartiQLValue> : CollectionValue<T> {
         if (this.isNull || other.isNull) return this.isNull == other.isNull
 
         // both not null, compare values
-        val lhs = this.elements!!
-        val rhs = other.elements!!
+        val lhs = this.toList()
+        val rhs = other.toList()
         // this is incorrect as it assumes ordered-ness, but we don't have a sort or hash yet
         return lhs == rhs
     }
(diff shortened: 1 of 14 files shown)


partiql-types/src/main/kotlin/org/partiql/value/PartiQLValue.kt
```diff
@@ -65,14 +65,14 @@ public sealed interface ScalarValue<T> : PartiQLValue {
 }
 
 @PartiQLValueExperimental
-public sealed interface CollectionValue<T : PartiQLValue> : PartiQLValue, Sequence<T> {
+public sealed interface CollectionValue<T : PartiQLValue> : PartiQLValue, Iterable<T> {
 
-    public val elements: Sequence<T>?
+    public val elements: Collection<T>?
```
johnedquinn, line 70: I feel as if this should only be `Iterable` as we may not know what is backing the CollectionValue and StructValue. AKA, with the implementations in this PR, we will be materializing all of this data using `toList`.

In any case, it seems like we should have:
```
override fun iterator(): Iterator<T>?
```

Is there a reason we need both `elements` and `iterator`?

RCHowell: Iterator gives consumers control over consumption/materialization, but elements is just a simpler interface which does the "toList" for you.

There are many places in the `org.partiql.value.io` which uses `elements` for writing data. Yes, these could all be replace with `.iterator().toList()`, but its less ergonomic imho.

Additionally, the `elements` backing field is nullable whereas `iterator()` will throw an NPE. I don't know if we need both, but elements is like a default way to consume the iterator

johnedquinn: True, but consumers of PartiQL should be using our APIs for checking nullability. AKA:
```kotlin
val bag = bagValue(someElements)
if (bag.isNull.not()) {
    // Materialize all underlying PartiQL Values
    bag.toList()

    // Loop through elements
    bag.forEach { element -> TODO() }
}
```

Also, notice above that you don't need to use: `.iterator().toList()`.

RCHowell: ack, updated

johnedquinn (approved):

PR #1306, opened 2023-12-14
Simplifies logical model of the path expression

johnedquinn:
## Relevant Issues
- In #1271, I mentioned the desire to move the typing of paths to the AST transformation. This PR does not do that exactly, however, it resolves the underlying issue.

## Description
- As part of the effort to model our logical plan as closely to the underlying data flow as possible, this PR models path expressions in a more representative manner. Specifically, a `Path`, in this PR, is a single step on some root. Therefore, we can directly compile logical plan nodes into physical plan nodes that represent the underlying operation (i.e. index lookup, sensitive lookup, insensitive lookup).
- To do this, upon transforming the AST to a Logical Plan (Unresolved), we take the `ExprPath` and attempt to grab the first steps to make a qualified identifier (Var Unresolved). This subtle change gives us the ability to attempt to resolve the qualified identifier in the planner without needing to maintain the syntactic metadata in the logical unresolved plan. See #1271 
- While resolving a qualified identifier, it may actually be that a subset of the identifier is a Global/Local. In that case, we use the remaining steps to construct a path. And, once we have a path, it's quite easy to type, since we model each path step as its own expression operator.
- Tests are unchanged (except when necessary for API changes).

## Other Information
- Any backward-incompatible changes? **NO** -- 0.14.0 hasn't been released yet

## License Information


RCHowell (changes requested):

partiql-planner/src/main/kotlin/org/partiql/planner/internal/ir/Nodes.kt
```diff
@@ -393,142 +393,117 @@ internal data class Rex(
...
-            internal val root: Rex,
-            @JvmField
-            internal val steps: List<Step>,
-        ) : Op() {
-            internal override val children: List<PlanNode> by lazy {
-                val kids = mutableListOf<PlanNode?>()
-                kids.add(root)
-                kids.addAll(steps)
-                kids.filterNotNull()
+        public sealed class Path : Op() {
```
alancai98, line 396: nit: for the internal plan code (under `planner/internal/ir/*`), `public` is used. It shouldn't matter since the top-level classes are `internal`, but it can be confusing. Would suggest removing the `public` or adding back the explicit `internal` visibility modifiers.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -1349,24 +1285,24 @@ internal class PlanTyper(
...
+//            when (it.bindingCase) {
+//                BindingCase.SENSITIVE -> rexOpPathKey(root, rexString(it.name))
+//                BindingCase.INSENSITIVE -> rexOpPathSymbol(root, it.name)
+//            }
+//        }
+//        return when (steps.isEmpty()) {
+//            true -> root.op
+//            false -> rexOpPath(root, steps)
+//        }
+//    }
```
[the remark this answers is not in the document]
johnedquinn: Oh wow, thanks for that.

RCHowell (approved):

partiql-planner/src/main/kotlin/org/partiql/planner/internal/ir/visitor/PlanBaseVisitor.kt
```diff
@@ -93,30 +93,23 @@ internal abstract class PlanBaseVisitor<R, C> : PlanVisitor<R, C> {
...
 
-    override fun visitRexOpPath(node: Rex.Op.Path, ctx: C): R = defaultVisit(node, ctx)
-
-    override fun visitRexOpPathStep(node: Rex.Op.Path.Step, ctx: C): R = when (node) {
-        is Rex.Op.Path.Step.Index -> visitRexOpPathStepIndex(node, ctx)
-        is Rex.Op.Path.Step.Key -> visitRexOpPathStepKey(node, ctx)
-        is Rex.Op.Path.Step.Symbol -> visitRexOpPathStepSymbol(node, ctx)
-        is Rex.Op.Path.Step.Wildcard -> visitRexOpPathStepWildcard(node, ctx)
-        is Rex.Op.Path.Step.Unpivot -> visitRexOpPathStepUnpivot(node, ctx)
+    public override fun visitRexOpPath(node: Rex.Op.Path, ctx: C): R = when (node) {
```
alancai98, line 96: still a few uses of `public` in this file

partiql-planner/src/main/kotlin/org/partiql/planner/internal/ir/visitor/PlanVisitor.kt
```diff
@@ -56,19 +56,13 @@ internal interface PlanVisitor<R, C> {
 
     fun visitRexOpGlobal(node: Rex.Op.Global, ctx: C): R
 
-    fun visitRexOpPath(node: Rex.Op.Path, ctx: C): R
+    public fun visitRexOpPath(node: Rex.Op.Path, ctx: C): R
```
alancai98, line 59: also in `PlanVisitor`

partiql-planner/src/main/kotlin/org/partiql/planner/internal/ir/Nodes.kt
```diff
@@ -393,142 +391,77 @@ internal data class Rex(
...
+                internal override fun <R, C> accept(visitor: PlanVisitor<R, C>, ctx: C): R =
+                    visitor.visitRexOpPathKey(this, ctx)
 
-                    internal companion object {
-                        @JvmStatic
-                        internal fun builder(): RexOpPathStepSymbolBuilder = RexOpPathStepSymbolBuilder()
-                    }
+                internal companion object {
+                    @JvmStatic
+                    public fun builder(): RexOpPathKeyBuilder = RexOpPathKeyBuilder()
```
alancai98, line 441: another `public` in `Nodes.kt`

alancai98 (approved):

PR #1311, opened 2023-12-15
Refactor modeling of `EXCLUDE` in `partiql-ast` and `partiql-plan`

alancai98:
## Description
Refactors the modeling of `EXCLUDE` in the `partiql-ast` and `partiql-plan`. Changes include
- `EXCLUDE` root modeled as a variable reference rather than an identifier symbol
- `EXCLUDE` wildcard steps modeled using an enum to differentiate between the two types (tuple vs collection)
- Change `EXCLUDE` class names in `partiql-ast` to align w/ the `partiql-plan` modeling

## Other Information

## License Information

### DIFF
--- a/partiql-ast/src/main/resources/partiql_ast.ion
+++ b/partiql-ast/src/main/resources/partiql_ast.ion
@@ -563,17 +563,17 @@ select::[
 ]
 
 exclude::{
-  exprs: list::[exclude_expr],
+  items: list::[item],
   _: [
-    exclude_expr::{
-      root: '.identifier.symbol',
+    item::{
+      root: '.expr.var',
       steps: list::[step],
     },
     step::[
-      exclude_tuple_attr::{ symbol: '.identifier.symbol' },
-      exclude_collection_index::{ index: int },
-      exclude_tuple_wildcard::{},
-      exclude_collection_wildcard::{},
+      struct_field::{ symbol: '.identifier.symbol' },
+      coll_index::{ index: int },
+      struct_wildcard::{},
+      coll_wildcard::{},
     ]
   ]
 }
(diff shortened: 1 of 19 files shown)


partiql-ast/src/main/resources/partiql_ast.ion
```diff
@@ -563,17 +563,22 @@ select::[
 ]
 
 exclude::{
-  exprs: list::[exclude_expr],
+  exprs: list::[item],
```
RCHowell, line 566:
```suggestion
  items: list::[item],
```

partiql-ast/src/main/resources/partiql_ast.ion
```diff
@@ -563,17 +563,22 @@ select::[
...
-      exclude_tuple_wildcard::{},
-      exclude_collection_wildcard::{},
+      attr::{
+        symbol: '.identifier.symbol'
+      },
+      pos::{
+        index: int
+      },
+      wildcard::{
+        type: [ STRUCT, COLLECTION ]
```
RCHowell, line 580: Seeing the bigger picture now I'm curious if we want to be consistent on struct verbiage or tuple verbiage. I think you had it great, just drop the extra `exclude_`

```suggestion
      tuple_attr:{ symbol: '.identifier.symbol' },
      tuple_wildcard::{},
      coll_index::{ index: int },
      coll_wildcard::{},
```

But from the customer's perspective a tuple is a struct and this is the public api so "binding tuple" is more a specification term whereas customers interact with "structs"

```
      struct_field:{ symbol: '.identifier.symbol' },
      struct_wildcard::{},
      coll_index::{ index: int },
      coll_wildcard::{},
```

Up to you!

RCHowell (changes requested):

partiql-ast/src/main/resources/partiql_ast.ion
alancai98: thanks for the catch. changed

partiql-ast/src/main/resources/partiql_ast.ion
alancai98: discussed offline. will change in `partiql-ast` and `partiql-plan` to use "struct" rather than "tuple" since customers may be more familiar with using "struct" in the public api.

Also discussed keeping thins consistent and using the following for the step variants:
```
struct_field:{ symbol: '.identifier.symbol' },
struct_wildcard::{},
coll_index::{ index: int },
coll_wildcard::{},
```

RCHowell (approved):

PR #1317, opened 2023-12-19
Initializes bindings in connectors and globals

RCHowell:
## Description

This PR introduces a getBindings() method to PartiQL Plugins. A plugin will return a PartiQLBindings object which returns a PartiQLValue given a ConnectorObjectPath.

I have implemented this for the MemoryPlugin which is now used to load globals from the conformance test suite. This PR also includes an implementation of a TestExecutor for the ongoing evaluator work.

Yes

## License Information

### DIFF
--- a/partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
+++ b/partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
@@ -13,10 +13,10 @@ import org.partiql.eval.internal.operator.rex.ExprCase
 import org.partiql.eval.internal.operator.rex.ExprCollection
 import org.partiql.eval.internal.operator.rex.ExprGlobal
 import org.partiql.eval.internal.operator.rex.ExprLiteral
-import org.partiql.eval.internal.operator.rex.ExprPivot
 import org.partiql.eval.internal.operator.rex.ExprPathIndex
 import org.partiql.eval.internal.operator.rex.ExprPathKey
 import org.partiql.eval.internal.operator.rex.ExprPathSymbol
+import org.partiql.eval.internal.operator.rex.ExprPivot
 import org.partiql.eval.internal.operator.rex.ExprSelect
 import org.partiql.eval.internal.operator.rex.ExprStruct
 import org.partiql.eval.internal.operator.rex.ExprTupleUnion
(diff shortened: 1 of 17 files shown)


partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
```diff
@@ -19,91 +23,129 @@ import org.partiql.plan.Rel
...
-            val constructor = visitRex(node.constructor, ctx)
-            return ExprSelect(rel, constructor)
-        }
+    override fun visitRexOpGlobal(node: Rex.Op.Global, ctx: Unit): Operator {
+        val catalog = plan.catalogs[node.ref.catalog]
+        val symbol = catalog.symbols[node.ref.symbol]
+        val path = ConnectorObjectPath(symbol.path)
+        val bindings = catalogs[catalog.name]!!
+        return ExprGlobal(path, bindings)
+    }
```
johnedquinn, line 69: Glad to see this!

plugins/partiql-memory/src/main/kotlin/org/partiql/plugins/memory/MemoryBindings.kt
```diff
@@ -0,0 +1,52 @@
...
+    private val bindings: Map<String, PartiQLValue>,
+) : ConnectorBindings {
+
+    @OptIn(PartiQLValueExperimental::class)
+    override fun getValue(path: ConnectorObjectPath): PartiQLValue {
+        val key = path.steps.joinToString(".")
+        return bindings[key] ?: missingValue()
+    }
+
+    internal fun iterator(): Iterator<PartiQLValue> = bindings.values.iterator()
```
johnedquinn, line 23: Is this used?

plugins/partiql-memory/src/main/kotlin/org/partiql/plugins/memory/MemoryBindings.kt
```diff
@@ -0,0 +1,52 @@
...
+
+        /**
+         * No bindings.
+         */
+        val empty = MemoryBindings(emptyMap())
+
+        /**
+         * Loads each declared global of the catalog from the data element.
+         */
+        fun load(metadata: MemoryConnector.Metadata, data: StructElement): MemoryBindings {
```
johnedquinn, line 35:
```suggestion
        fun load(metadata: MemoryConnector.Metadata, data: Map<String, PartiQLValue>): MemoryBindings {
```
While this is used for the conformance tests, I think this logic would more adequately fit there. The current implementation is specific to the conformance tests.

johnedquinn (changes requested):

plugins/partiql-memory/src/main/kotlin/org/partiql/plugins/memory/MemoryBindings.kt
RCHowell: Removed

plugins/partiql-memory/src/main/kotlin/org/partiql/plugins/memory/MemoryBindings.kt
RCHowell: Makes sense. I moved `load` to be a private function within the conformance test runner.

johnedquinn (approved):

PR #1319, opened 2023-12-20
Defines a PartiQL system plugin containing scalar and aggregation builtins.

RCHowell:
## Description
This leverages a plugin and extends connector metadata to "install" the partiql system builtins and aggregations to both the planner and an evaluation engine. The signatures and stubs were generated by PartiQLHeader used by the Planner. I've removed use of Headers in the PartiQLPlanner so now it only queries signatures via plugins. This means customers will be able to included UDFs into the planner via their plugins.

This commit includes generated stubs for currently defined partiql scalar and aggregation builtins.

## Other Information
No

  - < If YES, why? >
Yes, but this is unreleased.

No

Yes

## License Information


partiql-spi/src/main/kotlin/org/partiql/spi/connector/ConnectorMetadata.kt
```diff
@@ -16,12 +16,28 @@ package org.partiql.spi.connector
...
+    /**
+     * Scalar function signatures available via operator or special form syntax.
+     */
+    public val operators: List<FunctionSignature.Scalar>
+
+    /**
+     * Aggregation function signatures.
+     */
+    public val aggregations: List<FunctionSignature.Aggregation>
+
```
johnedquinn, line 40: This was already added in Plugin. What's the diff?

I've envisioned that the gathering of functions via the SQL Path would make use of the Connector Metadata. However, it would be more like `getFunction(path)`. Though, since we don't have the concept of the SQL Path yet, the functions exposed on Plugin would suffice.

RCHowell: The plugin returns signatures and implementations. This is signatures only and allows us to give the planner access to function signatures from connector metadata without having to implement SQL Path

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/FnRegistry.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Calculate a queryable map of scalar function signatures from special forms.
+     */
+    private val uOps: FnMap<FunctionSignature.Scalar> = metadata.flatMap { it.operators }.toFnMap()
+
+    /**
+     * Calculate a queryable map of aggregation function signatures
+     */
+    private val uAggs: FnMap<FunctionSignature.Aggregation> = metadata.flatMap { it.aggregations }.toFnMap()
```
yliuuuu, line 42: I feel like we can just load all functions and operators here instead of having a separate FnBuiltins. 

Also, not in the scope of this PR, but we might need a better model around user defined operators. Currently this is no way for user to load in customer operator implementation.

plugins/partiql-memory/src/main/kotlin/org/partiql/plugins/memory/MemoryConnector.kt
```diff
@@ -43,6 +44,8 @@ public class MemoryConnector(
      */
     class Metadata(private val map: Map<String, StaticType>) : ConnectorMetadata {
 
+        override val functions: ConnectorFunctions? = null
```
yliuuuu, line 47: May want to move this to the constructor.

plugins/partiql-plugin/src/main/kotlin/org/partiql/plugin/PartiQLPlugin.kt
```diff
@@ -0,0 +1,493 @@
...
+        Agg_AVG__INT32__INT32,
+        Agg_AVG__INT64__INT64,
+        Agg_AVG__INT__INT,
+        Agg_AVG__DECIMAL_ARBITRARY__DECIMAL_ARBITRARY,
+        Agg_AVG__FLOAT32__FLOAT32,
+        Agg_AVG__FLOAT64__FLOAT64,
+    )
+
+    @PartiQLFunctionExperimental
+    override val functions: List<PartiQLFunction> = scalars + operators + aggregations
```
yliuuuu, line 492: Had a offline discussion with this. Maybe consider move functions outside of plugin interface when we merge eval to main.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/FnBuiltins.kt
```diff
@@ -0,0 +1,30 @@
+package org.partiql.planner.internal.typer
+
+import org.partiql.plugin.PartiQLFunctions
+
+internal object FnBuiltins {
```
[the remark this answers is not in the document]
johnedquinn: It seems like this is a hard-coding of the SQL Path where PartiQL built-ins are always found first.

plugins/partiql-plugin/src/main/kotlin/org/partiql/plugin/PartiQLPlugin.kt
johnedquinn: IMO, it seems to me that the most logical place to put it would be in `ConnectorBindings` as the `bindings environment` really is just representative of the SQL environment (catalog and schema). It holds unique qualified names that represent some aspect of the environment (values/tables, views, functions, etc).

johnedquinn: Actually, I stand corrected on the uniqueness of names with regards to functions. Realistically, according to SQL:1999, all routines are placed in the `INFOMRATION_SCHEMA`'s `ROUTINES` table. Therefore, it really is separate from the bindings environment. The only relevant binding would be `<catalog>."INFORMATION_SCHEMA"."ROUTINES"` -- which we aren't getting into. Anyways, I've convinced myself that something separate from the `ConnectorBindings` would be adequate.

RCHowell: Feedback reflected in interfaces draft here https://github.com/partiql/partiql-lang-kotlin/pull/1332

RCHowell:
The major interface feedback for this PR is addressed in this draft https://github.com/partiql/partiql-lang-kotlin/pull/1332. Upon that PR review, I will rebase this.

PR #1280, opened 2023-12-02
Add exclude physical plan compiler impl

alancai98:
## Relevant Issues
- Spec issue for `EXCLUDE` https://github.com/partiql/partiql-lang/issues/27

## Description
Adds implementation of `EXCLUDE` to the plan compiler (different from #1249, which adds `EXCLUDE` to the `EvaluatingCompiler`). This PR also migrates more of the `partiql-lang` APIs to `partiql-eval`'s `internal` package.

## Other Information
- Updated Unreleased Section in CHANGELOG: **[YES]**

- Any backward-incompatible changes? **[NO]**

- Any new external dependencies? **[NO]**

- Do your changes comply with the [Contributing Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CONTRIBUTING.md)
  and [Code Style Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CODE_STYLE.md)? **[YES]**

## License Information

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1249, opened 2023-10-18
Implement EXCLUDE in `EvaluatingCompiler`

alancai98:
## Relevant Issues
https://github.com/partiql/partiql-lang/issues/27 (still experimental until the RFC is merged).

## Description
Implementation of `EXCLUDE` operator (https://github.com/partiql/partiql-lang/issues/27) in the `EvaluatingCompiler`. A future PR will implement `EXCLUDE` in the `PhysicalPlanCompiler`, targeting the [partiql-eval branch](https://github.com/partiql/partiql-lang-kotlin/tree/partiql-eval).

## Other Information
- Updated Unreleased Section in CHANGELOG: **[Yes]**

- Any backward-incompatible changes? **[NO]**

- Any new external dependencies? **[NO]**

- Do your changes comply with the [Contributing Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CONTRIBUTING.md)
  and [Code Style Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CODE_STYLE.md)? **[YES]**

## License Information

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1320, opened 2023-12-20
Add `EXCLUDE` to `partiql-eval`

alancai98:
## Relevant Issues
- `EXCLUDE` RFC -- https://github.com/partiql/partiql-docs/pull/51
- `EXCLUDE` implementation in `EvaluatingCompiler` -- https://github.com/partiql/partiql-lang-kotlin/pull/1249
- Closed PR to add `EXCLUDE` to `PhysicalPlanCompiler` -- https://github.com/partiql/partiql-lang-kotlin/pull/1280
  - Much of this PR's additions are based off that PR

## Description
- Adds `EXCLUDE` to the `partiql-eval` evaluator
- Different modeling of `EXCLUDE` in the plan representation. This change in modeling was done to better align with the redundant path elimination and evaluation approaches
- Adds some path subsumption tests that were previously commented out in 4ac1b3d1e49466e59f1f7e6573981dd2460150c2
  - More tests were added beyond the previous set of tests

## Other Information

## License Information


partiql-eval/src/test/kotlin/org/partiql/eval/internal/exclude/CompiledExcludeItemTest.kt
```diff
@@ -0,0 +1,251 @@
+package org.partiql.eval.internal.exclude
```
alancai98, line 1: These tests are the same as the commented out tests from `partiql-eval/src/test/kotlin/org/partiql/lang/eval/EvaluatingCompilerExcludeTests.kt`

partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEngineDefaultTest.kt
```diff
@@ -264,6 +264,35 @@ class PartiQLEngineDefaultTest {
         assertEquals(expected, output)
     }
 
+    @OptIn(PartiQLValueExperimental::class)
+    @Test
+    fun testExclude() {
```
alancai98, line 269: Test succeeds when `@Disabled` is removed from L29.

https://github.com/partiql/partiql-lang-kotlin/blob/672a09d354f1cd0a5615d72b4192e2b07d7ad2f2/partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEngineDefaultTest.kt#L29

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelExclude.kt
```diff
@@ -0,0 +1,212 @@
...
+            return newRecord
+        }
+    }
+
+    override fun close() {
+        input.close()
+    }
+}
+
+@OptIn(PartiQLValueExperimental::class)
```
RCHowell, line 48: Rather than having a bunch of top-level functions, could you please place these in the appropriate spots? Compile related functions in the Compiler, and evaluation related items wihtin the RelExclude operator?

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelExclude.kt
```diff
@@ -0,0 +1,212 @@
...
+import org.partiql.value.SexpValue
+import org.partiql.value.StructValue
+import org.partiql.value.bagValue
+import org.partiql.value.listValue
+import org.partiql.value.sexpValue
+import org.partiql.value.structValue
+
+internal class RelExclude(
+    val input: Operator.Relation,
+    private val compiledExcludeItems: List<CompiledExcludeItem>
```
RCHowell, line 26: both can be private
```suggestion
    private val input: Operator.Relation,
    private val exclusions: List<CompiledExcludeItem>
```

RCHowell (changes requested):

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelExclude.kt
alancai98: Changed locations of the top-level functions in https://github.com/partiql/partiql-lang-kotlin/pull/1320/commits/9e9a04032dc8c57a17bbbcb624bba166fd0ecdd0

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelExclude.kt
```diff
@@ -0,0 +1,178 @@
...
+    private fun excludeOnCollValue(
+        coll: CollectionValue<*>,
+        type: PartiQLValueType,
+        exclusions: ExcludeNode
+    ): PartiQLValue {
+        val leavesSteps = exclusions.leaves.map { leaf -> leaf.step }
+        val branches = exclusions.branches
+        if (leavesSteps.any { it is ExcludeStep.CollWildcard }) {
+            // collection wildcard at current level. return empty collection
+            return newCollValue(type, emptyList())
```
RCHowell, line 131: You could check the class on line 127 and short-circuit on the collection wildcard. This way you do not have to traverse the list twice.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelExclude.kt
```diff
@@ -0,0 +1,178 @@
...
+        } else {
+            val indexesToRemove = leavesSteps.filterIsInstance<ExcludeStep.CollIndex>()
+                .map { it.index }
+                .toSet()
```
RCHowell, line 135: Then this else branch would move to the top level and you wouldn't have to traverse yet again

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelExclude.kt
```diff
@@ -0,0 +1,178 @@
...
+            val collWithRemoved = when (coll) {
+                is BagValue -> coll
+                is ListValue, is SexpValue -> coll.filterIndexed { index, _ ->
+                    !indexesToRemove.contains(index)
+                }
+            }
+            val finalColl = collWithRemoved.mapIndexed { index, element ->
```
RCHowell, line 142: what's the difference between collWithRemoved and finalColl? I think you should be applying the exclusions with a mapIndexedNotNull and return null if the index is skipped otherwise apply the exlcusion. This way you don't traverse the collection more than once.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelExclude.kt
```diff
@@ -0,0 +1,178 @@
...
+                        expr = excludeOnPartiQLValue(element, it)
+                    }
+                }
+                // apply collection wildcard exclusions for lists, bags, and sexps
+                val collectionWildcardKey = ExcludeStep.CollWildcard
+                branches.find {
+                    it.step == collectionWildcardKey
+                }?.let {
+                    expr = excludeOnPartiQLValue(expr, it)
+                }
```
RCHowell, line 159: Do you have an example of what this is for?

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelExclude.kt
RCHowell: Similar comments can be applied to the struct exclusions.

RCHowell (changes requested):

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelExclude.kt
alancai98: True, changed to perform check on the exclusion leafs in one iteration.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelExclude.kt
alancai98: `collWithRemoved` would remove the indexes at that level (i.e. the leaf nodes with `ExcludeStep.CollIndex`). `finalColl` would remove the values at deeper nested levels (i.e. the branch nodes).

> I think you should be applying the exclusions with a mapIndexedNotNull and return null if the index is skipped otherwise apply the exlcusion. This way you don't traverse the collection more than once.

Good point. I will change to use `mapIndexedNotNull` so only one iteration of `coll` is necessary.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelExclude.kt
alancai98: I'll edit the comment to clarify. This basically applies all the collection wildcard exclude steps that have deeper exclude steps (e.g. `t.a.b.c[*].field_x` in
```
SELECT *
EXCLUDE
    t.a.b.c[*].field_x
FROM [{
    'a': {
        'b': {
            'c': [
                {                    -- c[0]; field_x to be removed
                    'field_x': 0, 
                    'field_y': 0
                },
                {                    -- c[1]; field_x to be removed
                    'field_x': 1,
                    'field_y': 1
                },
                {                    -- c[2]; field_x to be removed
                    'field_x': 2,
                    'field_y': 2
                }
            ]
        }
    }
}] AS t
```

When we get to the `c` list, the collection wildcard step will apply the remainder of the step (i.e. `.field_x`) on each of the elements of `c`.

alancai98: Ported above example to another test:

https://github.com/partiql/partiql-lang-kotlin/blob/50de0322624a06ecda07ea3c6bdfc1f3cefc5252/partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEngineDefaultTest.kt#L249-L296

partiql-eval/src/main/kotlin/org/partiql/eval/internal/exclude/CompiledExcludeItem.kt
```diff
@@ -0,0 +1,229 @@
...
+ */
+internal enum class ExcludeFieldCase {
+    INSENSITIVE, SENSITIVE
+}
+
+/**
+ * Represents all the compiled `EXCLUDE` paths that start with the same [CompiledExcludeItem.root]. This variant of
+ * [ExcludeNode] represents the top-level root node of the exclude tree.
+ *
+ * Notably, redundant paths (i.e. exclude paths that exclude values already excluded by other paths) will be removed.
```
RCHowell, line 24: I believe the planner should be responsible for removing redundant paths.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelExclude.kt
```diff
@@ -0,0 +1,184 @@
...
+        }.toSet()
+        val branches = exclusions.branches
+        val finalStruct = structValue.entries.mapNotNull { structField ->
+            if (attrsToRemove.contains(structField.first)) {
+                // struct attr is to be removed at current level
+                null
+            } else {
+                // deeper level exclusions
+                val name = structField.first
+                var expr = structField.second
```
RCHowell, line 84: The second is not an expression, it's a value. 


```suggestion
                val (key, value) = structField
```

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelExclude.kt
```diff
@@ -0,0 +1,184 @@
...
+                    expr = excludeOnPartiQLValue(expr, it)
+                }
+                // apply tuple wildcard exclusions at deeper levels
+                val tupleWildcardKey = ExcludeStep.StructWildcard
+                branches.find {
+                    it.step == tupleWildcardKey
+                }?.let {
+                    expr = excludeOnPartiQLValue(expr, it)
+                }
+                Pair(name, expr)
```
RCHowell, line 106: You're looping over the branches multiple times by using find. A set has O(1) lookup when you are using the hash function for the set, but in this case you are using `find` which is a loop over the set elements.

You should only have one loop of the branches. Does the order in which exclusions apply matter? If so, consider "pruning" early such applying wildcards before single fields as to eliminate unnecessary work. That being said, the planner should be responsible for reducing the exclusions.

RCHowell (changes requested):

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -377,11 +378,15 @@ internal class PlanTyper(
...
             }
-
-            val op = relOpExclude(input, newItems)
+            val subsumedPaths = newPaths
+                .groupBy(keySelector = { it.root }, valueTransform = { it.steps }) // combine exclude paths with the same resolved root before subsumption
+                .map { (root, allSteps) ->
+                    val nonRedundant = ExcludeRepr.toExcludeRepr(allSteps.flatten()).removeRedundantSteps()
+                    relOpExcludePath(root, nonRedundant.toPlanRepr())
+                }
+            val op = relOpExclude(input, subsumedPaths)
```
RCHowell, line 389: Is this in case the AST representations aren't equivalent, but perhaps they resolve to the same?

```
SELECT * EXCLUDE t.a.*, "t".a FROM { 'a': { 'x': 1 } ... } AS t
```

I _think_ the only time the root (which is a var) would differ is based on the case-sensitivity of the identifier.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeUtils.kt
```diff
@@ -146,70 +145,84 @@ private fun StaticType.asRuntimeType(): PartiQLValueType = when (this) {
...
- * @param steps
+ * @param step
  * @param lastStepOptional
  * @return
  */
-internal fun StructType.exclude(steps: List<Rel.Op.Exclude.Step>, lastStepOptional: Boolean = true): StaticType {
-    val step = steps.first()
-    val output = fields.map { field ->
-        val newField = if (steps.size == 1) {
+internal fun StructType.exclude(step: Rel.Op.Exclude.Step, lastStepOptional: Boolean = true): StaticType {
```
RCHowell, line 169: Could you please clarify in a comment why last step can be optional?

RCHowell (changes requested):

partiql-plan/src/main/resources/partiql_plan.ion
```diff
@@ -288,18 +288,23 @@ rel::{
...
+        path::{
           root: '.rex.op.var',
           steps: list::[step],
         },
-        step::[
-          struct_field::{ symbol: '.identifier.symbol' },
+        step::{
+          type: type,
+          substeps: list::[step],
+        },
```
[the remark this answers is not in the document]
alancai98: Yeah, definitely made the redundant path exclusion and evaluation more straightforward.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/exclude/CompiledExcludeItem.kt
alancai98: From offline discussion before the holiday break, I moved the redundant path elimination to the `PlanTyper` so the planner and not evaluator is responsible. The plan representation for `EXCLUDE` was also changed to better accommodate for the tree-like representation used by this elimination step and evaluation.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelExclude.kt
alancai98: Changed to store the lookups in a map. The order of exclusions should not matter. In latest revision, planner step now eliminates the redundant exclude paths.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelExclude.kt
```diff
@@ -0,0 +1,199 @@
...
+    ): Record {
+        val values = record.values
+        val value = values.getOrNull(path.root.ref)
+        val newValues = if (value != null) {
+            values[path.root.ref] = exclude(value, path.steps)
+            values
+        } else {
+            values
+        }
+        return Record(newValues)
```
RCHowell, line 55: We shouldn't have missing roots and the record values array is mutable so we don't need to fold / combine previous states. The fold is useful when we need to produce a new immutable state, but we already have a mutable state object (record values) so we just need to apply each

```
    override fun next(): Record? {
        val record = input.next() ?: return null
        for (path in exclusions) {
             val i = path.root.ref
             val value = record.values[i]
             record.values[i] = exclude(value, path)
        }
        return record
    }
```

RCHowell (changes requested):

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelExclude.kt
alancai98: Good catch. Removed the unnecessary fold

RCHowell (approved):
Nice work!

PR #1321, opened 2023-12-20
Remove redundancy by using ANY parameter for IS TYPE operators

RCHowell:
## Relevant Issues
N/A

## Description
Reduces redundancy of the IS TYPE operators as they do not need to be defined for each individual type.

## Other Information
No

  - < If YES, why? >
No, internal typer concern for 0.14

No

Yes

## License Information

### DIFF
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/PartiQLHeader.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/PartiQLHeader.kt
@@ -403,68 +403,61 @@ internal object PartiQLHeader : Header() {
     // To model type assertion, generating a list of assertion function based on the type,
     // and the parameter will be the value entered.
     //  i.e., 1 is INT2  => is_int16(1)
-    private fun isType(): List<FunctionSignature.Scalar> = types.all.filterNot { it == NULL || it == MISSING }.flatMap { element ->
-        types.all.filterNot { it == MISSING || it == ANY }.map { operand ->
+    private fun isType(): List<FunctionSignature.Scalar> =
+        types.all.filterNot { it == NULL || it == MISSING }.map { element ->
             FunctionSignature.Scalar(
                 name = "is_${element.name.lowercase()}",
                 returns = BOOL,
                 parameters = listOf(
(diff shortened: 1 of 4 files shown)


yliuuuu:
One thing that may cause us trouble is `missing` handling. 

In other functions, for example `1+missing`, will be rewrite to `plus(1, missing)` and the function would not be resolved. That is, in most of the functions where `missable` is true, the handling strategy is to not resolve the function at the first place. 

With this change, `MISSING IS INT2` will be resolved to `is_int2(missing)`, such function can be resolved due to the value argument being `ANY` in the header. In the typer we then raise a `expression always returns missing` error. 

This is more of a inconsistence than a problem, but it is worth to consider as change of errors may affect our customers.

RCHowell:
I wonder if this handles it though? https://github.com/partiql/partiql-lang-kotlin/blob/c73efbf3f1e96856c718a359771b2ffff91fb61c/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt#L654-L660

I feel the only issue here will be `is_missing` which I need to add a case for.

yliuuuu:
> I wonder if this handles it though?
> 
> https://github.com/partiql/partiql-lang-kotlin/blob/c73efbf3f1e96856c718a359771b2ffff91fb61c/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt#L654-L660
> 
> I feel the only issue here will be `is_missing` which I need to add a case for (spec page 30).

The code snippet you show raised a `alwaysMissing` problem and return a resolved function whose return type is `missing`. 

I tried to point out that for other functions, the issue raised is going to be a `unresolved function`. See 
https://github.com/partiql/partiql-lang-kotlin/blob/c73efbf3f1e96856c718a359771b2ffff91fb61c/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt#L619-L640

yliuuuu (approved):
Approving this PR based on offline discussion. We are going to further investigate on MISSING.

PR #1309, opened 2023-12-14
Replaces plugins in PartiQLPlanner with connector metadata

RCHowell:
## Relevant Issues
N/A

## Description
This simplifies the planner by using just ConnectorMetadata. It is not necessary to pass these plugins in and complicates the logic. The planner only needs catalogs with their respective type mappings. Additionally I cleaned up the parser builder to match the IR nodes builders and planner builder — ie a static .builder() method on the interface.

## Other Information
- Updated Unreleased Section in CHANGELOG: **[YES/NO]**
Yes

- Any backward-incompatible changes? **[YES/NO]**
Yes, just the builder method name.

- Any new external dependencies? **[YES/NO]**
No

- Do your changes comply with the [Contributing Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CONTRIBUTING.md)
  and [Code Style Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CODE_STYLE.md)? **[YES/NO]**
Yes

## License Information

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1323, opened 2023-12-20
Moves ConnectorMetadata to planner session for PartiQLPlanner re-use

RCHowell:
## Relevant Issues
N/A

## Description
This PR replaces https://github.com/partiql/partiql-lang-kotlin/pull/1309 as we need function information to come from the catalog (ConnectorMetadata).

## Other Information
Yes

  - < If YES, why? >
Technically yes, but these are unreleased. This will be in a 0.14.1 patch to replace lost functionality of 0.14.0-alpha used by the transpiler.

 No

Yes

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,8 @@ Thank you to all who have contributed!
 ## [Unreleased]
 
 ### Added
-- Adds the ability to pass a user-defined-function signature to the planner.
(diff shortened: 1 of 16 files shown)


johnedquinn (approved):
This looks good to me. While we know these new functions will eventually be replaced via some mechanism of the SQL Path, do you want to add an experimental tag? Non-blocking, but something to consider before release.

PR #1325, opened 2023-12-22
Adds unary, logical, binary arithmetic, and concat operators

johnedquinn:
## Relevant Issues
- None

## Description
- Adds:
  - Logical Operators (NOT, AND, OR)
  - Unary Operators (POS, NEG)
  - Binary Arithmetic Operators (PLUS, MINUS, TIMES, DIV, MOD, BITWISE-AND)
  - Concat Operator

## Other Information

## License Information


RCHowell (changes requested):

plugins/partiql-plugin/src/main/kotlin/org/partiql/plugin/internal/fn/scalar/BinaryOpHelpers.kt
```diff
@@ -0,0 +1,126 @@
...
+    return int8Value((op(lhsValue, rhsValue)).toByte())
+}
+
+@OptIn(PartiQLValueExperimental::class)
+@JvmName("binaryOpInt8ByteToByte")
+internal inline fun binaryOpInt8(lhs: PartiQLValue, rhs: PartiQLValue, op: (Byte, Byte) -> Byte): Int8Value {
+    val lhsValue = lhs.check<Int8Value>().value!!
+    val rhsValue = rhs.check<Int8Value>().value!!
+    return int8Value((op(lhsValue, rhsValue)))
+}
```
RCHowell, line 45: As discussed offline, we won't have these helpers. See 

```
        val l = args[0].check<Int16Value>()
        val r = args[1].check<Int16Value>()
        return if (l.isNull || r.isNull) {
            boolValue(null)
        } else {
            boolValue(l.value!! > r.value!!)
        }
```

Now the question is should the null check go before or after the type check?  Seems null checks should be first.

```
1 + stringValue(null)  <-> 1 + string.null <-> 1 + null  => null
```

RCHowell (changes requested):

RCHowell (approved):
This all looks good. Note that it does assume we are doing null call handling in the call operator within eval. This assumption is probably the right call so that every operator doesn't have to repeat logic

PR #1326, opened 2023-12-22
Adds IS TYPE and string builtin functions to PartiQL Plugin

RCHowell:
- is_type
- substring
- postition
- trim variants
- upper
- lower
- utcnow

## License Information


plugins/partiql-plugin/src/main/kotlin/org/partiql/plugin/internal/fn/scalar/FnIsDecimal.kt
```diff
@@ -33,15 +37,25 @@ internal object FnIsDecimal1 : PartiQLFunction.Scalar {
...
+        val s = args[1].check<Int32Value>().value
+        val v = args[2]
+        if (v !is DecimalValue) {
+            return boolValue(false)
+        }
+        val decimal = v.value
+        if (decimal == null) {
+            return boolValue(null)
+        }
+        return boolValue(decimal.precision() == p && decimal.scale() == s)
```
yliuuuu, line 59: `val decimal = v.value` would return a Java Big Decimal. 

The semantics of "precision" and "scale" are different in Big Decimal. 

From [Java doc](https://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html): 
> The value of the number represented by the BigDecimal is therefore (unscaledValue × 10^(-scale)).
> The precision is the number of digits in the unscaled value.
>  If zero or positive, the scale is the number of digits to the right of the decimal point. If negative, the unscaled value of the number is multiplied by ten to the power of the negation of the scale.

For example: 
SQL precision 3, scale 3 <- 0.011 === 11 x 10^-3 -> precision 2, scale 3

note that depending on how you construct the big decimal, there is no guarantee on the precision and scale of the object. 

For example: 0.011 = 11 x 10^-3 = 110 x 10^-4

The safe bet was to normalize the BigDecimal, then assert on the precision and scale. 

```
                // check scale
                val decimalPoint = if (d.scale() >= 0) d.scale() else 0
                if (decimalPoint > scale) {
                    return boolValue(false)
                }
                // check integer part
                val integerPart = d.setScale(0, RoundingMode.DOWN)
                val integerLength = if (integerPart.signum() != 0) integerPart.precision() - integerPart.scale() else 0
                // PartiQL precision semantics -> the maximum number of total digit (left of decimal place + right of decimal place)
                // PartiQL scale semantics -> the total number of digit after the decimal point.
                val expectedIntegerDigits = precision - scale

                return boolValue(expectedIntegerDigits >= integerLength)
```

plugins/partiql-plugin/src/main/kotlin/org/partiql/plugin/internal/fn/scalar/FnIsInt16.kt
```diff
@@ -21,6 +23,6 @@ internal object FnIsInt16 : PartiQLFunction.Scalar {
     )
 
     override fun invoke(args: Array<PartiQLValue>): PartiQLValue {
-        TODO("Function is_int16 not implemented")
+        return boolValue(args[0] is Int16Value)
```
yliuuuu, line 26: If lhs is a int8, should the function returns true?

i.e., `1 is INT16` should this be true?

More generally, I would think if the rhs is a super type of the value type, then the function should be true.

partiql-types/src/main/kotlin/org/partiql/value/datetime/DateTime.kt
```diff
@@ -528,6 +533,17 @@ public abstract class TimestampWithTimeZone : Timestamp {
...
+
+    public companion object {
+
+        @JvmStatic
+        public fun nowZ(): TimestampWithTimeZone {
+            val javaNowZ = OffsetDateTime.now(ZoneOffset.UTC)
+            val date = SqlDate.of(javaNowZ.year, javaNowZ.monthValue, javaNowZ.dayOfMonth)
+            val time = OffsetTimeLowPrecision.of(javaNowZ.hour, javaNowZ.minute, javaNowZ.second, javaNowZ.nano, TimeZone.UtcOffset.of(0))
+            return forDateTime(date, time)
+        }
```
yliuuuu, line 545: Why not just `OffsetLowPrecision.nowZ()`. https://github.com/partiql/partiql-lang-kotlin/blob/main/partiql-types/src/main/kotlin/org/partiql/value/datetime/impl/OffsetTimestampLowPrecision.kt#L136-L140

yliuuuu (changes requested):

plugins/partiql-plugin/src/main/kotlin/org/partiql/plugin/internal/fn/scalar/FnIsInt16.kt
RCHowell: Yes, considering range checks too

yliuuuu (approved):
LGTM

PR #1331, opened 2024-01-04
Adds interruption support for v0.2.8

johnedquinn:
## Relevant Issues
- See the same addition to main (for v0.13.1's release): #1211

## Description
- The base of this PR is branch `v0_2_8_SNAPSHOT` which is at the tag `v0.2.7-alpha`.
- I will be releasing this as `v0.2.8`. Please see each commit.
- Also, I needed to remove the `publish` workflow, or else the GH pages would've been overridden. And, I needed to update the `pull_request` workflow to use `ubuntu-latest` since 18.04 is now unsupported, see this [SO post](https://stackoverflow.com/a/70968478).

## Release Notes
As there isn't a CHANGELOG at this state of our codebase, here is what will be released:
> Release v0.2.8
>
> This release adds support for thread interruption of the CompilerPipeline during compilation and evaluation of long-running queries. To consume this capability, you must opt-in by setting `interruptible` to true in the `CompileOptions`. The changes ensure that evaluation of long-running scans and joins can be interrupted as well as the compilation of long-running aggregations.
>   - Note: the interruption of compilation of aggregations was specifically catered for as there is a bug in the compilation of aggregations that causes premature materialization of the data. Therefore, if you intend on interrupting your queries, it may also be a good idea to interrupt at the compilation phase. You may still interrupt at evaluation.
>   - Note: This addition is NOT added in later versions. If you plan on migrating to a later version, these changes are only present in v0.13.1+. To upgrade, please see the CHANGELOG of v0.13.1 which talks about the addition of thread interruption.

## Other Information
- Updated Unreleased Section in CHANGELOG: **NO** -- there is no changelog

## License Information


lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
```diff
@@ -66,6 +66,23 @@ internal class EvaluatingCompiler(
...
+     * of aggregations and joins, the "evaluation" of aggregations and joins, and the materialization of joins
+     * and from source scans.
+     *
+     * Note: This is essentially a way to avoid constantly checking [CompileOptions.interruptible]. By writing it this
+     * way, we statically determine whether to introduce checks. If the compiler has specified
+     * [CompileOptions.interruptible], the invocation of this function will insert a Thread interruption check. If not
+     * specified, it will not perform the check during compilation/evaluation/materialization.
+     */
+    private val interruptionCheck: () -> Unit = {
+            if (Thread.interrupted()) {
```
alancai98, line 81: Should this be behind a `CompileOptions` flag as the comment mentions? The interruption support added to `main` in #1211 puts the interruption check behind a compile option flag.

alancai98 (commented):
As a followup, could also update the CHANGELOG in `main` to include the v0.2.8 entry.

lang/src/org/partiql/lang/eval/EvaluatingCompiler.kt
johnedquinn: Agreed. I just pushed an update to add this. I originally wasn't sure due to adding new APIs in a patch. I decided to add a deprecated notice on the API.

alancai98 (approved):

PR #1332, opened 2024-01-05
[DO NOT MERGE] Draft of PartiQL SPI factoring for functions

RCHowell:
This PR is a draft of the interfaces for introducing user-defined functions to a catalog via a Connector. ConnectorMetadata can be queried (analogous to objects) for function types (function signatures). The connector also provides ConnectorFunctions which is responsible for getting function implementations from a handle.

YES

## License Information

### DIFF
--- a/partiql-spi/src/main/kotlin/org/partiql/spi/connector/ConnectorMetadata.kt
+++ b/partiql-spi/src/main/kotlin/org/partiql/spi/connector/ConnectorMetadata.kt
@@ -15,7 +15,6 @@
 package org.partiql.spi.connector
 
 import org.partiql.spi.BindingPath
-import org.partiql.types.StaticType
 
 /**
  * Aids in retrieving relevant Catalog metadata for the purpose of planning and execution.
@@ -23,26 +22,20 @@ import org.partiql.types.StaticType
 public interface ConnectorMetadata {
 
     /**
-     * Returns the descriptor of an object. If the handle is unable to produce a [StaticType], implementers should
-     * return null.
-     */
-    public fun getObjectType(handle: ConnectorObjectHandle): StaticType?
-
-    /**
      * Given a [BindingPath], returns a [ConnectorObjectHandle] that corresponds to the longest-available requested path.
      * For example, given an object named "Object" located within Catalog "AWS" and Namespace "a".b"."c", a user could
-     * call [getObjectHandle] with the [path] of "a"."b"."c"."Object". The returned [ConnectorObjectHandle] will contain
+     * call [getObject] with the [path] of "a"."b"."c"."Object". The returned [ConnectorObjectHandle] will contain
      * the object representation and the matching path: "a"."b"."c"."Object"
      *
      * As another example, consider an object within a Namespace that may be a Struct with nested attributes. A user could
-     * call [getObjectHandle] with the [path] of "a"."b"."c"."Object"."x". In the Namespace, only object "Object" exists.
+     * call [getObject] with the [path] of "a"."b"."c"."Object"."x". In the Namespace, only object "Object" exists.
      * Therefore, this method will return a [ConnectorObjectHandle] with the "Object" representation and the matching
      * path: "a"."b"."c"."Object". The returned [ConnectorObjectHandle.path] must be correct for correct
      * evaluation.
      *
-     * If the [path] does not correspond to an existing [ConnectorObject], implementers should return null.
+     * If the [path] does not correspond to an existing [ConnectorObjectType], implementers should return null.
      */
-    public fun getObjectHandle(path: BindingPath): ConnectorObjectHandle?
+    public fun getObject(path: BindingPath): ConnectorObjectHandle?
 
     /**
      * Returns all matching scalar functions (potentially overloaded) at the given path.
--- a/partiql-spi/src/main/kotlin/org/partiql/spi/connector/ConnectorMetadata.kt
+++ b/partiql-spi/src/main/kotlin/org/partiql/spi/connector/ConnectorMetadata.kt
@@ -16,7 +16,6 @@ package org.partiql.spi.connector
 
 import org.partiql.spi.BindingPath
 import org.partiql.types.StaticType
-import org.partiql.types.function.FunctionSignature
 
 /**
  * Aids in retrieving relevant Catalog metadata for the purpose of planning and execution.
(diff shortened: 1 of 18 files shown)


partiql-spi/src/main/kotlin/org/partiql/spi/connector/ConnectorMetadata.kt
```diff
@@ -56,10 +37,55 @@ public interface ConnectorMetadata {
...
      * call [getObjectHandle] with the [path] of "a"."b"."c"."Object"."x". In the Namespace, only object "Object" exists.
      * Therefore, this method will return a [ConnectorObjectHandle] with the "Object" representation and the matching
-     * path: "a"."b"."c"."Object". The returned [ConnectorObjectHandle.absolutePath] must be correct for correct
+     * path: "a"."b"."c"."Object". The returned [ConnectorObjectHandle.path] must be correct for correct
      * evaluation.
      *
      * If the [path] does not correspond to an existing [ConnectorObject], implementers should return null.
      */
-    public fun getObjectHandle(session: ConnectorSession, path: BindingPath): ConnectorObjectHandle?
+    public fun getObjectHandle(path: BindingPath): ConnectorObjectHandle?
```
johnedquinn, line 45: While we currently don't allow for users to change their authorization, almost all of the privileges in SQL1999 are described by the CURRENT_USER (distinct from the SESSION_USER). In the future if we need to support changing the current user, we might need to re-add the ConnectorSession as a parameter.

partiql-spi/src/main/kotlin/org/partiql/spi/connector/ConnectorMetadata.kt
```diff
@@ -56,10 +37,55 @@ public interface ConnectorMetadata {
...
+
+    /**
+     * Returns a list of scalar functions at the given path.
+     *
+     * @param path
```
johnedquinn, line 50: This might be confused with "give me all functions in this directory".
```suggestion
     * Returns all matching scalar functions (potentially overloaded) at the given path.
     * For example, if the [path] is `catalog.schema.foo`, and `foo` has two implementations, this should return both implementations in any order. If it only has a single implementation, this should return a list containing the single implementation.
     * If there is not a matching function, return an empty list.
     *
     * @param path : the [BindingPath] that ends with a function's name. Example: `catalog.schema.foo` where `foo` is the function name.
```

partiql-spi/src/main/kotlin/org/partiql/spi/connector/ConnectorMetadata.kt
```diff
@@ -56,10 +37,55 @@ public interface ConnectorMetadata {
...
+
+    /**
+     * Returns a list of aggregation functions at the given path.
+     *
+     * @param session
+     * @param path
+     * @return
+     */
+    public fun getAggregationFunctions(
+        session: ConnectorSession,
```
johnedquinn: Given that the others no longer use the session, this can be removed.

johnedquinn (commented):
This is a great step in the right direction! It's certainly what we want.

Somewhat related to this PR due to the refactoring of SPI APIs, I've been recently thinking about how we can maintain the separation of concerns for each stage in the pipeline. It seems like `PartiQLPlanner` (library) is the only thing that should use `ConnectorMetadata`. `PartiQLEval` (library) will only deal with loading in data/functions (aka `ConnectorBindings` and `ConnectorFunctions`). The `PartiQL CLI` (application) is the only thing that cares about the `Plugin`. It might be a good idea to extract those APIs, place them where they are needed (to maintain the separation of concerns), and flip the dependency tree so that SPI is a direct dependency of the CLI and the CLI only. Note: this isn't a comment about your PR as your PR is doing something different, but your work made me think of it.

Anyways, great work! I'm in line with the changes that you've made.

partiql-spi/src/main/kotlin/org/partiql/spi/connector/ConnectorMetadata.kt
johnedquinn: I see that your revision has that change.

partiql-spi/src/main/kotlin/org/partiql/spi/connector/ConnectorMetadata.kt
RCHowell: The session as a parameter is redundant. The factory method for ConnectorMetadata has a signature. The ConnectorMetadata would close over this.

```kotlin

// interface method in connector
fun getConnectorMetadata(session: ConnectorSession) : ConnectorMetadata

// session is associated to the metadata
class MyConnectorMetadata(private val session: ConnectorSession) : ConnectorMetadata {

     fun getObjectHandle(path: BindingPath) { ... do somethign with session }

}

```

How it's factored now is redundant, and it allows passing a different session that the one the metadata was created for which is potentially problematic.

```kotlin
val md = getConnectorMetadata(sessionFoo)

md.getObjectHandle(sessionBar, myPath) // !!! 
```

partiql-spi/src/main/kotlin/org/partiql/spi/connector/ConnectorObjectHandle.kt
```diff
@@ -15,9 +15,12 @@
...
+ * Holds a  Catalog and its location within the Catalog.
+ *
+ * @property path   The absolute path to this object in the catalog
+ * @property obj    Arbitrary metadata associated with this object.
  */
 public class ConnectorObjectHandle(
-    public val absolutePath: ConnectorObjectPath,
-    public val value: ConnectorObject
+    public val path: ConnectorPath,
+    public val obj: ConnectorObject,
```
yliuuuu, line 25: What is the need of a separate `ConnectorObject` interface, if this is just going to be holding metadata.

yliuuuu (commented):
My mental model says there are two entity modeled, functions and values. 

For functions, we use the `ConnectorFunction` interface, which holds the function signatures and concrete implementation. 

For values, we use `ConnectorObject`, which is an empty interface for metadata. and `ConnectorBinding` to hold the concrete value. 

I found this difference in modeling a little inconsistent. Could you please elaborate on the decision here?

partiql-spi/src/main/kotlin/org/partiql/spi/connector/ConnectorObjectHandle.kt
RCHowell: ConnectorObject is for arbitrary metadata about a catalog's values. ConnectorFunctionHandle is only for function metadata

RCHowell: I see your question is actually slightly different.

ConnectorObject is not analogous to ConnectorFunction. ConnectorObject would be better named as ConnectorObjectMetadata. It is most similar to FunctionSignature, but can be anything.

PR #1339, opened 2024-01-17
Generate internalized planning IR

RCHowell:
## Relevant Issues
N/A

## Description

Generates internalized planning IR. This simply replaces all public modifiers in the generated source with internal visibility modifiers. The only thing is that you cannot mark interface methods as internal (only the class) hence the skipping of override/fun visit.

## Other Information
No

No

 No

Yes

## License Information


partiql-planner/src/main/resources/builtins.sql
```diff
@@ -1,1681 +0,0 @@
--- [trim] ---------
```
alancai98, line 1: Just curious why are these *.sql files getting deleted?

partiql-planner/build.gradle.kts
```diff
@@ -77,3 +77,54 @@ publish {
     name = "PartiQL Planner"
     description = "PartiQL's Experimental Planner."
 }
+
+// Generate IR
```
alancai98, line 81: nit: could make more clear this is for the internal IR

```suggestion
// Generate internal IR
```

partiql-planner/src/main/resources/builtins.sql
RCHowell: They're inconsequential, outdated, and shouldn't have been pushed to begin with.

partiql-planner/build.gradle.kts
RCHowell: ack

alancai98 (approved):

PR #1340, opened 2024-01-17
Enable Eval Test Suites

yliuuuu:
## Relevant Issues
- [Closes/Related To] N/A

## Description
- Set up the conformance base line for Eval Engine. 
- The modeling are not finalized, many pieces are now hardcoded for the purpose of running the conformance test. Apis are subject to change. 

## Other Information
  - No.

  - N/A Not released. 

- No.

- Yes. 

## License Information


test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/ConformanceTestBase.kt
```diff
@@ -10,6 +11,7 @@ abstract class ConformanceTestBase<T, V> {
     abstract val runner: TestRunner<T, V>
 
     // Tests the eval tests with the Kotlin implementation
+    @Timeout(value = 100, threadMode = Timeout.ThreadMode.SEPARATE_THREAD)
```
yliuuuu, line 14: Unit is second. This is not a performance test. This is for stop long running tests during development process in eval engine. This number can be smaller, but to account for the cold start time and frustration of GitHub runner, decided to make this number a bit larger than needed.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/Env.kt
```diff
@@ -358,6 +360,11 @@ internal class Env(
                     matches.add(match)
                 }
             }
+            // TODO: Verify if this is the correct behavior
+            if (rootType is AnyType) {
+                val match = ResolvedVar.Local(StaticType.ANY, ordinal, rootType, listOf(pathPrefix) + path.steps)
+                matches.add(match)
```
yliuuuu, line 366: To temporarily unblock the path access on any type. This behavior is probably incorrect.

partiql-eval/src/main/kotlin/org/partiql/eval/PartiQLEngine.kt
```diff
@@ -19,7 +22,7 @@ import org.partiql.plan.PartiQLPlan
  */
 public interface PartiQLEngine {
 
-    public fun prepare(plan: PartiQLPlan): PartiQLStatement<*>
+    public fun prepare(plan: PartiQLPlan, session: Session): PartiQLStatement<*>
```
yliuuuu, line 25: making this more akin to the Planner APIs.

partiql-eval/src/main/kotlin/org/partiql/eval/PartiQLEngine.kt
```diff
@@ -31,4 +34,9 @@ public interface PartiQLEngine {
         @JvmStatic
         fun default() = PartiQLEngineBuilder().build()
     }
+
+    public class Session @OptIn(PartiQLFunctionExperimental::class) constructor(
+        var bindings: MutableMap<String, ConnectorBindings> = mutableMapOf(),
+        val functions: MutableMap<String, List<PartiQLFunction>> = mutableMapOf()
+    )
```
RCHowell, line 41: At some point we'll need session variables passed during statement execution.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
```diff
@@ -123,6 +131,33 @@ internal class Compiler(
...
+            else -> error("multiple math")
+        }
+    }
+
+    override fun visitRexOpCallDynamic(node: Rex.Op.Call.Dynamic, ctx: Unit): Operator {
+        val fn = node.candidates
+        fn.forEach {
+            println(it.fn.signature)
+        }
+        error("call dynamic not yet implemented")
```
RCHowell, line 158: remove the prints, perhaps those were for debugging?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/Env.kt
RCHowell: Interesting. I'm not sure we can resolve a variable if the type is ANY/UNKNOWN. I feel this should just be a path expression on the root

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/ConformanceTestBase.kt
RCHowell: Could you please add this comment in the code itself?

RCHowell (changes requested):

partiql-planner/src/main/kotlin/org/partiql/planner/internal/Env.kt
yliuuuu: We need a refactor on PartiQLValueType before doing this. 

Consider:

```
tbl : <<{'a' : 1, 'b': 2}>>
```

We now infer the `tbl` having `PartiQLValueType` of Bag, which is then converted to StaticType `BAG(ANY)`. 

So if we do `SELECT tbl.a FROM tbl`. Without this, the inferencer would fail.

partiql-eval/src/main/kotlin/org/partiql/eval/PartiQLEngine.kt
yliuuuu: Added a to-do note. Maybe we can do this after the work on Sql plugin is completed.

partiql-eval/src/main/kotlin/org/partiql/eval/PartiQLEngine.kt
```diff
@@ -31,4 +35,9 @@ public interface PartiQLEngine {
         @JvmStatic
         fun default() = PartiQLEngineBuilder().build()
     }
+
+    public class Session @OptIn(PartiQLFunctionExperimental::class) constructor(
+        var bindings: MutableMap<String, ConnectorBindings> = mutableMapOf(),
+        val functions: MutableMap<String, List<PartiQLFunction>> = mutableMapOf()
```
johnedquinn, line 41:
```suggestion
        val bindings: Map<String, ConnectorBindings> = emptyMap(),
        val functions: Map<String, List<PartiQLFunction>> = emptyMap()
```

If we don't intend on modifying the maps, can we make these immutable?

partiql-eval/src/main/kotlin/org/partiql/eval/PartiQLEngineBuilder.kt
```diff
@@ -1,36 +1,11 @@
...
-    /**
-     * Kotlin style method for assigning Catalog names to [ConnectorBindings].
-     *
-     * @param catalogs
-     * @return
-     */
-    public fun catalogs(vararg catalogs: Pair<String, ConnectorBindings>): PartiQLEngineBuilder = this.apply {
-        this.catalogs = mutableMapOf(*catalogs)
-    }
+    public fun build(): PartiQLEngine = PartiQLEngineDefault()
```
johnedquinn, line 10: Do we even need the Builder in this case? Could just be a:
```kotlin
companion object {
  @JvmStatic
  public fun default(): PartiQLEngine = PartiQLEngineDefault()
}
```

On the PartiQLEngine interface.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
```diff
@@ -123,8 +131,30 @@ internal class Compiler(
...
     }
 
-    // REL
+    @OptIn(PartiQLFunctionExperimental::class)
+    override fun visitRexOpCallStatic(node: Rex.Op.Call.Static, ctx: Unit): Operator {
+        val fn = node.fn.signature
+        val args = node.args.map { visitRex(it, ctx) }
+        val matches = session.functions
+            .flatMap { it.value }
+            .filter { it.signature == fn }
```
johnedquinn, line 140:
```suggestion
        val matches = session.functions
            .flatMap { it.value }
            .filterIsInstance<PartiQLFunction.Scalar>
            .filter { it.signature == fn }
```

Then, you can drop the check further below. We know that the `fn.signature` is a signature for a scalar function.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/test/TestRunner.kt
```diff
@@ -46,7 +46,7 @@ class TestRunner<T, V>(private val factory: TestExecutor.Factory<T, V>) {
             }
         } catch (e: Exception) {
             when (case.assertion) {
-                is Assertion.EvaluationSuccess -> error("Expected success but exception thrown: $e")
+                is Assertion.EvaluationSuccess -> error("Expected success but exception thrown: $e}")
```
johnedquinn, line 49: Accidental addition?

johnedquinn (changes requested):

partiql-planner/src/main/kotlin/org/partiql/planner/internal/Env.kt
johnedquinn: I think this would break the defeasible reasoning paper (PartiQL Variable Resolution) written by @RCHowell .

If R is NOT known and S is KNOWN, then we must choose S. In this scenario, we wouldn't choose either due to ambiguity. I believe the recently merged PR #1344 solves this.

partiql-eval/src/main/kotlin/org/partiql/eval/PartiQLEngineBuilder.kt
yliuuuu: I decided to keep the builder in this PR since those interfaces has not been finalized yet.

partiql-eval/src/main/kotlin/org/partiql/eval/PartiQLEngine.kt
yliuuuu: Sorry missed those. The original intention was to construct a builder interface for session. But you are right for now we don't need the mutable map.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/Env.kt
yliuuuu: Pick the #1344. Removed the temporary work around.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/executor/EvalExecutor.kt
```diff
@@ -54,11 +58,34 @@ class EvalExecutor(
...
+    // Value comparison of PartiQL Value that utilized Ion Hashcode.
+    // in here, null.bool is considered equivalent to null
+    // missing is considered different from null
+    // annotation::1 is considered different from 1
+    // 1 of type INT is considered the same as 1 of type INT32
+    // we should probably consider adding our own hashcode implementation
+    private fun valueComparison(v1: PartiQLValue, v2: PartiQLValue): Boolean {
+        if (v1.isNull && v2.isNull) {
+            return true
+        }
```
johnedquinn, line 75: `someAnnotation::null.bool` == `someAnnotation::null.int`

`someAnnotation::null.bool` != `otherAnnotation::null.int`

johnedquinn: I've already discussed with @yliuuuu, but for others who may be reading this comment, the KDoc says `annotation::1 is considered different from 1` and `null.bool is considered equivalent to null`. However, this implementation doesn't handle annotations on nulls. It should be a quick fix.

yliuuuu: Nice call out : )

PR #1342, opened 2024-01-19
Adds the unsafe DECIMAL to integer downcasts

RCHowell:
## Relevant Issues

The following explicit/unsafe casts were missing in the type definitions.

```sql
CAST(<decimal> AS INT2)     -- > int16l|missing
CAST(<decimal> AS INT4)     -- > int32|missing
CAST(<decimal> AS INT8)     -- > int64|missing
CAST(<decimal> AS BIGINT)   -- > int64|missing
CAST(<decimal> AS INT)      -- > int|missing
```

## Description

Adds the unsafe DECIMAL to integer downcasts.

## Other Information
No

No

No

Yes

## License Information


partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeLattice.kt
```diff
@@ -266,41 +266,57 @@ internal class TypeLattice private constructor(
...
-                INT32 to explicit(),
-                INT64 to explicit(),
                 BOOL to explicit(),
-                DECIMAL to explicit(),
-                DECIMAL_ARBITRARY to explicit(),
+                INT8 to unsafe(),
+                INT16 to unsafe(),
+                INT32 to unsafe(),
+                INT64 to unsafe(),
+                INT to unsafe(),
```
yliuuuu, line 274: I agree those should be unsafe, but let's hold off for now to avoid behavior change.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeLattice.kt
```diff
@@ -266,41 +266,57 @@ internal class TypeLattice private constructor(
...
+                DECIMAL to coercion(),
```
yliuuuu, line 275: This is an interesting one. It is probably fine now as this gets converted to StaticType.Decimal anyway. But when we have the notion of precision and scale, this will not be true.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeLattice.kt
```diff
@@ -266,41 +266,57 @@ internal class TypeLattice private constructor(
...
-            graph[FLOAT32] = relationships(
+            graph[DECIMAL_ARBITRARY] = relationships(
                 BOOL to explicit(),
-                DECIMAL to explicit(),
+                INT8 to unsafe(),
+                INT16 to unsafe(),
+                INT32 to unsafe(),
+                INT64 to unsafe(),
+                INT to unsafe(),
+                DECIMAL to unsafe(),
```
yliuuuu, line 289: Let's make those explicit as well, for consistent with DECIMAL since right now we are kind of inflating those two types during inferring.

yliuuuu (changes requested):

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeLattice.kt
RCHowell: Either way we have a behavior change. These did not exist before.

yliuuuu (approved):
LGTM.

PR #1343, opened 2024-01-20
Add `RelSort` and comparator between `PartiQLValue`

alancai98:
## Description
- Implements `RelSort` (ORDER BY) for new evaluator
- Adds internal `Comparator` between `PartiQLValue`s to partiql-eval
- Fixes `Rex.Op.Select` to output a list when input relation has the ordered property
- Fixes `Rex.Op.Collection` to output the appropriate collection rather than always outputting a bag
- Modifies `ClobValue` to implement `ScalarValue` rather than `TextValue`

## Other Information

  - Yes, modifies `ClobType` to implement `ScalarValue` rather than `TextValue`

## License Information


partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
```diff
@@ -31,127 +32,130 @@ import org.partiql.plan.Statement
...
-    override fun visitStatementQuery(node: Statement.Query, ctx: Unit): Operator.Expr {
+    override fun visitStatementQuery(node: Statement.Query, ctx: StaticType?): Operator.Expr {
         return visitRex(node.root, ctx)
     }
 
     // REX
 
-    override fun visitRex(node: Rex, ctx: Unit): Operator.Expr {
-        return super.visitRexOp(node.op, ctx) as Operator.Expr
+    override fun visitRex(node: Rex, ctx: StaticType?): Operator.Expr {
```
alancai98, line 72: In the current plan, the rex node's type is only accessible on the `Rex` and not the inner classes (e.g. `Rex.Op.Collection`). Passing it along in the context allows the visitor for the inner classes to access the type, which is needed by some nodes like `Rex.Op.Collection`.

partiql-eval/src/test/kotlin/org/partiql/eval/internal/util/PartiQLValueComparatorTest.kt
```diff
@@ -0,0 +1,429 @@
+package org.partiql.eval.internal.util
```
alancai98, line 1: Tests are taken from [NaturalExprValueComparatorsTest.kt](https://github.com/partiql/partiql-lang-kotlin/blob/partiql-eval-relsort/partiql-eval/src/test/kotlin/org/partiql/lang/eval/NaturalExprValueComparatorsTest.kt) but creates and compares the `PartiQLValue`s explictly rather than passing text strings and running the evaluator.

alancai98:
CI failures seem unrelated to this PR's changes. Will take another look

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/executor/EvalExecutor.kt
```diff
@@ -12,7 +12,6 @@ import org.partiql.eval.PartiQLStatement
 import org.partiql.lang.eval.CompileOptions
```
alancai98, line 12: The changes in this file were taken from [partiql-plugin-impl](https://github.com/partiql/partiql-lang-kotlin/tree/partiql-plugin-impl) to fix the build. Seems like the `partiql-eval` branch build was not working since the merge from `main` commit da03142909b808486e0f170adb52f55d0c0c1eb5.

partiql-planner/src/test/kotlin/org/partiql/planner/util/PlanNodeEquivalentVisitor.kt
```diff
@@ -125,13 +125,6 @@ class PlanNodeEquivalentVisitor : PlanBaseVisitor<Boolean, PlanNode>() {
         return true
     }
 
-    override fun visitRelOpExcludeStepCollIndex(node: Rel.Op.Exclude.Step.CollIndex, ctx: PlanNode): Boolean {
```
alancai98, line 128: This change from the `main` merge to `partiql-eval` commit da03142909b808486e0f170adb52f55d0c0c1eb5 also caused a build failure on `partiql-eval`. `EXCLUDE`'s modeling on `partiql-eval` was changed so this visitor function was not overriding anything.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/util/PartiQLValueComparator.kt
```diff
@@ -0,0 +1,272 @@
...
+ *  * [SexpValue] follows and compares within its type similar to `LIST`.
+ *  * [StructValue] values follow and compare lexicographically based on the *sorted*
+ *    (as defined by this definition) members, as pairs of field name and the member value.
+ *  * [BagValue] values come finally (except with [NullOrder.LAST]), and their values
+ *    compare lexicographically based on the *sorted* child elements.
+ *
+ *  @param nullOrder that places [NullValue], [MissingValue], and typed Ion null values first or last
+ */
+@OptIn(PartiQLValueExperimental::class)
+internal class PartiQLValueComparator(private val nullOrder: NullOrder) : Comparator<PartiQLValue> {
```
alancai98, line 55: TODO: keep internal + move to `partiql-value`; also add public static function to get back `Comparator<PartiQLValue>`

partiql-types/src/main/kotlin/org/partiql/value/PartiQLValueComparator.kt
```diff
@@ -0,0 +1,268 @@
...
+ */
+@OptIn(PartiQLValueExperimental::class)
+public class PartiQLValueComparator {
+    public companion object {
+        /**
+         * @param nullOrder that places [NullValue], [MissingValue], and typed Ion null values first or last
+         */
+        public fun comparator(nullOrder: NullOrder): Comparator<PartiQLValue> = PartiQLValueComparatorInternal(nullOrder)
+    }
+}
```
RCHowell, line 48: Let's put this on the base PartiQLValue and make it `@JvmStatic` so the usage looks like

```
PartiQLValue.comparator()
```

We can move the kdoc there as well. We don't want to be able to instantiate PartiQLValueComparator like we can here

RCHowell (changes requested):

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelSort.kt
```diff
@@ -0,0 +1,67 @@
...
+import org.partiql.value.PartiQLValueComparator
+import org.partiql.value.PartiQLValueExperimental
+
+@OptIn(PartiQLValueExperimental::class)
+internal class RelSort(
+    val input: Operator.Relation,
+    val specs: List<Pair<Operator.Expr, Rel.Op.Sort.Order>>
+
+) : Operator.Relation {
+    private var records: MutableList<Record> = mutableListOf()
```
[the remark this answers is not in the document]
alancai98: Changed to have `records` be an `Iterator<Record>`, which allows us to use `records.next()` rather than `removeAt`. I think `records` directly as an `Iterator` may be cleaner than having `records` be an `Iterable` with a backing collection, but let me know if there's some simpler representation.

partiql-types/src/main/kotlin/org/partiql/value/PartiQLValueComparator.kt
alancai98: Ack, moved the comparator companion object to `PartiQLValue` and added `@JvmStatic`.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
alancai98: Hmm I'd need to play around with the generated code a bit to see if there's a better way. Only concern I'd have is that the `type` field for `Rex` is a different type (i.e. `StaticType`) than the `type` field for `Rel` (i.e. `Rel.Type` which has schema and props). Probably that representation in the `PlanNode` base class would be an enum that could be either of those `type` definitions?

Anyways, perhaps we should tackle this in another issue/PR.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/executor/EvalExecutor.kt
RCHowell: partiql-eval is fixed now

partiql-types/src/main/kotlin/org/partiql/value/PartiQLValue.kt
```diff
@@ -48,6 +48,44 @@ public sealed interface PartiQLValue {
...
+         *  * [SexpValue] follows and compares within its type similar to `LIST`.
+         *  * [StructValue] values follow and compare lexicographically based on the *sorted*
+         *    (as defined by this definition) members, as pairs of field name and the member value.
+         *  * [BagValue] values come finally (except with [nullsFirst] == true), and their values
+         *    compare lexicographically based on the *sorted* child elements.
+         *
+         * @param nullsFirst whether [NullValue], [MissingValue], and typed Ion null values come first
+         */
+        @JvmStatic
+        public fun comparator(nullsFirst: Boolean): Comparator<PartiQLValue> = PartiQLValueComparatorInternal(nullsFirst)
```
RCHowell, line 87:
```suggestion
        @JvmStatic
        @JvmOverloads
        public fun comparator(nullsFirst: Boolean = true): Comparator<PartiQLValue> = PartiQLValueComparatorInternal(nullsFirst)
```

Let's make a non-argument factory method with the default behavior ie ASC NULLS FIRST

RCHowell (changes requested):

partiql-planner/src/test/kotlin/org/partiql/planner/util/PlanNodeEquivalentVisitor.kt
alancai98: Cool. I'll rebase

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/executor/EvalExecutor.kt
alancai98: I shall rebase

partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
```diff
@@ -40,73 +41,74 @@ import org.partiql.plan.Statement
...
 import org.partiql.value.PartiQLValueType
 import java.lang.IllegalStateException
 
 internal class Compiler(
     private val plan: PartiQLPlan,
-    private val session: PartiQLEngine.Session
-) : PlanBaseVisitor<Operator, Symbols>() {
+    private val session: PartiQLEngine.Session,
+    private val symbols: Symbols
+) : PlanBaseVisitor<Operator, StaticType?>() {
```
alancai98, line 53: (Following rebase of partiql-eval) Change `ctx` to have type `StaticType?`. Initialize `symbols` as part of `Compiler` constructor since it is not scope sensitive.

partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEngineDefaultTest.kt
```diff
@@ -203,6 +203,20 @@ class PartiQLEngineDefaultTest {
                     )
                 )
             ),
+            SuccessTestCase(
```
alancai98, line 206: (Following rebase of partiql-eval) This tests and the `EXCLUDE` tests from L250-315 were inadvertently deleted in #1362.

RCHowell (approved):
Nice work

partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
alancai98: > If you check out PlanTyper, you'll see we use two different visitors since the type is parameterized.
Oh I see. The separate visitors for Rel and Rex doesn't seem too cumbersome.

> The best situation however would be handwritten nodes with annotation based generation like Lombok. This would give us the most control.
Agree w/ a mix of code-generated nodes and handwritten nodes would give us the most flexibility when it comes to these interfaces.

PR #1344, opened 2024-01-23
Isolates local variable resolution logic

RCHowell:
## Description
While working towards unifying our resolution logic for catalog identifiers, I began cleaning up the Env.kt. This cleans up local resolution (isolating all logic to TypeEnv for easy testing). It also appropriately prioritizes known schema matches over open schema matches (known vs not known).

This matches what is defined in our specification for locals resolution, but there's a possible extension to discuss. We may consider matching longest locals paths. If you read p38 carefully, it indicates we only match the path root for locals (that's what in this PR) — but I believe we should consider matching longest local variable paths as well.

## Other Information
No

No

No

Yes

## License Information


johnedquinn (approved):
Very interesting. I like the consolidation of local typing into the TypeEnv while utilizing the normal visitor for the PlanTyper to type the rest of the produced path. It's a subtle difference, but it makes for a nice flow.

Looks good to me!

PR #1353, opened 2024-01-26
Adds support for PERMISSIVE vs STRICT

johnedquinn:
## Relevant Issues
- None

## Description
- Adds support for PERMISSIVE vs STRICT
- In short, this implementation attempts to reduce as much computation/checking as possible. In STRICT mode, we never attempt to recover from a TypeCheckException. In Permissive mode, given that MISSING represents a type-check exception and is largely propagated through operations, we attempt to catch TypeCheckExceptions as infrequently as possible (only when we need to). We specifically catch the errors (according to the specification) when:
  - the value of an attribute in a struct results in a TypeCheckException.
  - an element in an array results in a TypeCheckException
  - creating the elements of a SELECT VALUE (AKA the constructor results in a TypeCheckException)
  - when the inputs to a function (that takes in ANY/MISSING) results in a TypeCheckException
  - top-level expressions (AKA under StatementQuery)
- There are also multiple scenarios specified in the PartiQL Specification that handle mistyping cases that are edge-cases. For example: Iteration over a scalar value, iteration over an absent value, iteration over a tuple value, position variable on bags, out-of-bounds indexing, and more. In these scenarios, this implementation occasionally attempts to mitigate as many runtime checks as possible by compiling the nodes to permissive-specific nodes. See `RelScanPermissive` for example.
- This PR also adds support for `LIKE` by copying much of the logic from EvaluatingCompiler. If requested, I can remove these additions and include them in a separate PR.

## Testing
- This PR follows the specification to find guidance on when to use MISSING and when to error. See the naming of tests to see the relevant PartiQL Specification sections.

## Other Information

## License Information

### DIFF
--- a/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelScanIndexedPermissive.kt
+++ b/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelScanIndexedPermissive.kt
@@ -1,6 +1,5 @@
 package org.partiql.eval.internal.operator.rel
 
-import org.partiql.errors.TypeCheckException
 import org.partiql.eval.internal.Record
 import org.partiql.eval.internal.operator.Operator
 import org.partiql.value.BagValue
@@ -8,27 +7,29 @@ import org.partiql.value.CollectionValue
 import org.partiql.value.PartiQLValue
 import org.partiql.value.PartiQLValueExperimental
 import org.partiql.value.int64Value
+import org.partiql.value.missingValue
 
 @OptIn(PartiQLValueExperimental::class)
-internal class RelScanIndexed(
+internal class RelScanIndexedPermissive(
     private val expr: Operator.Expr
 ) : Operator.Relation {
 
     private lateinit var iterator: Iterator<PartiQLValue>
     private var index: Long = 0
+    private var isIndexable: Boolean = true
 
     override fun open() {
         val r = expr.eval(Record.empty)
         index = 0
         iterator = when (r) {
             is BagValue<*> -> {
-                close()
-                throw TypeCheckException()
+                isIndexable = false
+                r.iterator()
             }
             is CollectionValue<*> -> r.iterator()
             else -> {
-                close()
-                throw TypeCheckException()
+                isIndexable = false
+                iterator { yield(r) }
             }
         }
     }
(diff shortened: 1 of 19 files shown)


partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEngineDefaultTest.kt
```diff
@@ -38,6 +42,11 @@ class PartiQLEngineDefaultTest {
     @Execution(ExecutionMode.CONCURRENT)
     fun sanityTests(tc: SuccessTestCase) = tc.assert()
 
+    @ParameterizedTest
+    @MethodSource("typingModeTestCases")
+    @Execution(ExecutionMode.CONCURRENT)
+    fun typingModeTests(tc: TypingTestCase) = tc.assert()
```
yliuuuu, line 48: Thanks. Very nice sets of tests.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelScanIndexedPermissive.kt
```diff
@@ -0,0 +1,53 @@
...
+internal class RelScanIndexedPermissive(
+    private val expr: Operator.Expr
+) : Operator.Relation {
+
+    private lateinit var iterator: Iterator<PartiQLValue>
+    private var index: Long = 0
+    private var isIndexable: Boolean = true
+
+    override fun open() {
+        val r = expr.eval(Record.empty)
```
yliuuuu, line 22: Not related to this PR, but does this work for subquery?

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprPathKey.kt
```diff
@@ -19,6 +19,6 @@ internal class ExprPathKey(
         val rootEvaluated = root.eval(record).check<StructValue<PartiQLValue>>()
         val keyEvaluated = key.eval(record).check<StringValue>()
         val keyString = keyEvaluated.value ?: error("String value was null")
-        return rootEvaluated[keyString] ?: missingValue()
+        return rootEvaluated[keyString] ?: throw TypeCheckException()
```
yliuuuu, line 22: >  Tuple navigation in tuples with duplicate attributes When the tuple t has multiple attributes a, the tuple
path navigation t.a will return the first instance of a. Note that for tuples whose order is defined by schema, this is well-defined, for unordered tuples, it is implementation defined which attribute is returned in permissive mode or an error in type checking mode, which is described in Section 4.1

It seems like right now we don't have the notion of ordered vs unordered struct during runtime. Maybe just keep a todo note here?

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelScanIndexedPermissive.kt
johnedquinn: That's a good question. I have a subquery PR out, and when I rebase, I'll take a look at that.

PR #1356, opened 2024-01-29
Adds support for ROW and SCALAR subquery coercion

johnedquinn:
## Relevant Issues
- None

## Description
- Adds support for ROW and SCALAR subquery coercion
- NOTE: There is a bug with correlated subqueries (not subquery coercion which is this PR) that doesn't let us handle references to variables in the outer scope. See https://github.com/partiql/partiql-lang-kotlin/pull/1353#discussion_r1471748883. I've looked further into it, and it is likely going to require an overhaul of planning/evaluation. I've decided to exclude this from this PR, as this is specifically for coercion.

## How PR accomplishes this
- According to the PartiQL Specification, when SQL's SELECT is seen in a query (not on the RHS of an IN, not on the RHS of a FROM, and not as a top-level expression), it is coerced to a scalar. It does this by grabbing the first attribute of the first value (a tuple) of the SQL SELECT. However, if it is on the LHS or RHS of a literal array, then it gets coerced into a LIST. It does this by grabbing all attribute of the first value (a tuple) of the SQL SELECT.
- In order to accomplish this, we need to maintain the original query's semantics upon typing the plan. That being said, we cannot preemptively normalize a SQL SELECT to a PARTIQL SELECT VALUE. We need to first gather sufficient information to determine whether a SELECT should be coerced at all (and if so, whether it should be a SCALAR or ROW coercion). This PR moves the `NormalizeSelect` to occur **during** the `RelConverter`'s transformation.
- With this change, at planning time, we have a clear understanding of the user's intentions while coercing subqueries.

## Other Information
- Any backward-incompatible changes? **Potentially**
  - Potentially yes for behavioral changes. `NormalizeSelect` no longer occurs in the `org.partiql.ast.normalize` API.

## License Information

### DIFF
--- a/partiql-ast/src/main/kotlin/org/partiql/ast/normalize/Normalize.kt
+++ b/partiql-ast/src/main/kotlin/org/partiql/ast/normalize/Normalize.kt
@@ -23,7 +23,6 @@ public fun Statement.normalize(): Statement {
     // could be a fold, but this is nice for setting breakpoints
     var ast = this
     ast = NormalizeFromSource.apply(ast)
-    ast = NormalizeSelect.apply(ast)
     ast = NormalizeGroupBy.apply(ast)
     return ast
 }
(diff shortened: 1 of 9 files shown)


johnedquinn:
Closing in favor of #1365

PR #1355, opened 2024-01-29
Support Cast In Eval Engine

yliuuuu:
## Relevant Issues
- [Closes/Related To] Issue #XXX

## Description
In this PR, we model Cast as its own Plan Node. 

Added Plan Node: 
```
cast::{
  operand: partiql_value_type, // operand is a value.
  target: partiql_value_type, // TODO: Consider change this to Static Type
  cast_type: cast_type::[
    COERCION,
    EXPLICIT,
    UNSAFE
  ],
}

rex::{
  op:[
    cast_op::{
      arg: rex,
      cast: cast
    },
  ]
}    
```

Internal Plan Node is modeled exactly the same as public Plan Node for now. 

### Cast Table
The cast table are hard coded for now. Check [here](https://github.com/partiql/partiql-lang-kotlin/blob/2d493c625e09c2a0695a707e6b41841b9b73e2d6/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeCasts.kt) for the cast table. 

### Planning
During planning, the planner will check if there is a available cast relationship exist. If so, the planner store the cast relationship in cast Node, and the `castOp` node points to the cast Node. 

Note that if in planning time, we know that the arg is NULL, we reduce the cast op to a `RexOpLit($typedNullValue)` with type as the target type. If the arg is `missing` then we reduce the cast op to a `RexOpLit(missingValue())`

Note that for dynamic function, instead of storing a function signature that hints on the correct cast to invoke, we store a Cast ref to achieve the same functionality. 

Example: 
```
t_int2 : StaticType.INT2
t2: StaticType.unionOf(INT4, DECIMAL) 

t_int2 + t2 

⚬ PartiQLPlan[tag=Plan-9835b5f]
   ├──Catalog[name=test, tag=Plan-b2e2597c]
   │  ├──Symbol[type=int2, tag=Plan-90f65f9e]
   │  └──Symbol[type=union(decimal, int4), tag=Plan-1ba87674]
   └──Query[tag=Plan-810bfdb7]
      └──Rex[type=union(decimal, int4), tag=Plan-df596342]
         └──Dynamic[tag=Plan-8aa94980]
            ├──Rex[type=int2, tag=Plan-3ff7f113]
            │  └──Global[tag=Plan-7675096]
            │     └──Ref[catalog=0, symbol=0, tag=Plan-7412482]
            ├──Rex[type=union(decimal, int4), tag=Plan-c84905a8]
            │  └──Global[tag=Plan-ec8cbafe]
[...]

PR #1360, opened 2024-02-02
Merges partiql-spi work into partiql-plugin-impl

RCHowell:
## Description

At the start of the month, John, Yingtao, and Alan have been working off a branch of partiql-eval called partiql-plugin-impl. I have been working on the [partiql-spi](https://github.com/partiql/partiql-lang-kotlin/commits/partiql-spi/) branch which adds functions to catalogs and makes the necessary updates to the planner and plan representation.

This PR merges the partiql-spi work into partiql-plugin-impl. Once merged, we should merge partiql-plugin-impl into partiql-eval, and only work off of partiql-eval from here on out.

This PR should be followed-up by [Yingtao's CAST PR](https://github.com/partiql/partiql-lang-kotlin/pull/1355). I've run tests locally and we have a decrease of about 200 tests on the conformance report.

I have not had the time to look into the changes, but honestly I'm surprised at 25% passing given how extensive the changes the planner and plugin system are.

Yes

## License Information


partiql-spi/src/main/kotlin/org/partiql/spi/connector/sql/SqlConnector.kt
```diff
@@ -0,0 +1,48 @@
...
+     *
+     * @param session
+     * @return
+     */
+    abstract override fun getMetadata(session: ConnectorSession): SqlMetadata
+
+    override fun getBindings(): ConnectorBindings = SqlBindings(info)
+
+    @FnExperimental
+    override fun getFunctions(): ConnectorFnProvider = SqlFnProvider(info.functions)
```
yliuuuu, line 47: Per my understanding: 
I believe that the one of the reasons why those class are open is to allow customer to have those built in function available. 

However, with the currently way that the getFunction is implemented, the way to supply a UDF maybe to have your own implementation on ConnectorFunctionProvider. 


```
    @OptIn(FnExperimental::class)
    internal class TestFunctionProvider(val index: FnIndex, val info: FnIndex) : ConnectorFnProvider {
        @FnExperimental
        override fun getFn(path: ConnectorPath, specific: String): Fn? {
            val fn = index.get(path, specific)
            if (fn is Fn) return fn
            val buildin = info.get(path, specific)
            if (buildin is Fn) return buildin
            return null
        }
    }
```

I think it will be better if we provide APIs to combine function index? Or am I missing something here?

yliuuuu (commented):
Looks good to me in general. Thanks for the great work. 

Had one comment on SQL connector interface. 

I also fixed some build issue and unit test failure in #1361

PR #1365, opened 2024-02-07
Adds support for correlated subqueries & coercions via scoping

johnedquinn:
## Description
- This PR adds support for correlated subqueries and subquery coercions by:
  1. Introducing the concept of a stack to mimic the concept of a variables environment (see the PartiQL Spec).
- The stack keeps track of variables from outer scopes whenever a new scope is introduced. The `PlanTyper` now has the concept of a `Stack`, and the `Compiler` (along with the operator implementations) matches the functionality of the `PlanTyper`. The known instances of scope creation are:
  1. Top-level SFW queries (including PIVOTS)
  2. Subqueries (including PIVOTS)
  3. JOINs (according to the PartiQL Specification, a PartiQL JOIN's RHS can reference variables from the LHS).
- `Var`'s are just pointers to a location on our internal stack. Many SQL engines attempt to de-correlate all correlated subqueries during planning (to avoid using the stack), and therefore (almost) **all** data comes from the natural data flow of relational operators. Note that many engines do maintain a stack for variable resolution, but not all. PartiQL allows for referencing variables in many clauses where it is oftentimes disallowed in SQL. This implementation maintains the structure of our evaluator's use of the Volcano Model, however, we still allow accessing of the stack. This implementation of `Var` provides us with 4 benefits:
  1. It allows us to model what the PartiQL Specification says.
  2. It allows for quick access to bindings (with indexing).
  3. It helps us maintain the simplicity of the Volcano Model (using `Var`'s of depth 0) while also giving us flexibility to reference variables from outer variable environments (depths > 0).
  4. It allows us to KNOW which variables are coming from the outer variable environments **during** planning. Therefore, at planning time, we may attempt to remove these by de-correlating any correlated subqueries. I have a POC that does this.
- This PR also handles subquery coercion by modifying how we model subqueries. Now, instead of wrapping a `RexOpSelect`, the `RexOpSubquery` also has a constructor and input relation. This allows us the flexibility to call `next()` as we see fit to reduce any excess computation.

## Other (Small) Changes
- I've updated the PlanPrinter to be easier to read. No more printing of `Rel`'s and `Rex`'s. Though, we still print the types/schema.

## Other Information

## License Information

### DIFF
--- a/partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
+++ b/partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
@@ -133,15 +133,20 @@ internal class Compiler(
         }
     }
 
-    override fun visitRexOpVarOuter(node: Rex.Op.Var.Outer, ctx: StaticType?): Operator {
-        return ExprVarOuter(node.scope, node.ref, env)
+    /**
+     * All variables from the local scope have a depth of 0.
+     *
+     * All variables coming from the stack have a depth > 0. To slightly minimize computation at execution, we subtract
+     * the depth by 1 to account for the fact that the local scope is not kept on the stack.
+     */
+    override fun visitRexOpVar(node: Rex.Op.Var, ctx: StaticType?): Operator {
+        return when (node.depth) {
+            0 -> ExprVarLocal(node.ref)
+            else -> ExprVarOuter(node.depth - 1, node.ref, env)
+        }
     }
 
-    override fun visitRexOpVarLocal(node: Rex.Op.Var.Local, ctx: StaticType?): Operator {
-        return ExprVarLocal(node.ref)
-    }
-
-    override fun visitRexOpVarGlobal(node: Rex.Op.Var.Global, ctx: StaticType?): Operator = symbols.getGlobal(node.ref)
+    override fun visitRexOpGlobal(node: Rex.Op.Global, ctx: StaticType?): Operator = symbols.getGlobal(node.ref)
 
     override fun visitRexOpPathKey(node: Rex.Op.Path.Key, ctx: StaticType?): Operator {
         val root = visitRex(node.root, ctx)
--- a/partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
+++ b/partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
@@ -57,11 +57,7 @@ internal class Compiler(
     private val symbols: Symbols
 ) : PlanBaseVisitor<Operator, StaticType?>() {
 
-    /**
-     * This represents variable scopes in a PartiQL Query. This is in relation to [Rex.Op.Var.Upvalue]. For now,
-     * the three scope-creating expressions are: [Rex.Op.Select], [Rex.Op.Subquery], [Rel.Op.Join].
-     */
-    private val scopes: Stack<Record> = Stack<Record>()
+    private val env: Environment = Environment()
 
     fun compile(): Operator.Expr {
         return visitPartiQLPlan(plan, null)
@@ -116,15 +112,15 @@ internal class Compiler(
         val rel = visitRel(node.rel, ctx)
         val ordered = node.rel.type.props.contains(Rel.Prop.ORDERED)
         val constructor = visitRex(node.constructor, ctx).modeHandled()
-        return ExprSelect(rel, constructor, ordered, scopes)
+        return ExprSelect(rel, constructor, ordered, env)
     }
 
     override fun visitRexOpSubquery(node: Rex.Op.Subquery, ctx: StaticType?): Operator {
         val constructor = visitRex(node.constructor, ctx)
         val input = visitRel(node.rel, ctx)
         return when (node.coercion) {
-            Rex.Op.Subquery.Coercion.SCALAR -> ExprSubquery.Scalar(constructor, input, scopes)
-            Rex.Op.Subquery.Coercion.ROW -> ExprSubquery.Row(constructor, input, scopes)
+            Rex.Op.Subquery.Coercion.SCALAR -> ExprSubquery.Scalar(constructor, input, env)
+            Rex.Op.Subquery.Coercion.ROW -> ExprSubquery.Row(constructor, input, env)
         }
     }
 
@@ -139,7 +135,7 @@ internal class Compiler(
     }
 
     override fun visitRexOpVarUpvalue(node: Rex.Op.Var.Upvalue, ctx: StaticType?): Operator {
-        return ExprUpvalue(node.frameRef, node.valueRef, scopes)
+        return ExprUpvalue(node.frameRef, node.valueRef, env)
     }
 
     override fun visitRexOpVarLocal(node: Rex.Op.Var.Local, ctx: StaticType?): Operator {
@@ -252,10 +248,10 @@ internal class Compiler(
         val rhs = visitRel(node.rhs, ctx)
         val condition = visitRex(node.rex, ctx)
         return when (node.type) {
-            Rel.Op.Join.Type.INNER -> RelJoinInner(lhs, rhs, condition, scopes)
-            Rel.Op.Join.Type.LEFT -> RelJoinLeft(lhs, rhs, condition, scopes)
-            Rel.Op.Join.Type.RIGHT -> RelJoinRight(lhs, rhs, condition, scopes)
-            Rel.Op.Join.Type.FULL -> RelJoinOuterFull(lhs, rhs, condition, scopes)
+            Rel.Op.Join.Type.INNER -> RelJoinInner(lhs, rhs, condition, env)
+            Rel.Op.Join.Type.LEFT -> RelJoinLeft(lhs, rhs, condition, env)
+            Rel.Op.Join.Type.RIGHT -> RelJoinRight(lhs, rhs, condition, env)
+            Rel.Op.Join.Type.FULL -> RelJoinOuterFull(lhs, rhs, condition, env)
         }
     }
 
--- a/partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
+++ b/partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
@@ -23,6 +23,7 @@ import org.partiql.eval.internal.operator.rex.ExprCase
 import org.partiql.eval.internal.operator.rex.ExprCast
 import org.partiql.eval.internal.operator.rex.ExprCollection
 import org.partiql.eval.internal.operator.rex.ExprLiteral
+import org.partiql.eval.internal.operator.rex.ExprLocal
 import org.partiql.eval.internal.operator.rex.ExprPathIndex
 import org.partiql.eval.internal.operator.rex.ExprPathKey
 import org.partiql.eval.internal.operator.rex.ExprPathSymbol
@@ -31,8 +32,9 @@ import org.partiql.eval.internal.operator.rex.ExprPivot
 import org.partiql.eval.internal.operator.rex.ExprPivotPermissive
 import org.partiql.eval.internal.operator.rex.ExprSelect
 import org.partiql.eval.internal.operator.rex.ExprStruct
+import org.partiql.eval.internal.operator.rex.ExprSubquery
 import org.partiql.eval.internal.operator.rex.ExprTupleUnion
-import org.partiql.eval.internal.operator.rex.ExprVar
+import org.partiql.eval.internal.operator.rex.ExprUpvalue
 import org.partiql.plan.Catalog
 import org.partiql.plan.PartiQLPlan
 import org.partiql.plan.PlanNode
@@ -47,12 +49,19 @@ import org.partiql.types.StaticType
 import org.partiql.value.PartiQLValueExperimental
 import org.partiql.value.PartiQLValueType
 import java.lang.IllegalStateException
+import java.util.Stack
 
 internal class Compiler(
     private val plan: PartiQLPlan,
     private val session: PartiQLEngine.Session,
     private val symbols: Symbols
 ) : PlanBaseVisitor<Operator, StaticType?>() {
+
+    /**
+     * This represents variable scopes in a PartiQL Query. This is in relation to [Rex.Op.Var.Upvalue]. For now,
+     * the three scope-creating expressions are: [Rex.Op.Select], [Rex.Op.Subquery], [Rel.Op.Join].
+     */
+    private val scopes: Stack<Record> = Stack<Record>()
 
     fun compile(): Operator.Expr {
         return visitPartiQLPlan(plan, null)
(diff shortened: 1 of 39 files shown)


partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
```diff
@@ -23,6 +23,7 @@ import org.partiql.eval.internal.operator.rex.ExprCase
 import org.partiql.eval.internal.operator.rex.ExprCast
```
RCHowell, line 23: Wouldn't you need to add scopes to every Rel and Expr?

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprSubquery.kt
```diff
@@ -0,0 +1,83 @@
...
+        scopes.push(record)
+        input.open()
+        val firstRecord = input.next()
+        if (firstRecord == null) {
+            scopes.pop()
+            return null
+        }
+        val tuple = constructor.eval(firstRecord).check<StructValue<*>>()
+        val secondRecord = input.next()
+        scopes.pop()
```
RCHowell, line 77: Considering this is the second place we're seeing the scope push/pop it makes me wonder if the compiler should have an `env` class that encapsulates the scope logic, as well as scope push/pop macro using an inline lambda

Pretty much exactly the kotlin usage of inlines. Also I think we need to close the input. Similar to the open/close of input makes me think of try-with-resources and we may consider another macro-inline to avoid this accidental failure to close the input.

Notice too that we don't need the L72 and L77 pop. I encourage you to give it a go and check the decompiled Java

```suggestion
        env.scope(record) {
            input.open()
            val firstRecord = input.next() ?: return null
            val tuple = constructor.eval(firstRecord).check<StructValue<*>>()
            val secondRecord = input.next()
            input.close()
        }
```

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprSubquery.kt
```diff
@@ -0,0 +1,83 @@
...
+        if (secondRecord != null) {
+            throw TypeCheckException()
+        }
+        return tuple.values.iterator()
```
RCHowell, line 81: Why do we want a tuple.values.iterator()?

Why not just

```
private fun Operator.Relation.first(): StructValue<*>?

// let n = # rows
//   n = 0      ->  return null
//   n = 1       ->  return constructor.eval(row[0])
//   n > 1       -> throw CardinalityException()   
```

I don't think we want a TypeCheckException because that might be considered MISSING. The specification suggest we throw a cardinality exception for n > 1 in strict and ignore in permissive.

> An PartiQL extension with respect to SQL is that, in the permissive mode, subqueries that fail to coerce to the
required type (scalar or tuple) still run, as opposed to failing. They simply omit from the results the data that
correspond to the coercion failures.

Also, we'll need to throw/ignore for scalar/row coercions depending on the number of struct values.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprUpvalue.kt
```diff
@@ -0,0 +1,22 @@
...
+import org.partiql.value.PartiQLValue
+import org.partiql.value.PartiQLValueExperimental
+import java.util.Stack
+
+/**
+ * Returns the appropriate value from the stack.
+ */
+internal class ExprUpvalue(
+    private val frameIndex: Int,
+    private val varIndex: Int,
```
RCHowell, line 14:
```suggestion
    private val scope: Int,
    private val ref: Int,
```

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprUpvalue.kt
```diff
@@ -0,0 +1,22 @@
...
+import org.partiql.eval.internal.Record
+import org.partiql.eval.internal.operator.Operator
+import org.partiql.value.PartiQLValue
+import org.partiql.value.PartiQLValueExperimental
+import java.util.Stack
+
+/**
+ * Returns the appropriate value from the stack.
+ */
+internal class ExprUpvalue(
```
RCHowell, line 12: Perhaps we keep implementation names consistent with plan node names

- ExprVarGlobal
- ExprVarLocal
- ExprVarOuter

I'm only slightly hesitant on the upvalue name because I believe there is some lifetime implication, whereas we are more similar to the nested loop situation

```
for i in [0..n] {
   for j in [0...m] {
       // stack [ i, j ] <- top
   }
}
```

For us, we aren't concerned about lifetimes. In our case, we are calling "i" an upvalue when really it is just earlier in the traditional stack. The "upvalue" from Lua is more about keeping a value (that was previously on the stack) alive when its scope no longer exists (ie we returned, but had a closure so need to keep "i" alive). 

Our situation is exactly

```
for i in [0..n] {
   // record here [ i ]
   for j in [0...m] {
       // record here [ j ]   <-- record may have several values, this is what we call the "stack"
       // we do not concatenate records to form a singular stack, instead we do a stack-of-records aka stack-of-stacks, to create our scope
       //  scope [[i], [j]]
       //     outer^    ^local
   }
}
```

partiql-plan/src/main/resources/partiql_plan.ion
```diff
@@ -77,13 +77,21 @@ rex::{
...
+      },
+      // Refers to a value defined in outer scopes. See https://craftinginterpreters.com/closures.html#upvalues .
+      upvalue::{
+        frameRef: int,
+        valueRef: int,
+      },
+      // Refers to a value in the database environment.
+      global::{
+        ref: ref,
+      },
```
RCHowell, line 93:
```suggestion
      // Refers to a value coming from the input record.
      local::{
        ref: int
      },
      // Refers to a value defined in outer scopes.
      outer::{
        depth: int,
        ref: int,
      },
      // Refers to a value in the database environment.
      global::{
        ref: ref,
      },
```

Our implementation is more like the scope walking in the Java implementation of Lox rather than C implementation hence `depth`. https://craftinginterpreters.com/resolving-and-binding.html#interpreting-resolved-variables We are using the JVM call stack rather than a stack-based VM

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeEnv.kt
```diff
@@ -19,7 +20,10 @@ import org.partiql.value.stringValue
 /**
  * TypeEnv represents a variables type environment.
  */
-internal class TypeEnv(public val schema: List<Rel.Binding>) {
+internal data class TypeEnv(
+    public val schema: List<Rel.Binding>,
+    public val stack: List<TypeEnv>
```
RCHowell, line 25:
```
s/stack/scopes
s/height/depth
```

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeEnv.kt
```diff
@@ -57,7 +67,7 @@ internal class TypeEnv(public val schema: List<Rel.Binding>) {
      * @param name
      * @return
      */
-    private fun matchRoot(name: BindingName): Rex? {
+    private fun matchRoot(name: BindingName, level: Int): Rex? {
```
RCHowell, line 70: Let's keep the naming consistent with the plan node, why not call this local frameRef you know?

I suggest "depth" in favor of frameRef, height, level, etc.

partiql-planner/src/main/resources/partiql_plan_internal.ion
```diff
@@ -82,9 +82,19 @@ rex::{
...
+        valueRef: int
+      },
+      // Refers to a value coming from the input record.
+      local::{
         ref: int,
       },
+      // Refers to a value in the database environment.
+      global::{
+        ref: '.ref.obj',
+      },
```
RCHowell, line 97: Apply the same suggestions from partiql_plan.ion

RCHowell (changes requested):

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprUpvalue.kt
johnedquinn: Agreed. Will update!

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprSubquery.kt
johnedquinn: I like this. Will be adding it in upcoming revision.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprSubquery.kt
johnedquinn: Further on in the same section:
> In this example, in one instance the inner SELECT evaluates to a collection with more than one element.
Because the COLL_TO_SCALAR function produces a MISSING instead of failing, the query works

Either way, in permissive the error should be coerced to MISSING. I can make it a cardinality exception like the SQL:1999 spec says, but I'll just need to coerce that too.

johnedquinn: And as for why the iterator: it doesn't really matter. In both the Scalar and Row scenarios, you need access to the values, so it was just for de-duplicating one or two lines of code.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeEnv.kt
johnedquinn: In the section you referenced in crafting interpreters, the depth is the difference between the current scope and the variable being referenced. I'm moving away from that because we don't need to know the difference. We can just index the stack as-is.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
johnedquinn: Short answer -- no. This PR is an attempt to use both the Volcano Model as well as the Variables Environment (stack). 

In a traditional program (non-SQL), you might need to place values on the stack to invoke a function, but we're modeling data flow through the transfer of records -- volcano model. So, most expressions don't need access to this "global" environment of scopes. And, they don't need to push/pop scopes. Same goes for relations.

We can kind of think of new scopes as new concatenations of the current variables environment -- and we don't need to concatenate the variables environment for every Rel/Expr. The PartiQL Spec talks about it briefly:
> In particular, the WHERE’s condition is evaluated once for each input binding tuple b in Bin
WHERE. In general, each evaluation is done within the bindings environment (ρ0, ρ‖b), i.e., the concatenation of the binding tuple ρ (where ρ is the binding environment of the SFW query) with the binding tuple b that has the variables of the FROM clause

It kind of describes the situation where `ρ` is the scope of variables defined outside of the SFW query, and `b` is the binding tuple being passed from the FILTER's input (SCAN). Assuming this is a SELECT .. FROM .. WHERE, I'm going to make the assumption that the evaluation of the projection is done within the bindings environment `(ρ0, ρ‖b1)` (and not `(ρ0, ρ‖b‖b1)`) where `b1` is the binding tuple produced from the filter. This intuitively makes sense because of how you can't reference a table's columns in a projection list when there is an aggregation.

So, with the concept of both `Local` and `Outer`, it helps differentiate when you are referencing something from the input record (volcano) vs when things are coming from outer variable environments (stack).

So, for example:
```
SELECT
  (SELECT t1.a + t2.a FROM t2)
FROM t1
```

Will be transformed into:
```
Plan = Select(
  Project(                                                      // Call this Project_1
    projection = Coerce(SubPlan_1)
    input = Scan(t1)
  )
)

SubPlan_1 = Select(
  Project(                                                      // Call this Project_2
    projection = Outer(0,0).a + Local(0).a
    input = Scan(t2)
  )
)
```

We know that, at the moment we are evaluating the projection in Project_1, our variables environment is realistically `<set of variables defined outside of query_1> || <output of scan(t1)>`. But, upon evaluating Project_2's projection, the variables environment would look like: `<set of variables defined outside of query_2> || <output of scan(t2)>` == `<set of variables defined outside of query_1> || <output of scan(t1)> || <output of scan(t2)>`.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -1240,11 +1251,21 @@ internal class PlanTyper(
 
     // HELPERS
 
-    private fun Rel.type(locals: TypeEnv, strategy: Scope = Scope.LOCAL): Rel =
-        RelTyper(locals, strategy).visitRel(this, null)
+    private fun Rel.type(stack: List<TypeEnv>, strategy: Scope = Scope.LOCAL): Rel =
```
RCHowell, line 1254: "scopes" for consistency

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeEnv.kt
```diff
@@ -35,15 +41,21 @@ internal class TypeEnv(public val schema: List<Rel.Binding>) {
...
         }
+        val op = r.op as Rex.Op.Var.Outer
+        if (op.scope == scopeIndex) {
+            r = rex(r.type, rexOpVarLocal(op.ref))
+        }
         // Convert any remaining binding names (tail) to an untyped path expression.
         return if (tail.isEmpty()) r else r.toPath(tail)
     }
 
+    private val scopeIndex: Int = this.outer.size
```
RCHowell, line 57: This is interesting and deserves a comment explaining the calculation. I _almost_ think depth is a better value to use in the plan rather than the scope index because depth is relative to the VarOuter node rather than an absolute index if that makes sense.

Like a local var(0) has an implicit depth of 0, and an outer as a depth relative from 0. For an arbitrary var, what is the scope index? It depends because it's not relative to the variable .. hence the slight preference for depth

RCHowell (changes requested):

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeEnv.kt
johnedquinn: The only thing I don't like about *depth* is having to compute the index at each execution of an *outer*. We'd need to change our implementation of `ExprVarOuter` which currently looks like:
```kotlin
    override fun eval(record: Record): PartiQLValue {
        return env[scope][reference]
    }
```

to:

```kotlin
    override fun eval(record: Record): PartiQLValue {
        val scopeIndex = env.size - depth
        return env[scopeIndex][reference]
    }
```

partiql-eval/src/main/kotlin/org/partiql/eval/internal/Environment.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    internal inline fun <T> scope(record: Record, block: () -> T): T {
+        scopes.push(record)
+        val result = try {
+            block.invoke()
+        } catch (t: Throwable) {
+            scopes.pop()
+            throw t
+        }
+        scopes.pop()
```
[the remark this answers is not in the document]
johnedquinn: Ah, I forgot that Kotlin supported that. Thanks

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeEnv.kt
RCHowell: You shouldn't compute this in eval. This should be computed in the compiler class, then `scope: Int` is a field of ExprVarOuter

RCHowell: Both env.size and depth are known during compilation and are fixed values

partiql-eval/src/main/kotlin/org/partiql/eval/internal/Environment.kt
```diff
@@ -30,9 +30,10 @@ internal class Environment {
     }
 
     /**
-     * Gets the scope/record/variables-environment at the requested [index].
+     * Gets the scope/record/variables-environment at the requested [depth].
      */
-    operator fun get(index: Int): Record {
+    operator fun get(depth: Int): Record {
+        val index = scopes.lastIndex - depth
```
RCHowell, line 35: Scope size is known at compile time so the index can be precomputed in Compiler.kt.


So ExprVarOuter looks like before

```
class ExprVarOuter(
    val scope: Int,
    val ref: Int,
    val env: Env,
)
      // env[scope][ref]
```

And you compute `scope` in Compiler.kt

RCHowell (changes requested):

RCHowell (approved):

PR #1367, opened 2024-02-09
Adds support for aggregations

johnedquinn:
## Relevant Issues
- None

## Description
-  Adds support for aggregations by largely leveraging much of the existing functionality from the PlannerPipeline
- Adds support for COLL_AGGs by leveraging the above aggregations. We take advantage of argument coercion and only expect BAGs as our first parameter. We return ANY in these cases because we don't support parameterized types yet. Therefore, the dynamic case is the only case.
- This PR also takes into consideration the [direct use of grouping expressions in the projection list](https://partiql.org/partiql-lang/#sec:grouping-attributes).
- Fixes the `NormalizeGroupBy` normalization pass.
- An important thing to note is that SPI now has the ability to handle qualified aggregations, though, at the AST to Plan level, we are limiting the aggregations to the SQL:1999 aggregations (MAX, MIN, etc). We can change this whenever we want.
- Fixes dynamic dispatch as well.

## Unresolved Questions
- I've got a question regarding return types for SUM/AVG. I went ahead with what was left before me, but SQL:1999, specifically for SUM/AVG, says that we can choose the precision and scale (with some constraints):
> If SUM or AVG is specified, then:
a) DT shall be a numeric type or an interval type.
b) If SUM is specified and DT is exact numeric with scale S, then the declared type of the result is exact numeric with implementation-defined precision and scale S.
c) If AVG is specified and DT is exact numeric, then the declared type of the result is exact numeric with implementation-defined precision not less than the precision of DT and implementation-defined scale not less than the scale of DT.
d) If DT is approximate numeric, then the declared type of the result is approximate numeric with implementation-defined precision not less than the precision of DT.
e) If DT is interval, then the declared type of the result is interval with the same precision as DT.

## Other Information

## License Information


partiql-planner/src/main/kotlin/org/partiql/planner/internal/Env.kt
```diff
@@ -159,16 +175,107 @@ internal class Env(private val session: PartiQLPlanner.Session) {
...
-        outputCatalogPath: List<String>,
+        userInputPath: BindingPath,
+        pathSentToConnector: BindingPath,
+        actualAbsolutePath: List<String>,
     ): Int {
-        return originalPath.steps.size + outputCatalogPath.size - inputCatalogPath.steps.size
+        return userInputPath.steps.size + actualAbsolutePath.size - pathSentToConnector.steps.size
+    }
+
+    @OptIn(FnExperimental::class, PartiQLValueExperimental::class)
```
johnedquinn, line 217: From here till EOF, this is copied from `PathResolverAgg`.

partiql-spi/src/main/kotlin/org/partiql/spi/connector/sql/builtins/internal/Accumulator.kt
johnedquinn, line 1: Essentially this whole file comes from `org.partiql.lang.eval.physical.operators.Accumulator.kt`

yliuuuu (approved):
LGTM.

PR #1368, opened 2024-02-10
Flatten case when branch result type in PlanTyper

alancai98:
## Description
Typing of a `CASE WHEN` branch does not flatten the type in certain circumstances. Noticed this when I called `SELECT <table>.<struct>.* FROM <table>` and the an argument of the `TUPLEUNION` was an `AnyOfType` containing a single `STRUCT`.

## Other Information
- Updated Unreleased Section in CHANGELOG: **[Yes]**

## License Information

### DIFF
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
@@ -796,7 +796,7 @@ internal class PlanTyper(
                     }
 
                     // Replace the result's type
-                    val type = AnyOfType(ref.type.allTypes.filterIsInstance<StructType>().toSet())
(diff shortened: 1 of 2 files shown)


partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -796,7 +796,7 @@ internal class PlanTyper(
                     }
 
                     // Replace the result's type
-                    val type = AnyOfType(ref.type.allTypes.filterIsInstance<StructType>().toSet())
+                    val type = AnyOfType(ref.type.allTypes.filterIsInstance<StructType>().toSet()).flatten()
```
alancai98, line 799: In the other branch (L767-785), there is a `.flatten()` call on L782, but there is no `.flatten()` call within this branch.

RCHowell (approved):

PR #1370, opened 2024-02-13
Partiql eval continue

yliuuuu:
## Relevant Issues

## Description
- A collection of small fixes
   1. Port the change from #1366
   2. Dynamic function matches logic. 
   3. Comparator support for Big Integer. 
   4. Substring function name. 

## Other Information
  - < If NO, why? >

  - < If YES, why? >

## License Information

### DIFF
--- a/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
+++ b/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
@@ -71,6 +71,7 @@ internal class ExprCallDynamic(
                     val c = coercions[i]
                     when {
                         c == null -> return false
+                        c.input != args[i].type -> return false
                         c.target != types[i] -> return false
                     }
                 }
--- a/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
+++ b/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
@@ -48,9 +48,10 @@ internal class ExprCallDynamic(
      */
     internal class Candidate(
         val fn: Fn,
-        val types: Array<PartiQLValueType>,
         val coercions: Array<Ref.Cast?>
     ) {
+
+        private val signatureParameters = fn.signature.parameters.map { it.type }.toTypedArray()
 
         fun eval(originalArgs: Array<PartiQLValue>): PartiQLValue {
             val args = originalArgs.mapIndexed { i, arg ->
@@ -62,17 +63,27 @@ internal class ExprCallDynamic(
             return fn.invoke(args)
         }
 
-        internal fun matches(args: Array<PartiQLValue>): Boolean {
-            for (i in args.indices) {
-                if (types[i] == PartiQLValueType.ANY) {
-                    return true
-                }
-                if (args[i].type != types[i]) {
-                    val c = coercions[i]
-                    when {
-                        c == null -> return false
-                        c.input != args[i].type -> return false
-                        c.target != types[i] -> return false
+        internal fun matches(inputs: Array<PartiQLValue>): Boolean {
+            for (i in inputs.indices) {
+                val inputType = inputs[i].type
+                val parameterType = signatureParameters[i]
+                val c = coercions[i]
+                when (c) {
+                    // coercion might be null if one of the following is true
+                    // Function parameter is ANY,
+                    // Input type is null
+                    // input type is the same as function parameter
+                    null -> {
+                        if (!(inputType == parameterType || inputType == PartiQLValueType.NULL || parameterType == PartiQLValueType.ANY)) {
+                            return false
+                        }
+                    }
+                    else -> {
+                        // checking the input type is expected by the coercion
+                        if (inputType != c.input) return false
+                        // checking the result is expected by the function signature
+                        // this should branch should never be reached, but leave it here for clarity
+                        if (c.target != parameterType) return false
                     }
                 }
             }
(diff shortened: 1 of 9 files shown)


partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
```diff
@@ -64,7 +64,12 @@ internal class ExprCallDynamic(
                     return true
                 }
                 if (args[i].type != types[i]) {
-                    return false
+                    val c = coercions[i]
+                    when {
+                        c == null -> return false
+                        c.input != args[i].type -> return false
+                        c.target != types[i] -> return false
+                    }
```
RCHowell, line 72: Could you please add a one-line comment to clarify the conditional?

RCHowell (changes requested):

partiql-types/src/main/kotlin/org/partiql/value/util/NumberExtensions.kt
```diff
@@ -53,31 +56,48 @@ private val CONVERSION_MAP = mapOf<Set<Class<*>>, Class<out Number>>(
...
     Long::class.javaObjectType to Number::toLong,
     Float::class.javaObjectType to Number::toFloat,
     Double::class.javaObjectType to Number::toDouble,
+    BigInteger::class.javaObjectType to { num ->
+        when (num) {
+            is Int -> num.toBigInteger()
+            is Long -> num.toBigInteger()
+            is BigInteger -> num
+            else -> throw IllegalArgumentException(
+                "Unsupported number for decimal conversion: $num"
```
johnedquinn, line 89:
```suggestion
                "Unsupported number for BigInteger conversion: $num"
```

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/executor/EvalExecutor.kt
```diff
@@ -84,7 +84,7 @@ class EvalExecutor(
         if (v1.isNull && v2.isNull) {
             return true
         }
-        if (v1 == v2) {
+        if (comparator.compare(v1, v2) == 0) {
```
johnedquinn, line 87: NULL/MISSING according to your other comment would cause problems here, right?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/FnResolver.kt
```diff
@@ -147,7 +149,13 @@ internal object FnResolver {
...
 
     private fun buildArgumentPermutations(args: List<StaticType>): List<List<StaticType>> {
-        val flattenedArgs = args.map { it.flatten().allTypes }
+        val flattenedArgs = args.map {
+            if (it is AnyOfType) {
+                it.flatten().allTypes.filter { it !is NullType }
+            } else {
+                it.flatten().allTypes
+            }
+        }
```
johnedquinn, line 158: What happens when the StaticType is AnyType? Would the following suffice?
```suggestion
        val flattenedArgs = args.map { it.flatten().allTypes }.filter { it !is NullType }
```

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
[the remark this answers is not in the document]
yliuuuu: By bug you mean we can not return true if only one signature parameter if ANY right?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/FnResolver.kt
yliuuuu: Intuitively, if the input args is just a `StaticType.NULL`, then this will return an empty list, which might cause some trouble. 

I will look into this.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/executor/EvalExecutor.kt
yliuuuu: Yes.

We should not use the comparator here, we should have the equals or hash code method implemented in the value class. 

But the comparator is at the moment the only comparator, specifically when it comes to bag comparison. 

I think we can leave this line here but add a TODO.

johnedquinn: Is there a specific scenario where this was failing? I can see that all of the value impls are data classes with the exception of collections and structs.

johnedquinn (changes requested):

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
```diff
@@ -62,13 +63,28 @@ internal class ExprCallDynamic(
...
+                        if (!(inputType == parameterType || inputType == PartiQLValueType.NULL || parameterType == PartiQLValueType.ANY)) {
+                            return false
+                        }
+                    }
+                    else -> {
+                        // checking the input type is expected by the coercion
+                        if (inputType != c.input) return false
+                        // checking the result is expected by the function signature
+                        // this should branch should never be reached, but leave it here for clarity
+                        if (c.target != parameterType) return false
```
yliuuuu, line 86:
```suggestion
                        if (c.target != parameterType) error("Internal Error: Cast Target does not match Function Parameter")
```

yliuuuu: Let's keep the branch here for now, as we are still working towards finalize the type Semantic and cast semantic. 

Let me know if this is sufficient for you.

johnedquinn: Yeah, that works!

johnedquinn (approved):

RCHowell (approved):

PR #1371, opened 2024-02-14
Partiql eval limit offset

yliuuuu:
## Relevant Issues

## Description
- Remodel Limit / Offset

~~If limit is 0 or null, the query should behavior as the same as if there were no limit clause.~~
If offset is 0 or null, the query should behavior as the same as if there were no offset clause. 
If limit and offset clause both exist, the offset should get executed first. 

Update: 
When converting from AST to Plan, append the offset node first then the limit node. 

## Other Information
  - < If NO, why? >

  - < If YES, why? >

## License Information

### DIFF
--- a/partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
+++ b/partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
@@ -10,6 +10,7 @@ import org.partiql.eval.internal.operator.rel.RelJoinLeft
 import org.partiql.eval.internal.operator.rel.RelJoinOuterFull
 import org.partiql.eval.internal.operator.rel.RelJoinRight
 import org.partiql.eval.internal.operator.rel.RelLimit
+import org.partiql.eval.internal.operator.rel.RelOffset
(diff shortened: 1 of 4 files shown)


partiql-planner/src/main/resources/partiql_plan_internal.ion
```diff
@@ -278,10 +278,6 @@ rel::{
     limit::{
       input:  rel,
       limit:  rex,
-    },
-
-    offset::{
-      input:  rel,
       offset: rex,
     },
```
RCHowell, line 282: LIMIT NULL is not the same as LIMIT 0.

I had originally modeled LIMIT..OFFSET as the FETCH operator, because FETCH is defined in the SQL specification whereas LIMIT and OFFSET are not actually defined the SQL spec (92 or 99). 

However, PartiQL allows for OFFSET without LIMIT hence why I split them into their own operators. I don't think we should consolidate, but if that is agreed to be better then it should at least be called FETCH

RCHowell (changes requested):

RCHowell (approved):

PR #1374, opened 2024-02-19
Partiql eval wildcard

yliuuuu:
## Relevant Issues

## Description
- To be reviewed after PR#1365 is merged.
- Adding support for Path Wildcard. 
- This PR does so by rewriting the unresolved plan. 
```
Path Expressions with Wildcards PartiQL also provides multi-step path expressions, called path collection expressions.Their semantics is a generalization of the semantics of a path expression with a single [∗] or .∗. Consider the path collection expression:
ew1p1 . . . wnpn
where e is any expression; n > 0; each wildcard step wi is either [∗] or .∗; each series of plain path steps pi
is a sequence of zero or more tuple path navigations or array navigations (potentially mixed).
Then the path collection expression is equivalent to the SFW query
SELECT VALUE vn.pn
FROM
u1 e AS v1,
u2 @v1.p1 AS v2,
. . .,
un @vn−1.pn−1 AS vn,
where each vi is a fresh variable and each ui is UNPIVOT if wi is a .∗ and it is nothing if wi
is a [∗]. Intuitively vi corresponds to the i-th star.
```

## Other Information
  - < If NO, why? >

  - < If YES, why? >

## License Information


partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
```diff
@@ -249,41 +259,106 @@ internal object RexConverter {
...
+                                rexString(identifier.symbol)
+                            )
+
+                            Identifier.CaseSensitivity.INSENSITIVE -> rexOpPathSymbol(current, identifier.symbol)
+                        }
+                        op
+                    }
+
+                    is Expr.Path.Step.Unpivot -> {
+                        val op = rexOpVarLocal(1, -1)
```
johnedquinn, line 330: While this works, I think it may be subject to breakage, especially if we don't document why `-1` is being used here (I see its use in PlanTyper). Would it just be better to calculate the `ref` here? Seemingly, it's derived from the `fromList` schemas.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeEnv.kt
```diff
@@ -29,6 +29,7 @@ internal data class TypeEnv(
     internal fun getScope(depth: Int): TypeEnv {
         return when (depth) {
             0 -> this
+            // TODO: Consider resolve the variable to the correct index
```
johnedquinn, line 32: We calculate the correct index in `Compiler`. There are a couple reasons that we can discuss.

johnedquinn (changes requested):

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
yliuuuu: The `rexOpVarLocal(1,-1)` is used to indicate that the var is coming from the last element from the immediate outer scope. I think this is the correct modeling in this rewrite. 

From the spec: 

```
Consider the path collection expression:
ew1p1 . . . wnpn
where e is any expression; n > 0; each wildcard step wi is either [∗] or .∗; each series of plain path steps pi is a sequence of zero or more tuple path navigations or array navigations (potentially mixed).

Then the path collection expression is equivalent to : 
SELECT VALUE vn.pn
FROM
u1 e AS v1,
u2 @v1.p1 AS v2,
. . .,
un @vn−1.pn−1 AS vn,

where each vi is a fresh variable and each ui is UNPIVOT if wi is a .∗ and it is nothing if wi is a [∗]. Intuitively vi corresponds to the i-th star.
```

So this (relatively massive fold), would continue folding on non wild card steps, and upon encounter wildcard step, it would rewrite the tip of the fold to be the last var in the immediate outer scope.  (`rexOpVar(1,-1)`)

Once folding is completed, we check if the fromList is empty, if not, then we must rewrite. 
The from then is composite with one or more scan node (in case of more than one scan, they are combined using inner join). 

The local scope of the select construct can see all the bindings produced by each of the scan node, the one that it actually need to use the bonded to be the last one, i.e., `rexOpVar(0,-1)`. 

```
input: a.b.c 
  |_ current: a
     |_ b <- Not a wild card, current a.b 
       |_ c <- not a wild card, current a.b.c 

After fold: return a.b.c as a rex
```

```
input a[*]
   |_ current : a 
      |_ wild card:  add scan(a) to from list, current var(1, -1)

Note in this step, it should be identical to say this is var(1,0).

After fold: the op of path navi would be rexOpVar(1,-1)

We construct the select constructor to essentially rewrite this to be rexOpVar(0,-1), the last var in the local scope. 

That is equivalent to:

SELECT VALUE v1
FROM a as v1
```

```
input a[*].b
   |_ current : a 
      |_ wild card:  add scan(a) to from list, current var(1, -1)
         |_ b : var(1,-1).b

After fold: the op of path navi would be var(1,-1).b

We construct the select constructor to essentially rewrite this to be Var(0,-1).b, the last var in the local scope. 

That is equivalent to:

SELECT VALUE v1.b
FROM a as v1
```

```
input a.*
   |_ current: a 
     |_ unpivot: add unpivot(a) to from list, current var(1,-1)

Note that unpivot is interesting in terms of that it creates two bindings in the scope: key and value, where key is the first and value is the second. we use the -1 to points to second var. So in reality this is the same as var(1,1). 

That is equivalent to:

SELECT VALUE v1
FROM UNPIVOT a as v1
```

```
input a.*b
   |_ current: a 
     |_ unpivot: add unpivot(a) to from list, current var(1,-1)
       |_ b -> current var(1,-1).b

Notice that in this case, the from list contains 1 node, but because the node is unpivot, the number of bindings the select constructs holds locally should be two, So we currently when constructing var(0,-1) in the select constructor, the var would be resolved to var(0, 1) in this case instead of var(0,0). 

That is equivalent to:

SELECT VALUE v1.b
FROM UNPIVOT a as v1
```

yliuuuu: Summary: 
For individual FROM node: we can resolve the ref here, but I choose `-1` to model directly on `the last element of immediate out scope`. 

For select construct: we can also resolve the ref here, but the code is a little more completed as you would have to calculate the number of scan and unpivot to give the current index. This is why I use the -1, I believe it is always going to be the last var.

johnedquinn: We discussed offline, and I'm leaving some details.

The use of `-1` reminded me of [this use of negative indices](https://github.com/partiql/partiql-lang-kotlin/blob/1c7dce11cc9fad417e5ea97e1ec6df0283174334/partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt#L370). Since the contract of using negative indices is potentially weak and since we have access to the input schema, we can calculate the index at this stage.

partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEngineDefaultTest.kt
```diff
@@ -1289,4 +1289,56 @@ class PartiQLEngineDefaultTest {
...
+    }
+  ][*].books[*].title
+            """.trimIndent(),
+            expected = bagValue(listOf(stringValue("A"), stringValue("B"), stringValue("A"), stringValue("E")))
+        ).assert()
+
+    @Test
+    fun wildCard2() {
+        val query = """
+            {'a':1, 'b':2}.*
```
johnedquinn, line 1336: Can you add a query for `<someTuple>.*.a`?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
```diff
@@ -249,41 +259,123 @@ internal object RexConverter {
...
+
+                            Identifier.CaseSensitivity.INSENSITIVE -> rexOpPathSymbol(current, identifier.symbol)
+                        }
+                        op
+                    }
+
+                    is Expr.Path.Step.Unpivot -> {
+                        // Unpivot produces two binding, in this context we want the value,
+                        // which always going to be the second binding
+                        val op = rexOpVarLocal(1, 1)
```
johnedquinn, line 304: Does it matter at all what values you put in here? Same goes for the Wildcard below.

If so, does this work when you have more than two path expressions with wildcards/unpivots? Can you add a test to `PartiQLEngineDefaultTest` for an example like: `<someList>[*].books[*].author[*].name[*].middle_initial` and maybe something similar for unpivots?

yliuuuu: You can check the conformance report details [here](https://github.com/partiql/partiql-lang-kotlin/actions/runs/8056990189).

I think the conformance tests should be sufficient. But I can also port the tests to `PartiQLEngineDefaultTest`

johnedquinn: > If so, does this work when you have more than two path expressions with wildcards/unpivots?

This doesn't work when you have more than two path expressions. See this commit: https://github.com/partiql/partiql-lang-kotlin/commit/e8e31d90d161ec6dcc1758e9c134f6f1ce6118c6.

The example above doesn't compile.

johnedquinn: It has to do with the `1, 1` and `1, 0` you've used for `rexOpVarLocal` for unpivots and wildcards. Those only work for the test cases you've written. Since JOINs really concatenate the LHS schema with the RHS schema, you can't always assume that the item you'll be referencing will always be at some specific index. It needs to be calculated.

johnedquinn (changes requested):
See my other comments.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
yliuuuu: Oh I see what you mean. You are right.

johnedquinn (approved):
LGTM!

PR #1375, opened 2024-02-21
Additional Functions implementation for Eval Engine

yliuuuu:
## Relevant Issues

## Description
- exists
- cardinality
- bit_length
- octet_length
- Extract.
- overlay (rewrite)

## Other Information
  - < If NO, why? >

  - < If YES, why? >

## License Information


partiql-spi/src/main/kotlin/org/partiql/spi/connector/sql/builtins/FnExtract.kt
```diff
@@ -0,0 +1,368 @@
...
+            FnParameter("datetime", TIME),
+        ),
+        isNullCall = true,
+        isNullable = false,
+    )
+
+    override fun invoke(args: Array<PartiQLValue>): PartiQLValue {
+        val v = args[0].check<TimeValue>().value!!
+        return when (val tz = v.timeZone) {
+            TimeZone.UnknownTimeZone -> int32Value(0)
```
johnedquinn, line 291: QQ: You're the one most familiar with date time semantics, but should this be 0? While it's not a boolean value, in boolean logic, an unknown value gets coerced to a NULL value.

The Ion spec says:
> 2007-02-23T20:14:33.079Z         // The same instant in UTC ("zero" or "zulu")
2007-02-23T20:14:33.079+00:00    // The same instant, with explicit local offset
2007-02-23T20:14:33.079-00:00    // The same instant, with unknown local offset

Which shows that the timestamps themselves are equivalent. But, if you were to want to derive information from your data, I feel as if NULL would be returned upon extracting an unknown offset.

partiql-spi/src/main/kotlin/org/partiql/spi/connector/sql/exts/FnExists.kt
```diff
@@ -0,0 +1,107 @@
...
+        isNullable = false,
+    )
+
+    override fun invoke(args: Array<PartiQLValue>): PartiQLValue {
+        val container = args[0].check<BagValue<*>>()
+        val iter = container.iterator()
+        if (iter.hasNext()) {
+            return boolValue(true)
+        }
+        return boolValue(false)
```
johnedquinn, line 36:
```suggestion
        val exists = container.iterator().hasNext()
        return boolValue(exists)
```

Nit -- non-blocking, but just a recommendation

partiql-spi/src/main/kotlin/org/partiql/spi/connector/sql/exts/FnSize.kt
```diff
@@ -0,0 +1,115 @@
...
+        parameters = listOf(
+            FnParameter("container", PartiQLValueType.BAG),
+        ),
+        isNullCall = true,
+        isNullable = false,
+    )
+
+    override fun invoke(args: Array<PartiQLValue>): PartiQLValue {
+        val container = args[0].check<BagValue<*>>()
+        val iter = container.iterator()
```
johnedquinn, line 32: Same recommendations as CARDINALITY.

johnedquinn (commented):
Overall looks great. Some minor comments.

partiql-spi/src/main/kotlin/org/partiql/spi/connector/sql/builtins/FnExtract.kt
yliuuuu: This is a good question. 

While I don't know what should the decision be, for the time being I decided to follow the behavior from the existing pipelines, which returns zero (in form of decimal for some reasons)...

```
EXTRACT(TIMEZONE_HOUR FROM `2024-01-01T23:12:59-00:00`)
   | 
===' 
0.
```

I think for the time being we can stay with that behavior, but I will add a todo note for this line.

Great call out. Thanks.

johnedquinn (approved):
LGTM!

PR #1382, opened 2024-03-05
Add additional APIs for async physical plan evaluation

alancai98:
## Description
Creates an asynchronous version of the existing physical plan evaluator APIs. This PR differs from the other attempt to make the physical plan evaluator async (https://github.com/partiql/partiql-lang-kotlin/compare/main...plan-eval-async-make-statement-eval-async), which had **changed** the existing synchronous APIs to be async. Performance-wise, we see about a 10-20% drop in performance when using the async APIs on a single query.

This PR chooses to have both APIs to be compatible with semver. Due to the performance drop, we choose to not just wrap the async evaluator w/ a `runBlocking` call. Including both versions also makes it easier to test the synchronous and asynchronous API performance more easily. The previous synchronous versions have been marked as deprecated and are expected to be removed in the next major version.

For reviewers, I recommend starting to look at
1. [AsyncOperatorTests.kt](https://github.com/partiql/partiql-lang-kotlin/pull/1382/files/4a1a79c5b2d1c5c59e019322dfe4328a9859850a#diff-d5434bdafbfc181054b7cfe03f0ec30e2fd96b28603276cfea815a7a13e4fde2) -- shows the use case for the async evaluator
2. [PartiQLCompilerAsync.kt](https://github.com/partiql/partiql-lang-kotlin/pull/1382/files/4a1a79c5b2d1c5c59e019322dfe4328a9859850a#diff-3c8489d73a91e103ecbc89990f0d3c7337ab6df512bb0860dd85fc55b90fac53) -- shows the public API for calling the async evaluator

The remaining changes were pretty straightforward (other than what's noted in the self-review comments). Essentially they were
- Creating an async version (usually w/ `Async` added to the end of the previous class/interface)
- Making the functions within the classes/interfaces `suspend fun`

## Other Information

  - No. Previous synchronized physical plan evaluator APIs are the same but have been marked as deprecated. In a future major version, all of these deprecated APIs will be removed.

  - Kotlin co-routine libraries to enable async
    - partiql-lang
      - implementation dep on org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.0
      - test dep on org.jetbrains.kotlinx:kotlinx-coroutines-test:1.6.0
    - partiql-examples
      - implementation dep on org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.0
      - implementation dep on org.jetbrains.kotlinx:kotlinx-coroutines-jdk8:1.6.0 (for converting coroutine to Java's `CompletableFuture`)
      - test dep on org.jetbrains.kotlinx:kotlinx-coroutines-test:1.6.0
    - partiql-cli
      - implementation dep on org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.0

## License Information

### DIFF
--- a/examples/src/main/kotlin/org/partiql/examples/PartiQLCompilerPipelineExample.kt
+++ b/examples/src/main/kotlin/org/partiql/examples/PartiQLCompilerPipelineExample.kt
@@ -1,6 +1,7 @@
 package org.partiql.examples
 
 import com.amazon.ion.system.IonSystemBuilder
+import kotlinx.coroutines.runBlocking
 import org.partiql.annotations.ExperimentalPartiQLCompilerPipeline
 import org.partiql.examples.util.Example
 import org.partiql.lang.compiler.PartiQLCompilerPipelineAsync
@@ -13,7 +14,6 @@ import org.partiql.lang.planner.EvaluatorOptions
 import org.partiql.lang.planner.GlobalResolutionResult
 import org.partiql.lang.planner.GlobalVariableResolver
 import java.io.PrintStream
-import kotlinx.coroutines.runBlocking
 
 /**
  * This example demonstrate how to use PartiQLCompilerPipeline. This feature is currently in experimental stage.
--- a/examples/src/main/kotlin/org/partiql/examples/PartiQLCompilerPipelineExample.kt
+++ b/examples/src/main/kotlin/org/partiql/examples/PartiQLCompilerPipelineExample.kt
@@ -3,7 +3,7 @@ package org.partiql.examples
 import com.amazon.ion.system.IonSystemBuilder
 import org.partiql.annotations.ExperimentalPartiQLCompilerPipeline
 import org.partiql.examples.util.Example
-import org.partiql.lang.compiler.PartiQLCompilerPipeline
+import org.partiql.lang.compiler.PartiQLCompilerPipelineAsync
 import org.partiql.lang.eval.Bindings
 import org.partiql.lang.eval.EvaluationSession
 import org.partiql.lang.eval.ExprValue
@@ -13,6 +13,7 @@ import org.partiql.lang.planner.EvaluatorOptions
 import org.partiql.lang.planner.GlobalResolutionResult
 import org.partiql.lang.planner.GlobalVariableResolver
 import java.io.PrintStream
+import kotlinx.coroutines.runBlocking
 
 /**
  * This example demonstrate how to use PartiQLCompilerPipeline. This feature is currently in experimental stage.
@@ -59,7 +60,7 @@ class PartiQLCompilerPipelineExample(out: PrintStream) : Example(out) {
         .build()
 
     @OptIn(ExperimentalPartiQLCompilerPipeline::class)
-    private val partiQLCompilerPipeline = PartiQLCompilerPipeline.build {
+    private val partiQLCompilerPipeline = PartiQLCompilerPipelineAsync.build {
         planner
             .globalVariableResolver(globalVariableResolver)
         compiler
@@ -71,7 +72,10 @@ class PartiQLCompilerPipelineExample(out: PrintStream) : Example(out) {
 
         print("PartiQL query:", query)
         @OptIn(ExperimentalPartiQLCompilerPipeline::class)
-        val exprValue = when (val result = partiQLCompilerPipeline.compile(query).eval(session)) {
+        val result = runBlocking {
+            partiQLCompilerPipeline.compile(query).eval(session)
+        }
+        val exprValue = when (result) {
             is PartiQLResult.Value -> result.value
             is PartiQLResult.Delete,
             is PartiQLResult.Explain.Domain,
--- a/partiql-lang/src/main/kotlin/org/partiql/lang/eval/ThunkAsync.kt
+++ b/partiql-lang/src/main/kotlin/org/partiql/lang/eval/ThunkAsync.kt
@@ -16,8 +16,6 @@ package org.partiql.lang.eval
 
 import com.amazon.ionelement.api.MetaContainer
 import kotlinx.coroutines.flow.flow
-import kotlinx.coroutines.flow.forEach
-import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.toList
 import org.partiql.errors.ErrorBehaviorInPermissiveMode
 import org.partiql.errors.ErrorCode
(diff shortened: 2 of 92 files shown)


partiql-lang/src/main/kotlin/org/partiql/lang/eval/physical/PhysicalPlanCompilerAsyncImpl.kt
```diff
@@ -0,0 +1,1908 @@
...
+            false -> RelationType.BAG
+        }
+
+        return thunkFactory.thunkEnvAsync(expr.metas) { env ->
+            // we create a snapshot for currentRegister to use during the evaluation
+            // this is to avoid issue when iterator planner result
+            val currentRegister = env.registers.clone()
+            val elements: Flow<ExprValue> = flow {
+                env.load(currentRegister)
+                val relItr = bexprThunk(env)
```
alancai98, line 324: This thunk evaluation is an async call. In the synchronous code, this call was within a sequence -- https://github.com/partiql/partiql-lang-kotlin/blob/plan-eval-async-keep-sync/partiql-lang/src/main/kotlin/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt#L335-L341. But since the evaluation is an async call, we need to use a coroutine `Flow` (similar to a sequence), which allows for an async call within the block.

More on `Flow`s can be found here: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/.

partiql-lang/src/test/kotlin/org/partiql/lang/compiler/IntegrationTests.kt
```diff
@@ -30,15 +32,24 @@ class TestContext {
         assertEquals(expectedIon, result.toIonValue(ION))
     }
 
+    // Executes query on async evaluator
```
alancai98, line 35: Tests in partiql-lang that tested the synchronous physical plan evaluator will also test the async physical plan evaluator. I tried to limit the amount of copied code, where possible, while also making it easy to remove the synchronous tests when we remove the synchronous evaluator in an upcoming major version.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/physical/PhysicalBexprToThunkConverterAsync.kt
```diff
@@ -0,0 +1,362 @@
...
+import org.partiql.lang.eval.physical.operators.SortOperatorFactoryAsync
+import org.partiql.lang.eval.physical.operators.UnpivotOperatorFactoryAsync
+import org.partiql.lang.eval.physical.operators.WindowRelationalOperatorFactoryAsync
+import org.partiql.lang.eval.physical.operators.valueExpressionAsync
+import org.partiql.lang.eval.physical.window.createBuiltinWindowFunctionAsync
+import org.partiql.lang.util.toIntExact
+
+/** Converts instances of [PartiqlPhysical.Bexpr] to any [T]. A `suspend` version of the physical plan converter
+ * interface is added since PIG currently does not output async functions.
+ */
```
alancai98, line 37: The physical plan to async thunk converter implements this `Converter`, which is the same as the PIG-generated `Converter` but the functions within the interface are all async (i.e. `suspend fun`s). We could consider adding some configurability to PIG to support generating async versions.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/ThunkAsync.kt
```diff
@@ -0,0 +1,613 @@
...
+
+/**
+ * A thunk with no parameters other than the current environment.
+ *
+ * See https://en.wikipedia.org/wiki/Thunk
+ *
+ * @param TEnv The type of the environment.  Generic so that the legacy AST compiler and the new compiler may use
+ * different types here.
+ */
+internal typealias ThunkAsync<TEnv> = suspend (TEnv) -> ExprValue
```
alancai98, line 36: File is basically the same as `Thunk.kt`. Notable changes include
- Making any functions/lambda/types async (i.e. adding `suspend`)
- Converting sequence constructors that make calls to async functions use coroutine `Flow`s

dlurton (commented):
Only concern I have is about the `.toList` calls in the thunk returned from `compileBindingsToValues`.  We need to decide if we should try to avoid the eagerness now or if I should file a feature request later.

examples/src/main/kotlin/org/partiql/examples/PartiQLCompilerPipelineExample.kt
```diff
@@ -1,9 +1,10 @@
 package org.partiql.examples
```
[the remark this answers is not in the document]
alancai98: Thanks for the catch. Class and file should include the `Async` suffix. I will change it.

---

> Also, what's the plan when the sync pipeline is removed? Will the new classes keep their Async suffix or will the suffix be removed?

The new classes will keep the `Async` suffix.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/physical/PhysicalPlanCompilerAsyncImpl.kt
```diff
@@ -0,0 +1,1908 @@
...
+                while (relItr.nextRow()) {
+                    emit(mapThunk(env))
+                }
+            }
+            when (relationType) {
+                RelationType.LIST -> ExprValue.newList(elements.toList())
+                RelationType.BAG -> ExprValue.newBag(elements.toList())
```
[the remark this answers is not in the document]
alancai98: Discussed offline and including a summary of the discussion. To summarize, making the async evaluator lazy would necessitate a larger refactor of the evaluator. We will keep the async evaluator as it is and revisit if we find we need the previous, lazy materialization behavior.

For more context, we use `Flow`s in this block (as well as other parts of the async evaluator) rather than `Sequence`s because code within a `Flow` block can call suspend functions. `Flow`s act somewhat differently than sequences in that to get the computed value back from a `Flow`, you need to "collect" the flow using a function like [`Flow.collect`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/collect.html), which appears to eagerly get all items. From the docs on [`Flow`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/):
> Terminal operators on the flow are either suspending functions such as [collect](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/collect.html), [single](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/single.html), [reduce](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/reduce.html), [toList](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/index.html), etc. or [launchIn](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/launch-in.html) operator that starts collection of the flow in the given scope. They are applied to the upstream flow and trigger execution of all operations. Execution of the flow is also called collecting the flow and is always performed in a suspending manner without actual blocking. Terminal operators complete normally or exceptionally depending on successful or failed execution of all the flow operations in the upstream.

Making an `ExprValue` that accepts a `Flow` rather than a list/sequence will not work either. `ExprValue` implements `Iterable<ExprValue>`, which necessitates overriding `iterator()`. Since `iterator()` is synchronous and calls to `Flow.collect()` (or similar terminal operators) are async, this approach is not suitable.

One possible solution could be to make `PartiQLResult.Value`'s parameter `value` a `Flow<ExprValue>` rather than an `ExprValue`. This could ensure the materialization is left to the consuming application when they call `.collect` or `.toList()` rather than within the async evaluator. But this change would require a larger refactor and possible duplication of `PartiQLResult`.

dlurton (approved):

partiql-lang/src/main/kotlin/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
```diff
@@ -1886,14 +1885,6 @@ internal class PhysicalPlanCompilerImpl(
         )
 }
 
-internal val MetaContainer.sourceLocationMeta get() = this[SourceLocationMeta.TAG] as? SourceLocationMeta
-internal val MetaContainer.sourceLocationMetaOrUnknown get() = this.sourceLocationMeta ?: UNKNOWN_SOURCE_LOCATION
-
-internal fun StaticType.getTypes() = when (val flattened = this.flatten()) {
-    is AnyOfType -> flattened.types
-    else -> listOf(this)
-}
```
alancai98, line 1895: When I remove these top-level internal functions, this is a breaking change from Java. From the Java to Kotlin interop [docs](https://kotlinlang.org/docs/java-to-kotlin-interop.html#visibility):

> `internal` declarations become `public` in Java. Members of `internal` classes go through name mangling, to make it harder to accidentally use them from Java and to allow overloading for members with the same signature that don't see each other according to Kotlin rules

So from Java code, `internal` classes will have some name mangling. But for top-level `internal` functions, there is no name mangling. For instance, `StaticType.getTypes()` is directly callable from Java by: 
```
PhysicalPlanCompilerAsyncImplKt.getTypes(StaticType.ANY);
```

IntelliJ will give an error like "Usage of Kotlin internal declaration from different module " but these are just errors in the IDE that can be disabled. Gradle and the Java compiler can still run the internal Kotlin code.

I'll cut an issue to track if we should allow for these top-level functions (both `internal` and `public`).

alancai98:
Ran the Java API compliance checker between this PR and v0.14.3. The only breaking changes were the removal of the internal functions from `PhysicalPlanCompilerImpl.kt`, as pointed out in https://github.com/partiql/partiql-lang-kotlin/pull/1382#discussion_r1522281665. I'll cut an issue and discuss w/ team regarding what we should do w/ these internal top-level functions.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/physical/PhysicalPlanCompilerImpl.kt
alancai98: GitHub issue: https://github.com/partiql/partiql-lang-kotlin/issues/1387

PR #1383, opened 2024-03-06
Fixes aggregations of attribute references to values of union types

johnedquinn:
## Relevant Issues
- None

## Description
- Fixes aggregations of attribute references to values of union types
- Before, in `TypeEnv`, we would assert, when looking up an attribute on a value, that the value were a `StructType` -- even though it could be a union of StructTypes. Or a union of other types. This PR handles this scenario.
- This PR does NOT handle heterogeneous data for aggregation functions. See the new, disabled test in PlanTyperTestsPorted.
- For the purposes of better error reporting, modified the definition of UndefinedVariable. See the CHANGELOG.

## Open Questions (Not addressed in this PR)
- What should `AVG` return? I've pasted a snippet from SQL:1999 which shows that we are conformant, however, I'm not sold.

## Other Information
  - Yes, but not API incompatible changes. See the CHANGELOG.

## License Information

### DIFF
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/Errors.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/Errors.kt
@@ -3,6 +3,7 @@ package org.partiql.planner
 import org.partiql.errors.ProblemDetails
 import org.partiql.errors.ProblemSeverity
 import org.partiql.plan.Identifier
+import org.partiql.planner.internal.utils.PlanUtils
 import org.partiql.types.StaticType
 
 /**
@@ -31,7 +32,8 @@ public sealed class PlanningProblemDetails(
     ) : PlanningProblemDetails(
         ProblemSeverity.ERROR,
         {
-            "Variable ${pretty(name)} does not exist in the database environment and is not an attribute of the following in-scope variables $inScopeVariables." +
+            val humanReadableName = PlanUtils.identifierToString(name)
+            "Variable $humanReadableName does not exist in the database environment and is not an attribute of the following in-scope variables $inScopeVariables." +
                 quotationHint(isSymbolAndCaseSensitive(name))
         }
     ) {
@@ -74,16 +76,6 @@ public sealed class PlanningProblemDetails(
             private fun isSymbolAndCaseSensitive(id: Identifier): Boolean = when (id) {
                 is Identifier.Symbol -> id.caseSensitivity == Identifier.CaseSensitivity.SENSITIVE
                 is Identifier.Qualified -> false
-            }
-
-            private fun pretty(id: Identifier): String = when (id) {
-                is Identifier.Symbol -> pretty(id)
-                is Identifier.Qualified -> (listOf(id.root) + id.steps).joinToString(".") { pretty(it) }
-            }
-
-            private fun pretty(id: Identifier.Symbol): String = when (id.caseSensitivity) {
-                Identifier.CaseSensitivity.INSENSITIVE -> id.symbol
-                Identifier.CaseSensitivity.SENSITIVE -> "\"${id.symbol}\""
             }
         }
     }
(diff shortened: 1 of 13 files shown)


partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -1218,8 +1218,8 @@ internal class PlanTyper(
         fun resolveAgg(agg: Agg.Unresolved, arguments: List<Rex>): Pair<Rel.Op.Aggregate.Call, StaticType> {
             var missingArg = false
             val args = arguments.map {
-                val arg = visitRex(it, null)
-                if (arg.type.isMissable()) missingArg = true
+                val arg = visitRex(it, it.type)
+                if (arg.type is MissingType) missingArg = true
```
RCHowell, line 1222: I'm wondering why the `missingArg` matters now. Aggregates filter out all null/missing afaik so we kind of ignore them in typing?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -1349,15 +1349,42 @@ internal class PlanTyper(
...
 
+    private fun Identifier.Symbol.toPlan(): org.partiql.plan.Identifier.Symbol = org.partiql.plan.Identifier.Symbol(
+        this.symbol,
+        this.caseSensitivity.toPlan()
+    )
+
+    private fun Identifier.Qualified.toPlan(): org.partiql.plan.Identifier.Qualified = org.partiql.plan.Identifier.Qualified(
+        this.root.toPlan(),
+        this.steps.map { it.toPlan() }
+    )
```
RCHowell, line 1386: I've seen these functions and written some myself all over the place. Perhaps we can get some identifier utilities in their own internal object?

All the .toString() .. toNormalizedString() .. etc.?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeEnv.kt
```diff
@@ -152,4 +152,39 @@ internal class TypeEnv(public val schema: List<Rel.Binding>) {
...
+     * @param name
+     * @return
+     */
+    private fun StaticType.containsKey(name: BindingName): Boolean? {
+        return when (val type = this.flatten()) {
+            is StructType -> type.containsKey(name)
+            is AnyOfType -> {
+                val anyKnownToContainKey = type.allTypes.any { it.containsKey(name) == true }
+                val anyKnownToNotContainKey = type.allTypes.any { it.containsKey(name) == false }
+                val anyNotKnownToContainKey = type.allTypes.any { it.containsKey(name) == null }
```
RCHowell, line 173: reduce to a single loop

```kotlin
var anyKnownToContainKey = false
var anyKnownToNotContainKey = false
var anyNotKnownToContainKey = false
for (t in type.allTypes) {
   val containsKey = it.containsKey(name)
   anyKnownToContainKey = anyKnownToContainKey || (containsKey == true)
   anyKnownToNotContainKey = anyKnownToNotContainKey || (containsKey == false)
   anyNotKnownToContainKey = anyNotKnownToContainKey || (containsKey == null)
}
```

RCHowell (changes requested):

partiql-planner/src/main/kotlin/org/partiql/planner/Errors.kt
```diff
@@ -94,6 +149,14 @@ public sealed class PlanningProblemDetails(
         "Unknown function `$identifier($types)"
     })
 
+    public data class UnknownAggregateFunction(
+        val identifier: String,
```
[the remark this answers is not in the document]
johnedquinn: Good call out. Updated.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
johnedquinn: Yeah, that makes sense to me. Removed.

RCHowell (approved):

PR #1385, opened 2024-03-11
PartiQL Eval - Planner Mode

yliuuuu:
## Relevant Issues

====== Update ========
The previous approach have the error/missing node to be a wrapper around the original problematic node.

Based on the discussion, we don't want to increase the public api surface.

So in this PR, we change the error/missing node to contains a message and a trace, which is a list of rex.op. We replace the original problematic node with the error/missing node, and add their children to the trace.

=========================

## Description
- Add the ability to control over Missing Behavior during planning.
- Consolidate Problem reporting mechanism. 
- Note: The error reporting on `REL` operation were not complete, specifically, the aggregation and exclude operation. 

The gist of the PR was quite straightforward. 
In the internal Plan: 
```
rex::{
  op:[
    ....
    err::{
      input: '.rex.op',
      problem: problem,
    },

    missing::{
      input: '.rex.op',
      problem: problem,
    },
  ]
}
```

During typing, if we find a type error / an operation that always returns missing -> we put the input rexOp underneath `rexOpMissing`. 

If on the other hand, the operation is an error, i.e., `not_a_function(1)`, we put the input rexOp underneath the `rexOpErr`. 

During the conversion between the internal plan to the public plan, we added a flag to control the planner behavior, and add all problem in the problem callback. 

At the moment: the choice we have are `QUIET` vs `SIGNAL`. 
In `QUIET` mode: We turns the top most RexOpMissing in to a literal missing. 
In `SIGNAL` mode: We turns the top most RexOpMissing into a ErrorNode. 

During the process, we still visit the children of the missing node to gather the problems. That said, in some complex queries, it would be nice to trace the missing ops in the plan for the purpose of debugging. 

To do so, this PR also included a debugging planner. 

If using a debugging compiler: 
In `QUIET` mode: upon encoutner a `RexOpMissing` node, instead of turning it into a missing literal, we turn it into a `RexOpDebugProblemMissing`. The `RexOpDebugProblemMissing` node is a new node we added in that mirrors the internal `RexOpMissing` node. Similar node was added for `SIGNAL` mode. 

[partiql-planner/src/test/kotlin/org/partiql/planner/PlannerErrorReportingTests.kt](https://github.com/partiql/partiql-lang-kotlin/pull/1385/files#diff-8f6f4f6bb347bdef0ebff5c7fd886df952d8c47661f369ef9cf217ce26709a64) can be a good entry point to this review. 

## Other Information
  - < If NO, why? >

  - < If YES, why? >

## License Information


buildSrc/src/main/kotlin/partiql.conventions.gradle.kts
```diff
@@ -50,6 +50,7 @@ java {
 tasks.test {
     useJUnitPlatform() // Enable JUnit5
     jvmArgs.addAll(listOf("-Duser.language=en", "-Duser.country=US"))
+    jvmArgs.add("-Djunit.jupiter.execution.timeout.mode=disabled_on_debug")
```
yliuuuu, line 53: Disable timeout on debug mode for all junit5 tests.

partiql-planner/src/main/kotlin/org/partiql/planner/PartiQLPlannerDefault.kt
```diff
@@ -31,12 +31,12 @@ internal class PartiQLPlannerDefault(
...
 
         // 3. Resolve variables
-        val typer = PlanTyper(env, onProblem)
+        val typer = PlanTyper(env)
         val typed = typer.resolve(root)
         val internal = org.partiql.planner.internal.ir.PartiQLPlan(typed)
 
         // 4. Assert plan has been resolved — translating to public API
-        var plan = PlanTransform.transform(internal, onProblem)
+        var plan = PlanTransform.transform(internal, session.missingOpBehavior, onProblem, false)
```
RCHowell, line 39: here, session.flags

partiql-planner/src/main/kotlin/org/partiql/planner/internal/PathResolver.kt
```diff
@@ -61,15 +61,15 @@ internal abstract class PathResolver<T>(
             when (n) {
                 0 -> return null
                 1 -> return get(absPath)
-                2 -> return get(absPath) ?: get(path)
+                2 -> return get(absPath) ?: get(path) ?: search(path)
```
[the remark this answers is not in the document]
RCHowell: I don't think this is correct. If N =2 we only check absPath then path. I think absPath and path should be given better names IMO.

Also, now N=2 and N=3 are the same

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/executor/EvalExecutor.kt
```diff
@@ -40,6 +41,11 @@ class EvalExecutor(
     override fun prepare(statement: String): PartiQLStatement<*> {
         val stmt = parser.parse(statement).root
         val plan = planner.plan(stmt, plannerSession)
+        println(
+            buildString {
+                PlanPrinter.append(this, plan.plan)
+            }
+        )
```
RCHowell, line 48: Remove this? Maybe print the plan on a failure

RCHowell (changes requested):

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelOffset.kt
```diff
@@ -33,11 +33,12 @@ internal class RelOffset(
...
     override fun hasNext(): Boolean {
         if (!init) {
-            for (record in input) {
+            while (input.hasNext()) {
                 if (_seen >= _offset) {
                     break
                 }
                 _seen = _seen.add(BigInteger.ONE)
+                input.next()
             }
```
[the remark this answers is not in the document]
yliuuuu: We switched to iterator model for record. 

Previously: 
```
for (record in input) {
     if (_seen >= _offset) {
             break
      }
     _seen = _seen.add(BigInteger.ONE)
} 
```
Which is equivalent to: 
```
while(input.hasNext()) {
   // in the last iteration before break, this accidentally move the pointer one position further.  
    val record = input.next()
    if (_seen >= _offset) {
             break
      }
    _seen = _seen.add(BigInteger.ONE)
}
```

See the decompiled java code for complete reference: 

Before: 
```
public boolean hasNext() {
      if (!this.init) {
         Iterator var3 = (Iterator)this.input;

         BigInteger var10001;
         for(Iterator var2 = var3; var2.hasNext(); this._seen = var10001) {
            Record record = (Record)var2.next();
            if (this._seen.compareTo(this._offset) >= 0) {
               break;
            }

            var10001 = this._seen.add(BigInteger.ONE);
            Intrinsics.checkNotNullExpressionValue(var10001, "_seen.add(BigInteger.ONE)");
         }

         this.init = true;
      }

      return this.input.hasNext();
   }
```

After
```
  public boolean hasNext() {
      if (!this.init) {
         while(true) {
            if (!this.input.hasNext() || this._seen.compareTo(this._offset) >= 0) {
               this.init = true;
               break;
            }

            BigInteger var10001 = this._seen.add(BigInteger.ONE);
            Intrinsics.checkNotNullExpressionValue(var10001, "_seen.add(BigInteger.ONE)");
            this._seen = var10001;
            this.input.next();
         }
      }

      return this.input.hasNext();
   }
```

partiql-plan/src/main/resources/partiql_plan.ion
```diff
@@ -197,6 +197,26 @@ rex::{
...
+        missing::{
+          trace: '.rex.op',
+          problem: string
+        },
+        error::{
+          trace: '.rex.op',
+          problem: string
+        },
+      ]
+    ]
```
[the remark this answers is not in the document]
yliuuuu: Essentially: 

Consider the query `1 + t.f2` where t is `t : {f1: INT2}`

The design was to have a trace of missing: 

```
Query[tag=Plan-780ab923]
      └── Missing[problem=-1:-1:<unknown>: int4, missing is/are incompatible data types for the 'PLUS' operator., tag=Plan-39c9df92, static_type=missing]
         └── Dynamic[tag=Plan-7452fb19]
            ├── Lit[value=Int32ValueImpl(value=1, annotations=[]), tag=Plan-6063c078, static_type=int4]
            ├── Missing[problem=-1:-1:<unknown>: Expression always returns null or missing: caused by Path Navigation always returns MISSING, tag=Plan-3defad5b, static_type=missing]
            │  └── Symbol[key=f2, tag=Plan-ccbc991]
            │     └── Global[tag=Plan-e021a9f7, static_type=struct(f1: int2, [Open(value=false), UniqueAttrs(value=true), Ordered])]
            │        └── Ref[catalog=0, symbol=8, tag=Plan-cfc1a326]
            ├── Candidate[tag=Plan-7e3c9eab]
            │  └── Ref[catalog=0, symbol=0, tag=Plan-6263a65a]
            | ....... 
```

Breaking it down a little bit: 
```
Query[tag=Plan-780ab923]
      └── Missing[problem=-1:-1:<unknown>: int4, missing is/are incompatible data types for the 'PLUS' operator., tag=Plan-39c9df92, static_type=missing]
```

Top level missing -> because 1 + MISSING is an unresolved function. 
Note that the idea is to have 1 + MISSING report type mismatch error, NOT plus function propagate missing. 
This is because the `isMissingCall` flag is set on each FnSignature.  
During function resolution, we first get all the variants based on function name. We never prohibit the possibility to have say two variants of the same function name, but has different `isMissingCall` flag. 

Next: 
```
└── Missing[problem=-1:-1:<unknown>: int4, missing is/are incompatible data types for the 'PLUS' operator., tag=Plan-39c9df92, static_type=missing]
         └── Dynamic[tag=Plan-7452fb19]
```

The RexOpMissing should have put all the candidate function to help debugging. This is essentially to show the user what is the actually input type and all the possible legal static type. 

Next: 
```
└── Dynamic[tag=Plan-7452fb19]
            ├── Lit[value=Int32ValueImpl(value=1, annotations=[]), tag=Plan-6063c078, static_type=int4]
            ├── Missing[problem=-1:-1:<unknown>: Expression always returns null or missing: caused by Path Navigation always returns MISSING, tag=Plan-3defad5b, static_type=missing]
            │  └── Symbol[key=f2, tag=Plan-ccbc991]
            │     └── Global[tag=Plan-e021a9f7, static_type=struct(f1: int2, [Open(value=false), 
```
The interesting part is the second arg, which contains another rexOpMissing. The idea is to have the trace (what caused this missing), again for debugging purpose. 

====

Now those trace are nice for debugging, but the evaluator does not really care about those. 
So instead of having the evaluator to deal with rexOpMissing, it could be nice, before we send the plan to evaluator, does the conversion ourselves. 

```
// Plan for evaluator to consume
└── Query[tag=Plan-951622ec]
      └── Lit[value=MissingValueImpl(annotations=[]), tag=Plan-33b21fcd, static_type=missing]
```

Hence we have essentially two planner build for that purpose. 

We can put all those information to Rex.Op.MISSING in the public plan, but I having those as debugging node avoids backward incompatible changes.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/PathResolver.kt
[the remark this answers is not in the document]
yliuuuu: This is where I could use a little help. 

Could you please elaborate the modeling here?  

Case 1: 
   inserting an obj (`t`) into catalog (`test`). 
   session does not specify current directory. 
   search for `t` 

In this case, the resolution succeeds.  
This is expected as it should search for <catalog>.<path>

Case 2: 
   inserting an obj (`t`) into catalog (`test`). 
   session does not specify current directory. 
   search for `test.t` 

In this case, the resolution failed.   

This is unexpected as it should search for <catalog>.<path>

Case 3: 
   inserting an obj (`t`) into catalog (`test`). 
   session specify current directory to be `dir`
   search for `t` 

In this case, the resolution failed. 

Not sure if this is expected. Based on the comment, it says it should search for <catalog>.<schema>.<path>. But why this should not search for <catalog>.<path> ? 

Case 4: 
   inserting an obj (`t`) into catalog (`test`). 
   session specify current directory to be `dir`
   search for `test.t` 

In this case, the resolution failed. 

This is unexpected as it should search for <catalog>.<path>.

partiql-plan/src/main/resources/partiql_plan.ion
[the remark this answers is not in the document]
RCHowell: I can see how attaching this information is useful, but why is this **debug** in the public API plan? This makes sense for the internal plan, and we can dump these detailed messages during the PlanTransform.

I don't believe we should be increasing the public plan surface with debug information — error node is already a stretch..

partiql-planner/src/main/kotlin/org/partiql/planner/internal/Env.kt
```diff
@@ -88,14 +88,28 @@ internal class Env(private val session: PartiQLPlanner.Session) {
     }
 
     @OptIn(FnExperimental::class, PartiQLValueExperimental::class)
-    fun resolveFn(path: BindingPath, args: List<Rex>): Rex? {
+    fun resolveFn(path: BindingPath, args: List<Rex>, unresolvedFunc: Rex.Op.Call.Unresolved): Rex? {
```
RCHowell, line 91: Where is the unresolvedFunc used?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt
```diff
@@ -23,9 +28,9 @@ import org.partiql.value.PartiQLValueExperimental
  */
 internal object PlanTransform {
 
-    fun transform(node: PartiQLPlan, onProblem: ProblemCallback): org.partiql.plan.PartiQLPlan {
+    fun transform(node: PartiQLPlan, missingOpBehavior: PartiQLPlanner.Session.MissingOpBehavior, onProblem: ProblemCallback, debugMode: Boolean): org.partiql.plan.PartiQLPlan {
```
RCHowell, line 31: Again, let's get MissingOpBehavior and DebugMode into a single compiler flags construct. These shouldn't be passed independently like this. All compilation configuration should be contained in a single type

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt
```diff
@@ -362,20 +403,31 @@ internal object PlanTransform {
...
-            )
-
-            override fun visitRelOpExcludePath(node: Rel.Op.Exclude.Path, ctx: Unit): org.partiql.plan.Rel.Op.Exclude.Path {
-                val root = when (node.root) {
-                    is Rex.Op.Var.Unresolved -> org.partiql.plan.Rex.Op.Var(-1, -1) // unresolved in `PlanTyper` results in error
-                    is Rex.Op.Var.Local -> visitRexOpVarLocal(node.root, ctx)
-                    is Rex.Op.Var.Global -> error("EXCLUDE only disallows values coming from the input record.")
+                paths = node.paths.mapNotNull {
+                    val root = when (val root = it.root) {
+                        is Rex.Op.Var.Unresolved -> org.partiql.plan.Rex.Op.Var(-1, -1) // unresolved in `PlanTyper` results in error
```
RCHowell, line 408: We should constructor an error and call visitRexOpErr. We don't want this to go unreported.

partiql-plan/src/main/resources/partiql_plan.ion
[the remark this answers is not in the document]
RCHowell: My comment is really asking, **could we please consolidate any error/debug messaging to the single err node**.

partiql-planner/src/main/resources/partiql_plan_internal.ion
```diff
@@ -335,7 +342,7 @@ rel::{
       paths: list::[path],
       _: [
         path::{
-          root: '.rex.op.var',
```
RCHowell, line 338: Check with @alancai98 on this .. I know that using a rex.op makes it possible to insert an err node. In this case, I think you should be inserting an error node on the unresolved variable rather than a silent `var(-1,-1)`

RCHowell (changes requested):

partiql-plan/src/main/resources/partiql_plan.ion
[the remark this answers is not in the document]
yliuuuu: The blocker here really comes from unresolved function: 

if you have `not_a_func(arg1, arg2)`

The internal plan is: 

To faithfully preserve the problem trace in the public plan as a PlanNode, we the have to wrap this unresolvedFunction in to a node. 

```
// What to turn that into the public node? 
internal data class Unresolved(
                @JvmField internal val identifier: Identifier,
                @JvmField internal val args: List<Rex>,
            )
```

The best option I see is to have a dedicated debug node: i.e., you have a debugging node for unresolved function and wrapper that node as the input for Rex.Op.Err. This way we fully preserved the node information and it is very easy to present to the customer about what exactly the error is.

I am sure we can hack around the issues, for example, we can construct some rex.op for all the args and left the function name in the error message, but this is a little hacky in my opinion.  

------- 

If we need the unresolved function node as a debug node, I don't see a reason why you should not have the missingOp and errOp as debug node. Separating those two in debug node makes debugging a little easier in my opinion.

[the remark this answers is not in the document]
yliuuuu: This is assumed that you want to use the plan not just the problem to debug: 

Take `not_a_function(atomic.no_exist_field)` for example 
Plan: 
```
--------Plan---------
⚬ PartiQLPlan[tag=Plan-9d5bede7]
   ├── Catalog[name=mode_test, tag=Plan-5b7090a4]
   │  └── Value[type=int2, tag=Plan-7097c826]
   └── Query[tag=Plan-4028b20a]
      └── Error[problem=-1:-1:<unknown>: Unknown function `not_a_function(<missing>), tag=Plan-70e40d7a, static_type=any]
         └── Unresolved[tag=Plan-2c91e120]
            ├── Symbol[caseSensitivity=INSENSITIVE, symbol=not_a_function, tag=Plan-2e518174]
            └── Missing[problem=-1:-1:<unknown>: Expression always returns null or missing: caused by Path Navigation failed - root is not a struct, tag=Plan-aa832a34, static_type=missing]
               └── Symbol[key=no_exist_field, tag=Plan-2ba2fcac]
                  └── Global[tag=Plan-ecd64321, static_type=int2]
                     └── Ref[catalog=0, symbol=0, tag=Plan-d4929e03]
```
Problem: 
```
-1:-1:<unknown>: Expression always returns null or missing: caused by Path Navigation failed - root is not a struct
-1:-1:<unknown>: Unknown function `not_a_function(<missing>)
```

In my opinion the plan is much easier.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/Env.kt
yliuuuu: Removed.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt
yliuuuu: I believe this is a dead branch. Given the the resolution happened in typing time. Going to throw an error.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt
yliuuuu: https://github.com/partiql/partiql-lang-kotlin/pull/1385#discussion_r1523729759

I think MissingOpBehavior is a session mode, whereas the Debugging mode vs production mode should be a planner mode. 

>  let's get MissingOpBehavior and DebugMode into a single compiler flags construct. These shouldn't be passed independently like this. All compilation configuration should be contained in a single type

Could you please elaborate on why you think this is a better approach?

partiql-plan/src/main/resources/partiql_plan.ion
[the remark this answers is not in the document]
RCHowell: Each plan node has a unique `tag`. I suggest we return error nodes only, and any details are in a `Problem` which which is associated with the error node via the tag.

This would mean you don't need to stuff a bunch of debug stuff in the plan. And then we can expand upon the errors without breaking the plan API. Does this make sense?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/PathResolver.kt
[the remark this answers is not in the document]
yliuuuu: The original look-up logic extends SQL's three level structure, but preserve the fact that we can not create a table descriptor under catalog. This is a valid assumption.

I rolled back the lookup logic and modified the test.

johnedquinn:
BTW, this should be updated to point to the v1 branch.

partiql-planner/src/main/kotlin/org/partiql/planner/PartiQLPlanner.kt
```diff
@@ -49,7 +49,26 @@ public interface PartiQLPlanner {
...
+             *  The result plan will turn the problemetic operation into an error node.
+             */
+            QUIET,
+            /**
+             * The problometic operation will be tracked in problem callback as a error.
+             * The result plan will turn the problemetic operation into an missing node.
+             */
+            SIGNAL
+        }
+    }
```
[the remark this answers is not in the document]
yliuuuu: Based on the offline discussion, switching to flag.

partiql-planner/src/main/kotlin/org/partiql/planner/PartiQLPlannerBuilder.kt
```diff
@@ -41,6 +46,20 @@ public class PartiQLPlannerBuilder {
...
+    public fun signalMode(): PartiQLPlannerBuilder = this.apply {
+        this.flags.add(PlannerFlag.SIGNAL_MODE)
+    }
+
+    /**
+     * Java style method for setting the planner to quite mode
+     */
+    public fun quiteMode(): PartiQLPlannerBuilder = this.apply {
+        this.flags.remove(PlannerFlag.SIGNAL_MODE)
+    }
```
RCHowell, line 61: Let's only have one method here, as remember we are setting a boolean. The comment should be very clear which is the default. 

Something as straightforward as this,

```kotlin
public fun warningsAsErrors(): PartiQLPlannerBuilder = this.apply {
        this.flags.add(PlannerFlag.SIGNAL_MODE)
 }
```

We could get more sophisticated if we want to generalize, but we have no need for that now. I don't care about the internal names, but signal/quiet is a bit too vague and associated with values/NaN's whereas gcc uses warn/error lingo in regards to the compiler behavior.

I'm referencing `-Werror` https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

```
-Werror

    Make all warnings into errors.
```

partiql-planner/src/main/kotlin/org/partiql/planner/internal/PlannerFlag.kt
```diff
@@ -0,0 +1,20 @@
...
+     *
+     *    The result plan will turn the problematic operation into an error node.
+     *
+     * Otherwise:
+     *
+     *    The problematic operation will be tracked in problem callback as a missing.
+     *
+     *    The result plan will turn the problematic operation into a missing node.
+     */
+    SIGNAL_MODE
```
RCHowell, line 19: This is ok and we don't have to change anything because it is internal.

We are introducing two concepts, compiler _warnings_ and a warning for "always missing". What we are calling "signal" is really emit warnings for always missing + treat warnings as errors -> resulting in a throw

RCHowell (changes requested):

RCHowell (approved):

PR #1388, opened 2024-03-13
Adds performance optimizations for ExprCallDynamic

johnedquinn:
## Relevant Issues
- None

## Description
- Adds performance optimizations for ExprCallDynamic
  - Creates a `CandidateIndex.All` that preserves the initial order of the candidate functions, however, it internally wraps a list of `CandidateIndex.Direct` and `CandidateIndex.Indirect`. `CandidateIndex.Direct` holds a map between runtime parameter types and the associated candidate to invoke. `CandidateIndex.Indirect` holds an ordered list of any non-runtime parameter types (aka `ANY`). With this, `CandidateIndex.All` can reduce the number of checks by taking advantage of these two subclasses.
  - Fixes ordering of the dynamic candidates. Before, we would lose the original ordering of candidates, and for some functions, it would cause extremely expensive computation. For example, when profiling `SELECT ... WHERE <ANY> = 2`, the lack of ordering would for some reason **always** place `EQ(ANY, ANY)` first -- which would invoke PartiQL's comparator -- which is seemingly very slow (taking up 25% of the entire test's CPU usage). By re-ordering the candidates, it significantly impacted our performance for the better (the matched `EQ` function did not even show up on the profiler, indicating that it took up less than 1% of the CPU usage).
- Since CASTs are really function calls in SQL:1999, I've updated the definition of `Ref.Cast` to contain `isNullable` -- at least until we truly represent casts with function calls.
```
<user-defined cast definition> ::=
    CREATE CAST <left paren> <source data type> AS <target data type> <right paren>
    WITH <cast function>
    [ AS ASSIGNMENT ]
```
- I needed to update the tests since the typing of functions like `ADD(NULL, INT)` were initially returning the type `NULL`, but it really should have been returning `INT?`. This is modeled slightly different in SQL:1999 since types are nullable, but until we add this, we'll need to support this.
- By examining the profiler, in permissive mode, our runtime exceptions that get converted to `MISSINGs` are extremely expensive (causing up to 33.2% of our CPU usage for evaluation) due to the `fillInStackTrace` method. Since we are programming a language, we can/should be architecting our own error messaging system.

## Note
- The conformance tests that now fail are from `NULL = MISSING`. See the Summary. I think these tests are wrong. See PartiQL Spec: "Equality never fails in the type-checking mode and never returns MISSING in the permissive mode."

## Other Information
- Any backward-incompatible changes? **YES** -- the `isNullable` flag, which is required

## License Information

### DIFF
--- a/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
+++ b/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
@@ -21,20 +21,21 @@ import org.partiql.value.PartiQLValueType
  */
 @OptIn(PartiQLValueExperimental::class, FnExperimental::class)
 internal class ExprCallDynamic(
-    private val candidates: List<Candidate>,
+    candidates: List<Candidate>,
     private val args: Array<Operator.Expr>
 ) : Operator.Expr {
 
+    private val candidateIndex = CandidateIndex.All(candidates)
+
     override fun eval(env: Environment): PartiQLValue {
         val actualArgs = args.map { it.eval(env) }.toTypedArray()
-        candidates.forEach { candidate ->
-            if (candidate.matches(actualArgs)) {
-                return candidate.eval(actualArgs, env)
-            }
+        val actualTypes = actualArgs.map { it.type }
+        candidateIndex.get(actualTypes)?.let {
+            return it.eval(actualArgs, env)
         }
         val errorString = buildString {
             val argString = actualArgs.joinToString(", ")
-            append("Could not dynamically find function for arguments $argString in $candidates.")
+            append("Could not dynamically find function (${candidateIndex.name}) for arguments $argString.")
         }
         throw TypeCheckException(errorString)
     }
@@ -47,7 +48,7 @@ internal class ExprCallDynamic(
      *
      * @see ExprCallDynamic
      */
-    internal class Candidate(
+    internal data class Candidate(
         val fn: Fn,
         val coercions: Array<Ref.Cast?>
     ) {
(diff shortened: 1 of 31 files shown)


yliuuuu:
If we know that the function input will be always null, in my opinion we should type the function to `NULL`. 

For two reasons: 
1. Consider a unary function, i.e., `-null`, the typer now returns `UnionOf(INT2, NULL)`, which can be fine, but a little confused. 
2. The information that the function is always going to return `NULL` can help with optimization, i.e., `1 + NULL IS NULL`, we can constant fold that to `true`.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
```diff
@@ -21,20 +21,21 @@ import org.partiql.value.PartiQLValueType
  */
 @OptIn(PartiQLValueExperimental::class, FnExperimental::class)
 internal class ExprCallDynamic(
-    private val candidates: List<Candidate>,
+    candidates: List<Candidate>,
     private val args: Array<Operator.Expr>
 ) : Operator.Expr {
 
+    private val candidateIndex = CandidateIndex.All(candidates)
```
RCHowell, line 28: I say create the index in the Compiler and pass to the ExprCallDynamic. In this scenario, we're passing an arg into the constructor just to be passed to another constructor.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
```diff
@@ -63,32 +64,114 @@ internal class ExprCallDynamic(
...
+         */
+        data class Direct private constructor(val directCandidates: Map<List<PartiQLValueType>, Candidate>) : CandidateIndex {
+
+            companion object {
+                internal fun of(candidates: List<Pair<List<PartiQLValueType>, Candidate>>) = Direct(candidates.toMap())
+            }
+
+            override fun get(args: List<PartiQLValueType>): Candidate? {
+                return directCandidates[args]
+            }
```
RCHowell, line 156: perhaps we choose array over list? Then its on the stack rather than a reference type. Might have marginally better lookup too

johnedquinn:
> If we know that the function will be always null, in my opinion we should type the function to `NULL`.
> 
> For two reasons:
> 
>     1. Consider a unary function, i.e., `-null`, the typer now returns `UnionOf(INT2, NULL)`, which can be fine, but a little confused.
> 
>     2. The information that the function is always going to return `NULL` can help with optimization, i.e., `1 + NULL IS NULL`, we can constant fold that to `true`.

True. Good call out. Seemingly our typing of casts is wrong. From the SQL:1999 Spec Section 6.22:
> If the <cast operand> specifies NULL, then TV is the null value.
> If SV is the null value, then the result is the null value.

Since we model NULL as a distinct type, we should still insert the casts, but the typer really should fold the CAST (since we know the input is null). If you're aligned, I can fix the typing of rexOpCast.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
```diff
@@ -63,32 +64,114 @@ internal class ExprCallDynamic(
...
+
+            override fun get(args: List<PartiQLValueType>): Candidate? {
+                return this.lookups.firstNotNullOfOrNull { it.get(args) }
+            }
+        }
+
+        /**
+         * An O(1) structure to quickly find directly matching dynamic candidates.
+         */
+        data class Direct private constructor(val directCandidates: Map<List<PartiQLValueType>, Candidate>) : CandidateIndex {
```
RCHowell, line 148: what's the difference between All and Direct? Why two?

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt
```diff
@@ -729,7 +729,7 @@ class PlanTyperTestsPorted {
             SuccessTestCase(
                 name = "BITWISE_AND_NULL_OPERAND",
                 query = "1 & NULL",
-                expected = StaticType.NULL,
+                expected = StaticType.unionOf(StaticType.INT4, StaticType.NULL),
```
RCHowell, line 732: Why not StaticType.NULL?

partiql-plan/src/main/resources/partiql_plan.ion
```diff
@@ -41,6 +41,7 @@ ref::{
     cast::{
       input: partiql_value_type,
       target: partiql_value_type,
+      isNullable: bool
     }
```
RCHowell, line 45: >  I've updated the definition of Ref.Cast to contain isNullable -- at least until we truly represent casts with function calls

Well, we did have them as function calls but that was a pain to actually work with. While the SQL specification may define them like function calls, we don't necessarily have to implement them as such if it's more trouble than its worth.

Why add isNullable here? This is the cast specification, and every cast propagates null. So the nullability of a cast is entirely determined by the input argument, unless I'm missing something related to CAST(.. AS NULL)

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
```diff
@@ -63,32 +64,114 @@ internal class ExprCallDynamic(
...
-                val c = coercions[i]
-                when (c) {
-                    // coercion might be null if one of the following is true
-                    // Function parameter is ANY,
-                    // Input type is null
-                    // input type is the same as function parameter
-                    null -> {
-                        if (!(inputType == parameterType || inputType == PartiQLValueType.NULL || parameterType == PartiQLValueType.ANY)) {
-                            return false
+            init {
```
alancai98, line 86: Curious if we see any notable CPU utilization jump from this `init` block since there is more work when an `ExprCallDynamic` is initialized. Wondering if there are any situations in which this construction of the `CandidateIndex` results in worse evaluation performance/utilization (e.g. if `candidates` is just a list of two elements)? I assume this change is better in the general case in which there are more than just a few candidate elements though.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
```diff
@@ -63,32 +64,114 @@ internal class ExprCallDynamic(
...
+                val lookupsMutable = mutableListOf<CandidateIndex>()
+                val accumulator = mutableListOf<Pair<List<PartiQLValueType>, Candidate>>()
+
+                // Indicates that we are currently processing dynamic candidates that accept ANY.
+                var activelyProcessingAny = true
+
+                candidates.forEach { candidate ->
```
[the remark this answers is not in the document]
alancai98: Also is there any way to verify/test that the lookup is returning the correct candidate? Perhaps we could add some unit tests for this lookup logic.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt
johnedquinn: The value will be NULL, however, what is the type of the null value? I'm using our normal rules for type/function precedence to resolve the output types. See the following [PostgreSQL DB Fiddle Example](https://www.db-fiddle.com/f/jxMVQnmDXvRnc1tHCzhK96/0).

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
johnedquinn: Good call out -- in the future, I'd say it'd be a good idea for the `Compiler` to dictate what version of `ExprCallDynamic` to create (the indexing O(1) vs iteration O(n)) depending on the number of candidates (and our own testing). As for `init` by itself, it is only invoked at compilation.

For now, we know dynamic is quite expensive for our customers.

RCHowell (approved):

Issue #1390, opened 2024-03-14
CASE-WHEN branches are not typed per SQL-99 rules.


RCHowell:
## Description

The current implementation of PartiQL (0.14) added static typing to queries, but the CASE-WHEN does not follow SQL-99 when branches have a _minimal common supertype_.

We expect CASE-WHEN to follow [SQL-99 9.3 Data types of results of aggregations](https://web.cecs.pdx.edu/~len/sql1999.pdf#page=359) which defines how a CASE-WHEN should be typed.

## To Reproduce

Consider two exact-numeric return values

```sql
 CASE
   WHEN a = 'hello' THEN 0
   ELSE 1.0
END

-- type
-- decimal | int32
```

## Expected Behavior

SQL-99 defines the coercion rules of the output types of a CASE-WHEN. We expect this to be typed as just `decimal`

> If all of the data types in DTS are exact numeric, then the result data type is exact numeric
with implementation-defined precision and with scale equal to the maximum of the scales of
the data types in DTS

For heterogenous data, PartiQL uses the `dynamic` type as the minimal common supertype

> Otherwise, there shall exist a subtype family STF such that each data type in DTS is a
member of STF. The result data type is the minimal common supertype of each data type in
DTS.

This enables PartiQL to behave like ANSI SQL while still handling heterogenous branch types.

## Additional Context
- Java version: 11
- PartiQL version: 0.14.3
- Add any other context about the problem here.

PR #1391, opened 2024-03-14
Apply SQL 9.3 typing rules for CASE-WHEN

RCHowell:
## Relevant Issues
#1390 

## Description

This PR introduces the typing rules of a CASE-WHEN which types branches to the minimum common supertype. In the case of types from different sub-type families (STF) then we go up to the ANY type (known now as `dynamic`).

## Other Information
No

This is a behavioral change due to a bug.

No

## License Information

### DIFF
--- a/partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
+++ b/partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
@@ -15,7 +15,6 @@
 
 package org.partiql.cli
 
-import AstPrinter
 import com.amazon.ion.system.IonSystemBuilder
 import com.amazon.ion.system.IonTextWriterBuilder
 import org.partiql.cli.pico.PartiQLCommand
--- a/partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
+++ b/partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
@@ -17,13 +17,16 @@ package org.partiql.cli
 
 import AstPrinter
 import com.amazon.ion.system.IonSystemBuilder
+import com.amazon.ion.system.IonTextWriterBuilder
 import org.partiql.cli.pico.PartiQLCommand
 import org.partiql.cli.shell.info
 import org.partiql.lang.eval.EvaluationSession
 import org.partiql.parser.PartiQLParser
+import org.partiql.plan.Statement
 import org.partiql.plan.debug.PlanPrinter
 import org.partiql.planner.PartiQLPlanner
 import org.partiql.plugins.local.LocalConnector
+import org.partiql.plugins.local.toIon
 import picocli.CommandLine
 import java.io.PrintStream
 import java.nio.file.Paths
@@ -65,8 +68,8 @@ object Debug {
 
         // Parse
         val statement = parser.parse(input).root
-        out.info("-- AST ----------")
-        AstPrinter.append(out, statement)
+        // out.info("-- AST ----------")
+        // AstPrinter.append(out, statement)
 
         // Plan
         val sess = PartiQLPlanner.Session(
@@ -79,6 +82,16 @@ object Debug {
         val result = planner.plan(statement, sess).plan
         out.info("-- Plan ----------")
         PlanPrinter.append(out, result.statement)
+
+        when (val plan = result.statement) {
+            is Statement.Query -> {
+                out.info("-- Schema ----------")
+                val outputSchema = java.lang.StringBuilder()
+                val ionWriter = IonTextWriterBuilder.minimal().withPrettyPrinting().build(outputSchema)
+                plan.root.type.toIon().writeTo(ionWriter)
+                out.info(outputSchema.toString())
+            }
+        }
 
         return "OK"
     }
--- a/partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
+++ b/partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
@@ -67,8 +67,8 @@ object Debug {
 
         // Parse
         val statement = parser.parse(input).root
-        // out.info("-- AST ----------")
-        // AstPrinter.append(out, statement)
+        out.info("-- AST ----------")
+        AstPrinter.append(out, statement)
 
         // Plan
         val sess = PartiQLPlanner.Session(
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
@@ -56,6 +56,7 @@ internal typealias SuperGraph = Array<Array<PartiQLValueType?>>
 internal class DynamicTyper {
 
     private var supertype = NULL
+
     private var nullable = false
     private var missable = false
     private val allTypes = mutableSetOf<StaticType>()
@@ -85,16 +86,20 @@ internal class DynamicTyper {
         val modifiers = mutableSetOf<StaticType>()
         if (nullable) modifiers.add(StaticType.NULL)
         if (missable) modifiers.add(StaticType.MISSING)
+        // If at top supertype, then return union of all accumulated types
         if (supertype == ANY) {
             return StaticType.unionOf(allTypes + modifiers)
         }
+        // If a collection, then return union of all accumulated types as these coercion rules are not defined by SQL.
+        if (supertype == STRUCT || supertype == BAG || supertype == LIST || supertype == SEXP) {
+            return StaticType.unionOf(allTypes + modifiers)
+        }
+        // Otherwise, return the supertype
         val st = supertype.toNonNullStaticType()
         return if (modifiers.isEmpty()) {
-            // simple
             st
         } else {
-            // add in the modifiers
-            StaticType.unionOf(setOf(st) + modifiers)
+            StaticType.unionOf(setOf(st) + modifiers).flatten()
         }
     }
 
@@ -105,9 +110,9 @@ internal class DynamicTyper {
         // Lookup and set the new minimum common supertype
         val t = type.toRuntimeType()
         supertype = when {
-            supertype == NULL -> t                                  // initialize
-            t == NULL || t == MISSING || supertype == t -> return   // skip
-            else -> graph[supertype][t] ?: ANY                      // lookup, if missing then go to top.
+            supertype == NULL -> t // initialize
+            t == NULL || t == MISSING || supertype == t -> return // skip
+            else -> graph[supertype][t] ?: ANY // lookup, if missing then go to top.
         }
     }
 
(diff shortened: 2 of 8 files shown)


partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
```diff
@@ -0,0 +1,323 @@
...
+            return StaticType.unionOf(types + modifiers) to null
+        }
+        // Otherwise, return the supertype along with the coercion mapping
+        val type = supertype.toNonNullStaticType()
+        val mapping = args.map { it to supertype }
+        return if (modifiers.isEmpty()) {
+            type to mapping
+        } else {
+            StaticType.unionOf(setOf(type) + modifiers).flatten() to mapping
+        }
```
yliuuuu, line 109: If one would keep accumulating `MISSING`, does not seem like the NULL type will be dropped.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
```diff
@@ -0,0 +1,323 @@
...
+
+    private var supertype = NULL
+    private var args = mutableListOf<PartiQLValueType>()
+
+    private var nullable = false
+    private var missable = false
+    private val types = mutableSetOf<StaticType>()
+
+    fun accumulate(type: StaticType) {
+        val arg = type.toRuntimeType()
```
yliuuuu, line 66: StaticType.toRunTimeType() does not support for AnyOfType.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -681,34 +686,75 @@ internal class PlanTyper(
...
+            for (i in oldBranches.indices) {
+
+                // Type the branch
+                var branch = oldBranches[i]
+                branch = visitRexOpCaseBranch(branch, branch.rex.type)
+
+                // Check if branch condition is a literal
+                val condition = branch.condition
+                when (boolOrNull(condition.op)) {
+                    true -> return branch.rex // fold
```
johnedquinn, line 702: I think there's a bug here. Let's say a CASE has three branches (the first two being booleans (but not constants) and the third being a boolean of literal true). With this logic, it seems like we'd accumulate the first two visited branches, but for the third, we'd immediately return it's result.

yliuuuu (changes requested):
Had a high level question: 

Consider: `a: unionOf(int2, int4, null)`. 

Given a query: 
```
Case when a IS NULL then a
ELSE a END;
```

Should the common super type be 
`unionOf(int2, int4, null)`  // a dynamic with constraint being int2, int4 and null
or 
`unionOf(int4, null)`. // a dynamic with constraint being int4, and null

If the later, should the query `a` also has the type `unionOf(int4, null)`? 

I understand that the common super type rules applied for an aggregation over values of compatible data types. 

We might need to clarify what is an aggregation in PartiQL. That is: one can argue that a simple project is an aggregation. In SQL that is an aggregation over values of same data type, in PartiQL it might be an aggregation over any possible data type.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
```diff
@@ -0,0 +1,323 @@
...
+                STRING to STRING,
+                SYMBOL to STRING,
+                CLOB to CLOB,
+            )
+            graph[TIME] = edges(
+                TIME to TIME,
+            )
+            graph[TIMESTAMP] = edges(
+                TIMESTAMP to TIMESTAMP,
+            )
```
alancai98, line 298: SQL:99 also defines some datetime typing rules:

> f) If some data type in DTS is a datetime data type, then every data type in DTS shall be a
datetime data type having the same datetime fields. The result data type is a datetime data
type having the same datetime fields, whose fractional seconds precision is the largest of the
fractional seconds precisions in DTS. If some data type in DTS has a timezone displacement
value, then the result has a timezone displacement value; otherwise, the result does not
have a timezone displacement value.

If we aren't tackling those rules in this PR, could add a TODO

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/functions/NullIfTest.kt
```diff
@@ -8,27 +8,34 @@ import org.partiql.planner.util.cartesianProduct
 import org.partiql.types.StaticType
 import java.util.stream.Stream
 
-// TODO: Model handling of Truth Value in typer better.
+/**
```
alancai98, line 11: iirc, SQL spec defines a rewrite for `COALESCE` using case-when. So we'd probably need to test `COALESCE`'s typing as well.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
RCHowell: I'm going to leave constant folding `true` out of this considering the urgency

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/functions/NullIfTest.kt
RCHowell: Yes which we already rewrite.

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
```diff
@@ -65,8 +67,8 @@ object Debug {
 
         // Parse
         val statement = parser.parse(input).root
-        out.info("-- AST ----------")
-        AstPrinter.append(out, statement)
+        // out.info("-- AST ----------")
```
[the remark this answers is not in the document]
RCHowell: I can revert this section. It's just debugging info.

partiql-planner/src/testFixtures/resources/inputs/basics/case.sql
```diff
@@ -1,54 +1,309 @@
...
+
+-- -----------------------------
+--  Approximate Numeric
+-- -----------------------------
+
+-- TODO model approximate numeric
+-- We do not have the appropriate StaticType for this.
+
+--#[case-when-12]
+-- type: (float64|null)
```
alancai98, line 100: shouldn't have `null` in comment. test in `PlanTyperTestsPorted` looks to be correct

```suggestion
-- type: (float64)
```

partiql-planner/src/testFixtures/resources/inputs/basics/case.sql
```diff
@@ -1,54 +1,309 @@
...
+--  Heterogeneous Branches
+-- -----------------------------
+
+--#[case-when-24]
+-- type: (int32|int64|string)
+CASE t_item.t_string
+    WHEN 'a' THEN t_item.t_int32
+    WHEN 'b' THEN t_item.t_int64
+    ELSE 'default'
+END;
```
alancai98, line 211: So for the heterogeneous cases, we're opting to keep all of the return types in the minimal common subtype and not doing any sort of "flattening" (for this case, returning `(int64, string)`)?

partiql-planner/src/testFixtures/resources/catalogs/default/pql/t_item.ion
```diff
@@ -0,0 +1,184 @@
...
+    },
+    // Exact Numeric
+//    {
+//      name: "t_int8",
+//      type: "int8",
+//    },
+//    {
+//      name: "t_int8_null",
+//      type: ["int8", "null"],
+//    },
```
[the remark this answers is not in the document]
alancai98: nice set of tests btw!

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
```diff
@@ -0,0 +1,362 @@
...
+                INT to FLOAT64,
+                DECIMAL to FLOAT64,
+                DECIMAL_ARBITRARY to FLOAT64,
+                FLOAT32 to FLOAT64,
+                FLOAT64 to FLOAT64,
+            )
+            graph[CHAR] = edges(
+                CHAR to CHAR,
+                STRING to STRING,
+                SYMBOL to STRING,
```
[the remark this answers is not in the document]
RCHowell: More or less "string" as far as our ability to express types goes..

```
a) If any of the data types in DTS is character string, then all data types in DTS shall be
character string, and all of them shall have the same character repertoire.

...

i) If any of the data types in DTS is character large object string, then the result data
type is character large object string with maximum length in characters equal to the
maximum of the lengths in characters and maximum lengths in characters of the data
types in DTS.

ii) If any of the data types in DTS is variable-length character string, then the result data
type is variable-length character string with maximum length in characters equal to the
maximum of the lengths in characters and maximum lengths in characters of the data
types in DTS.

iii) Otherwise, the result data type is fixed-length character string with length in characters
equal to the maximum of the lengths in characters of the data types in DTS.
```

Based off of SQL it's pretty-much `CLOB > VARCHAR > CHAR` but this implementation doesn't include the length constraints.

partiql-planner/src/testFixtures/resources/catalogs/default/pql/t_item.ion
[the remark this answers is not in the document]
RCHowell: We don't have int8 with StaticType. Only int16->INT2

partiql-planner/src/testFixtures/resources/inputs/basics/case.sql
RCHowell: Exactly. Once the minimal common supertype hits the top ie `dynamic`, we don't coerce/cast and branches. We carry all types to the result.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
RCHowell: Could you please elaborate more?

partiql-planner/src/testFixtures/resources/catalogs/default/pql/t_item.ion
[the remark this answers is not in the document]
alancai98: ahh.. forgot about that

alancai98 (approved):

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
yliuuuu:
```
val typer = DynamicTyper() // at this stage, super Type is NULL
typer.accumulate(MISSING) // no touch on super type, missable is true

// modifier contains missing, 
// type is supertype.toNonNullStaticType() -> NULL
// return is StaticType.unionOf(setOf(type) + modifiers).flatten() to mapping
typer.mapping() // UnionOf( NULL, MISSING) 
```

yliuuuu (approved):
LGTM. Thanks.

PR #1404, opened 2024-03-28
Add `COALESCE` and `NULLIF` to the logical plan

alancai98:
## Relevant Issues
Related to https://github.com/partiql/partiql-scribe/issues/26.

## Description
Previous rewrite according to SQL-99 spec was creating large logical plans with nested `COALESCE` and `NULLIF` expressions. This change makes dedicated `COALESCE` and `NULLIF` nodes within the logical plan.

Other than that,
- fixes a prior typing issue related to `COALESCE` (context in https://github.com/partiql/partiql-lang-kotlin/pull/1404#discussion_r1543461964)
- fixes a slight bug in dynamic typer w/ `StaticType.ANY`, where it was previously getting expanded to all of its types

## Other Information

Adding new nodes to the plan is technically a breaking change for the `PlanVisitor` interface. Tracking the sprout generation to add a warning on the interface as a separate issue -- https://github.com/partiql/partiql-lang-kotlin/issues/1405.

## License Information

### DIFF
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
@@ -30,8 +30,10 @@ import org.partiql.planner.internal.ir.identifierQualified
 import org.partiql.planner.internal.ir.identifierSymbol
 import org.partiql.planner.internal.ir.rex
 import org.partiql.planner.internal.ir.rexOpCallStatic
+import org.partiql.planner.internal.ir.rexOpCoalesce
 import org.partiql.planner.internal.ir.rexOpCollection
 import org.partiql.planner.internal.ir.rexOpLit
+import org.partiql.planner.internal.ir.rexOpNullif
 import org.partiql.planner.internal.ir.rexOpPathIndex
 import org.partiql.planner.internal.ir.rexOpPathKey
 import org.partiql.planner.internal.ir.rexOpPathSymbol
@@ -107,7 +109,7 @@ internal object RexConverter {
         private fun visitExprCoerce(node: Expr, ctx: Env, coercion: Rex.Op.Subquery.Coercion = Rex.Op.Subquery.Coercion.SCALAR): Rex {
             val rex = super.visitExpr(node, ctx)
             return when (rex.op is Rex.Op.Select) {
-                true -> rex(StaticType.ANY, rexOpSubquery(rex.op as Rex.Op.Select, coercion))
+                true -> rex(StaticType.ANY, rexOpSubquery(rex.op, coercion))
                 else -> rex
             }
         }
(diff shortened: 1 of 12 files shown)


partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
```diff
@@ -439,44 +441,21 @@ internal object RexConverter {
             return rex(type, call)
         }
 
-        // coalesce(expr1, expr2, ... exprN) ->
-        //   CASE
-        //     WHEN expr1 IS NOT NULL THEN EXPR1
-        //     ...
-        //     WHEN exprn is NOT NULL THEN exprn
-        //     ELSE NULL END
```
alancai98, line 447: Prior modeling of `COALESCE` was slightly incorrect here. Previously, there was an explicit `ELSE NULL` added here, which would always add `NULL` to the return type. The correct equivalence would actually be:
```
COALESCE(v1, v2,..., vN)
==
CASE
    WHEN v1 IS NOT NULL THEN v1
    WHEN v2 IS NOT NULL THEN v2
    ... -- similarly for v3..vN-1
    ELSE vN
END
```

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -760,6 +762,54 @@ internal class PlanTyper(
...
+        //     WHEN v1 = v2 THEN NULL -- WHEN branch always a boolean
+        //     ELSE v1
+        // END
+        // --> minimal common supertype of (NULL, <type v1>)
+        override fun visitRexOpNullif(node: Rex.Op.Nullif, ctx: StaticType?): Rex {
+            val v1 = visitRex(node.v1, node.v1.type)
+            val v2 = visitRex(node.v2, node.v2.type)
+            val typer = DynamicTyper()
+            typer.accumulate(NULL)
+            typer.accumulate(v1.type)
```
alancai98, line 807: Only add `v1`'s type here. I found that posgresql actually does something different here. Per [their docs](https://www.postgresql.org/docs/current/functions-conditional.html#FUNCTIONS-NULLIF):
> The result has the same type as the first argument — but there is a subtlety. What is actually returned is the first argument of the implied = operator, and in some cases that will have been promoted to match the second argument's type. For example, NULLIF(1, 2.2) yields numeric, because there is no integer = numeric operator, only numeric = numeric.

Per my read of [SQL-99](https://web.cecs.pdx.edu/~len/sql1999.pdf#page=205), I can't find any mention of this promotion behavior, so leaving the result type as just the minimal common supertype of (\<type v1\>, null). Postgresql would do something like the minimal common supertype of (\<type v1\>, \<type v2\>, null).

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
RCHowell: Nice catch

RCHowell (changes requested):
Overall looks good and is very straightforward. Only suggestion is to update the nullif names and simplify the typing logic.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
alancai98: Yeah I'll simplify it. I initially included the dynamic typer in case we wanted to follow what postgresql did, which is easy to do w/ the dynamic typer.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
```diff
@@ -439,44 +441,21 @@ internal object RexConverter {
...
-            val fn = fnUnresolved(id, true)
-            val call = rexOpCallStatic(fn, listOf(expr1, expr2))
-            val branches = listOf(
-                rexOpCaseBranch(rex(type, call), rex(type = StaticType.NULL, op = rexOpLit(value = nullValue()))),
-            )
-            val op = rexOpCase(branches.toMutableList(), expr1)
-            rex(type, op)
+            val expr1 = visitExprCoerce(node.value, ctx)
+            val expr2 = visitExprCoerce(node.nullifier, ctx)
+            val op = rexOpNullif(expr1, expr2)
```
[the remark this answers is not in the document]
alancai98: [SQL99](https://web.cecs.pdx.edu/~len/sql1999.pdf#page=205) used `v1` and `v2` when going over the rewrite to `CASE WHEN`, so that's probably where `v1` and `v2` came from. I agree w/ keeping the plan consistent w/ the ast. I'll change it in the next revision.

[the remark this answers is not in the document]
RCHowell: This was actually Vladimir's original suggestion to me. https://github.com/partiql/partiql-lang-kotlin/pull/1054#discussion_r1186750877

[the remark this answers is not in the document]
alancai98: Ah I see. Agree w/ the suggestion

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
alancai98: Followup from offline discussion, will keep the dynamic typer here to ensure consistency with the typing of `CASE WHEN` and `COALESCE`.

partiql-planner/src/testFixtures/resources/inputs/basics/coalesce.sql
```diff
@@ -0,0 +1,71 @@
...
+-- type: (int64 | string)
+COALESCE(t_item.t_int64, t_item.t_string);
+
+--#[coalesce-12]
+-- type: (int64 | null | string)
+COALESCE(t_item.t_int64_null, t_item.t_string);
+
+--#[coalesce-13]
+-- type: (int16 | int32 | int64 | int | decimal)
+COALESCE(t_item.t_num_exact, t_item.t_int32);
```
RCHowell, line 55: Would this not be `type: decimal`?

partiql-planner/src/testFixtures/resources/inputs/basics/nullif.sql
```diff
@@ -0,0 +1,77 @@
...
+-- type: (string | null)
+NULLIF(t_item.t_string, t_item.t_int32);
+
+--#[nullif-15]
+-- type: (int32 | null)
+NULLIF(t_item.t_int32, t_item.t_num_exact);
+
+--#[nullif-16]
+-- type: (int16 | int32 | int64 | int | decimal | null)
+NULLIF(t_item.t_num_exact, t_item.t_int32);
```
RCHowell, line 69: Same as previous comment. Intuitively I would think this is typed to a decimal

partiql-planner/src/testFixtures/resources/inputs/basics/coalesce.sql
alancai98: The union of all the numeric types is output here to follow the existing `CASE WHEN` behavior. I believe this was brought up in Yingtao's comment about a union type for one of the `THEN` branches https://github.com/partiql/partiql-lang-kotlin/pull/1391#pullrequestreview-1943767991. For instance, the current behavior of `CASE WHEN` will include all of the types within the union/AnyOfType:
```
CASE <type string>
    WHEN 'a' THEN <type union(int4, int8)>
    ELSE <type int4>
END;
```
will have return type of `union(int4, int8)`.

This is also the case for `COALESCE` which is equivalent to `CASE WHEN` with each argument corresponding to a `THEN` branch:
```
COALESCE(<type union(int4, int8)>, <type int4>)
==
CASE WHEN <type union(int4, int8)> IS NOT NULL THEN <type union(int4, int8)>
     ELSE <type int4>
END
```
will have return type of `union(int4, int8)`.

One could make the argument that the minimal common supertype is actually just `int8`; however, this `union` type argument behavior has not been defined yet.

RCHowell (approved):

PR #1430, opened 2024-04-12
Adds compilation/evaluation support for UNION/INTERSECT/EXCEPT ALL/DISTINCT

johnedquinn:
## Relevant Issues
- None

## Description
- Adds compilation/evaluation support for UNION/INTERSECT/EXCEPT ALL/DISTINCT
- I have used Substrait's modelling of [Set Operations](https://substrait.io/relations/logical_relations/#set-operation) as a guide.

## Specification and RFC Insight
- This PR does NOT pass some of the conformance tests in regards to coercion of scalar values. I'm unsure of what to do here. According to the PartiQL Specification, Section 5.1.1, "Mistyping Cases":
> In the following cases the expression in the FROM clause item has the wrong type. Under the type checking option, all of these cases raise an error and the query fails. Under the permissive option, the cases proceed as follows:
>
> Iteration over a scalar value: Consider the query:
> FROM s AS v AT p
> or the query:
> FROM s AS v
>
> where s is a scalar value. Then s coerces into the bag \<\< s \>\>, i.e., the bag that has a single element, the s. The
> rest of the semantics is identical to what happens when the lhs of the FROM item is a bag.
- The above essentially states that, in strict mode, the query will FAIL when the RHS of a FROM is a scalar. In permissive, it'll be coerced to a bag holding the scalar.
- As I was implementing [RFC-0007](https://github.com/partiql/partiql-lang/blob/main/RFCs/0007-rfc-bag-operators.md), I remembered the above scenario since I recently implemented the scalar coercion of FROM. I looked to RFC-0007 as a guide, however, the wording isn't clear when it comes to typing modes:
> A scalar value is coerced into a singleton bag; a list is coerced to a bag by discarding ordering; NULL and MISSING are coerced into the empty bag.
> ...
> ... PartiQL has chosen to produce heterogeneous bags in permissive mode and homogenous bags or error in conventional typing mode.
- For this PR, I am assuming the SAME rules of coercion apply for both the RHS of a FROM and the operands of the OUTER UNION.
- **ALSO**, for naming of output bindings, from RFC-0007:
> If the i-th column of T1 and T2 does not have the same name, then the name of i-th column of TR is implementation-dependent and unique among all other column names.

## Examples
- If you'd like some insight into the ops, feel free to play with this [DB Fiddle Example](https://www.db-fiddle.com/f/sFKd5xs8oewDH7ho1bswPc/0).

## Other Information
- Any backward-incompatible changes? **YES** -- plan representation now more closely matches Substrait's model

## License Information


partiql-plan/src/main/resources/partiql_plan.ion
```diff
@@ -261,19 +261,19 @@ rel::{
...
       lhs: rel,
       rhs: rel,
+      type: [
+        UNION_ALL,
+        UNION_DISTINCT,
+        INTERSECT_ALL,
+        INTERSECT_DISTINCT,
+        EXCEPT_ALL,
+        EXCEPT_DISTINCT
+      ]
```
RCHowell, line 276: We've gone back and forth for this modeling. I believe I did "set" initially, but PIG modeled as `(union ...) (intersect ...) ...` etc. hence them being their own operators. 

I'm keen on keeping them separate mostly for staying similar to PartiQL 0.14, but this is entirely preference. However, I would like the set quantifier to be split.

```
   union::{
      setq: set_quantifier,
      lhs: rel,
      rhs: rel,
    },

    intersect::{
      setq: set_quantifier,
      lhs: rel,
      rhs: rel,
    },

    except::{
      setq: set_quantifier,
      lhs: rel,
      rhs: rel,
   }
```

Interestingly, substrait uses a list of inputs rather than lhs, rhs. Seems like an optimization

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -223,16 +224,47 @@ internal class PlanTyper(private val env: Env) {
...
+                    if (lhsBindingType != rhsBindingType) {
+                        return Rel(Rel.Type(emptyList(), emptySet()), Rel.Op.Err("LHS and RHS of SET OP do not have the same type."))
+                    }
+                }
+            }
 
-        override fun visitRelOpExcept(node: Rel.Op.Except, ctx: Rel.Type?): Rel {
-            TODO("Type RelOp Except")
+            // Compute Output Schema
+            val size = max(lhs.type.schema.size, rhs.type.schema.size)
```
RCHowell, line 250: Would this not depend on both the OUTER boolean and the operator type? 

Like OUTER UNION size may not be computable with open tuples.

RCHowell (changes requested):

partiql-plan/src/main/resources/partiql_plan.ion
```diff
@@ -261,20 +261,29 @@ rel::{
...
-    union::{
-      lhs: rel,
-      rhs: rel,
-    },
+    set::[
+      union::{
+        quantifier: quantifier,
+        lhs: rel,
+        rhs: rel,
+      },
```
RCHowell, line 269: Nice. This seems like a good middle ground.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelPeeking.kt
```diff
@@ -2,44 +2,20 @@ package org.partiql.eval.internal.operator.rel
...
-            ?: error("There was not a record to be produced, however, next() was called. Please use hasNext() beforehand.")
-        this._next = null
-        return next
+        next = null
     }
 
     override fun close() {
-        _next = null
+        next = null
     }
```
RCHowell, line 20: nit, but it would be better to have the abstract method hold the close control so that inheritors don't need to call super.close(). In some languages you can do something like [@mustCallSuper](https://github.com/vandadnp/flutter-tips-and-tricks/blob/main/tipsandtricks/mustcallsuper-in-dart/mustcallsuper-in-dart.md) but we can do better with a concrete close that does

```
override fun close() {
     close1()
     next = null
}

abstract fun close1()
```

It's a nit, but I wouldn't want to track down a bug because an inheritor didn't call super. Perhaps we just add a comment to keep the change small.

johnedquinn: Agreed. I was going to do that in this PR, but I didn't want the diff to affect implementers of RelPeeking. Seeing that you've reviewed the rest of the PR, I'll update. Thanks!

RCHowell (approved):

PR #1436, opened 2024-04-18
Handles null calls in dynamic dispatch

RCHowell:
## Description

Handles null calls in dynamic dispatch

## Tests Addressed

```


    OVERLAY mismatched type, compileOption: LEGACY

    null value on any of the 3 inputs returns false{likeExpr:" d.sid LIKE null ESCAPE null "}, compileOption: PERMISSIVE

    null value on any of the 3 inputs returns false{likeExpr:" d.sid LIKE null ESCAPE null "}, compileOption: LEGACY

    null value on any of the 3 inputs returns false{likeExpr:" d.sid LIKE null ESCAPE '[' "}, compileOption: PERMISSIVE

    null value on any of the 3 inputs returns false{likeExpr:" d.sid LIKE null ESCAPE '[' "}, compileOption: LEGACY

    null value on any of the 3 inputs returns false{likeExpr:" d.sid LIKE 'S1' ESCAPE null "}, compileOption: PERMISSIVE

    null value on any of the 3 inputs returns false{likeExpr:" d.sid LIKE 'S1' ESCAPE null "}, compileOption: LEGACY

    null value on any of the 3 inputs returns false{likeExpr:" d.sid LIKE null "}, compileOption: PERMISSIVE

    null value on any of the 3 inputs returns false{likeExpr:" d.sid LIKE null "}, compileOption: LEGACY

```

Yes

## License Information

### DIFF
--- a/partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
+++ b/partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
@@ -60,7 +60,7 @@ import java.lang.IllegalStateException
 internal class Compiler(
     private val plan: PartiQLPlan,
     private val session: PartiQLEngine.Session,
-    private val symbols: Symbols
+    private val symbols: Symbols,
 ) : PlanBaseVisitor<Operator, StaticType?>() {
 
     fun compile(): Operator.Expr {
@@ -102,6 +102,7 @@ internal class Compiler(
         val type = ctx ?: error("No type provided in ctx")
         return ExprCollection(values, type)
     }
+
     override fun visitRexOpStruct(node: Rex.Op.Struct, ctx: StaticType?): Operator {
         val fields = node.fields.map {
             val value = visitRex(it.v, ctx).modeHandled()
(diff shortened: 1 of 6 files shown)


partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
```diff
@@ -220,13 +221,37 @@ internal class Compiler(
...
+                arity = fnArity
+                name = fnName
+            } else {
+                if (fnArity != arity) {
+                    error("Dynamic call candidate had different arity than others; found $fnArity but expected $arity")
+                }
+                if (fnName != name) {
+                    error("Dynamic call candidate had different name than others; found $fnName but expected $name")
+                }
+            }
```
johnedquinn, line 251: Is it necessary to perform these checks?

Also, we shouldn't be checking the name this way due to case-insensitive lookup. If we need these checks, then I'd consider making `var name` maintain a normalized function name.

johnedquinn (changes requested):

partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
RCHowell: Yes, as it's basic one-time input validation. Not strictly necessary if plans come from the internal partiql-planner — but that might not always be where we receive plans from. 

Ack on the normalized names.

johnedquinn (approved):

PR #1438, opened 2024-04-19
Simplifies joins and fixes bugs

johnedquinn:
## Relevant Issues
- None

## Description
- Increases conformance by 4 tests (.07%). See [report](https://github.com/partiql/partiql-lang-kotlin/actions/runs/9877982679?pr=1438).
- Simplifies joins by using Kotlin's helpful `iterator` inline function. This substantially simplifies our code at the cost of some performance. The reason I'm making this move is that there are bugs in the current implementation, and for the sake of our goal (conformance, not performance), I'm moving towards a simpler implementation. Once we pass all join tests, we can re-evaluate our implementation for performance reasons.
- This PR now passes 8 additional tests, however, there are now 4 failing that were not previously failing. This is due to our inability to correctly coerce nulls to other types. Please disregard this issue as this is being handled in #1451.

## Other Information
- Any backward-incompatible changes? **Maybe?**
  - There is one change that is worth noting. `<struct evaluating to null>.a` returns null. Is this right? IMO, it makes the most sense, and it makes the modelling of JOINs very very easy.
- Do your changes comply with the [Contributing Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CONTRIBUTING.md) and [Code Style Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CODE_STYLE.md)? **YES**

## License Information


RCHowell:
Why draft?

johnedquinn:
> Why draft?

Currently incorporating the change for `check()` that is causing some of the conformance tests for joins to fail.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelJoinOuterRight.kt
```diff
@@ -0,0 +1,82 @@
...
+     * ```
+     *
+     * Another Note: For some databases, RIGHT OUTER JOIN cannot be used with LATERAL. This should be taken care of
+     * at planning time. TODO: Should we support LATERAL here? Can we deal with this at planning time?
+     */
+    @OptIn(PartiQLValueExperimental::class)
+    private fun implementation() = iterator {
+        for (rhsRecord in rhs) {
+            var rhsMatched = false
+            lhs.open(env.push(rhsRecord))
```
yliuuuu, line 67: I found this feature kind of counter-intuitive (as in SQL most stuff reads from left to right), and arguably incompatible with the PartiQL Semantics (as in PartiQL the "inner loop" usually can access binding environment in the "outer loop")

I believe that in set theory:`A LEFT OUTER JOIN B` is the same as `B RIGHT OUTER JOIN A`. 

In SQL implementation, the difference might be the order of attribute in the produced tuple....

I think this implementation is trying to achieve the same result.... But I am NOT convinced that we should allow the left hand side relation to have access to the binding env of rhs relation....

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelJoinOuterFull.kt
```diff
@@ -1,61 +1,105 @@
...
+     *       yield(lhsRecord + rhsRecord)
+     * for lhsRecord, lhsIndex in lhs_sorted:
+     *   if lhsMatches[lhsIndex] = false:
+     *     yield(lhsRecord, null)
+     * for rhsRecord, rhsIndex in rhs_sorted:
+     *   if rhsMatches[rhsIndex] = false:
+     *     yield(null, rhsRecord)
+     * ```
+     *
+     * Note: The LHS and RHS must be sorted. TODO: We need to add sorting to the LHS and RHS
```
yliuuuu, line 71: Could you elaborate on this please?

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelJoinOuterFull.kt
```diff
@@ -1,61 +1,105 @@
...
+        lhs.close()
+        rhs.close()
+        iterator = emptyList<Record>().iterator()
     }
 
     /**
-     * Specifically, for FULL OUTER JOIN, when the JOIN Condition ([condition]) is TRUE, we need to return the
-     * rows merged (without modification). When the JOIN Condition ([condition]) is FALSE, we need to return
-     * the LHS padded (and merged with RHS not padded) and the RHS padded (merged with the LHS not padded).
+     * FULL OUTER JOIN (CANNOT BE LATERAL)
```
yliuuuu, line 53: Worth to call out spec section 5.5 here: 

The FROM clause expression:
`l FULL JOIN r ON *$c$*`

replicates SQL’s FULL JOIN functionality. It assumes that (alike SQL) the lhs of r does not use variables defined from l.

partiql-plan/src/main/resources/partiql_plan.ion
```diff
@@ -300,6 +300,8 @@ rel::{
       projections:  list::[rex],
     },
 
+    // TODO: Specify that this is a LATERAL JOIN. Create a separate JOIN. Also, determine the allowable types of JOIN.
+    // For context: Oracle SQL doesn't allow ... FULL OUTER JOIN LATERAL ... or ... RIGHT OUTER JOIN LATERAL ...
```
yliuuuu, line 304: PartiQL Spec section 5.7

SQL 2003 used the LATERAL keyword to correlate FROM clause items. In the interest of compatibility with SQL, PartiQL also allows the use of the keyword LATERAL, though it does not do anything more than the comma itself would do.
That is “l , LATERAL r” is equivalent to “l , r”.


This is: I don't think we need to support LATERAL JOIN. The spec is quite clear about there is no difference between left join and left lateral join, inner join and inner lateral join. 

The spec is also quite clear about the fact that Full Join can not be lateral. 

I think the only decision we have to make here is if we want right join to be lateral. 

I vote no : )

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelJoinOuterFull.kt
johnedquinn: Updated KDoc.

partiql-plan/src/main/resources/partiql_plan.ion
johnedquinn: Discussed offline a while back. I'd like to have different plan nodes for both types -- which corresponds to a different implementation in the evaluator. While yes, syntactically, one doesn't need to use the LATERAL keyword, it has semantic implementations. Upon analyzing the plan and seeing no references to bindings from the LHS, we can potentially use the normal join operator (and not the lateral join).

If we ever want to support that in the future, what would we call this plan node in the future? `non_lateral_join`? Since `join` is already taken and is used to semantically denote a lateral join?

Especially given that we are a library where DB implementers choose which features to support. Does their execution allow for joins? DynamoDB doesn't. What about lateral joins? Does every implementation support lateral joins in their execution? I'm not sure -- but I do know that lateral joins are expensive due to their correlated nature. I believe it limits the implementations that one can use. If I'm not mistaken, you are forced to use a variation of a nested loop join.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelJoinOuterFull.kt
```diff
@@ -1,61 +1,107 @@
...
- *     store their values in-memory. Then we return a merged Record where the LHS is padded and the RHS is not (equivalent
- *     to result of RIGHT OUTER JOIN).
- *   - If they aren't null, then we pad the RHS with NULLS (we assume we've already padded the LHS) and return (equivalent
- *     to result of LEFT OUTER JOIN). We also make sure [previousLhs] and [previousRhs] are now null.
- *
- * Performance Analysis: Assume that [lhs] has size M and [rhs] has size N.
- * - Time: O(M * N)
- * - Space: O(1)
- */
 internal class RelJoinOuterFull(
```
yliuuuu, line 10: Nit: For this and other relational operators, could be nice to add some kdoc: 

```
Full Outer Join returns all the rows from both the table, with non-matching attributes being null-padded. 

Full Outer Join Cannot be lateral according to Spec section 5.5. 
```

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelJoinOuterRight.kt
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    override fun closePeeking() {
+        lhs.close()
+        rhs.close()
+        iterator = emptyList<Record>().iterator()
+    }
+
+    /**
+     * RIGHT OUTER JOIN (CANNOT BE LATERAL)
```
yliuuuu, line 44: Thank you!

partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEngineDefaultTest.kt
```diff
@@ -155,6 +160,123 @@ class PartiQLEngineDefaultTest {
...
+                    FROM << 0, 1, 2 >> lhs
+                    RIGHT OUTER JOIN << 3, 4, 5 >> rhs
+                    ON lhs = rhs
+                """.trimIndent(),
+                expected = bagValue(
+                    listValue(int32Value(null), int32Value(3)),
+                    listValue(int32Value(null), int32Value(4)),
+                    listValue(int32Value(null), int32Value(5)),
+                )
+            ),
```
yliuuuu, line 277: Perhaps add a couple tests for full outer join as well? 

Also, would be nice if we can test the lateral properties of Left Join, and set up baseline for non-lateral behavior for Right and Full Join. Not sure if those should go to the evaluation test or the planning test (if not already exist)

yliuuuu (commented):
Implementation looks good to me. 

Had some small comments around test cases and kdoc.

partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEngineDefaultTest.kt
johnedquinn: I've been using the conformance tests for full outer join -- but I just updated and pushed additional tests for planning.

yliuuuu (approved):
Thanks for the additional test cases.

Issue #1407, opened 2024-04-01
Shadow the ANTLRv4 runtime dependency to avoid conflicts.


RCHowell:
## Relevant Issue/Bug

The partiql-parser package uses the ANTLRv4 runtime (4.10.x) which conflicts with other projects that require different minor versions of this dependency. A reasonable solution is to shade (aka re-package) this dependency.

## Requested Solution/Feature

The partiql-parser repackages the antlr runtime dependency so customers who need to consume an incompatible minor version of it can consume both.

## Describe Alternatives
- jarjar
- shadow
- vend the runtime internally + ctrl+f/r generated-src (perhaps too hacky)

## Additional Context
N/A

## DoD (Definition of Done)
1. The partiql-parser library does not have an antlr runtime dependency*
2. The antlr runtime dependency is shadowed (aka move the package name to avoid conflicts)
3. The generated sources use the shaded names

alancai98:
Closed by #1439.

PR #1439, opened 2024-04-19
Shade antlr dependency for partiql-parser and partiql-lang

alancai98:
## Relevant Issues
- #1407
- #1440 

## Description
- Shades the antlr dependency for `partiql-parser` and `partiql-lang`
- Upgrades Gradle to 8.7 and a few other dependencies
- Testing
  - publishing to Maven local puts the `antlr` dependencies for `partiql-parser` and `partiql-lang` behind `org.partiql.thirdparty`
  - published `.pom` does not include `antlr` dependency
  - published `partiql-parser` and `partiql-lang` generated code uses `antlr` from `org.partiql.thirdparty` dependency:

```
// PartiQLParser.class
package org.partiql.parser.antlr;

import java.util.ArrayList;
import java.util.List;
import org.partiql.thirdparty.antlr.v4.runtime.FailedPredicateException;
import org.partiql.thirdparty.antlr.v4.runtime.NoViableAltException;
import org.partiql.thirdparty.antlr.v4.runtime.Parser;
import org.partiql.thirdparty.antlr.v4.runtime.ParserRuleContext;
import org.partiql.thirdparty.antlr.v4.runtime.RecognitionException;
import org.partiql.thirdparty.antlr.v4.runtime.RuleContext;
import org.partiql.thirdparty.antlr.v4.runtime.RuntimeMetaData;
import org.partiql.thirdparty.antlr.v4.runtime.Token;
import org.partiql.thirdparty.antlr.v4.runtime.TokenStream;
import org.partiql.thirdparty.antlr.v4.runtime.Vocabulary;
import org.partiql.thirdparty.antlr.v4.runtime.VocabularyImpl;
import org.partiql.thirdparty.antlr.v4.runtime.atn.ATN;
import org.partiql.thirdparty.antlr.v4.runtime.atn.ATNDeserializer;
import org.partiql.thirdparty.antlr.v4.runtime.atn.ParserATNSimulator;
import org.partiql.thirdparty.antlr.v4.runtime.atn.PredictionContextCache;
import org.partiql.thirdparty.antlr.v4.runtime.dfa.DFA;
import org.partiql.thirdparty.antlr.v4.runtime.tree.ParseTreeListener;
import org.partiql.thirdparty.antlr.v4.runtime.tree.ParseTreeVisitor;
import org.partiql.thirdparty.antlr.v4.runtime.tree.TerminalNode;

public class PartiQLParser extends Parser {
...
```

## Other Information
  - No public api changes

  - com.github.johnrengelman:shadow:8.1.1
  - gradle-wrapper version bump 7.6.2 -> 8.7
  - detekt version bump 1.20.0-RC1 -> 1.20.0-RC2
  - jacoco version bump 0.8.8 -> 0.8.11

## License Information


alancai98:
Draft PR in its current state appears to do the proper shading (putting the `antlr` dependencies behind a different directory). Still looking at the issues related to publishing. It looks like I'll need a bugfix https://github.com/johnrengelman/shadow/pull/798 that's only in `shadow` 8.0+, which only works w/ Gradle 8.0+.

Will try updating the Gradle version to 8.0+ https://github.com/partiql/partiql-lang-kotlin/issues/1440 and upgrade the `shadow` version to 8.0 to see if that fixes the publishing issue.

partiql-lang/build.gradle.kts
```diff
@@ -64,6 +64,24 @@ dependencies {
...
+    }
+}
+
+// Workaround for https://github.com/johnrengelman/shadow/issues/651
+components.withType(AdhocComponentWithVariants::class.java).forEach { c ->
+    c.withVariantsFromConfiguration(project.configurations.shadowRuntimeElements.get()) {
+        skip()
+    }
+}
+
```
RCHowell, line 84: Why does partiql-lang have an ANTLR runtime dependency to begin with? Same with the CSV dependency (that's a CLI concern no?). I'm wondering if these deps were mistakenly not removed during package split.

RCHowell (changes requested):

partiql-lang/build.gradle.kts
alancai98: Discussed offline. Adding the notes here:
- https://github.com/partiql/partiql-lang-kotlin/blob/shade-antlr/partiql-lang/src/main/kotlin/org/partiql/lang/eval/io/DelimitedValues.kt#L73-L105 is where the CSV dependency is used
- https://github.com/partiql/partiql-lang-kotlin/blob/shade-antlr/partiql-lang/src/main/kotlin/org/partiql/lang/syntax/impl/PartiQLPigParser.kt uses the ANTLR runtime

alancai98:
Confirmed that all of the subprojects published to Maven have the same POM dependencies with all of the `ANTLR` + `ANTLR-runtime` dependencies removed.

partiql-parser/build.gradle.kts
```diff
@@ -23,7 +23,35 @@ dependencies {
...
+}
+
+// Workaround for https://github.com/johnrengelman/shadow/issues/651
+components.withType(AdhocComponentWithVariants::class.java).forEach { c ->
+    c.withVariantsFromConfiguration(project.configurations.shadowRuntimeElements.get()) {
+        skip()
+    }
+}
+
+// `antlr` dependency configuration adds the ANTLR API configuration (and Maven `compile` dependency scope on publish)
```
alancai98, line 49: This is a known issue w/ the ANTLR gradle plugin. Workaround here removes the `antlr` Gradle `api` (equivalent to Maven `compile`) dependency.

alancai98: There are some build issues following this change. I'll take another look

johnedquinn:
I see that you've published locally (`publishToMavenLocal`) to look at the POM. Have you created a new project to test that you can use the `partiql-parser` package? I'm having trouble in my own build. I had previously locally published 0.13.3-SNAPSHOT and I've just now locally published 0.14.6-SNAPSHOT (including these changes). For some reason, I can't resolve 0.14.6-SNAPSHOT when I specify it in build.gradle.kts, but when I specify the 0.13.3-SNAPSHOT, it works.

For context, I was going to run some tests to make sure I can use `org.partiql.parser.PartiQLParser`. I was also going to check that `org.antlr.*` would not be able to be used. Then, I was going to add another dependency on a different version of the ANTLR runtime to see if the resolution would be okay when actually using some of the runtime classes (`org.antlr.v4.runtime.Token` for instance).

Here's my build script:
```kotlin
plugins {
    id("java")
}

group = "org.partiql.shade-test"
version = "1.0-SNAPSHOT"

repositories {
    mavenCentral()
    mavenLocal()
}

dependencies {
    // For testing
    testImplementation("org.partiql:partiql-parser:0.13.3-SNAPSHOT")
    // testImplementation("org.partiql:partiql-parser:0.14.6-SNAPSHOT")

    testImplementation(platform("org.junit:junit-bom:5.9.1"))
    testImplementation("org.junit.jupiter:junit-jupiter")
}

tasks.test {
    useJUnitPlatform()
}
```

alancai98:
> I see that you've published locally (`publishToMavenLocal`) to look at the POM. Have you created a new project to test that you can use the `partiql-parser` package? I'm having trouble in my own build. I had previously locally published 0.13.3-SNAPSHOT and I've just now locally published 0.14.6-SNAPSHOT (including these changes). For some reason, I can't resolve 0.14.6-SNAPSHOT when I specify it in build.gradle.kts, but when I specify the 0.13.3-SNAPSHOT, it works.
> 
> For context, I was going to run some tests to make sure I can use `org.partiql.parser.PartiQLParser`. I was also going to check that `org.antlr.*` would not be able to be used. Then, I was going to add another dependency on a different version of the ANTLR runtime to see if the resolution would be okay when actually using some of the runtime classes (`org.antlr.v4.runtime.Token` for instance).
> ...

Thanks for testing the local publishing on your end. My `build.gradle.kts` was pretty similar to yours except for specifying the Kotlin plugin.

I was able to reproduce some build errors you're probably experiencing when pulling in the local published version of this branch (0.14.6-SNAPSHOT) and using an **empty** Maven local cache. I had not seen those errors in previous testing since I had locally published 0.14.6-SNAPSHOT before the latest commit of this PR. My guess is that some combination of the previous local published Maven artifacts and the latest PR's local publishing allowed my local build to succeed.

I'll try fixing the publishing task and performing further testing.

alancai98:
Did some more local testing on my end after the most recent commits.

Process I followed:
1. delete local `~/.m2` directory for partiql
2. ensure `DOKKA` environment is set to `true`
3. published the `shade-antlr` branch to maven local (`./gradlew publishToMavenLocal`) as `0.14.6-SNAPSHOT`
4. created a local project with a gradle build similar to https://github.com/partiql/partiql-lang-kotlin/pull/1439#issuecomment-2103567854

Testing done:
1. Confirmed I could pull in `org.partiql:partiql-parser:0.14.6-SNAPSHOT` from maven local and run code from `parser` package
2. Confirmed I could pull in `org.partiql:partiql-lang-kotlin:0.14.6-SNAPSHOT` from maven local and run code from `lang` package
3. With just the `partiql-parser:0.14.6-SNAPSHOT` dependency, confirmed that ANTLRv4 dependencies are under the `org.partiql.thirdparty` namespace. Confirmed that I could not import APIs from `org.antlr.v4` directly.
4. With just the `partiql-parser:0.14.6-SNAPSHOT` dependency and other ANTLR runtime versions (4.5.1, 4.13.1), confirmed there were no version conflicts when building and running parser code.
5. Took a look at the POM file diffs with v0.14.5. Confirmed that for `partiql-parser` and `partiql-lang` that the ANTLR dependencies are removed; other dependencies are the same.

partiql-lang/build.gradle.kts
```diff
@@ -64,6 +64,24 @@ dependencies {
     }
 }
 
+val relocations = mapOf(
+    "org.antlr" to "org.partiql.thirdparty.antlr"
```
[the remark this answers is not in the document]
alancai98: I had thought about this issue previously and tested locally. Seems like there's no collision:

1. Defining a project w/ just the `partiql-lang-kotlin` dependency, `org.partiql.thirdparty.antlr.v4.runtime.Token` would resolve to use `partiql-lang-kotlin`'s definition.
2. Defining a project w/ `partiql-lang-kotlin` and `partiql-parser` dependency (in both orderings in the build.gradle.kts file), `org.partiql.thirdparty.antlr.v4.runtime.Token` resolves to use `partiql-lang-kotlin`'s defintion.

Based on above, it seems to always resolve to the `lang` dependency.

Not sure if there's a use-case in which we would have different versions of ANTLR for `lang` and `parser` since the `lang` dependency should match the generated `parser`'s dependency. To prevent any potential issues w/ namespacing, I'll add an additional package suffix before `thirdparty`.

partiql-planner/build.gradle.kts
```diff
@@ -112,19 +112,46 @@ tasks.register<Copy>("copyUtils") {
...
+//     into("src/main/kotlin")
+// }
 
 tasks.register("generate") {
     dependsOn("codegen", "copyUtils")
 }
 
 tasks.compileKotlin {
     dependsOn("generate")
+    dependsOn(tasks.withType<Copy>())
```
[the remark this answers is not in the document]
alancai98: These additions to other packages (i.e. everything not `lang` and `planner`) were added following the upgrade to gradle 8+. In our previous gradle version, 7+, there were some task dependencies that weren't deterministic, which became errors in gradle 8+. So these additional `dependsOn` blocks were needed for this PR.

The shadowjar plugin is added to all of the published packages to keep the publishing configuration the same for all packages. The published artifacts for the other packages (not `lang` or `parser`) will be the same since there's no `shadowJar` configuration in their gradle build files.

partiql-lang/build.gradle.kts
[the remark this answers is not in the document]
alancai98: With the above change to namespacing, users can be sure where the ANTLR runtime dependency is coming from:

```
import org.partiql.lang.thirdparty.antlr.v4.runtime.Token as PartiQLParserToken
import org.partiql.parser.thirdparty.antlr.v4.runtime.Token as PartiQLLangToken
```

johnedquinn (approved):

PR #1442, opened 2024-04-22
Support parsing for attribute and tuple level constraint

yliuuuu:
## Relevant Issues

## Description
- Supports parsing of Attribute-Level Constraint (CHECK, NULL, NOT NULL, UNIQUE, PRIMARY KEY)
- Supports parsing of Table-Level Constraints (CHECK, UNIQUE, PRIMARY KEY)
- It contains breaking changes on the AST modeling. in particular remodeling of constraints to be unnested from table definition. 
- This is because the constraint can be reused in multiple parts of the DDL operation, (in this PR, attribute level constraint and table level constraints; in the future, we might needs to use constraint in ALTER Statement). 

### Added
- Supports parsing of Attribute-Level Constraint (CHECK, NULL, NOT NULL, UNIQUE, PRIMARY KEY)
- Supports parsing of Table-Level Constraints (CHECK, UNIQUE, PRIMARY KEY)
### Changed
- Modeling of DDL node. 
    - DDL node is modeled as a wrapper around DDLOp. For example: replace `statementDDLCreateTable(...)` to `statementDDL(ddlOpCreateTable(....))`

## Other Information
  -No. Targeting V1 branch. 

  -  Yes. AST modeling. 
  
  - No. 
  
- Yes. 

## License Information

### DIFF
--- a/partiql-ast/src/main/resources/partiql_ast.ion
+++ b/partiql-ast/src/main/resources/partiql_ast.ion
@@ -784,14 +784,12 @@ returning::{
   ],
 }
 
-// `<column_name> <type> <column_constraint>*`
-// `( CONSTRAINT <column_constraint_name> )?  <column_constraint_def>`
 table_definition::{
-  columns: list::[column],
+  attributes: list::[attribute],
   // table level constraints
   constraints: list::[constraint],
   _: [
-    column::{
+    attribute::{
       name:        '.identifier.symbol',
       type:        '.type',
       constraints: list::[constraint],
@@ -801,13 +799,13 @@ table_definition::{
 
 constraint::{
   name: optional::string,
-  body: [
+  definition: [
     nullable::{},
     not_null::{},
     check::{ expr: expr },
     unique::{
       // for attribute level constraint, we can set this attribute to null
-      columns: optional::list::['.identifier.symbol'],
+      attributes: optional::list::['.identifier.symbol'],
       is_primary_key: bool,
     },
   ],
--- a/partiql-ast/src/main/resources/partiql_ast.ion
+++ b/partiql-ast/src/main/resources/partiql_ast.ion
@@ -106,32 +106,9 @@ statement::[
   ],
 
   // Data Definition Language
-  d_d_l::[
-
-    // CREATE TABLE <identifier> [<table_def>]
-    create_table::{
-      name:       identifier,
-      definition: optional::table_definition,
-    },
-
-    // CREATE INDEX [<identifier>] ON <identifier> (<path> [, <path>]...)
-    create_index::{
-      index:  optional::identifier,
-      table:  identifier,
-      fields: list::[path],
-    },
-
-    // DROP TABLE <identifier>
-    drop_table::{
-      table: identifier,
-    },
-
-    // DROP INDEX <identifier> ON <identifier>
-    drop_index::{
-      index:  identifier, // <identifier>[0]
-      table:  identifier, // <identifier>[1]
-    },
-  ],
+  d_d_l::{
+    op: ddl_op
+  },
 
   // EXEC <symbol> [<expr>.*]
   exec::{
@@ -148,6 +125,32 @@ statement::[
         format:     optional::string,
       },
     ],
+  },
+]
+
+ddl_op::[
+  // CREATE TABLE <identifier> [<table_def>]
+  create_table::{
+    name:       identifier,
+    definition: optional::table_definition,
+  },
+
+  // CREATE INDEX [<identifier>] ON <identifier> (<path> [, <path>]...)
+  create_index::{
+    index:  optional::identifier,
+    table:  identifier,
+    fields: list::[path],
+  },
+
+  // DROP TABLE <identifier>
+  drop_table::{
+    table: identifier,
+  },
+
+  // DROP INDEX <identifier> ON <identifier>
+  drop_index::{
+    index:  identifier, // <identifier>[0]
+    table:  identifier, // <identifier>[1]
   },
 ]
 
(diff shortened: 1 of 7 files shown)


partiql-ast/src/main/resources/partiql_ast.ion
```diff
@@ -785,23 +788,27 @@ returning::{
...
-          ],
-        },
-      ],
+    }
+  ],
+}
+
+constraint::{
+  name: optional::string,
+  body: [
```
alancai98, line 804: (naming) not sure if `body` is a good name. Perhaps could use `type` or something else?

partiql-parser/src/test/kotlin/org/partiql/parser/internal/PartiQLParserDDLTests.kt
```diff
@@ -77,27 +107,232 @@ class PartiQLParserDDLTests {
                     null
                 )
             ),
+
+            //
```
alancai98, line 111: nice set of tests. could also consider adding a few more:
- multiple constraints
- search condition w/ an `OR` (like what was mentioned in the ANTLR grammar
- other parsing error edge cases
  - missing parens
  - table constraint used where attribute constraint should be placed (and vice-versa)

partiql-lang/src/test/kotlin/org/partiql/lang/syntax/PartiQLParserDDLTest.kt
```diff
@@ -35,7 +35,47 @@ internal class PartiQLParserDDLTest : PartiQLParserTestBase() {
                 query = "DROP Table foo.bar",
                 code = ErrorCode.PARSE_UNEXPECTED_TOKEN,
                 context = mapOf(),
-            )
+            ),
```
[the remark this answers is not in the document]
yliuuuu: Probably not, as the old parser is getting deprecated. 

Those tests are set to ignore context, hence the context here is an empty map. But in Pig visitor code an error message is included.

partiql-parser/src/test/kotlin/org/partiql/parser/internal/PartiQLParserDDLTests.kt
yliuuuu: Good suggestion, we can add more tests once the parser portion of DDL implementation is completed. Since we are going to add more features, the modeling will change hence the test code.

partiql-ast/src/main/resources/partiql_ast.ion
yliuuuu: I do not like the name either. 

To address this and other naming suggestion in the comments: 

The name are reused from the previous implementation. I am more than happy to change those: 

"column" -> "attribute" 
"body" -> "definition"

partiql-ast/src/main/resources/partiql_ast.ion
```diff
@@ -785,23 +788,27 @@ returning::{
...
-            check::{ expr: expr },
-          ],
-        },
-      ],
+    }
+  ],
+}
+
+constraint::{
+  name: optional::string,
```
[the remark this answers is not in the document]
yliuuuu: This is a good call out. 

Sql's grammar is constructed this way because they treat constraint as an object... 

The semantics with constraint name is not clear at the moment, but constraint name was modeled as a string previously in our grammar.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -588,62 +592,94 @@ internal class PartiQLParserDefault : PartiQLParser {
...
 
         override fun visitCreateTable(ctx: GeneratedParser.CreateTableContext) = translate(ctx) {
             val table = visitQualifiedName(ctx.qualifiedName())
             val definition = ctx.tableDef()?.let { visitTableDef(it) }
-            statementDDLCreateTable(table, definition)
+            ddlOpCreateTable(table, definition)
         }
 
         override fun visitCreateIndex(ctx: GeneratedParser.CreateIndexContext) = translate(ctx) {
             // TODO add index name to ANTLR grammar
```
alancai98, line 617: Is this TODO still relevant based on the updated DDL modeling?

partiql-parser/src/main/antlr/PartiQL.g4
```diff
@@ -100,17 +99,43 @@ tableDef
...
 
 columnConstraint
-    : ( CONSTRAINT columnConstraintName )?  columnConstraintDef
+    : ( CONSTRAINT constraintName )?  columnConstraintDef
     ;
 
 columnConstraintDef
-    : NOT NULL                                  # ColConstrNotNull
-    | NULL                                      # ColConstrNull
+    : NOT NULL                                     # ColConstrNotNull
```
alancai98, line 115: Since the partiql-ast now uses "attribute" in the table definition, should the ANTLR definitions also use "attribute" rather than "column"? I think we should keep the parser and ast terminology consistent.

Related -- would the PIG parser also need to be updated to use "attribute"?

partiql-ast/src/main/resources/partiql_ast.ion
```diff
@@ -785,23 +788,27 @@ returning::{
...
-            not_null::{},
-            check::{ expr: expr },
-          ],
-        },
-      ],
+    }
+  ],
+}
+
+constraint::{
```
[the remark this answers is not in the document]
alancai98: Based on offline discussion, there's a lot of overlap between the "table" and "attribute" constraints hence why they're modeled as one "constraint" node here. @yliuuuu had also mentioned that this unified modeling made the conversion to the plan a bit simpler. As we continue to work on the rest of DDL modeling in the AST and plan we can adjust the modeling if needed.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
yliuuuu: Yes. Current Index grammar is untouched.

partiql-parser/src/main/antlr/PartiQL.g4
yliuuuu: I think this is a valid call out. But given the fact that Antlr grammar is an internal concern, perhaps we can treat this as a backlog item? 

I don't think we want to introduce changes to the PIG parser.

yliuuuu: At some point we would want to update the ANTLR grammar on our [website](https://partiql.org/syntax/antlr.html). Perhaps then would be a good time to do those changes in batch?

alancai98: > I think this is a valid call out. But given the fact that Antlr grammar is an internal concern, perhaps we can treat this as a backlog item?

Ah, I hadn't realized the current ANTLR-generated parse code was meant to be "internal" despite the visibility modifiers indicating it was "public". Created an issue to track the internalization of the ANTLR-generated sources -- https://github.com/partiql/partiql-lang-kotlin/issues/1446.

> At some point we would want to update the ANTLR grammar on our [website](https://partiql.org/syntax/antlr.html). Perhaps then would be a good time to do those changes in batch?

Since we'll likely have other changes to the ANTLR grammar before the DDL syntax is finalized, I'm fine w/ punting on the changes. Could you create an issue to track updating the website grammar and grammar within `partiql-parser`? That should make it easier to track what'll need changed before the next release.

alancai98 (approved):

PR #1443, opened 2024-04-23
Initializes PartiQLData and PartiQLValueLoader

johnedquinn:
## Relevant Issues
- None

## Description
- Initializes `PartiQLData` and `PartiQLValueLoader`.
- I modeled this with [JDBC's ResultSet](https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html) and [Ion's IonReader](https://www.javadoc.io/doc/software.amazon.ion/ion-java/latest/software/amazon/ion/IonReader.html) in mind.
- I also updated `PartiQLResult.Value` to hold `PartiQLData` -- NOT a `PartiQLValue`.
- Note: I decided to write this in Java.

## Reviewer Notice
- The second commit removes all CLI tests as the KOW branch's CLI has been completely overhauled causing this PR to fail the build. For the reviewer, I'd recommend reviewing the first commit by itself and then checking to see that I didn't accidentally delete any other files in the second commit.

## Other Information

## License Information


partiql-eval/src/main/kotlin/org/partiql/eval/PartiQLResult.kt
```diff
@@ -1,12 +1,12 @@
...
 
-import org.partiql.value.PartiQLValue
+import org.partiql.value.PartiQLData
 import org.partiql.value.PartiQLValueExperimental
 
 public sealed interface PartiQLResult {
 
     @OptIn(PartiQLValueExperimental::class)
-    public data class Value(public val value: PartiQLValue) : PartiQLResult
+    public data class Value(public val value: PartiQLData) : PartiQLResult
```
RCHowell, line 9: We should probably have a little doc on naming now. The Value/value and type not being a PartiQLValue is interesting, but perhaps to get the diff small. I do think we should incorporate "cursor" (a'la SQL) or "reader" (a'la Ion) such as

```
public data class Value(public val cursor: ResultCursor) : PartiQLResult
public data class Value(public val reader: PartiQLReader) : PartiQLResult
```

PartiQLReader / PartiQLCursor / PartiQLData / etc etc

partiql-types/src/main/java/org/partiql/value/PartiQLData.java
```diff
@@ -0,0 +1,158 @@
...
+     * there are more children remaining.
+     */
+    public void stepIn();
+
+    /**
+     * Positions the iterator after the current parent's value, moving up one level in the data hierarchy. There's no
+     * current value immediately after stepping out, so the next thing you'll want to do is call {@link #hasNext()} and
+     * {@link #next()} to move onto the following value.
+     */
+    public void stepOut();
```
RCHowell, line 41: Why no container type arg?

partiql-types/src/main/java/org/partiql/value/PartiQLData.java
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * This is only applicable when the current value's type is {@link PartiQLValueType#FLOAT32}.
+     */
+    public float getFloatValue();
+
+    /**
+     * This is only applicable when the current value's type is {@link PartiQLValueType#FLOAT64}.
+     */
+    public double getDoubleValue();
```
RCHowell, line 143: Interesting seeing the Java type names appear in the method names. Let's discuss. I'd prefer the partiql type names to be the method names since this is the interface (and may appear in other language implementations). 

I advise to say `public long getInt64Value()` or something similar just using the PartiQLValue names exactly

RCHowell (changes requested):

partiql-types/src/main/java/org/partiql/value/PartiQLValueLoader.java
```diff
@@ -0,0 +1,30 @@
...
+     * must be set <b>before</b> the value that you'd like to load.
+     * <p>
+     * This method will invoke {@link PartiQLData#next()}. This method will <b>not</b> throw an error is there is
+     * more data to be processed after the value immediately following the cursor.
+     *
+     * @param data the PartiQL data to load.
+     * @return a materialized, in-memory instance of a {@link PartiQLValue} containing the contents of the {@code data}.
+     */
+    @NotNull
+    PartiQLValue loadSingleValue(@NotNull PartiQLData data);
```
RCHowell, line 22: Why single value? Would we ever have an Iterator<PartiQLValue> rather than a CollectionValue. Should just be `load`

partiql-eval/src/main/kotlin/org/partiql/eval/PartiQLResult.kt
johnedquinn: Updated the PartiQLData to be PartiQLCursor. Yeah, I opted to not rename `PartiQLResult.Value` to something else (maybe `PartiQLResult.Data` or `PartiQLResult.Cursor`) due to the large diff that would occur. That can be a follow-up to this PR.

partiql-eval/src/main/kotlin/org/partiql/eval/PartiQLEngineDefault.kt
```diff
@@ -33,7 +34,8 @@ internal class PartiQLEngineDefault : PartiQLEngine {
         return when (statement) {
             is PartiQLStatement.Query -> try {
                 val value = statement.execute()
```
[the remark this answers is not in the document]
johnedquinn: Yes -- see my other response.

partiql-types/src/main/java/org/partiql/value/PartiQLData.java
johnedquinn: Would there be a need to pass a type? We already receive enough information from `next()`, `getType()`, and `isNullValue()` to determine whether we can/should invoke `stepIn()` and `stepOut()`.

RCHowell (approved):

Issue #1446, opened 2024-04-25
Internalize ANTLR-generated parser sources


alancai98:
As of v0.14.5, all the ANTLR-generated parser sources in the `partiql-parser` subproject are marked as public. See the docs -- https://javadoc.io/doc/org.partiql/partiql-parser/0.14.5/partiql-parser/org.partiql.parser.antlr/index.html. We currently do not intend for these to be implemented outside of other packages within our codebase (ideally just `partiql-parser`).

We should try to do one or a combination of the following:
1. Make the ANTLR-generated parser package-private. It's currently Java code and not Kotlin code so package-private rather than `internal`.
2. Add a comment that the ANTLR-generated sources are intended to be implemented only by other partiql packages (similar to what we had done for sprout-generated interfaces -- https://github.com/partiql/partiql-lang-kotlin/pull/1413).
3. Move the ANTLR-generated sources behind an `internal` namespace and add a comment that we do not intend for APIs under the `internal` namespace are not intended to be implemented.

alancai98:
Likely can't do the first point (make package-private) until we deprecate and remove the PIG-parser from `partiql-ast`.

alancai98:
#1452 was merged to `v1` branch. Change will be part of the next major version release.

PR #1447, opened 2024-04-25
run apiDump

yliuuuu:
## Relevant Issues

## Description
- Running `./gradlew apiDump` to fix build. 

## Other Information
  - < If NO, why? >

  - < If YES, why? >

## License Information

### DIFF
--- a/partiql-ast/api/partiql-ast.api
+++ b/partiql-ast/api/partiql-ast.api
@@ -1,4 +1,13 @@
 public final class org/partiql/ast/Ast {
+	public static final fun constraint (Ljava/lang/String;Lorg/partiql/ast/Constraint$Definition;)Lorg/partiql/ast/Constraint;
+	public static final fun constraintDefinitionCheck (Lorg/partiql/ast/Expr;)Lorg/partiql/ast/Constraint$Definition$Check;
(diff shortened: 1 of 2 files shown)


yliuuuu:
One observation is the that API dumps captures changes in ANTLR generated files. It would be nice to filter out those. 

Perhaps we want to: 
1. package ANTRL code in an package called `xxx.internal.antlr` 
2. filter out anything source under `internal` package when running `apiDump`

alancai98 (approved):

alancai98:
> One observation is the that API dumps captures changes in ANTLR generated files. It would be nice to filter out those. 
>
> Perhaps we want to: 
> 1. package ANTRL code in an package called `xxx.internal.antlr` 
> 2. filter out anything source under `internal` package when running `apiDump`

Agree that it'll be good to do both of those.

1. is tracked in https://github.com/partiql/partiql-lang-kotlin/issues/1446.
2. we should do as part of the package audit process before v1. For reference, the compatibility checker gradle plugin is pretty easy to configure to ignore certain class paths -- https://github.com/Kotlin/binary-compatibility-validator#optional-parameters. Let's tackle the filtering/excluding in subsequent PRs.

PR #1449, opened 2024-04-27
struct subfield and list element type

yliuuuu:
## Relevant Issues

## Description
- Adding support for parsing Struct type with field declaration and List with Element type in DDL. 
     - Note that for operator that takes in type as parameter: `IS`, `CAST`, etc, we do not allow struct type with field declaration or list type with element type declaration at the moment. 
- Removing support for using NULL and MISSING as data type in DDL. 

## Other Information
  - < If NO, why? >

  - < If YES, why? >

## License Information


partiql-ast/src/main/resources/partiql_ast.ion
```diff
@@ -196,10 +196,29 @@ type::[
...
   bag::{},                                                     // BAG
+  // Keep the list type for backward compatible reason
+  // user should migrate to the array type
   list::{},                                                    // LIST
+  array::{ type: optional::type },                             // ARRAY
   sexp::{},                                                    // SEXP
+  // Keep the tuple type for backward compatible reason
+  // migration: in v0.15.0: struct modeling should NOT be Introduced,
+  //  users of struct type should be migrate to tuple type.
+  //  in v1.0.0: introduce the struct modeling change, all users from tuple type should migrate to struct type.
```
alancai98, line 207: Comment is a little confusing. If I understand correctly,
1. `struct` is added in v0.15.0
2. `tuple` is kept in v0.15.0 but "deprecated"; users are advised to use `struct`
3. `tuple` is removed in v1.0.0 and users have to use `struct`?

partiql-ast/src/main/resources/partiql_ast.ion
```diff
@@ -196,10 +196,29 @@ type::[
...
   tuple::{},                                                   // TUPLE
-  struct::{},                                                  // STRUCT
+  struct::{
+    fields: list::[field],
```
alancai98, line 210: So is the list of fields always required for structs? It appears somewhat different than the other types with parameters, which are all optional parameters.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -636,15 +636,43 @@ internal class PartiQLParserDefault : PartiQLParser {
...
+            else -> Unit
+        }
+
+        private fun isValidTypeParameterOrThrow(type: Type, ctx: GeneratedParser.TypeContext) = when (type) {
+            is Type.Struct -> {
+                if (type.fields.isNotEmpty()) throw error(ctx, "use parameterized struct as type parameter is not supported")
+                else Unit
+            }
+            is Type.Array -> {
+                if (type.type != null) throw error(ctx, "use parameterized list as type parameter is not supported")
```
alancai98, line 664:
```suggestion
                if (type.fields.isNotEmpty()) throw error(ctx, "using parameterized struct as type parameter is not supported")
                else Unit
            }
            is Type.Array -> {
                if (type.type != null) throw error(ctx, "using parameterized array as type parameter is not supported")
```

partiql-ast/src/main/resources/partiql_ast.ion
alancai98: Wondering if there is a distinction between a struct with unspecified fields and a struct with no specified fields (i.e. empty list)?

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -2189,6 +2214,39 @@ internal class PartiQLParserDefault : PartiQLParser {
             typeCustom(ctx.text.uppercase())
         }
 
+        override fun visitTypeComplexLegacy(ctx: GeneratedParser.TypeComplexLegacyContext) = translate(ctx) {
+            when (ctx.datatype.type) {
+                GeneratedParser.STRUCT, GeneratedParser.TUPLE -> typeStruct(emptyList())
+                GeneratedParser.ARRAY, GeneratedParser.LIST -> typeArray(null)
```
alancai98, line 2220: Based on the grammar for `TypeComplexLegacy`, array won't be a possible type:

```
datatype=(STRUCT|TUPLE|LIST)                                                                                     # TypeComplexLegacy
```

Perhaps `ARRAY` would need added to the `TypeComplexLegacy` node?

partiql-parser/src/test/kotlin/org/partiql/parser/internal/PartiQLParserDDLTests.kt
```diff
@@ -312,6 +312,60 @@ class PartiQLParserDDLTests {
                     ),
                 )
             ),
+
+            SuccessTestCase(
+                "CREATE TABLE with CASE SENSITIVE Identifier as column name",
+                """
```
alancai98, line 318: Could include a few more tests:
- struct with no attributes
- struct with multiple attributes
- struct with a field that has constraints
- array with no element type
- nested cases
  - array with a struct element type
  - struct type with an array field type

partiql-ast/src/main/resources/partiql_ast.ion
```diff
@@ -196,10 +196,29 @@ type::[
...
+      field :: {
+        name: '.identifier.symbol',
+        type: '.type',
+        // This could be a boolean flag since we only support NOT NULL constraint
+        // for struct subfield. But modeling this to be a list of constraints
+        // to prevent future breaking changes.
+        constraints: list::[constraint],
+      }
+    ],
+  },                                                           // STRUCT
```
alancai98, line 221: similarly for `STRUCT`

```suggestion
  },                                                           // STRUCT <fields>
```

partiql-ast/src/main/resources/partiql_ast.ion
yliuuuu: `Struct<>` and `LIST<>` is not permitted by the ANTLR Grammar to avoid this confusion.

partiql-ast/src/main/resources/partiql_ast.ion
yliuuuu: Tuple/Struct is a bit more difficult, let's first talk about Array/List

Consider: 

```
// in 0.14.x, we only have typeList
// LIST get routed to typeList()
typeList()

// In 0.15.x 
@Deprecated, migrate to new APIs 
typeList()
typeArray(elementType = ...)

// in 1.0.x 
typeArray(elementType = ...)
```
The List/Array migration is pretty clean due to the added typeArray()

Consider the Tuple/Struct Migration.... We could do something similar, but since this PR targets the 1.0, I just directly modified the API. 

The comment purposed a different migration strategy: 
```
// v0.14.x 
typeStruct()
typeTuple() 

// v0.15.x
@Deprecated, move the usage to typeTuple()
typeStruct()

// v0.16.x -- intermediate to avoid confusion, we don't really need this
typeTuple()

// v0.17.x
@Deprecated
typeTuple()
typeStruct(fields = ....)

// v1.0.x 
typeStruct(fields = ...)
```

The goal for this PR is to demonstrate the APIs (`typeStruct` and `typeArray`) in v1 release, without actually deleting other APIs.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
yliuuuu: Array is not previously allowed, Array here should be a dead branch. I will remove it.

partiql-ast/src/main/resources/partiql_ast.ion
```diff
@@ -196,10 +196,29 @@ type::[
   timestamp_with_tz::{ precision: optional::int },             // TIMESTAMP [(<int>)] WITH TIMEZONE
   interval::{ precision: optional::int },                      // INTERVAL
   bag::{},                                                     // BAG
+  // Keep the list type for backward compatible reason
+  // user should migrate to the array type
   list::{},                                                    // LIST
+  array::{ type: optional::type },                             // ARRAY
   sexp::{},                                                    // SEXP
```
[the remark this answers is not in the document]
yliuuuu: SEXP was treated separately as of today, I would be happy to see it get deprecated but this should not be within the scope of this PR.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
yliuuuu: I see what you saying is that parse of ARRAY with out any element type will not be allowed if we do not put it there. Let me change the name legacy to unparameterized.

partiql-parser/src/main/antlr/PartiQL.g4
```diff
@@ -812,12 +812,19 @@ type
...
+        | STRING | SYMBOL | BLOB | CLOB | DATE | SEXP | BAG | ANY
       )                                                                                                                # TypeAtomic
+    | datatype=(STRUCT|TUPLE|LIST|ARRAY)                                                                               # TypeComplexUnparameterized
     | datatype=DOUBLE PRECISION                                                                                        # TypeAtomic
     | datatype=(CHARACTER|CHAR|FLOAT|VARCHAR) ( PAREN_LEFT arg0=LITERAL_INTEGER PAREN_RIGHT )?                         # TypeArgSingle
     | CHARACTER VARYING ( PAREN_LEFT arg0=LITERAL_INTEGER PAREN_RIGHT )?                                               # TypeVarChar
     | datatype=(DECIMAL|DEC|NUMERIC) ( PAREN_LEFT arg0=LITERAL_INTEGER ( COMMA arg1=LITERAL_INTEGER )? PAREN_RIGHT )?  # TypeArgDouble
     | datatype=(TIME|TIMESTAMP) ( PAREN_LEFT precision=LITERAL_INTEGER PAREN_RIGHT )? (WITH TIME ZONE)?                # TypeTimeZone
+    | datatype=(STRUCT|TUPLE) (ANGLE_LEFT structAttr ( COMMA structAttr )* ANGLE_RIGHT)                                # TypeStruct
```
am357, line 823: Since this is going to the `1.0`, do we want to flesh out the open/close and other struct constraints in the syntax now; if not, how can we avoid backward incompatibility when we introduce the syntax for those?

am357, line 823: Since this is going to the `1.0`, do we want to flesh out the open/close and other struct constraints in the syntax now; if not, how can we avoid backward incompatibility when we introduce the syntax for those?

partiql-ast/src/main/resources/partiql_ast.ion
alancai98: The deprecation plan for ARRAY/LIST makes sense to me.

---

Followup to the offline discussion, I think condensing the changes across fewer major version bumps can be easier to understand. Consider:

```
// v0.14.x -- the current state in `main`
typeTuple()

// v0.15.x -- the next release
@Deprecated // recommend using `typeStruct`
typeTuple()
typeStruct()

// v1.0.x
typeStruct() // `typeTuple` now deleted
```

alancai98: Whatever release process we choose though, I think this comment in the `partiql_ast.ion` file should be updated.

partiql-ast/src/main/resources/partiql_ast.ion
alancai98: From offline discussion, the semantics for `STRUCT<>` and `LIST<>` is currently undefined. Can decide in the future whether this
- is allowed
- if allowed, does it have different semantics from `STRUCT` and `LIST` (without any fields or element specified within the angle brackets)

partiql-parser/src/main/antlr/PartiQL.g4
```diff
@@ -812,12 +812,19 @@ type
...
+    | datatype=(LIST|ARRAY) ANGLE_LEFT type ANGLE_RIGHT                                                                # TypeList
     | symbolPrimitive                                                                                                  # TypeCustom
     ;
+
+structAttr
```
alancai98, line 828: nit: maybe should be called `structField` for consistency with the ast definition?

partiql-parser/src/test/kotlin/org/partiql/parser/internal/PartiQLParserDDLTests.kt
```diff
@@ -371,7 +555,69 @@ class PartiQLParserDDLTests {
...
+                    CREATE TABLE TBL(
+                        a LIST<>
+                    )
+                """.trimIndent()
+            ),
+
+            // TODO: Fix Me
+            //  Potentially modify the bag grammar to use
+            //  ANGLE_LEFT ANGLE_LEFT ( expr ( COMMA expr )* )? ANGLE_RIGHT ANGLE_RIGHT
+            //  We should look into the multi-character token in ANTLR grammar
```
alancai98, line 595: Could cut an issue if we're not tackling it in the short-term

yliuuuu: Issue is fixed in the new commit.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -1546,13 +1574,27 @@ internal class PartiQLParserDefault : PartiQLParser {
...
+
+        private fun convertBinaryMathOp(token: Token) = when (token.type) {
+            GeneratedParser.AMPERSAND -> Expr.Binary.Op.BITWISE_AND
+            GeneratedParser.CONCAT -> Expr.Binary.Op.CONCAT
+            GeneratedParser.PLUS -> Expr.Binary.Op.PLUS
+            GeneratedParser.MINUS -> Expr.Binary.Op.MINUS
+            GeneratedParser.PERCENT -> Expr.Binary.Op.MODULO
+            GeneratedParser.ASTERISK -> Expr.Binary.Op.TIMES
+            GeneratedParser.SLASH_FORWARD -> Expr.Binary.Op.DIVIDE
+            else -> throw error(token, "Invalid binary math operator")
```
alancai98, line 1597: nit: error message can be confusing since `AMPERSAND` and `CONCAT` aren't math operators. Could consider reusing the `convertBinaryOp` error message.

```suggestion
            else -> throw error(token, "Invalid binary operator")
```

alancai98: Also naming for the function can be confusing (not that big an issue since it's private). Could just use the name `convertBinaryOp` since the first argument is different. And add a comment that this is used for `mathOp00`, `mathOp01`, and `mathOp02`.

partiql-parser/src/test/kotlin/org/partiql/parser/internal/PartiQLParserDDLTests.kt
```diff
@@ -312,6 +313,220 @@ class PartiQLParserDDLTests {
...
+                        ),
+                        emptyList()
+                    ),
+                )
+            ),
+            SuccessTestCase(
+                "CREATE TABLE no space between angle right",
+                """
+                    CREATE TABLE tbl(
+                        a LIST<STRUCT<b:INT2>>
```
alancai98, line 481: nice that the double right angle parses now. should this use `ARRAY` since `LIST` will be deprecated/removed?

partiql-parser/src/test/kotlin/org/partiql/parser/internal/PartiQLParserDDLTests.kt
```diff
@@ -312,6 +313,220 @@ class PartiQLParserDDLTests {
...
+                                emptyList(),
+                            )
+                        ),
+                        emptyList()
+                    ),
+                )
+            ),
+
+            SuccessTestCase(
+                "CREATE TABLE with LIST without element type",
```
alancai98, line 510:
```suggestion
                "CREATE TABLE with ARRAY without element type",
```

partiql-parser/src/test/kotlin/org/partiql/parser/internal/PartiQLParserDDLTests.kt
```diff
@@ -371,7 +586,57 @@ class PartiQLParserDDLTests {
...
+            ErrorTestCase(
+                "NULL not allowed as type in type decalration",
+                """
+                    CREATE TABLE TBL(
+                        a NULL
+                    )
+                """.trimIndent()
+            ),
+            ErrorTestCase(
+                "MISSING not allowed as type in type decalration",
```
alancai98, line 599:
```suggestion
                "NULL not allowed as type in type declaration",
                """
                    CREATE TABLE TBL(
                        a NULL
                    )
                """.trimIndent()
            ),
            ErrorTestCase(
                "MISSING not allowed as type in type declaration",
```

partiql-ast/src/main/resources/partiql_ast.ion
```diff
@@ -196,10 +196,46 @@ type::[
...
+  // @Deprecated // recommend using `typeArray()`
+  // typeList()
+  // typeArray(...)
+  //
+  // v1.0.x
+  // typeArray(...) // `typeList` now deleted
   list::{},                                                    // LIST
+  array::{ type: optional::type },                             // ARRAY [<type>]
   sexp::{},                                                    // SEXP
+  // Potential migration stretegy:
```
alancai98, line 213:
```suggestion
  // Potential migration strategy:
  // v0.14.x -- the current state in `main`
  // typeList()
  //
  // v0.15.x -- the next release
  // @Deprecated // recommend using `typeArray()`
  // typeList()
  // typeArray(...)
  //
  // v1.0.x
  // typeArray(...) // `typeList` now deleted
  list::{},                                                    // LIST
  array::{ type: optional::type },                             // ARRAY [<type>]
  sexp::{},                                                    // SEXP
  // Potential migration strategy:
```

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -1546,13 +1574,27 @@ internal class PartiQLParserDefault : PartiQLParser {
...
             GeneratedParser.ANGLE_RIGHT -> Expr.Binary.Op.GT
             GeneratedParser.GT_EQ -> Expr.Binary.Op.GTE
-            GeneratedParser.NEQ -> Expr.Binary.Op.NE
+            GeneratedParser.BANG -> Expr.Binary.Op.NE
             GeneratedParser.EQ -> Expr.Binary.Op.EQ
-            else -> throw error(token, "Invalid binary operator")
+            else -> throw error(ctx.start, "Invalid binary operator")
+        }
+
+        private fun convertBinaryMathOp(token: Token) = when (token.type) {
```
alancai98, line 1589: Taking another look at this, do we need to separate the binary ops between two separate functions? The [PartiQLPigVisitor](https://github.com/partiql/partiql-lang-kotlin/blob/v1-struct-list/partiql-lang/src/main/kotlin/org/partiql/lang/syntax/impl/PartiQLPigVisitor.kt#L1736) just has one binary op conversion function?

partiql-parser/src/test/kotlin/org/partiql/parser/internal/PartiQLParserDDLTests.kt
yliuuuu: I think we still permit parsing of LIST type, it is just that the AST node we use for LIST/ARRAY should be the same.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
yliuuuu: PIG AST needs to do the get all tokens to have the right error location, hence the list and the fold. 

Error location functionality is not finalized and tested in the new AST. Hence the two separate function to reduce the diff.

partiql-parser/src/main/antlr/PartiQL.g4
yliuuuu: In my mental model, I think that using SQL's `CREATE TABLE` get us a bag of closed struct. For creating shapes that is unique to PartiQL, a bag of opened struct, a bag of integers, etc. we can consider vend out a PartiQL specific syntax (i.e., CREATE SHAPE ?) for those.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
alancai98: Thanks for the clarification

alancai98 (approved):

partiql-parser/src/main/antlr/PartiQL.g4
am357: Just like other deviations from Hive DDL, we may have to support such use-cases within the same syntax; let's discuss further about this.

am357: Adding the summary of our discussion in an example here for posterity:

```SQL
CREATE TABLE [OPEN] Foo (
   s STRUCT<a: INT, b:STRING> OPEN,
)
```

PR #1451, opened 2024-04-27
Initializes Datum and adds functionality to evaluator

johnedquinn:
## Relevant Issues
- None

## Description
- Initializes Datum, the engine's representation of a value
-  Also and adds functionality to evaluator. Currently, Datum is all over the evaluator, however, we convert to PartiQLValue when calling UDFs and doing comparisons.
- While working on the JOIN PR (#1438) and PartiQL Cursor PR (#1443), it became apparent to me that our modelling of PartiQLValue is tightly coupled with Java's type semantics. In many places, we rely on Java's types (example:` x is StructValue<*>`) rather than relying on our plumbing: (`x.type == PartiQLValueType.STRUCT`). On top of that, as part of the JOIN PR, I found it incredibly difficult (and sub-optimal performance-wise) to coerce nulls of a particular type to a null of another type. This PR takes these learnings and completely overhauls `PartiQLValue` to a representation that we are all familiar with: Ion's `AnyElement`. In a way, this modelling is as close as we can get to C's Union type, and by modelling it this way, we gain control over our type system.
- Notice that the API proposed is called `PQLValue`. I intend for its core to be copied and pasted directly into `PartiQLValue`. I have also written this in Java, as I've noticed our inability to control our public API by writing in Kotlin. On top of that, `PQLValue` is written in Java to get us _closer_ to using primitives directly. That being said, `PQLValue` will NOT exist during the `V1` release. Its contents will replace the contents of `PartiQLValue` -- and `PartiQLValue` will be written in Java.

## Internal Notes
- Notice that I attempt to keep the null-safety that Kotlin provides by marking all applicable API's with `@NotNull`.
- Similarly, I attempt to use primitives as much as possible -- limiting our memory footprint and maintaining the null-safety.
- I've left remaining action items to fully replace PartiQLValue in the KDocs of `PQLValue`. These include adding the comparator, equals, hashCode, etc. These should be able to be added fairly quickly.

## Other Information
- Any backward-incompatible changes? **NO** -- not really, however, PQLValue intends to replace PartiQLValue

## License Information


partiql-eval/src/main/java/org/partiql/eval/StructValue.java
```diff
@@ -0,0 +1,37 @@
...
+    public Iterator<StructField> getStructFields() {
+        return _value.iterator();
+    }
+
+    @NotNull
+    @Override
+    public PartiQLValueType getType() {
+        return PartiQLValueType.STRUCT;
+    }
+}
```
RCHowell, line 37: We should have a map and multi-map implementation. When a "partiql struct" does not have duplicate keys, then we use the map which is key->value, but when there are duplicates, we store linked-lists as the map entries.

So we have `key->list<field>` where the field is as earlier which preserves the original field name case. Keys are all case-normalized

partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
```diff
@@ -380,7 +381,7 @@ internal class Compiler(
 
     @OptIn(PartiQLValueExperimental::class)
     override fun visitRexOpLit(node: Rex.Op.Lit, ctx: StaticType?): Operator {
-        return ExprLiteral(node.value)
+        return ExprLiteral(PQLValue.of(node.value))
```
RCHowell, line 384: Perfect. Here is one of the bridges from public API to engine internals.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/helpers/ValueUtility.kt
```diff
@@ -0,0 +1,97 @@
...
+    fun PQLValue.getInt32Coerced(): Int {
+        return when (this.type) {
+            PartiQLValueType.INT8 -> this.int8Value.toInt()
+            PartiQLValueType.INT16 -> this.int16Value.toInt()
+            PartiQLValueType.INT32 -> this.int32Value
+            PartiQLValueType.INT64 -> this.int64Value.toInt()
+            PartiQLValueType.INT -> this.intValue.toInt()
+            else -> throw TypeCheckException()
+        }
+    }
```
RCHowell, line 96: I suggest adding all of these coerces to the PQLValue implementations.

```java
class Int32Value ... {

    private Int value;

    @Override
    short toShort() { 
       // throw DataException on overflow  etc.
      return value as short;
     }

    @Override
    int getInt() { return value };

    @Override
    long getLong() { return value };
}

**Review by RCHowell — CHANGES_REQUESTED** [review]

File: partiql-eval/src/main/java/org/partiql/eval/BagValue.java [hunk 4]
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Collections;
+import java.util.Iterator;
+
+/**
+ * This shall always be package-private (internal).
+ */
+class BagValue implements PQLValue {
+
+    @NotNull
+    final Iterable<PQLValue> _value;
```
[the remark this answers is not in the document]
johnedquinn: Class is package-private

partiql-eval/src/main/java/org/partiql/eval/CharValue.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.math.BigInteger;
+
+/**
+ * This shall always be package-private (internal).
+ */
+class CharValue implements PQLValue {
+
+    @NotNull
+    final String _value;
```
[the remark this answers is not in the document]
johnedquinn: The character data type in SQL doesn't represent the char in Java. It's more like a char array

partiql-eval/src/main/java/org/partiql/eval/ListValue.java
```diff
@@ -0,0 +1,37 @@
...
+    @Override
+    public boolean isNull() {
+        return false;
+    }
+
+    @Override
+    @NotNull
+    public Iterator<PQLValue> getListValues() {
+        return _value.iterator();
+    }
```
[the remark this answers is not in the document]
johnedquinn: From previous discussions on PartiQLCursor, it seemed that we wanted to be explicit about the value's type when calling getters. AKA getInt8Value and getByteValue return the same primitive, but a decision was made to keep the APIs distinct

partiql-eval/src/main/java/org/partiql/eval/PQLValue.java
```diff
@@ -0,0 +1,615 @@
...
+
+    @NotNull
+    static PQLValue missingValue() {
+        return new MissingValue();
+    }
+
+    @NotNull
+    static PQLValue nullValue(@NotNull PartiQLValueType type) {
+        return new NullValue(type);
+    }
```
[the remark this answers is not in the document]
johnedquinn: In order to pass existing tests, this is necessary. I can mark it as deprecated, but for this PR I'd like to avoid removing PartiQLValueType.NULL as it touches many things

partiql-eval/src/main/java/org/partiql/eval/Int32Value.java
```diff
@@ -0,0 +1,31 @@
+package org.partiql.eval;
```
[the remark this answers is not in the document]
johnedquinn: Mentioned in other comment, but the lack of a visibility modifier implies the "default" visibility (which is package private). Therefore, these classes need to be placed in the same package as PQLValue. That being said, I'm going to release an iteration of this where it is placed in `partiql-types` -- based on our offline discussion.

partiql-eval/src/main/java/org/partiql/eval/PQLValue.java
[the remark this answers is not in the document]
johnedquinn: See PR #1463. This will have a pretty significant impact on this comment.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/helpers/ValueUtility.kt
johnedquinn: See https://github.com/partiql/partiql-lang-kotlin/pull/1451/commits/7f345443ec0d3479fcba47f1931eb0fec3370ef2#diff-7273fc22dadb098a56a8f9e14afb134d7fdce01af8709eff841f91576bee9d3e

partiql-eval/src/main/java/org/partiql/eval/CharValue.java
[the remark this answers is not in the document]
RCHowell: Oh the fixed-length character. In my head I had

```
char <-> char(1) -- fixed-length character with length 1
```

partiql-eval/src/main/java/org/partiql/eval/value/DecimalArbitraryValue.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.jetbrains.annotations.NotNull;
+import org.partiql.value.PartiQLValueType;
+
+import java.math.BigDecimal;
+
+/**
+ * This shall always be package-private (internal).
+ */
+class DecimalArbitraryValue implements PQLValue {
```
RCHowell, line 11: I don't think we need both Decimal and DecimalArbitrary for runtime values.

partiql-eval/src/main/java/org/partiql/eval/value/Float32Value.java
```diff
@@ -0,0 +1,32 @@
...
+    @Override
+    public boolean isNull() {
+        return false;
+    }
+
+    @Override
+    public float getFloat() {
+        return _value;
+    }
+
```
RCHowell, line 26: This is starting to look exactly like PartiQLValue just in Java and with the `getFloat()` .. `getLong()` etc. I have a couple of questions.

- What's the difference now? (besides flat class hierarchy)
- Why don't we have the coercions in the PQLValue? (in this case `getDouble()`)

Perhaps its been too long since I looked, but I don't think I understand the value prop here.

partiql-eval/src/main/java/org/partiql/eval/value/PQLValue.java
```diff
@@ -0,0 +1,546 @@
...
+ * <p></p>
+ * This is intended to completely replace {@link org.partiql.value.PartiQLValue} in the future. As it stands, this
+ * implementation will initially be used solely for the evaluator. However, the scope of this can be expanded by copying
+ * and pasting its contents to completely replace {@link org.partiql.value.PartiQLValue}.
+ * <p></p>
+ * There are some pre-requisites to actually replacing {@link PartiQLValue} including, but not limited to, ...:
+ * - The comparator for ordering and aggregations
+ * - Adding support for annotations
+ */
+public interface PQLValue extends Iterable<PQLValue> {
```
RCHowell, line 34: I don't think it should be public if it's not a replacement. I also thought we discussed them solving different problems?

We shouldn't introduce another duplication of something existing when PartiQLValue is _still_ under experimental. We either replace it completely or change it however we need to.

partiql-eval/src/main/java/org/partiql/eval/value/PQLValue.java
```diff
@@ -0,0 +1,546 @@
...
+     *                                       {@link #isNull()} returns false before attempting to invoke this method.
+     */
+    default short getShort() {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * @return the underlying value applicable to the types:
+     * {@link PartiQLValueType#INT32}
+     * @throws UnsupportedOperationException if the operation is not applicable to the type returned from
```
RCHowell, line 190: I'm confused why coercions wouldn't fit like Int16Value

partiql-eval/src/main/java/org/partiql/eval/value/BagValue.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.jetbrains.annotations.NotNull;
+import org.partiql.value.PartiQLValueType;
+
+import java.util.Iterator;
+
+/**
+ * This shall always be package-private (internal).
+ */
+class BagValue implements PQLValue {
```
RCHowell, line 11: I don't think we need Bag/List/Sexp classes if we have a type enum. Because this is always internal, we can add additional classes later, but it doesn't provide additional value. Same for Decimal/DecimalArbitrary.

partiql-eval/src/main/java/org/partiql/eval/value/CharValue.java
```diff
@@ -0,0 +1,34 @@
+package org.partiql.eval.value;
+
+import org.jetbrains.annotations.NotNull;
+import org.partiql.value.PartiQLValueType;
+
+/**
+ * This shall always be package-private (internal).
+ */
+class CharValue implements PQLValue {
```
RCHowell, line 9: Similar comment, but seems we don't need Char and String considering their identical implementations.

partiql-eval/src/main/java/org/partiql/eval/value/ClobValue.java
```diff
@@ -0,0 +1,34 @@
+package org.partiql.eval.value;
+
+import org.jetbrains.annotations.NotNull;
+import org.partiql.value.PartiQLValueType;
+
+/**
+ * This shall always be package-private (internal).
+ */
+class ClobValue implements PQLValue {
```
RCHowell, line 9: Clob/Blob -> ByteArray or Lob

RCHowell (changes requested):

partiql-eval/src/main/java/org/partiql/eval/value/Float32Value.java
johnedquinn: Yes, this PR aims to replace PartiQLValue incrementally. I'd suggest:
1. Add PQLValue without releasing.
2. Update functions to use PQLValue. (subsequent PR)
3. Move the existing PartiQLValue to the plan package (maybe called PlanLiteral) (since it's use is more for the DOM)
4. Rename PQLValue to PartiQLValue

The difference is that we are moving away from using Java's type semantics in favor of our own type system. With this, we can/should be using our own type information to determine how to extract the underlying data. The idea is that this logic should be exposed by `partiql-eval` instead of `partiql-types`.

There shouldn't be a need for coercions that are not represented in the plan. What I'd like to see is that when the executor receives a plan, without any modification, the executor shall execute the plan. If a JVM double is expected, then the planner should have inserted the appropriate coercion.

Consider `Rex.Op.Path.Index`. The planner currently allows all integer types as the index, and then the executor is actually under the hood coercing any integer value into a JVM int. In my opinion, the "function signature" of the path index should have required that the index be a PartiQL int_32. With this mindset, the expressions representing the index (say its typed to be an INT16) would be wrapped with a coercion by the planner. We're currently making up for our lack of planning by adding all of these additional coercions in the execution.

RCHowell: Discussed other points offline to create more distance from DOM.

> There shouldn't be a need for coercions that are not represented in the plan. What I'd like to see is that when the executor receives a plan, without any modification, the executor shall execute the plan. If a JVM double is expected, then the planner should have inserted the appropriate coercion.

Agreed, makes sense now. Thank you.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/helpers/RecordValueIterator.kt
```diff
@@ -1,15 +1,13 @@
...
- * An [Iterator] over a [CollectionValue] lazily producing [Record]s as you call [next].
+ * An [Iterator] over an [Iterator] lazily producing [Record]s as you call [next].
  */
-@OptIn(PartiQLValueExperimental::class)
 internal class RecordValueIterator(
-    collectionValue: CollectionValue<*>
+    collectionValue: Iterator<Datum>
 ) : Iterator<Record> {
 
     private val collectionIter = collectionValue.iterator()
```
RCHowell, line 13: Nit, but now we can have `private val iterator: Iterator<Dataum>`

RCHowell (changes requested):
Only change is to fix the test/partiql-tests submodule otherwise ship it.

johnedquinn:
> Only change is to fix the test/partiql-tests submodule otherwise ship it.

I believe there is a bug in GitHub's rendering of the submodule diff. For some reason, the commit is the same on `v1` as on `johnedquinn:v1-eval-value-rebased`. Checked the files being "changed" and they're all already in `v1`. I pushed twice to move to some older submodule commit and then back to what is currently on v1, but it's still not rendering appropriately.

RCHowell (approved):

PR #1453, opened 2024-05-02
Eradicates premature failures when visiting error nodes

johnedquinn:
## Relevant Issues
- None

## Description

According to the PartiQL Specification:
> For example, in the presence of schema validation, an PartiQL query processor can throw a compile-time error
> when given the path expression {a:1, b:2}.c. In a more important and common case, an PartiQL implementation can
> utilize the input data schema to prove that a path expression always returns MISSING and thus throw a
> compile-time error.

So, this translates to deferring failure until evaluation. While it is sub-optimal, it complies with the specification. We can always change the testing mechanism in the future.

Beyond that, I found that we had some bugs due to throwing errors when we saw unresolved nodes in the PlanTransform. The root cause is that we were appending unresolved nodes to the "causes" list of Rex.Op.Err. Therefore, with the introduction of Rex.Op.Err, yes -- we would temporarily handle errors, however, when we'd visit those errors, we would prematurely fail compilation.

This PR fixes both of the above situations.

## Other Information

## License Information

### DIFF
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt
@@ -123,8 +123,9 @@ internal class PlanTransform(
         override fun visitRexOpVar(node: Rex.Op.Var, ctx: Unit) =
             super.visitRexOpVar(node, ctx) as org.partiql.plan.Rex.Op
 
-        override fun visitRexOpVarUnresolved(node: Rex.Op.Var.Unresolved, ctx: Unit) =
-            error("The Internal Plan Node Rex.Op.Var.Unresolved should be converted to an MISSING Node during type resolution if resolution failed")
+        override fun visitRexOpVarUnresolved(node: Rex.Op.Var.Unresolved, ctx: Unit): org.partiql.plan.Rex.Op.Err {
+            return org.partiql.plan.Rex.Op.Err("Unresolved variable: ${node.identifier}", emptyList())
+        }
 
         override fun visitRexOpVarGlobal(node: Rex.Op.Var.Global, ctx: Unit) = org.partiql.plan.Rex.Op.Global(
             ref = visitRef(node.ref, ctx)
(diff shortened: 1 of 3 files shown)


partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt
```diff
@@ -157,11 +158,9 @@ internal class PlanTransform(
...
             super.visitRexOpPath(node, ctx) as org.partiql.plan.Rex.Op.Path
 
-        override fun visitRexOpCast(node: Rex.Op.Cast, ctx: Unit) =
-            super.visitRexOpCast(node, ctx) as org.partiql.plan.Rex.Op.Cast
-
-        override fun visitRexOpCastUnresolved(node: Rex.Op.Cast.Unresolved, ctx: Unit): PlanNode {
-            error("Unresolved cast $node")
+        @OptIn(PartiQLValueExperimental::class)
+        override fun visitRexOpCastUnresolved(node: Rex.Op.Cast.Unresolved, ctx: Unit): org.partiql.plan.Rex.Op {
+            return rexOpErr("Unresolved cast: ${node.arg.type} to ${node.target}.", emptyList())
```
RCHowell, line 163: I don't think the public plan should have error nodes btw... I suggest PlanTransform use the antlr style error listeners. It's a bit silly to hide errors in a tree, then make the compiler later throw them.

We have _two_ error nodes at the moment, but should only have _one_.

We have,
 - rexOpError
 - rexOpMissing

The _rexOpError_ should NOT exist in the public plan IR, but _rexOpMissing_ should. The _rexOpMissing_ operation represents a typing error that either throws or becomes MISSING in the evaluator.

The _rexOpError_ is an outright error that will _never_ compile.

RCHowell (changes requested):

johnedquinn:
> > So, this translates to deferring failure until evaluation.
> 
> The specification says most commonly you want to error, but _can_ defer. This is what [Yingtao's PR for planner modes](https://github.com/partiql/partiql-lang-kotlin/pull/1385) handles. I urge that signal mode is the default.

The focus of this PR is to bridge the conformance gap (this PR passes an additional 20 tests). The existing conformance tests defer on erroring as does the existing evaluator. To accommodate your suggestion, there are two things I could do:
1. Update the tests to allow for erroring during compilation.
2. Or, add a flag to the engine (not default) to defer on erroring. Then, in the conformance tests runs, I can use the non-default flag to match the functionality of the existing evaluator.

RCHowell:
I still don't quite understand the purpose here other than conformance tests being wrong. I believe Yingtao's planner work and the compiler implementation are correct.

**HOWEVER**

I think perhaps we should be making changes in the _planner_ not the compiler.

**Here are the Options afaik** 

1. If we can confirm that the tests are **correct**, then the fix should go into the planner.
2. If the tests are **incorrect**, then we can close this and open an issue in partiql-tests

PR #1455, opened 2024-05-02
V1 ddl extended keyword

yliuuuu:
## Relevant Issues

## Description
- Extended DDL Features: 
    -  OPTIONAL Keyword: Declared after the attribute name to mark the attribute as optional. 
    -  COMMENT Keyword: Declared at the end of attribute declaration to add an description to the attribute. 
    -  TBLPROPERTIES Keyword: Table Level properties (arbitrary key value pair).
    -  PARTITION BY keyword: Partition By Clause, for now we only support list of attribute for Partition Expression. 

## Other Information
  - < If NO, why? >

  - < If YES, why? >

## License Information


partiql-ast/src/main/resources/partiql_ast.ion
```diff
@@ -829,6 +837,9 @@ table_definition::{
       name:        '.identifier.symbol',
       type:        '.type',
       constraints: list::[constraint],
+      // Should this be a constraint?
+      is_optional: bool,
```
am357, line 841: Maybe we can prepend this with a `TODO` to make discoverable later. Also, on syntax, do this matter to be a constraint or not?

partiql-parser/src/main/antlr/PartiQL.g4
```diff
@@ -136,6 +137,19 @@ uniqueConstraintDef
 // but we at least can eliminate SFW query here.
 searchCondition : exprOr;
 
+// SQL Extension, Support additional table metadatas such as partition by, tblProperties, etc.
```
am357, line 140:
```suggestion
// SQL/HIVE DDL Extension, Support additional table metadatas such as partition by, tblProperties, etc.
```

partiql-parser/src/main/antlr/PartiQL.g4
```diff
@@ -136,6 +137,19 @@ uniqueConstraintDef
...
+tableExtension
+    : PARTITION BY partitionExpr                                                           # PartitionBy
+    | TBLPROPERTIES PAREN_LEFT keyValuePair (COMMA keyValuePair)* PAREN_RIGHT              # TblProperties
+    ;
+
+keyValuePair : key=LITERAL_STRING EQ value=literal;
```
am357, line 146: Not sure if the full `literal` is required. Maybe it can get limited to  `LITERAL_STRING` for now?

partiql-parser/src/test/kotlin/org/partiql/parser/internal/PartiQLParserDDLTests.kt
```diff
@@ -521,10 +596,284 @@ class PartiQLParserDDLTests {
...
+                    null,
+                    emptyList()
+                )
+            ),
+
+            // Partition BY
+            SuccessTestCase(
+                "CREATE TABLE with Partition by single attribute",
+                """
+                    CREATE TABLE tbl
```
am357, line 814: Shouldn't this be syntax error?

yliuuuu: I think this should be semantic error, i.e., referring a non-existing attribute in Partition By expression.

partiql-parser/src/main/antlr/PartiQL.g4
yliuuuu: I do not think we should permit Literal, i.e., key to be a string or timestamp does not make much sense to me. 

I think (and forget to leave a comment here), the question is more towards `identifier` vs `string`? 

i.e., 

```
TBLPROPERTIES(myKey = 'some value')
```

vs 
```
TBLPROPERTIES('myKey' = 'someValue' )
```

Reason why this PR leans towards using string is 
1. this syntax are based from HIVE and HIVE uses string. 
2. Supporting string as key value seems reasonable to me and I don't think we need to deprecate this later. 
3. Supporting identifier as string seems to be a ergonomic feature and most likely we will process the identifier as string in subsequent process (as opposite to attempting to bind the identifier to a variable...). If you agree this is the case then we can always add the support later.

partiql-ast/src/main/resources/partiql_ast.ion
yliuuuu: I don't think it matters on the syntax. 

The only thing is that Attribute Constraint in SQL is often after the data type declaration, but this is more of a terminology things.

partiql-parser/src/main/antlr/PartiQL.g4
am357: I also agree with the key being `string`, but what I meant was do we need `value=literal`:

```
literal
    : NULL                                                                                # LiteralNull
    | MISSING                                                                             # LiteralMissing
    | TRUE                                                                                # LiteralTrue
    | FALSE                                                                               # LiteralFalse
    | LITERAL_STRING                                                                      # LiteralString
    | LITERAL_INTEGER                                                                     # LiteralInteger
    | LITERAL_DECIMAL                                                                     # LiteralDecimal
    | ION_CLOSURE                                                                         # LiteralIon
    | DATE LITERAL_STRING                                                                 # LiteralDate
    | TIME ( PAREN_LEFT LITERAL_INTEGER PAREN_RIGHT )? (WITH TIME ZONE)? LITERAL_STRING   # LiteralTime
    | TIMESTAMP ( PAREN_LEFT LITERAL_INTEGER PAREN_RIGHT )? (WITH TIME ZONE)? LITERAL_STRING   # LiteralTimestamp
    ;

```

partiql-parser/src/test/kotlin/org/partiql/parser/internal/PartiQLParserDDLTests.kt
am357: But is the following in a valid syntax considering it lacks the parenthesis and column definitions?

```sql
CREATE TABLE tbl PARTITION BY (a)
```
See:
https://www.db-fiddle.com/f/6PztiiVM6uEzpvdWnMyKX7/0

yliuuuu: I think so because 
```
CREATE TABLE tbl
```
has been a valid syntax for us. 

Based on this I am treating the grammar as three parts: 

```
CREATE TABLE tbl  -- header
(a INT2) -- data type definition
PARTITION BY (a) -- table extension
```

partiql-parser/src/main/antlr/PartiQL.g4
yliuuuu: Oh I see. 

I am not against limiting the scope to string value only, but I don't see any drawback to support all other values. It could be useful for customer who wants to specify like `TBLPROPERTIES('isActive' = false)`.

am357: Perhaps would be good for us to avoid widening the scope for now :)

partiql-parser/src/main/antlr/PartiQL.g4
```diff
@@ -136,6 +137,19 @@ uniqueConstraintDef
...
+
+// For now: just support a list of column name
+// In the future, we might support common partition expression such as Hash(), Range(), etc.
+partitionExpr
+    : PAREN_LEFT columnName (COMMA columnName)* PAREN_RIGHT                   #PartitionColList
```
alancai98, line 151: Based on your above comment, I'd assume the other parts of HIVE DDL for `PARTITION BY`/`PARTITIONED BY` are out of scope

> `  [PARTITIONED BY (col_name data_type [COMMENT col_comment], ...)]`

namely the requirement to specify a data type along with an optional comment

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -610,7 +613,27 @@ internal class PartiQLParserDefault : PartiQLParser {
...
+                .filterIsInstance<GeneratedParser.PartitionByContext>()
+                .let {
+                    if (it.size > 1) throw error(ctx, "EXPECT ONE PARTITION BY CLAUSE")
+                    it.firstOrNull()?.let { visitPartitionBy(it) }
+                }
+            val tblProperties = ctx
+                .tableExtension()
+                .filterIsInstance<GeneratedParser.TblPropertiesContext>()
+                .let {
+                    if (it.size > 1) throw error(ctx, "EXPECT ONE TBLPROPERTIES CLAUSE")
```
alancai98, line 627: very nit: any reason these error messages are in caps? SQL does have a lot of all-caps, but it feels a little out of place given the other error messages aren't in all-caps

partiql-parser/src/test/kotlin/org/partiql/parser/internal/PartiQLParserDDLTests.kt
```diff
@@ -521,10 +596,284 @@ class PartiQLParserDDLTests {
...
+                    TBLPROPERTIES ('k1' = 'v1', 'k2' = 2)
+                """.trimIndent(),
+                ddlOpCreateTable(
+                    identifierSymbol("tbl", Identifier.CaseSensitivity.INSENSITIVE),
+                    null,
+                    null,
+                    listOf(
+                        tableProperty("k1", stringValue("v1")),
+                        tableProperty("k2", int32Value(2))
+                    )
```
alancai98, line 876: Could also add a test using both a `PARTITION BY` clause along w/ a `TBLPROPERTIES`

partiql-parser/src/main/antlr/PartiQL.g4
```diff
@@ -136,6 +137,19 @@ uniqueConstraintDef
 // but we at least can eliminate SFW query here.
 searchCondition : exprOr;
 
+// SQL Extension, Support additional table metadatas such as partition by, tblProperties, etc.
+tableExtension
+    : PARTITION BY partitionExpr                                                           # PartitionBy
```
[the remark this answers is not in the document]
alancai98: Also noticed that `PARTITION` was already an existing keyword that's used for window functions:

> ```windowPartitionList
>   : PARTITION BY expr (COMMA expr)*
>   ;```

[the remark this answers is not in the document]
yliuuuu: A partitioned table is not in the SQL spec, hence the syntax is an extension made by various DB implementation. 

The reason we choice to not following HIVE but PostgreSQL is to avoided the cognitive overhead. 

`PARTITION BY (a)` vs `PARTITIONED BY (a INT2)`. 

 That is: The `PARTITION BY ` clause can take in a Partition Expression, and the `PARTITIONED BY` clause (with is not supported yet) can take in attribute declaration.

partiql-parser/src/main/antlr/PartiQL.g4
yliuuuu: I would treat `PARTITION BY` and `PARTITIONED BY` as two different clauses.

partiql-parser/src/main/antlr/PartiQL.g4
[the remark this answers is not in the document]
alancai98: I see. Thanks for the added context. Could be helpful that context in a comment for why we choose PostgreSQL's syntax here rather than HIVE's

alancai98 (approved):

PR #1457, opened 2024-05-06
Add to non-reserved keywords; rework `functionName` and `symbolPrimitive` parse rules

alancai98:
## Description
- Adds non-reserved keywords from SQL-99 (https://ronsavage.github.io/SQL/sql-99.bnf.html#non-reserved%20word)
- Reworks the following ANTLR rules
  - `functionName` -- gets rid of hard-coded reserved keywords as function names block
  - `symbolPrimitive` -- adds `nonReserved` keywords to this rule

## Other Information
  - No, v1 branch

- Any backward-incompatible changes? **[NO?]**
  - Will have some behavioral changes related to some invalid queries not resulting in a parse error (will still be an evaluation error)

## License Information

### DIFF
--- a/partiql-lang/src/test/kotlin/org/partiql/lang/errors/ParserErrorsTest.kt
+++ b/partiql-lang/src/test/kotlin/org/partiql/lang/errors/ParserErrorsTest.kt
@@ -15,6 +15,7 @@
 package org.partiql.lang.errors
 
 import com.amazon.ion.Timestamp
+import org.junit.Ignore
 import org.junit.Test
 import org.partiql.errors.ErrorCode
 import org.partiql.errors.Property
@@ -655,6 +656,7 @@ class ParserErrorsTest : PartiQLParserTestBase() {
     }
 
     @Test
+    @Ignore("No longer a parser error; evaluation error")
     fun callTrimNoArgs() {
         checkInputThrowingParserException(
             "trim()",
@@ -683,6 +685,7 @@ class ParserErrorsTest : PartiQLParserTestBase() {
     }
 
     @Test
+    @Ignore("No longer a parser error; evaluation error")
     fun callTrimZeroArguments() {
         checkInputThrowingParserException(
             "trim()",
@@ -795,6 +798,7 @@ class ParserErrorsTest : PartiQLParserTestBase() {
     }
 
     @Test
+    @Ignore("No longer a parser error; evaluation error")
     fun aggregateWithNoArgs() {
         checkInputThrowingParserException(
             "SUM()",
@@ -809,6 +813,7 @@ class ParserErrorsTest : PartiQLParserTestBase() {
     }
 
     @Test
+    @Ignore("No longer a parser error; evaluation error")
     fun aggregateWithTooManyArgs() {
         checkInputThrowingParserException(
             "SUM(a, b)",
@@ -1672,6 +1677,7 @@ class ParserErrorsTest : PartiQLParserTestBase() {
     }
 
     @Test
+    @Ignore("No longer a parser error; evaluation error")
     fun callExtractMissingFromWithComma() {
         checkInputThrowingParserException(
             "extract(year, b)",
@@ -1717,6 +1723,7 @@ class ParserErrorsTest : PartiQLParserTestBase() {
     }
 
     @Test
+    @Ignore("No longer a parser error; evaluation error")
     fun callExtractOnlySecondArgument() {
         checkInputThrowingParserException(
             "extract(b)",
@@ -1746,6 +1753,7 @@ class ParserErrorsTest : PartiQLParserTestBase() {
     }
 
     @Test
+    @Ignore("No longer a parser error; evaluation error")
     fun callExtractOnlyDateTimePart() {
         checkInputThrowingParserException(
             "extract(year)",
--- a/partiql-parser/src/main/antlr/PartiQL.g4
+++ b/partiql-parser/src/main/antlr/PartiQL.g4
@@ -763,12 +763,7 @@ dateFunction
 
 // SQL-99 10.4 — <routine invocation> ::= <routine name> <SQL argument list>
 functionCall
-    : functionName PAREN_LEFT ( expr ( COMMA expr )* )? PAREN_RIGHT
-    ;
-
-// SQL-99 10.4 — <routine name> ::= [ <schema name> <period> ] <qualified identifier>
-functionName
-    : (qualifier+=symbolPrimitive PERIOD)* name=symbolPrimitive
+    : qualifiedName PAREN_LEFT ( expr ( COMMA expr )* )? PAREN_RIGHT
     ;
 
 pathStep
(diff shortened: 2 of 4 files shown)


partiql-parser/src/main/antlr/PartiQL.g4
```diff
@@ -40,7 +40,9 @@ byIdent
     : BY symbolPrimitive;
 
 symbolPrimitive
-    : ident=( IDENTIFIER | IDENTIFIER_QUOTED )
+    : IDENTIFIER            # UnquotedIdentifier
+    | IDENTIFIER_QUOTED     # QuotedIdentifier
+    | nonReserved           # UnquotedIdentifier
```
RCHowell, line 45: The reverse naming sorts the long method names together

```suggestion
    : IDENTIFIER            # IdentifierUnquoted
    | IDENTIFIER_QUOTED     # IdentifierQuoted
    | nonReserved           # IdentifierUnquoted
```

partiql-lang/src/test/kotlin/org/partiql/lang/errors/ParserErrorsTest.kt
```diff
@@ -655,6 +656,7 @@ class ParserErrorsTest : PartiQLParserTestBase() {
     }
 
     @Test
+    @Ignore("No longer a parser error; evaluation error")
```
RCHowell, line 659: Exactly

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -557,29 +557,29 @@ internal class PartiQLParserDefault : PartiQLParser {
...
+                Identifier.CaseSensitivity.SENSITIVE
+            )
+        }
+
+        override fun visitUnquotedIdentifier(ctx: org.partiql.parser.internal.antlr.PartiQLParser.UnquotedIdentifierContext): Identifier.Symbol = translate(ctx) {
+            identifierSymbol(
+                ctx.text,
+                Identifier.CaseSensitivity.INSENSITIVE
+            )
         }
```
RCHowell, line 578: Hmm, why do we need the `visitAs` when we have explicit return types? I wonder if adding the two alternatives caused this. I wonder if we should define our own `getIdentifier` that is more efficient than visitAs.

So it would be like
```kotlin
fun identifier(ctx: ParserRuleContext): Identifier = when (ctx) {
     is IdentifierQuotedContext -> identifier(ctx)
     is IdentifierUnquotedContext -> identifier(ctx)
}

fun identifier(ctx: IdentifierQuotedContext): Identifier = ....

fun identifier(ctx: IdentifierUquotedContext): Identifier.Symbol = ....
```

We also should take care where qualified names appear and make sure our grammar rules follow this.

RCHowell (changes requested):

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
RCHowell: Like, I think we should be replacing `visitSymbolPrimitive` with `visitIdentifierUnquoted` and it shouldn't be anymore than a name change. Is there a reason this did not work?

alancai98: There wasn't a reason this did not work previously. I was just following the `visitAs<...> ()` pattern I often saw in the visitors. I'll update to use a `getIdentifier` (or the like) rather than `visitAs`.

partiql-parser/src/main/antlr/PartiQL.g4
```diff
@@ -766,8 +768,7 @@ functionCall
 
 // SQL-99 10.4 — <routine name> ::= [ <schema name> <period> ] <qualified identifier>
 functionName
-    : (qualifier+=symbolPrimitive PERIOD)* name=( CHAR_LENGTH | CHARACTER_LENGTH | OCTET_LENGTH | BIT_LENGTH | UPPER | LOWER | SIZE | EXISTS | COUNT | MOD )  # FunctionNameReserved
```
alancai98, line 769: (self-review) -- this is the rule we were aiming to remove from the ANTLR grammar in this PR

partiql-parser/src/main/antlr/PartiQL.g4
```diff
@@ -789,11 +790,52 @@ parameter
...
     : qualifier=AT_SIGN? ident=(IDENTIFIER|IDENTIFIER_QUOTED)   # VariableIdentifier
-    | qualifier=AT_SIGN? key=nonReservedKeywords                # VariableKeyword
-    ;
-
-nonReservedKeywords
-    : EXCLUDED
+    | qualifier=AT_SIGN? key=nonReserved                # VariableKeyword
+    ;
+
+nonReserved
```
alancai98, line 796: (self-review) -- the [non-reserved words](https://ronsavage.github.io/SQL/sql-99.bnf.html#non-reserved%20word) were taken directly from the SQL-99 grammar. At the end are some, PartiQL specific-keywords marked as non-reserved.

If you'll notice, [postgresql](https://www.postgresql.org/docs/current/sql-keywords-appendix.html) along with other SQL dialects use their own set of non-reserved keywords that diverges from the SQL standards. Also between different SQL standards, there are some changes to the reserved vs non-reserved keywords. We choose to follow SQL-99 as we do in other places in the project.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
alancai98: New commit [445be01](https://github.com/partiql/partiql-lang-kotlin/pull/1457/commits/445be01db417657469c8b833b93c6b7b4b63965c) uses the private `visitSymbolPrimitive` function which calls the corresponding `visitIdentifierQuoted` and `visitIdentifierUnquoted`. Also changed some existing usages of `visitAs<Identifier.Symbol>` to use `visitSymbolPrimitive`.

johnedquinn (changes requested):

partiql-parser/src/main/antlr/PartiQL.g4
```diff
@@ -766,8 +768,7 @@ functionCall
 
 // SQL-99 10.4 — <routine name> ::= [ <schema name> <period> ] <qualified identifier>
 functionName
-    : (qualifier+=symbolPrimitive PERIOD)* name=( CHAR_LENGTH | CHARACTER_LENGTH | OCTET_LENGTH | BIT_LENGTH | UPPER | LOWER | SIZE | EXISTS | COUNT | MOD )  # FunctionNameReserved
-    | (qualifier+=symbolPrimitive PERIOD)* name=symbolPrimitive                                                                                         # FunctionNameSymbol
+    : (qualifier+=symbolPrimitive PERIOD)* name=symbolPrimitive
```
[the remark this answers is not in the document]
alancai98: Ah you're right. Didn't realize the definition was the same. Replaced usages in https://github.com/partiql/partiql-lang-kotlin/pull/1457/commits/7b615f893f360636d72e552148a59c0a1a4bb361

johnedquinn (approved):

PR #1461, opened 2024-05-10
DDL Plan Model. 

yliuuuu:
## Relevant Issues

## Description
- Experimental DDL Plan Modeling. 

### Reviewer Note: 
I recommend taking a look at the [test cases](https://github.com/partiql/partiql-lang-kotlin/pull/1461/files#diff-98d286f5d58ead03984774ec93c091713fc250bacae185355420876a0676630c) first. Then the [internal plan modeling](https://github.com/partiql/partiql-lang-kotlin/pull/1461/files#diff-9528ffa80596b1ac019d7674040f290258e005d9557e8ca0f72492f5659b8158). 

Plan Vs Internal Plan: the diff between the two is large in this PR, this is because we want to continue experimenting with model of the ddl plan and type system, while maintain a relatively stable public plan. 

#### DDL: 
```
  d_d_l::{
    shape: static_type,
    op: ddl_op,
  },
```
The DDL Operation carries a shape properties and an op properties.  

The shape properties describe the resulting shape of data after performed the DDL Operation. This split between `Op` and `shape` can be helpful when we incorporate other DDL operation, like `ALTER` 

#### DdlOpCreateTable
```
  create_table::{
    name: identifier,
    shape: '.type.collection',
    partition_by: optional::partition_by,
    table_properties: list::[table_property],
}
```
- Name: Table name; support qualified and unqualified identifier. In this PR, no validation on table name is performed. This is, we do not attempts to validate on potential problems such as table already exists, no existing scope, etc. 
- Shape: The shape of the data stored in the table. The internal plan comes with a helper type definition that helps to better traverse and manipulate the constraints putting on the type. Such type definition is validated, normalized, resolved, and converted to `StaticType`. In the public plan, the type definition is not kept to reduced the public API surface. 
- Partition By: Despite PartiQL does not support partitioned table at this stage, the plan will carry this information to subsequent processor. 
- Table_Properties: A arbitrary key value pair. PartiQL Planner at this stage does not use those information, the plan will carry this information to subsequent processor. 

### Public Plan: 
```
  d_d_l::{
    op: ddl_op,
  }

ddl_op::[
  create_table::{
    name: identifier,
    shape: 'static_type',
    partition_by: optional::partition_by,
    table_properties: list::[table_property],
  },
]
```

The public plan is designed to have minimum interference with the existing public APIs. Hence we switch the shape in create table to use static type. 

Notice that DDL node will not carry the static type in the public plan, this is a decision that can be revisited. But I opt to not included this properties in this PR. Depending on the status of the type work do over, we can potentially just use the new Type APIs instead of the migration plan. Because we are only working on Create Table, no information will be lost. 

### Constraints

We should distinguish those from the constraint concept from typed lambda calculus and the constraint concept from SQL Spec. 

- SQL constraints: 
    - A named database object, the primary purpose is to enforce data integrity. 
    - SQL (spec) allows for query constraints object like other SQL objects. 
    - SQL (spec) defines syntax/semantics for defining, manipulating and dropping constraints objects.
- Type constraints: Conditions or restrictions placed on the types of expressions or variables to ensure type correctness. 
   - Type constraints are used in type inference and type checking processes, where the system if an expression conforms to a particular type or if a type assignment is valid.  

#### Not Null Constraints
- Modeled as a type constraints. i.e., changes the result typing of an **type expression**.  
- `a INT2` -> the type of `a` will be `unionOf(null, int2)`. 
- `a INT2 NOT NULL` -> the type of `a` will be `int2`. 
- The type expression in the first case is **INT2**, the type expression in the second case is **INT2 NOT NULL**. 

#### Check constraints. 
- Modeled as a SQL constraint. 
- a INT2 CHECK(a > 0)
- The type expression in this case is **INT2**. The check constraint is NOT part of the type expression. 
- Semantics rule with check constraint: 
     - The expression of check constraints needs to be inferred as a boolean. 
     - If the constraint is originally set as a attribute level constraints, the scope of check constraint expression is limited to such attribute. 
     - If the constraint is originally set as a tuple level constraint, the scope of check constraint expression is limited to the attributes defined in the tuple. 
     - We normalized attribute level check constraints to tuple level check constraints. This is, all check constraints are kept under type struct (`record` in this pr). 

### Primary key constraint: 
- Primary key constraint is associated with collection level. 
- If an attribute is declared to be primary key, then for all tuples within the collection, such attribute needs to be unique and not null. 
- In this PR, if an attribute is declared as not null, then we will impose a not null constraint on the given attribute (notice that is a type constraint), and keep track of the primary key constraint at the collection level, to allow dropping of the primary key constraints (not yet supported). 

This is, given
```
CREATE TABLE tbl (
     a INT2 PRIMARY KEY
)
```
The type will be: 
```
Bag (
   STRUCT (
       a : INT2
   )
   PK : ["a"]
)
```

#### Other aspect of the PR
- ShapeNormalizer: Normalize constraints to the associated type. 
       -  Consider `a INT2 PRIMARY KEY`. The SQL syntax places primary key constraint, which is a collection level constraint, on attribute level. 
       - The internal plan converted from AST will also associated this primary key constraints with `Record.Field` (struct field) to keep the conversion code simple. 
       - The shape normalizer will "lift" the constraint to collection level. 

- ConstraintResolver: Convert the internal type definition to Static Type. 

- DDLFeatureGate: Gate certain features from customer usage. This is to unblock us from continuously doing internal experiments without the risk of breaking our customer.  

## Other Information
  - < If NO, why? >

  - < If YES, why? >

## License Information


partiql-plan/src/main/resources/partiql_plan.ion
```diff
@@ -52,8 +52,37 @@ statement::[
...
+]
+
+// DDL is an experimental feature.
+// We are actively working on finializing the semantics of DDL operation.
+// As of today,
+// the public plan aims to expose as minium implementation details as possible to avoid breaking changes in the feature.
+ddl_op::[
+  create_table::{
+    // Table Name, unresolved
+    name: identifier,
```
johnedquinn, line 67: I would think this would be fully-resolved for the public plan.

partiql-plan/src/main/resources/partiql_plan.ion
```diff
@@ -52,8 +52,37 @@ statement::[
   query::{
     root: rex,
   },
+  d_d_l::{
+    op: ddl_op,
+  },
```
johnedquinn, line 57: This is more of a nit, but the generated code would be quite nice to keep the cascading heirarchy of: `DDL.Op`. AKA:
```suggestion
  d_d_l::{
    op: op::[
      create_table::{ ... },
      ...
    ]
  },
```

Or, if `op` is always going to be the only property, then:
```suggestion
  define::[
    create_table::{ ... },
    ...
  ],
```

partiql-planner/src/test/kotlin/org/partiql/planner/internal/ddl/DDLTests.kt
```diff
@@ -0,0 +1,106 @@
...
+            ((plan.statement as Statement.DDL).op as DdlOp.CreateTable).shape
+        val expected = BagType(
+            StructType(
+                fields = listOf(
+                    StructType.Field("A", StaticType.INT2),
+                    StructType.Field("B", StaticType.INT2.asNullable()),
+                ),
+                contentClosed = true,
+                primaryKeyFields = emptyList(),
+                constraints = setOf(TupleConstraint.Open(false), TupleConstraint.UniqueAttrs(true)),
```
johnedquinn, line 77: Should you include the Ordered constraint too?

am357 (commented):
Submitting my initial comments; I will review the tests in more details and submit another review if necessary.

partiql-plan/src/main/resources/partiql_plan.ion
yliuuuu: I want to delegate the validation logic to connector at the moment, until we fully flush out the unix style path resolution for create command. 

At the moment, the connector shall receive a identifier, and it needs to implement its own logic as of where to persist the table. 

This is, the planner at this stage will not complain about problems like `Table already exists`, `schema not exist`, etc.

The comment is confusing, I will change it.

partiql-plan/src/main/resources/partiql_plan.ion
yliuuuu: In the feature, I hope that the every DDL node will carry the shape of the resulting table for ease of processing. 

This is not needed now as the only operation we support is `CREATE TABLE`, but it could be helpful when we want to support ALTER, etc.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/ddl/DDLTests.kt
```diff
@@ -0,0 +1,106 @@
...
+                metas = mapOf(
+                    "check_constraints" to ionStructOf(field("\$_my_catalog.my_schema.tbl_0", ionString("a <> b"))),
```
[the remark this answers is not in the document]
yliuuuu: Yes. I don't want to leak the modeling of constraints yet. 

I want to flush out more between the modeling of type constraints, and SQL constraints before we expose the constraint modeling. 

At the moment, constraints are information we pass along but do not process, which makes meta a good fit.

partiql-plan/src/main/resources/partiql_plan.ion
```diff
@@ -52,8 +52,37 @@ statement::[
...
+]
+
+// DDL is an experimental feature.
+// We are actively working on finializing the semantics of DDL operation.
+// As of today,
+// the public plan aims to expose as minium implementation details as possible to avoid breaking changes in the feature.
+ddl_op::[
+  create_table::{
```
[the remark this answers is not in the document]
yliuuuu: I think break down based on action might be a little over complicated. 

For example: Create Table vs Create Index: They don't really share common properties as far as I can see.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt
```diff
@@ -85,6 +94,44 @@ internal class PlanTransform(
...
+
+        override fun visitDdlOp(node: DdlOp, ctx: Unit) =
+            when (node) {
+                is DdlOp.CreateTable -> visitDdlOpCreateTable(node, ctx)
+            }
+
+        override fun visitDdlOpCreateTable(node: DdlOp.CreateTable, ctx: Unit): org.partiql.plan.DdlOp.CreateTable {
+            return ddlOpCreateTable(
+                visitIdentifier(node.name, ctx),
+                StaticType.ANY,
```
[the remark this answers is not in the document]
yliuuuu: No, this is a hack so we don't have to pass along static type as context

For internal Node: 
https://github.com/partiql/partiql-lang-kotlin/blob/84ae882dcf3e72954d00d7561a6b834a8da342fe/partiql-planner/src/main/resources/partiql_plan_internal.ion#L117-L125

The `DdlOp.CreateTable` node uses an internal representation of type. The converted static type is kept at `DdlOp` node. 

https://github.com/partiql/partiql-lang-kotlin/blob/84ae882dcf3e72954d00d7561a6b834a8da342fe/partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt#L104-L107

In the public plan, the `DdlOp.CreateTable` node replaces the internal type representation to use static type. 

In the plan transform (which convert the internal plan to the public plan), when processing the `DdlOp` node, we first go down to convert the `DdlOp.CreateTable` node, which use ANY type as a place holder here. 

Then we populate the resolved StaticType into the the `DdlOp.CreateTable` node.

partiql-plan/src/main/resources/partiql_plan.ion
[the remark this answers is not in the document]
am357: how about?
```
ddl_op 
 |- table [create, alter, ..]
 |- index [create, ...]
```

partiql-planner/src/test/kotlin/org/partiql/planner/internal/ddl/DDLTestBase.kt
```diff
@@ -0,0 +1,1206 @@
...
+                ),
+                tableInternal(
+                    FIELD_A_INT4.first,
+                    collectionConstraint = listOf(tuplePk("\$_tbl_0", listOf("a")))
+                ),
+                table(StructType.Field("A", StaticType.INT4), tableConstraint = setOf(CollectionConstraint.PrimaryKey(setOf("A")))),
+            ),
+
+            TestCase.success(
+                "CREATE TABLE tbl (a OPTIONAL INT4)",
```
johnedquinn, line 275: I'm beginning to think OPTIONAL as a modifier shouldn't be allowed in the column definitions of CREATE TABLE. We think of relations as bags of structs. Consider the following SQL table definition:
```
CREATE TABLE tbl(
  a INT
)
```
In PartiQL, a conformant "relation" might look like:
```
<<
  { 'a': 4 },
  { 'a': NULL }
>>
```

So, in PartiQL, a "relation" that isn't conformant to this table might look like:
```
<<
  { 'a': 4 },
  {  }
>>
```

Though, does this conform to the definition of the following table?
```
CREATE TABLE(
  a OPTIONAL INT
)
```
The PartiQL Spec describes the missing value in a struct to be stripped of the struct so that the key no longer exists. With tables being abstracted as a bag of structs, can we say that the column gets occasionally stripped out too?

That being said, I think the following is appropriate:
```
CREATE TABLE tbl(
  a STRUCT<b: OPTIONAL INT>
)
```

Maybe in a `CREATE VALUE/OBJECT/BINDING` we could create a bag of structs with optional fields, but `CREATE TABLE` specifically makes me wary of optional columns due to its semantics in SQL.

partiql-plan/src/main/resources/partiql_plan.ion
johnedquinn: I'm not entirely convinced by this since the connector is distinct from our public plan representation. In the connector (metadata), it doesn't necessarily need to check for whether the table already exists, but, the plan should be unambiguous. The executor can complain about the table already existing at the runtime, but still, it needs to know where to place the table (in an abstract sense) and in my opinion shouldn't leave it up to the evaluator's discretion.

If there is a concern regarding future compatibility, could we only allow creation of tables in the current schema? AKA, non-qualified table names. This would comply with our intention of the "relative path" as well as SQL:1999's semantics. Also, all of the tests are written this way currently. And, this would result in a disambiguated plan.

johnedquinn: And, on top of this, we gain control over our language. If we propagate case insensitivity and non-qualified identifiers in the plan (say table `tbl`), one evaluator might create a table named `"TBL"` in the current schema, and another evaluator could create the table named `"tbl"` in some other schema. In other words, by propagating the ambiguity, we rely on our consumers to know the internal, subtle details of the PartiQL language, and our semantics are not enforced.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/ddl/DDLTestBase.kt
johnedquinn: Discussed offline, but to reiterate: I actually like the usefulness of OPTIONAL as a column description modifier. But just worth it to ask around (maybe a PartiQL Discussion) and make sure that others can look past the *potential* oddities that come with a fragmented "table".

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -1442,6 +1473,173 @@ internal class PlanTyper(private val env: Env) {
...
+                node.name, normalizedShape, partitionBy, node.tableProperties
+            )
+        }
+
+        override fun visitTypeRecord(node: Type.Record, ctx: List<Type.Record.Field>): Type.Record {
+            val fields = node.fields.map { field ->
+                val constraints = field.constraints.map { constr ->
+                    visitConstraint(constr, listOf(field))
+                }
+                val type = visitType(field.type, ctx)
```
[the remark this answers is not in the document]
johnedquinn: Discussed offline. I see that the logic is outside of the PlanTyper. Nice.

yliuuuu:
Close this PR for now. 

We'd like to merge DDL as a post-V1 Feature.

PR #1463, opened 2024-05-15
Deprecates absent types and removes associated planner logic

johnedquinn:
## Description
- As discussed with the rest of the PartiQL maintainers, the null and missing values have leaked into our type semantics -- causing issues when it comes to function resolution, unsafe casts, and dynamic functions.
- This PR deprecates NullType and MissingType and removes their usage in the planner. The implication is that ALL types may contain the null and missing values. See the CHANGELOG for all deprecated APIs.
- In this PR, literal NULL and MISSING values are typed as ANY (dynamic). In the future, we may want to pursue the "unknown" type that PostgreSQL internally uses.
- Smaller changes include:
  - I deprecated `AnyOfType`'s primary constructor. We've been plagued for too long by forgetting to use `.flatten()` .
  - The majority of this diff is me getting rid of "null" and "missing" in the testFixtures.

## Other Information
  - Not an API breaking change, but a behavioral one.

## License Information

### DIFF
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
@@ -4,6 +4,8 @@ package org.partiql.planner.internal.typer
 
 import org.partiql.planner.internal.ir.Rex
 import org.partiql.types.StaticType
+import org.partiql.value.MissingValue
+import org.partiql.value.NullValue
 import org.partiql.value.PartiQLValueExperimental
 import org.partiql.value.PartiQLValueType
 import org.partiql.value.PartiQLValueType.ANY
@@ -71,7 +73,7 @@ internal class DynamicTyper {
      */
     private fun Rex.isLiteralAbsent(): Boolean {
         val op = this.op
-        return op is Rex.Op.Lit && (op.value.type == PartiQLValueType.MISSING || op.value.type == PartiQLValueType.NULL)
+        return op is Rex.Op.Lit && (op.value is MissingValue || op.value is NullValue)
     }
 
     /**
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
@@ -2,8 +2,7 @@
 
 package org.partiql.planner.internal.typer
 
-import org.partiql.types.MissingType
-import org.partiql.types.NullType
+import org.partiql.planner.internal.ir.Rex
 import org.partiql.types.StaticType
 import org.partiql.value.PartiQLValueExperimental
 import org.partiql.value.PartiQLValueType
@@ -27,8 +26,6 @@ import org.partiql.value.PartiQLValueType.INT64
 import org.partiql.value.PartiQLValueType.INT8
 import org.partiql.value.PartiQLValueType.INTERVAL
 import org.partiql.value.PartiQLValueType.LIST
-import org.partiql.value.PartiQLValueType.MISSING
-import org.partiql.value.PartiQLValueType.NULL
 import org.partiql.value.PartiQLValueType.SEXP
 import org.partiql.value.PartiQLValueType.STRING
 import org.partiql.value.PartiQLValueType.STRUCT
@@ -57,51 +54,55 @@ internal class DynamicTyper {
     private var supertype: PartiQLValueType? = null
     private var args = mutableListOf<PartiQLValueType>()
 
-    private var nullable = false
-    private var missable = false
     private val types = mutableSetOf<StaticType>()
 
     /**
-     * This primarily unpacks a StaticType because of NULL, MISSING.
-     *
-     *  - T
-     *  - NULL
-     *  - MISSING
-     *  - (NULL)
-     *  - (MISSING)
-     *  - (T..)
-     *  - (T..|NULL)
-     *  - (T..|MISSING)
-     *  - (T..|NULL|MISSING)
-     *  - (NULL|MISSING)
-     *
+     * Adds the [rex]'s [Rex.type] to the typing accumulator (if the [rex] is not a literal NULL/MISSING).
+     */
+    fun accumulate(rex: Rex) {
+        when (rex.isLiteralAbsent()) {
+            true -> accumulateUnknown()
+            false -> accumulate(rex.type)
+        }
+    }
+
+    /**
+     * Checks for literal NULL or MISSING.
+     */
+    private fun Rex.isLiteralAbsent(): Boolean {
+        val op = this.op
+        return op is Rex.Op.Lit && (op.value.type == PartiQLValueType.MISSING || op.value.type == PartiQLValueType.NULL)
+    }
+
+    /**
+     * When a literal null or missing value is present in the query, its type is unknown. Therefore, its type must be
+     * inferred. This function ignores literal null/missing values, yet adds their indices to know how to return the
+     * mapping.
+     */
+    private fun accumulateUnknown() {
+        args.add(ANY)
+    }
+
+    /**
+     * This adds non-unknown types (aka not NULL / MISSING literals) to the typing accumulator.
      * @param type
      */
-    fun accumulate(type: StaticType) {
-        val nonAbsentTypes = mutableSetOf<StaticType>()
+    private fun accumulate(type: StaticType) {
         val flatType = type.flatten()
         if (flatType == StaticType.ANY) {
-            // Use ANY runtime; do not expand ANY
             types.add(flatType)
             args.add(ANY)
             calculate(ANY)
             return
         }
-        for (t in flatType.allTypes) {
-            when (t) {
-                is NullType -> nullable = true
-                is MissingType -> missable = true
-                else -> nonAbsentTypes.add(t)
-            }
-        }
-        when (nonAbsentTypes.size) {
+        val allTypes = flatType.allTypes
+        when (allTypes.size) {
             0 -> {
-                // Ignore in calculating supertype.
-                args.add(NULL)
+                error("This should not have happened.")
             }
             1 -> {
                 // Had single type
-                val single = nonAbsentTypes.first()
+                val single = allTypes.first()
                 val singleRuntime = single.toRuntimeType()
                 types.add(single)
                 args.add(singleRuntime)
@@ -109,7 +110,7 @@ internal class DynamicTyper {
             }
             else -> {
                 // Had a union; use ANY runtime
-                types.addAll(nonAbsentTypes)
+                types.addAll(allTypes)
                 args.add(ANY)
                 calculate(ANY)
             }
@@ -124,31 +125,20 @@ internal class DynamicTyper {
      * @return
      */
     fun mapping(): Pair<StaticType, List<Pair<PartiQLValueType, PartiQLValueType>>?> {
-        val modifiers = mutableSetOf<StaticType>()
-        if (nullable) modifiers.add(StaticType.NULL)
-        if (missable) modifiers.add(StaticType.MISSING)
         // If at top supertype, then return union of all accumulated types
         if (supertype == ANY) {
-            return StaticType.unionOf(types + modifiers).flatten() to null
+            return StaticType.unionOf(types).flatten() to null
         }
         // If a collection, then return union of all accumulated types as these coercion rules are not defined by SQL.
         if (supertype == STRUCT || supertype == BAG || supertype == LIST || supertype == SEXP) {
-            return StaticType.unionOf(types + modifiers) to null
+            return StaticType.unionOf(types) to null
         }
         // If not initialized, then return null, missing, or null|missing.
-        val s = supertype
-        if (s == null) {
-            val t = if (modifiers.isEmpty()) StaticType.MISSING else StaticType.unionOf(modifiers).flatten()
-            return t to null
-        }
+        val s = supertype ?: return StaticType.ANY to null
         // Otherwise, return the supertype along with the coercion mapping
-        val type = s.toNonNullStaticType()
+        val type = s.toStaticType()
         val mapping = args.map { it to s }
-        return if (modifiers.isEmpty()) {
-            type to mapping
-        } else {
-            StaticType.unionOf(setOf(type) + modifiers).flatten() to mapping
-        }
+        return type to mapping
     }
 
     private fun calculate(type: PartiQLValueType) {
@@ -163,7 +153,7 @@ internal class DynamicTyper {
         // Lookup and set the new minimum common supertype
         supertype = when {
             type == ANY -> type
-            type == NULL || type == MISSING || s == type -> return // skip
+            s == type -> return // skip
             else -> graph[s][type] ?: ANY // lookup, if missing then go to top.
         }
     }
@@ -206,8 +196,6 @@ internal class DynamicTyper {
                 graph[type] = arrayOfNulls(N)
             }
             graph[ANY] = edges()
-            graph[NULL] = edges()
-            graph[MISSING] = edges()
             graph[BOOL] = edges(
                 BOOL to BOOL
             )
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
@@ -86,7 +86,7 @@ internal class DynamicTyper {
     }
 
     /**
-     * This adds non-unknown types (aka not NULL / MISSING literals) to the typing accumulator.
+     * This adds non-absent types (aka not NULL / MISSING literals) to the typing accumulator.
      * @param type
      */
     private fun accumulate(type: StaticType) {
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/FnResolver.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/FnResolver.kt
@@ -408,9 +408,9 @@ internal class FnResolver(private val header: Header) {
         // This does not imply the ability to CAST; this defines function resolution behavior.
         private val precedence: Map<PartiQLValueType, Int> = listOf(
             @Suppress("DEPRECATION")
-            PartiQLValueType.NULL, // TODO: Remove
+            PartiQLValueType.NULL, // TODO: Remove once functions no longer specify parameter/return types with the NULL type.
             @Suppress("DEPRECATION")
-            PartiQLValueType.MISSING, // TODO: Remove
+            PartiQLValueType.MISSING, // TODO: Remove once functions no longer specify parameter/return types with the MISSING type.
             BOOL,
             INT8,
             INT16,
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/FnResolver.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/FnResolver.kt
@@ -5,9 +5,6 @@ import org.partiql.planner.internal.ir.Agg
 import org.partiql.planner.internal.ir.Fn
 import org.partiql.planner.internal.ir.Identifier
 import org.partiql.planner.internal.ir.Rex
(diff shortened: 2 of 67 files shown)


partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt
```diff
@@ -2144,15 +2136,13 @@ class PlanTyperTestsPorted {
                     ) FROM <<
                         { 'a': { 'b': 1 } },
                         { 'a': { 'b': 'hello' } },
-                        { 'a': NULL },
```
johnedquinn, line 2147: I updated this only because the literal NULL's type is unknown. Which leads to the inferred type to be ANY when used as an argument to a literal struct.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt
```diff
@@ -3127,55 +3104,8 @@ class PlanTyperTestsPorted {
...
-                        contentClosed = true,
-                        constraints = setOf(
-                            TupleConstraint.Open(false),
-                            TupleConstraint.UniqueAttrs(true),
-                            TupleConstraint.Ordered
-                        )
-                    )
-                )
-            ),
-            SuccessTestCase(
```
johnedquinn, line 3142: This is deleted now since ALL types are nullable.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/FnResolver.kt
```diff
@@ -440,8 +407,10 @@ internal class FnResolver(private val header: Header) {
         // This is not explicitly defined in the PartiQL Specification!!
         // This does not imply the ability to CAST; this defines function resolution behavior.
         private val precedence: Map<PartiQLValueType, Int> = listOf(
-            NULL,
-            MISSING,
+            @Suppress("DEPRECATION")
+            PartiQLValueType.NULL, // TODO: Remove
+            @Suppress("DEPRECATION")
+            PartiQLValueType.MISSING, // TODO: Remove
```
alancai98, line 413: should be removed per the TODO?

partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
```diff
@@ -20,7 +20,7 @@ public sealed class StaticType {
          */
         @JvmStatic
         public fun unionOf(vararg types: StaticType, metas: Map<String, Any> = mapOf()): StaticType =
-            unionOf(types.toSet(), metas)
+            unionOf(types.toSet(), metas).flatten()
```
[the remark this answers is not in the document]
alancai98: Side-note: there are probably a bunch of other places we now call `.flatten()` redundantly (especially in `PlanTyper`), perhaps we could fix or cut an issue to track fixing.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -515,20 +511,31 @@ internal class PlanTyper(
...
-            }.toSet()
-            val finalType = unionOf(pathTypes + toAddTypes).flatten()
-            return rex(finalType.swallowAny(), rexOpPathKey(root, key))
+            }
+            if (elementType.isEmpty()) {
+                handleAlwaysMissing()
+                return rex(ANY, rexOpPathKey(root, key))
+            }
+            // TODO: SwallowAny should happen by default
+            return rex(unionOf(elementType).swallowAny(), rexOpPathKey(root, key))
```
alancai98, line 520: perhaps we could get rid of `.swallowAny()` function overall? I think this PR https://github.com/partiql/partiql-lang-kotlin/pull/1395 fixes the flatten call when one of the types is `AnyType`.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -537,11 +544,21 @@ internal class PlanTyper(
...
-            val type = unionOf(paths.map { it.type }.toSet()).flatten()
+            // Determine output type
+            val type = when (paths.size) {
+                // Escape early since no inference could be made
+                0 -> {
+                    handleAlwaysMissing()
+                    return rex(ANY, Rex.Op.Path.Symbol(root, node.key))
+                }
+                // TODO: Flatten() should occur by default
+                else -> unionOf(paths.map { it.type }.toSet()).flatten()
```
alancai98, line 555: yea with the changes in `StaticType.unionOf()`, `flatten()` can be removed here

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -562,15 +579,6 @@ internal class PlanTyper(
 
         private fun rexString(str: String) = rex(STRING, rexOpLit(stringValue(str)))
 
-        override fun visitRexOpPath(node: Rex.Op.Path, ctx: StaticType?): Rex {
-            val path = super.visitRexOpPath(node, ctx) as Rex
-            if (path.type == MISSING) {
```
alancai98, line 567: should this override be deleted? Should we check if the `path.type == ANY`?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -621,71 +625,29 @@ internal class PlanTyper(
...
-            // 7.1 All functions return MISSING when one of their inputs is MISSING (except `=`)
-            newArgs.forEach {
-                if (it.type == MissingType && !isNotMissable) {
-                    handleAlwaysMissing()
-                    return rex(MISSING, rexOpCallStatic(newFn, newArgs))
-                }
+            // Check literal missing inputs
+            val argAlwaysMissing = args.any {
+                val op = it.op as? Rex.Op.Lit ?: return@any false
+                op.value.type == PartiQLValueType.MISSING
```
alancai98, line 636: similar to other comment, should check if something is a missing literal by checking if it's the literal missing value?

```suggestion
                op.value == missingValue()
```

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -1225,26 +1163,30 @@ internal class PlanTyper(
                 isClosed && isOrdered -> {
                     struct.fields.firstOrNull { entry -> binding.isEquivalentTo(entry.key) }?.let {
                         (sensitive(it.key) to it.value)
-                    } ?: (key to MISSING)
+                    } ?: run {
+                        return null
+                    }
```
alancai98, line 1168: think you don't need this `run`?

```suggestion
                    } ?: return null
```

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -621,71 +625,29 @@ internal class PlanTyper(
...
-                        if (arg.type.isNullable()) {
-                            isNullable = true
-                            break
-                        }
-                    }
+            if (argAlwaysMissing) {
+                // TODO: The V1 branch has support for isMissable and isMissingCall. This codebase, however, does not
+                //  have support for these concepts yet. This specific commit (see Git blame) does not seek to add this
+                //  functionality. Below is a work-around for the lack of "isMissable" and "isMissingCall"
+                if (match.signature.name !in listOf("is_null", "is_missing", "eq")) {
```
alancai98, line 642: Do we also need to include the logical ops here (and, or, not)? When `AND` and `OR` are provided with a `NULL`/`MISSING` argument, they may still return a boolean:

```
PartiQL> NULL AND false;
==='
false
---
OK!
PartiQL> NULL OR true;
==='
true
---
OK!
```

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt
```diff
@@ -635,26 +637,23 @@ class PlanTyperTestsPorted {
             SuccessTestCase(
                 name = "BITWISE_AND_NULL_OPERAND",
                 query = "1 & NULL",
-                expected = StaticType.NULL,
+                expected = unionOf(INT4, INT8, INT),
```
alancai98, line 640: Hmm, curious why this will now give a union of these int types? Similarly for the below case w/ `MISSING`.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt
```diff
@@ -2506,37 +2494,27 @@ class PlanTyperTestsPorted {
...
-            ),
-            SuccessTestCase(
-                key = PartiQLTest.Key("basics", "case-when-22"),
-                catalog = "pql",
-                expected = unionOf(StaticType.INT4, StaticType.NULL, StaticType.MISSING),
-            ),
-            SuccessTestCase(
-                key = PartiQLTest.Key("basics", "case-when-23"),
-                catalog = "pql",
-                expected = StaticType.INT4,
```
alancai98, line 2539: Any reason these tests are deleted?

alancai98 (commented):
nice work in removing the absent types from the static type modeling. I think it helps simplify a lot of the typing and test fixture involving typing logic quite a bit. comments were mostly minor though I did have some questions about some behavioral changes in tests.

partiql-planner/src/testFixtures/resources/catalogs/default/pql/t_item.ion
```diff
@@ -10,7 +10,7 @@
     },
     {
       name: "t_bool_nul",
-      type: ["bool","null"],
+      type: "bool",
```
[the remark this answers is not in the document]
johnedquinn: Yeah, I thought to do that, but I opted to avoid doing that due to the diff on tests. I can either include that change or follow up with a clean up of the duplicate tests.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
johnedquinn: The suggestion is a method invocation resulting in a whole new Java class instance and uses the equals method to check the Java class and properties. The original is a check on an enum (integer). Also, this is similar to the reasoning behind PQLValue -- we should rely on our own type constructs rather than Java's.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
```diff
@@ -57,59 +54,63 @@ internal class DynamicTyper {
...
+            false -> accumulate(rex.type)
+        }
+    }
+
+    /**
+     * Checks for literal NULL or MISSING.
+     */
+    private fun Rex.isLiteralAbsent(): Boolean {
+        val op = this.op
+        return op is Rex.Op.Lit && (op.value.type == PartiQLValueType.MISSING || op.value.type == PartiQLValueType.NULL)
```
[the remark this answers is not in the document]
alancai98: followup to not call the function constructor for `missingValue` and `nullValue`,

```suggestion
        return op is Rex.Op.Lit && (op.value is NullValue || op.value is MissingValue)
```

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
alancai98: I should have linked the comment for clarity -- https://github.com/partiql/partiql-lang-kotlin/pull/1463#discussion_r1604083403. We're calling `PartiQLValueType.MISSING` here but since it's deprecated in this PR and will be removed in a future release, I'm thinking we should not call the enum variant. Here's a version not calling the function invocation:

```suggestion
                op.value is MissingValue
```

partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
```diff
@@ -170,6 +244,11 @@ public sealed class StaticType {
     /**
      * Type is optional if it is Any, or Missing, or an AnyOfType that contains Any or Missing type
      */
+    @Suppress("DEPRECATION")
+    @Deprecated(
+        message = "This will be removed in a future major-version bump.",
+        replaceWith = ReplaceWith("true")
+    )
```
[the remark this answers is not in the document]
johnedquinn: True -- it was largely just to highlight the usages across the codebase.

partiql-types/src/main/kotlin/org/partiql/types/StaticType.kt
```diff
@@ -621,7 +708,14 @@ public data class GraphType(
 /**
  * Represents a [StaticType] that's defined by the union of multiple [StaticType]s.
  */
-public data class AnyOfType(val types: Set<StaticType>, override val metas: Map<String, Any> = mapOf()) : StaticType() {
+public data class AnyOfType @Deprecated(
+    message = "This will be removed in a future major-version bump.",
```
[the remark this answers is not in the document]
johnedquinn: Realistically, given our intention to eventually deprecate/delete this entire file, I don't really care to remove it. People should just not be using the constructor.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt
johnedquinn: Yeah, these tests originally are as follows:
```
--#[case-when-22]
-- type: (null|missing|int32)
CASE t_item.t_string
    WHEN 'a' THEN t_item.t_absent
    ELSE -1
END;

--#[case-when-23]
-- type: int32
-- false branch is pruned
CASE
    WHEN false THEN t_item.t_absent
    ELSE -1
END;
```

Since the t_absent type is now just ANY instead of `["null", "missing"]`, the result is just ANY -- and we have tests that already test this. That being said, I'm going to go one step further and delete the Ion tests and remove `t_absent` as well as `t_null` and `t_missing`.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt
johnedquinn: Discussed offline -- but it has to do with the fact that a literal NULL has an unknown type, which gets interpreted as ANY for our type inference. Therefore, the applicable functions for `bitwise_and` are only for the integer types that correspond to the LHS being an int_32.

Before this PR, `1 & NULL` would be constant folded in the typer, which is wrong, because it doesn't actually give insight to the type being returned (just the value). This change narrows down the type. Though, in the future (or, if you'd like in this PR), we can attempt to find the highest precedence function when the argument is the literal NULL/MISSING value. This is kind of what PostgreSQL does. As it would result in more in-depth logic changes, I'd prefer a follow-up to this PR.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt
alancai98: Not too sure why the empty struct in the input binding tuples doesn't cause `ANY` for the type. Perhaps that binding tuple is filtered out?

alancai98: Not sure if there's a test showing this but if there's not, could still be helpful to include a test showing that the inferred type for a path expression on a literal struct with a `NULL`/`MISSING` value is `ANY`. E.g. `{ 'a': NULL }.a` -> `ANY` and `{ 'a': MISSING }.a` -> `ANY`

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeUtils.kt
```diff
@@ -177,7 +195,7 @@ internal fun StructType.exclude(steps: List<Rel.Op.Exclude.Step>, lastStepOption
     val output = fields.map { field ->
         val newField = if (steps.size == 1) {
             if (lastStepOptional) {
-                StructType.Field(field.key, field.value.asOptional())
+                StructType.Field(field.key, field.value) // TODO: double check this
```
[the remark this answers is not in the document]
alancai98: cool, thanks for cutting the issue. could also link the issue in the code for easier discoverability

RCHowell (changes requested):

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -301,14 +296,7 @@ internal class PlanTyper(
...
 
             // Calculate output schema given JOIN type
-            val l = lhs.type.schema
-            val r = rhs.type.schema
-            val schema = when (node.type) {
-                Rel.Op.Join.Type.INNER -> l + r
-                Rel.Op.Join.Type.LEFT -> l + r.pad()
-                Rel.Op.Join.Type.RIGHT -> l.pad() + r
-                Rel.Op.Join.Type.FULL -> l.pad() + r.pad()
-            }
```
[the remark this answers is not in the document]
johnedquinn: All types are considered nullable, therefore, there isn't a need to pad the types with the NULL type.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -515,20 +511,30 @@ internal class PlanTyper(
                     // we might improve upon this with some constant folding prior to typing
                     ANY
                 }
-            }.toSet()
-            val finalType = unionOf(pathTypes + toAddTypes).flatten()
-            return rex(finalType.swallowAny(), rexOpPathKey(root, key))
+            }
+            if (elementType.isEmpty()) {
+                handleAlwaysMissing()
+                return rex(ANY, rexOpPathKey(root, key))
```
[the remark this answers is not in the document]
johnedquinn: Yes, in the `v1` branch we can do that. Main doesn't have that concept yet.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -515,20 +511,30 @@ internal class PlanTyper(
...
 
         override fun visitRexOpPathSymbol(node: Rex.Op.Path.Symbol, ctx: StaticType?): Rex {
             val root = visitRex(node.root, node.root.type)
 
-            val paths = root.type.allTypes.map { type ->
-                val struct = type as? StructType ?: return@map rex(MISSING, rexOpLit(missingValue()))
+            // Check Root Type
+            if (!root.type.mayBeType<StructType>()) {
+                handleAlwaysMissing()
+                return rex(ANY, rexOpErr("Symbol lookup may only occur on structs, not ${root.type}."))
```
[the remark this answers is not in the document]
johnedquinn: https://github.com/partiql/partiql-lang-kotlin/pull/1463#discussion_r1610375092

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -537,40 +543,28 @@ internal class PlanTyper(
                     )
                 }
             }
-            val type = unionOf(paths.map { it.type }.toSet()).flatten()
+            // Determine output type
+            val type = when (paths.size) {
+                // Escape early since no inference could be made
+                0 -> {
+                    handleAlwaysMissing()
+                    return rex(ANY, Rex.Op.Path.Symbol(root, node.key))
```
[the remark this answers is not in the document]
johnedquinn: https://github.com/partiql/partiql-lang-kotlin/pull/1463#discussion_r1610375092

RCHowell (approved):

Issue #1460, opened 2024-05-08
Remove hard-coded aggregations within the AST


alancai98:
There are several places within the partiql-parser and partiql-ast where we specify dedicated aggregation rules and nodes.

In the ANTLR grammar:
https://github.com/partiql/partiql-lang-kotlin/blob/2879f3a0a1d3197cc00fda6f1613490f88d19353/partiql-parser/src/main/antlr/PartiQL.g4#L727-L730
A couple problems w/ above:
- Aggregate function names are hard-coded
- Designated `aggregate` rule -- perhaps can be consolidated with the existing function call rule

In the partiql-ast:
https://github.com/partiql/partiql-lang-kotlin/blob/2879f3a0a1d3197cc00fda6f1613490f88d19353/partiql-ast/src/main/resources/partiql_ast.ion#L351-L356
Similar issue to above.

To allow for aggregate UDFs in the future, we should not distinguish between the aggregate and scalar functions till planning time. This will require modification of the AST to plan [conversion code](https://github.com/partiql/partiql-lang-kotlin/blob/2879f3a0a1d3197cc00fda6f1613490f88d19353/partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt#L332-L415), which currently assumes the resolution occurs during parsing.

alancai98:
This task can be broken down into multiple components:
1. removing hard-coded aggregate functions from AST + parser
2. figuring out scalar vs aggregation resolution in the planner. As a temporary workaround, could hard-code the logic in the AST to plan code. A future PR could integrate this resolution w/ the SPI + connector metadata interfaces.
3. (future) creating a UDF aggregation function -- might require further specification. E.g. name conflicts between scalar and aggregate functions.

alancai98:
Resolved by #1464.

PR #1464, opened 2024-05-15
Remove hard-coded aggregations from parser and ast

alancai98:
## Relevant Issues
- Resolves https://github.com/partiql/partiql-lang-kotlin/issues/1460

## Description
- Removes the hard-coded ANTLR parse rule for aggregation functions
- Removes the dedicated aggregation node from AST. Set quantifier added to function call node.
- Preserves legacy PIG parsing and AST behavior
- Slight differences in modeling of `COUNT(*)`
  - Previously, would map to the function named `COUNT_STAR` with no args; changed to map to `COUNT` with no args

## Other Information
  - No, v1 branch

  - Yes, AST modeling changes -- AST's call now represents both scalar and aggregate functions

## License Information


partiql-spi/src/main/kotlin/org/partiql/spi/connector/sql/builtins/AggCountStar.kt
```diff
@@ -1,25 +0,0 @@
-// ktlint-disable filename
```
alancai98, line 1: (self-review) `COUNT(*)` builtin function deleted in favor of mapping `COUNT(*)` (`COUNT()` in ast) to `COUNT(1)`

partiql-spi/src/main/kotlin/org/partiql/spi/connector/sql/builtins/AggCount.kt
```diff
@@ -13,7 +13,7 @@ import org.partiql.value.PartiQLValueType.ANY
 import org.partiql.value.PartiQLValueType.INT64
 
 @OptIn(PartiQLValueExperimental::class, FnExperimental::class)
-public object Agg_COUNT__ANY__INT32 : Agg {
+public object Agg_COUNT__ANY__INT64 : Agg {
```
alancai98, line 16: (self-review) I noticed the return type was changed to `INT64` in a recent PR but the name wasn't changed

partiql-lang/src/main/kotlin/org/partiql/lang/syntax/impl/PartiQLPigVisitor.kt
```diff
@@ -1325,16 +1325,48 @@ internal class PartiQLPigVisitor(
...
+        val args = ctx.expr()
+        if (args.size != 1) {
+            error("expect only one argument to aggregate function call")
+        }
+        val arg = visitExpr(args.first())
+        val metas = nameCtx.name.getSourceMetaContainer()
+        return@build callAgg(strategy, name, arg, metas)
+    }
+
+    private fun String.isAggregateCall(): Boolean {
```
alancai98, line 1366: (self-review) in the legacy PIG AST and conversion from `partiql-ast` to PIG AST, keep the same behavior as before with respect to supported aggregations. E.g.
- `SUM(*)` and `F(*)` will still error
- Aggregations require exactly `1` argument

Hard-coding the aggregations in the conversion to PIG AST since this is legacy code that will be removed by the v1 release.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -607,13 +617,32 @@ internal object RelConverter {
...
+                        caseSensitivity = org.partiql.ast.Identifier.CaseSensitivity.INSENSITIVE
+                    }
+                    ctx.aggregations += node
+                    exprVar(id, Expr.Var.Scope.DEFAULT)
+                }
+                else -> node
+            }
+        }
+
+        private fun String.isAggregateCall(): Boolean {
```
alancai98, line 636: (self-review) since the function resolution specification for scalar vs aggregate functions has not yet been defined, opt to hard-code the aggregation logic to the planner for now. Once we formalize the function resolution, we can remove this hard-coded logic.

partiql-ast/src/test/kotlin/org/partiql/ast/sql/SqlDialectTest.kt
```diff
@@ -599,51 +599,51 @@ class SqlDialectTest {
...
                     function = id("FOO")
                     setq = SetQuantifier.DISTINCT
                     args += exprVar(id("x"), Expr.Var.Scope.DEFAULT)
                     args += exprVar(id("y"), Expr.Var.Scope.DEFAULT)
                 }
             },
             expect("COUNT(*)") {
-                exprAgg {
-                    function = id("COUNT_STAR")
+                exprCall {
```
alancai98, line 645: (self-review) modeling difference in `partiql-ast` for `COUNT(*)`. Previously it was an `exprAgg` with a function name of `COUNT_STAR` and no arguments. Changed to just be an `exprCall` with function name of `COUNT` and no arguments, which I think is slightly cleaner.

This is how Trino opts to represent `COUNT(*)` in their AST -- https://github.com/trinodb/trino/blob/429f8a8257d38e09b11a69571456c346f973373e/core/trino-parser/src/main/java/io/trino/sql/ExpressionFormatter.java#L436-L438.

alancai98:
PR should be ready for review. Regarding future work related to user-defined aggregations, function resolution between scalars and aggregations, and qualified paths for aggregations, I'll need to discuss further w/ the PartiQL team on these topics when I'm back online at the end of the month.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -607,13 +617,32 @@ internal object RelConverter {
...
+            // currently hard-coded
+            return listOf("count", "avg", "sum", "min", "max", "any", "some", "every").contains(this)
```
RCHowell, line 638: Similar to the legacy ast converter, let's make this a @JvmStatic set

RCHowell (changes requested):
Perfect! Very clean and I'm happy to see this item tackled.

RCHowell (approved):

PR #1467, opened 2024-05-16
Initializes PartiQLCursor and PartiQLValueLoader

johnedquinn:
## Relevant Issues
- See #1443.

## Description
- This PR adds a PartiQLCursor to represent a stream of PartiQL data.
- This has already been approved for the KOW branch. This is a cherry-pick plus the `apiDump`. See the above linked PR for the original description.

## Other Information
- See linked PR

## License Information


partiql-eval/src/main/kotlin/org/partiql/eval/PartiQLResult.kt
```diff
@@ -1,12 +1,12 @@
...
 
-import org.partiql.value.PartiQLValue
+import org.partiql.value.PartiQLCursor
 import org.partiql.value.PartiQLValueExperimental
 
 public sealed interface PartiQLResult {
 
     @OptIn(PartiQLValueExperimental::class)
-    public data class Value(public val value: PartiQLValue) : PartiQLResult
+    public data class Value(public val value: PartiQLCursor) : PartiQLResult
```
RCHowell, line 9:
```suggestion
    public data class Value(public val cursor: PartiQLCursor) : PartiQLResult
```

partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEngineDefaultTest.kt
```diff
@@ -1176,7 +1177,7 @@ class PartiQLEngineDefaultTest {
                     throw returned.cause
                 }
             }
-            val output = result.value
+            val output = PartiQLValueLoader.standard().load(result.value)
```
RCHowell, line 1180: Is there anything other than "standard"?

partiql-types/src/main/java/org/partiql/value/PartiQLCursor.java
```diff
@@ -0,0 +1,187 @@
...
+     * This is only applicable when the current value's type is {@link PartiQLValueType#STRING}.
+     */
+    @NotNull
+    String getStringValue();
+
+    /**
+     * This is only applicable when the current value's type is {@link PartiQLValueType#CHAR}.
+     */
+    @NotNull
+    String getCharValue();
```
RCHowell, line 76:
```suggestion
    Char getCharValue();
```

partiql-types/src/main/java/org/partiql/value/PartiQLCursor.java
```diff
@@ -0,0 +1,187 @@
...
+
+    /**
+     * This is only applicable when the current value's type is {@link PartiQLValueType#BLOB}.
+     */
+    public byte[] getBlobValue();
+
+    /**
+     * This is only applicable when the current value's type is {@link PartiQLValueType#CLOB}.
+     */
+    public byte[] getClobValue();
```
RCHowell, line 102: Two comments,

1. Let's follow postgres where we follow the host language (C) for method names and matching the return type. Ie `getBytes() -> byte[]` and `getLong() -> Long` rather than `getInt64`.
2. Drop the redundant value

partiql-types/src/main/java/org/partiql/value/PartiQLValueLoader.java
```diff
@@ -0,0 +1,30 @@
...
+     * @param data the PartiQL data to load.
+     * @return a materialized, in-memory instance of a {@link PartiQLValue} containing the contents of the {@code data}.
+     */
+    @NotNull
+    PartiQLValue load(@NotNull PartiQLCursor data);
+
+    /**
+     * @return a basic implementation of {@link PartiQLValueLoader}.
+     */
+    static PartiQLValueLoader standard() {
```
RCHowell, line 27:
```suggestion
    static PartiQLValueLoader default() {
```

RCHowell (changes requested):

partiql-types/src/main/java/org/partiql/value/PartiQLValueLoader.java
johnedquinn: "default" is a Java keyword. Found this out recently when I was writing Java and was trying to use a method named default.

partiql-types/src/main/java/org/partiql/value/PartiQLCursor.java
johnedquinn: This method will be deleted in accordance with your other suggestion of following the JVM names. Also, side note, the CHAR datatype in SQL acts more like Java's String.

partiql-types/src/main/java/org/partiql/value/PartiQLCursor.java
```diff
@@ -0,0 +1,250 @@
...
+     * This is applicable to the following types:
+     * {@link PartiQLValueType#CLOB}
+     * @return a value representing the applicable PartiQL value
+     * @throws UnsupportedOperationException when this method is not applicable to the type returned by {@link PartiQLCursor#getType()}
+     * @throws NullPointerException if this method is invoked when {@link PartiQLCursor#isNull()} returns true
+     * @apiNote <b>! ! ! EXPERIMENTAL ! ! !</b> This is an experimental API under development by the PartiQL maintainers.
+     * Please abstain from using this API until given notice otherwise. This may break between iterations without prior notice.
+     */
+    @NotNull
+    public Clob getClob() throws UnsupportedOperationException, NullPointerException;
```
RCHowell, line 125: I'm not sure these are necessary considering you can use `getBytes()`. I would remove them for the time being so this package has to java.sql dependency.

partiql-types/src/main/java/org/partiql/value/PartiQLValueLoader.java
RCHowell: Makes me wonder if we should be changing the 1.0 builder names ...

RCHowell (changes requested):
One minor change, otherwise looks good.

partiql-types/src/main/java/org/partiql/value/PartiQLCursor.java
johnedquinn: ACK. Just updated to remove for now.

partiql-types/src/main/java/org/partiql/value/PartiQLCursor.java
RCHowell: Same as PQLValue I was thinking

```sql
char --> char(1) -- fixed-length char 1
```

It should return `char[]` be called `getChars()` — at a minimum it should be plural.

johnedquinn:
I'd like PartiQLCursor to exist -- however, after some thought, I'd like it to exist in another package (maybe `partiql-client`). Since users of `partiql-eval` are supposed to be implementing a PartiQL Server per-se, I figured that the consumption of data over the wire should be exposed by a PartiQL Client. Closing this for now, though this logic can be used in the future.

PR #1469, opened 2024-05-17
Merges main into V1

johnedquinn:
## Description
- Since I'll be performing a pretty large change on main (#1463) that should interfere with V1, I'm preemptively reducing the current diff to zero before I will need to perform the larger merge.

## Other Information
- Just a merge.

## License Information


alancai98 (changes requested):
Seems to be missing a few of commits from `main`. See https://github.com/johnedquinn/partiql-lang-jvm/compare/v1-main-merge...partiql%3Apartiql-lang-kotlin%3Amain

alancai98 (approved):

PR #1476, opened 2024-05-30
struct type field meta

yliuuuu:
## Relevant Issues
- [Closes/Related To] Issue N/A

## Description
- Ability to add meta in Field of StructType. 
- The purpose was to store information on field (attribute) directly, instead of push the metadata into type of field, which is not ergonomic and carries different semantic meaning. 
- To make the change backward compatible: instead of adding a parameter in the data class construct, we come up with a secondary constructor and override the equals and hash code function. We also introduced a function called copyWithMeta() to mimic the data class copy behavior.  

## Other Information
  - Yes.

  -  No
 
- No

Yes.

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ Thank you to all who have contributed!
 ## [Unreleased]
 
 ### Added
-
+- Ability to append Meta in Field of Struct Type. 
 ### Changed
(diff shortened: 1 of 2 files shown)


yliuuuu:
Closed in favor of PartiQL Type work.

PR #1480, opened 2024-06-03
[DRAFT] Initializes PartiQL-system interfaces and document.

RCHowell:
## Description

https://github.com/partiql/partiql-lang-kotlin/blob/v1-path/partiql-planner/README.adoc

This PR initializes the interfaces for the 1.0 planner interfaces which model the _PartiQL-system_. This review is meant to collect feedback on the high-level public interfaces and documentation before updating internals.

Details can be found in the [README.adoc.](https://github.com/partiql/partiql-lang-kotlin/blob/v1-path/partiql-planner/README.adoc)

## License Information


partiql-planner/src/main/kotlin/org/partiql/planner/metadata/Scope.kt
```diff
@@ -0,0 +1,61 @@
...
+     */
+    public fun getName(): String
+
+    /**
+     * Get a table by name.
+     *
+     * @param name  The case-sensitive [Table] name.
+     * @return      The [Table] or null if not found.
+     */
+    public fun getTable(name: String): Table? = null
```
johnedquinn, line 31: For this method (as well as the ones below), I'm wondering how we can determine the case-sensitive names for third-party consumers (that may not consistently upper-or-lower-case their table names upon storage).

partiql-planner/README.adoc
```diff
@@ -0,0 +1,76 @@
...
+
+=== Routine Lookup
+
+Let _N_ be a routine name with steps _n~1~ . n~2~ ... n~i~_ and let _S_ be the session scope with name _s~1~_ ... _s~j~_.
+
+. If n = 1 then the session path is searched left-to-right for _n~1~_.
+. If n > 1, then let _R_ be the fully-qualified routine name.
+.. If _N_ is a relative name, then _R = s~1~ . s~2~ ... s~j~ . n~1~ . n~2~. ... n~i~_ .
+.. If _N_ is an absolute name, then _R = n~1~ . n~2~ ... n~i~_ .
+. Lookup the fully-qualified name _R_ and fail compilation if not found.
```
johnedquinn, line 64: Regarding "fully-qualified" vs "absolute", I think the following suggestion is a bit more clear.

```suggestion
Let _N_ be the routine invocation name (absolute or relative) with steps _n1_, _n2_, …​, _ni_. Let _S_ be the session scope with name _.s1.s2. …​ sj_. Let _R_ by the absolute name of the invoked routine.

1. If _N_ is an absolute name, then _R = . n1 . n2 …​ ni_ . Return _R_.
2. If n = 1 then the function path is searched left-to-right for _n1_.
3. If n > 1, then _R = . s1 . s2 …​ sj . n1 . n2. …​ ni_ .
4. Lookup the absolute name _R_. Return _R_ if found. Fail compilation if not found.
```

Maybe the "function path" should be SQL Path or invocation path. Up to you.

RCHowell: Typo, it should have been "session path"

partiql-planner/src/main/kotlin/org/partiql/planner/metadata/Name.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import java.util.Spliterator
+import java.util.function.Consumer
+
+/**
+ * Thin wrapper over a list of strings.
+ *
+ * @property steps
+ */
+public data class Name(public val steps: List<String>) : Iterable<String> {
```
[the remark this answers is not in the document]
RCHowell: Neither, it's just a name.

partiql-planner/README.adoc
```diff
@@ -0,0 +1,76 @@
...
+| Relative Name |  `n~1~.---.n~n~`
+| Absolute Name | `.n~1~.---.n~n~`
+
+|===
+
+=== Table Resolution
+
+Let _N_ be a name with steps _n~1~ . n~2~ ... n~i~_ and let _S_ be the session scope with name _s~1~_ ... _s~j~_.
+
+. Let _T_ be the fully-qualified table name.
```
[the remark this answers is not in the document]
RCHowell: Yes, you can have a relative-qualified path.

```sql
x.y.z     -- relative-qualified
.a.b.c    -- absolute-qualified
```

I don't think "qualified" adds any value here though, and should just be "relative name" and "absolute name".

partiql-planner/src/main/kotlin/org/partiql/planner/metadata/Function.kt
```diff
@@ -0,0 +1,55 @@
...
+        @JvmField public val type: TempType,
+    )
+
+    /**
+     * PartiQL-function properties.
+     *
+     * @property isNullCall
+     */
+    public data class Properties(
+        @JvmField public val isNullCall: Boolean = false,
```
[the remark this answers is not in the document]
RCHowell: No I don't believe it's necessary for function calls, only come very special operators (which will come later).

partiql-planner/src/main/kotlin/org/partiql/planner/metadata/Scope.kt
RCHowell: We don't determine the case-sensitive names; I think you have consumer-implementer backwards. We case-normalize to lowercase and it's up to the implementer how to handle.

In other words, all usage of the Scope comes from the planner; so the planner is only "consumer" of Scope — implementers are responsible for handling their own mapping from _PartiQL's case-normalized_

PR #1481, opened 2024-06-04
Merges main into v1

johnedquinn:
## Description
- Merges `main` into `v1`.
- These changes notably contain the removal of NullType and MissingType. I'd mostly look at the final commit (https://github.com/partiql/partiql-lang-kotlin/commit/cc3654a69562d9e2a3231f7456ea196f3f796281) to see what I needed to change.

## Other Information
- This is unreleased

  and [Code Style Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CODE_STYLE.md)? **NO**

## License Information


partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -566,113 +561,153 @@ internal class PlanTyper(private val env: Env) {
...
+            if (elementTypes.isEmpty()) {
+                return ProblemGenerator.missingRex(
+                    rexOpPathIndex(root, key),
+                    ProblemGenerator.expressionAlwaysReturnsMissing("Only lists and s-expressions can be indexed with integers, found ${root.type}")
+                )
+            }
+            return rex(unionOf(elementTypes), rexOpPathIndex(root, key))
         }
 
+        private fun Rex.isLiteralMissing(): Boolean = this.op is Rex.Op.Lit && this.op.value.withoutAnnotations() == missingValue()
```
alancai98, line 600: Wondering if we could just do:

```suggestion
        private fun Rex.isLiteralMissing(): Boolean = this.op is Rex.Op.Lit && this.op.value is MissingValue
```

or are the annotations part of this check?

partiql-planner/src/test/kotlin/org/partiql/planner/PlannerErrorReportingTests.kt
```diff
@@ -182,15 +181,19 @@ internal class PlannerErrorReportingTests {
                 assertOnProblemCount(0, 1)
             ),
             // Chained, demostrate missing trace.
+            // TODO: We currently don't have a good way to retain missing value information. The following test
+            //  should have 2 errors.
```
alancai98, line 185: Should this have 2 **warnings** rather than 2 **errors**? Also what would be the two warnings/errors?

alancai98: Might be worth tracking in a GH issue if there are more warnings/errors expected from the planner here.

partiql-planner/src/test/kotlin/org/partiql/planner/PlannerErrorReportingTests.kt
```diff
@@ -212,27 +215,7 @@ internal class PlannerErrorReportingTests {
...
-            ),
-            TestCase(
-                """
-                    -- both branches are missing, problem
-                    CASE WHEN  
-                        1 = 1 THEN MISSING
-                        ELSE MISSING END
-                """.trimIndent(),
-                true,
-                assertOnProblemCount(0, 1),
```
alancai98, line 235: Why are these 2 tests deleted?

partiql-planner/src/test/kotlin/org/partiql/planner/PlannerErrorReportingTests.kt
```diff
@@ -264,28 +247,30 @@ internal class PlannerErrorReportingTests {
...
                 "not_a_function(1)",
                 true,
                 assertOnProblemCount(0, 1),
-                StaticType.MISSING
+                StaticType.ANY
             ),
 
             // 1 + not_a_function(1)
             //  The continuation will return all numeric type
+            // TODO: Should the warning count be 1? Does it matter if it is zero?
```
alancai98, line 261: Probably shouldn't matter that the warning count is 0 if an error is already being reported.

partiql-planner/src/test/kotlin/org/partiql/planner/PlannerErrorReportingTests.kt
```diff
@@ -264,28 +247,30 @@ internal class PlannerErrorReportingTests {
...
             TestCase(
                 "1 + not_a_function(1)",
                 false,
-                assertOnProblemCount(1, 1),
-                StaticType.MISSING,
+                assertOnProblemCount(0, 1),
+                StaticType.unionOf(StaticType.INT4, StaticType.INT8, StaticType.INT, StaticType.FLOAT, StaticType.DECIMAL),
             ),
+            // TODO: Should the warning count be 1? Does it matter if it is zero?
```
alancai98, line 268: Not sure why the tests are repeated. Seems like one can be deleted? Or maybe `signal` should be `true` for this one?

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt
```diff
@@ -760,22 +760,19 @@ class PlanTyperTestsPorted {
             ErrorTestCase(
                 name = "BITWISE_AND_MISSING_OPERAND",
                 query = "1 & MISSING",
-                expected = unionOf(INT4, INT8, INT),
+                expected = ANY, // TODO: Is this unionOf(INT4, INT8, INT) ?
```
alancai98, line 763: Hmm thought it should also be the `union(INT4, INT8, INT)` #1463. May need to check this again?

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt
```diff
@@ -3827,15 +3821,26 @@ class PlanTyperTestsPorted {
...
+                        StructType(
+                            fields = listOf(
+                                StructType.Field("_1", ANY)
+                            ),
+                            contentClosed = true,
+                            constraints = setOf(
+                                TupleConstraint.Open(false),
+                                TupleConstraint.UniqueAttrs(true),
+                            )
+                        ),
```
alancai98, line 3842: So why are there two possible struct types returned for this test and a couple below?
- one that is open content and empty set of fields
- one that is closed content w/ one field, `_1` of type `ANY`

partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEngineDefaultTest.kt
```diff
@@ -1385,6 +1333,73 @@ class PartiQLEngineDefaultTest {
             )
         ).assert()
 
+    @Test
+    @Disabled(
+        """
+            We currently do not have support for consolidating collections containing MISSING/NULL. The current
+            result (value) is correct. However, the types are slightly wrong due to the SUM__ANY_ANY being resolved.
```
alancai98, line 1340: What should the correct type be then? Does this need corrected for just `t_c_sum`?

alancai98 (commented):
Reviewed just the last commit fixing the tests [`cc3654a` (#1481)](https://github.com/partiql/partiql-lang-kotlin/pull/1481/commits/cc3654a69562d9e2a3231f7456ea196f3f796281)

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt
johnedquinn: Agreed -- however, I'd like to just get this in. Realistically, if we want to narrow down the type, this can be accomplished using the "unknown" type that PostgreSQL exposes.

partiql-planner/src/test/kotlin/org/partiql/planner/PlannerErrorReportingTests.kt
johnedquinn: Updating the comment to give more context, however I'd say that this is inconsequential and not worthy of a GH issue to track. What we really care about are warnings/errors being populated in the first place. I'm honestly not entirely sure why these tests are strict on the number of warnings/errors.

If you debug a C program to fix an error, oftentimes recompiling it will show plenty more errors.

partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEngineDefaultTest.kt
johnedquinn: Well, we don't have solidified type semantics when you see a literal NULL in a collection of tuples. Therefore, there is no correct type. I just know that the SUM that gets resolved takes in dynamic and returns a `null.null` rather than `null.int`.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/ProblemGenerator.kt
```diff
@@ -40,16 +40,16 @@ internal object ProblemGenerator {
...
-        rex(StaticType.MISSING, rexOpMissing(problem, listOf(causes)))
+        rex(StaticType.ANY, rexOpMissing(problem, listOf(causes)))
 
     fun errorRex(causes: List<Rex.Op>, problem: Problem): Rex =
-        rex(StaticType.MISSING, rexOpErr(problem, causes))
+        rex(StaticType.ANY, rexOpErr(problem, causes))
 
     fun errorRex(trace: Rex.Op, problem: Problem): Rex =
-        rex(StaticType.MISSING, rexOpErr(problem, listOf(trace)))
+        rex(StaticType.ANY, rexOpErr(problem, listOf(trace)))
```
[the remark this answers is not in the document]
johnedquinn: Hmm. Good question. I'm unsure. `unknown` is used for literal values whose types need to be inferred based on their context (and original value).

Just thinking out loud here, but the goal of the planner is to produce a plan that can be "almost" directly executed. Given our gradual nature, the execution must execute the same way as a fully typed plan would have. Therefore, if `unknown` is treated in such a way that it may be handled in function resolution, then I think it would be appropriate.

Maybe that is appropriate in some scenarios. But I'm also thinking about functions that always receive the missing value as one of its inputs. I imagine that, currently, this probably collapses as a problem to always return missing -- though, it should be typed as the original function's type.

partiql-parser/build.gradle.kts
```diff
@@ -90,6 +90,8 @@ tasks.processResources {
     from("src/main/antlr") {
         include("**/*.g4")
     }
+    // TODO remove in next major version release.
+    rename("PartiQLParser.g4", "PartiQL.g4")
```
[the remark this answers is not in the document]
johnedquinn: ACK. I'm going to keep it since we know that we plan on merging this to main to perform a release before V1. This should be removed once all of our deprecation notices are eventually deleted.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt
alancai98: Could consider capturing the future addition of an "unknown" type in a GH issue. Since this is just a merge, we can punt on the behavior on this (and similar tests).

partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEngineDefaultTest.kt
alancai98: I see. Test comment makes sense to me now.

partiql-planner/src/test/kotlin/org/partiql/planner/PlannerErrorReportingTests.kt
alancai98: Makes sense w/ me. I think the original context for the "always returns missing" planning error was for data type mismatches (e.g. adding a string with an int). The literal `MISSING` doesn't denote a data type mismatch so we could say this case isn't actually an error.

alancai98 (approved):
Merge and updated tests look good. Left a minor suggestion to track adding an `unknown` type since there are a few tests that will have a different result type.

johnedquinn:
As this was pre-approved, and there are no changes between this revision and 145991eae8b4d5e0eb5e1d1ec2985228b72a43ea, I'm overriding and merging. The conformance test fix is actively being worked on by @alancai98 on the main branch and will be ported to `v1` after.

Issue #1482, opened 2024-06-05
Negative numeric literals are wrapped in unary operators.


RCHowell:
## Description

Given a numeric literal like `-1`, this becomes an AST like `unary_minus(literal(1))` which has strange downstream implications. The desired behavior is to have the sign as part of the literal like `literal(-1)`.

## To Reproduce

I noticed when pretty-printing I got `-(1)` back rather than `-1`.

## Expected Behavior

```sql
-1

-- AST literal(-1)
```

## Additional Context
- Java version: 11
- PartiQL version: 0.14.5

PR #1484, opened 2024-06-06
Fixes parsing of signed numeric literals

RCHowell:
## Relevant Issues

#1482 

## Description

Applies +/- signs to numeric literals.

## Other Information
No

No

No

Yes

## License Information

### DIFF
--- a/partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
+++ b/partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
@@ -209,6 +209,14 @@ import org.partiql.parser.SourceLocation
 import org.partiql.parser.SourceLocations
 import org.partiql.parser.antlr.PartiQLParserBaseVisitor
 import org.partiql.parser.internal.util.DateTimeUtils
+import org.partiql.value.DecimalValue
+import org.partiql.value.Float32Value
+import org.partiql.value.Float64Value
+import org.partiql.value.Int16Value
+import org.partiql.value.Int32Value
+import org.partiql.value.Int64Value
+import org.partiql.value.Int8Value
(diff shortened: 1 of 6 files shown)


partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -1488,9 +1501,48 @@ internal class PartiQLParserDefault : PartiQLParser {
...
+            val op = when (ctx.sign.type) {
+                GeneratedParser.NOT -> Expr.Unary.Op.NOT
+                GeneratedParser.PLUS -> Expr.Unary.Op.POS
+                GeneratedParser.MINUS -> Expr.Unary.Op.NEG
+                else -> throw error(ctx.sign, "Invalid unary operator")
+            }
+            // If argument is not a literal, then return the op.
+            val arg = visit(ctx.rhs) as Expr
+            return when (arg) {
+                is Expr.Lit -> arg.negate(op)
```
yliuuuu, line 1518: I think this potentially can cause overflow: 

```
-(-2147483648) => 2147483648 which can not be represented by 32 bit integer. 
```

yliuuuu (commented):
Quote from the issue #1482 
> Given a numeric literal like -1, this becomes an AST like unary_minus(literal(1)) which has strange downstream implications. The desired behavior is to have the sign as part of the literal like literal(-1).

Could you please elaborate why you think `unary_minus(literal(1))` is a bad idea? 

I don't know if this is worth to attempt to lower some nodes during the parsing stage. 

I see that you listed pretty printer as an example, but I feel like outputting -(1) is fine and if we want to change this, it might be better to modify the pretty printer instead of the parsing logic.

yliuuuu (approved):

Issue #1491, opened 2024-06-24
`PartiQLValueTextWriter` currently does not implement printing datetime values


alancai98:
## Description
[`PartiQLValueTextWriter`](https://github.com/partiql/partiql-lang-kotlin/blob/68e4849b152debc3bf5964c9dccd0082d80ed26e/partiql-types/src/main/kotlin/org/partiql/value/io/PartiQLValueTextWriter.kt#L50) does not print datetime values. Currently it does not implement the writer for datetime values, so attempting to print out a datetime value will result in an `IllegalArgumentException`.

This is also pointed out in the tests:

https://github.com/partiql/partiql-lang-kotlin/blob/68e4849b152debc3bf5964c9dccd0082d80ed26e/partiql-types/src/test/kotlin/org/partiql/value/io/PartiQLValueTextWriterTest.kt#L39-L40.

## To Reproduce
1. Call `PartiQLValueTextWriter`'s `append`, which calls the `ToString` visitor with a datetime `PartiQLValue`
2. Get an `IllegalArgumentException` -- `Caused by: java.lang.IllegalArgumentException: Cannot write value of type DATE`

## Expected Behavior
- Expect `PartiQLValueTextWriter` to print out datetime values

## Additional Context
- Java version: 11
- PartiQL version: tested w/ 0.14.6-SNAPSHOT and 0.14.5

PR #1492, opened 2024-06-25
Add implementation of date, time, and timestamp values for `PartiQLValueTextWriter`

alancai98:
## Relevant Issues
- Implements #1491

## Description
- Adds an implementation of datetime values for the `PartiQLValueTextWriter`
- With this change, `SqlDialect` will also be able to print datetime literals.

## Other Information

## License Information

### DIFF
--- a/partiql-types/src/main/kotlin/org/partiql/value/io/PartiQLValueTextWriter.kt
+++ b/partiql-types/src/main/kotlin/org/partiql/value/io/PartiQLValueTextWriter.kt
@@ -18,6 +18,7 @@ import org.partiql.value.BagValue
 import org.partiql.value.BoolValue
 import org.partiql.value.CharValue
 import org.partiql.value.CollectionValue
+import org.partiql.value.DateValue
 import org.partiql.value.DecimalValue
 import org.partiql.value.Float32Value
 import org.partiql.value.Float64Value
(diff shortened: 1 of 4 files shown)


RCHowell (approved):

partiql-types/src/main/kotlin/org/partiql/value/io/PartiQLValueTextWriter.kt
```diff
@@ -192,6 +201,74 @@ public class PartiQLValueTextWriter(
...
+        }
+
+        private fun sqlString(t: Timestamp): String {
+            val yyyy = padZeros(t.year, 4)
+            val mon = padZeros(t.month, 2)
+            val dd = padZeros(t.day, 2)
+            val hh = padZeros(t.hour, 2)
+            val min = padZeros(t.minute, 2)
+            val ss = padZeros(t.decimalSecond.toInt(), 2)
+            val frac = t.decimalSecond.remainder(BigDecimal.ONE).toString().substring(1) // drop leading 0
```
yliuuuu, line 267: What is the reason to separate the big decimal in term of second and fraction second in print?

yliuuuu (commented):
One small comment, otherwise looks good to me.

partiql-types/src/main/kotlin/org/partiql/value/io/PartiQLValueTextWriter.kt
alancai98: I broke down the big decimal representing the seconds into full seconds and fractional to keep the full seconds portion of the string two characters.

E.g. if the seconds was `9.123`, I want to ensure the resulting string of the seconds was `09.123`. Perhaps I can simplify this logic further by just casing if the seconds is less than 10.

alancai98: Possible simplification:

```
val ss = if (t.decimalSecond < BigDecimal.TEN) {
    "0${t.decimalSecond}"
} else {
    t.decimalSecond.toString()
}
```

yliuuuu: Oh I see. Yeah I missed the padding zero part. Thanks.

yliuuuu (approved):

Issue #1496, opened 2024-06-27
PartiQL-Environment Tracking Issue


RCHowell:
## What?

_This issue is to help the review of the SPI dependency switch; the review will be broken down into several PRs targeting [v1-metadata](https://github.com/partiql/partiql-lang-kotlin/tree/v1-metadata)_

## Reviews
- [ ] Define Interfaces
    - Reviewer Alan
- [ ] Define prelude (in parallel to 1)
    * Reviewers Josh and Arash
    * https://github.com/partiql/partiql-lang-kotlin/pull/1490
- [ ] Update plugins (partiql-local and partiql-memory)
- [ ] Update partiql-planner
- [ ] Fix build / update tests

Branches

- [v1-metadata](https://github.com/partiql/partiql-lang-kotlin/tree/v1-metadata) — destination branch
- [v1-prelude](https://github.com/partiql/partiql-lang-kotlin/tree/v1-prelude) — dump of functions to .ion files
- v1-fns — current working/hacking branch

PR #1497, opened 2024-06-27
Defines the V1 bindings and metadata interfaces (updating SPI)

RCHowell:
## Relevant Issues

https://github.com/partiql/partiql-lang-kotlin/issues/1496

## Description

_This PR introduces the V1 interfaces for data and metadata access._

In PartiQL 0.x implementations, all packages depended on SPI for access to data and catalog information. With the additions of Datum and PType, we are flipping the dependency such that SPI depends upon the package-specific interfaces.

Current customers of partiql-lang-kotlin (eval) use the lang.eval.Bindings and lang.eval.ExprValue interfaces which are inside of the lang.eval package. However, there is not an equivalent metadata layer. We initially used Trino's SPI approach to the metadata layer (early 2023), but it results in two disparate extension models. This PR is an effort to extend the existing partiql-lang-kotlin model to include metadata while defining the SPI as a layer on top of both eval.Bindings and planner.Metadata. The end result is to make the changes from 0.x to 1.0 more of an extension to current models rather than a different design approach altogether.

## Other Information
- Do your changes comply with the [Contributing Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CONTRIBUTING.md) and [Code Style Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CODE_STYLE.md)? **YES**

## License Information


partiql-planner/src/main/kotlin/org/partiql/planner/metadata/Table.kt
```diff
@@ -0,0 +1,40 @@
...
+     */
+    public fun getSchema(): PType = PType.typeDynamic()
+
+    /**
+     * A [Table] can be one of several [Kind]s.
+     */
+    public enum class Kind {
+        TABLE,
+        VIEW,
+        INDEX,
```
alancai98, line 37: Are we sure we want to add these other `Kind`s of `Table`s at this time? Perhaps we could just have `TABLE` and `OTHER` for now?

partiql-planner/src/main/kotlin/org/partiql/planner/metadata/Metadata.kt
```diff
@@ -0,0 +1,12 @@
+package org.partiql.planner.metadata
+
+/**
+ * Top-level metadata interface for access to object descriptors.
```
alancai98, line 4: Slightly confused on the distinction between `Metadata` and `Catalog`. Here we define the `Metadata` interface to be "Top-level metadata interface for access to object descriptors."

while in `Namespace.kt`, it's mentioned "top-level namespaces are called "Catalogs"".

Is there a distinction between the two? If not, then could we utilize the same terminology?

partiql-spi/src/test/kotlin/org/partiql/spi/connector/sql/SqlHeader.kt
```diff
@@ -1,831 +0,0 @@
-package org.partiql.spi.connector.sql
```
alancai98, line 1: I'm assuming all of the headers will be generated as part of https://github.com/partiql/partiql-lang-kotlin/pull/1490?

alancai98 (commented):
Some minor comments/questions. Otherwise refactor looks good.

partiql-spi/src/test/kotlin/org/partiql/spi/connector/sql/SqlHeader.kt
RCHowell: Yes

partiql-planner/src/main/kotlin/org/partiql/planner/metadata/Table.kt
RCHowell: I've removed these now.

partiql-planner/src/main/kotlin/org/partiql/planner/catalog/Catalog.kt
```diff
@@ -0,0 +1,56 @@
...
+     */
+    public fun listNamespaces(namespace: Namespace)
+
+    /**
+     * Get a function's variants by name.
+     *
+     * @param name  The case-sensitive [Routine] name.
+     * @return A collection of all [Routine]s in the current namespace with this name.
+     */
+    public fun getFunctions(name: Name): Collection<Routine> = emptyList()
```
alancai98, line 55: nit: for naming consistency, should call `getRoutines`?

```suggestion
    /**
     * Get a routines' variants by name.
     *
     * @param name  The case-sensitive [Routine] name.
     * @return A collection of all [Routine]s in the current namespace with this name.
     */
    public fun getRoutines(name: Name): Collection<Routine> = emptyList()
```

partiql-planner/src/main/kotlin/org/partiql/planner/catalog/Catalog.kt
```diff
@@ -0,0 +1,56 @@
...
+     * List top-level namespaces from the catalog.
+     */
+    public fun listNamespaces(): Collection<Namespace> = emptyList()
+
+    /**
+     * List all child namespaces from the namespace.
+     *
+     * @param namespace
+     */
+    public fun listNamespaces(namespace: Namespace)
```
alancai98, line 47: Return type seems off for this function perhaps should be a `Collection<Namespace>`?

```suggestion
    /**
     * List all child namespaces from the namespace.
     *
     * @param namespace
     */
    public fun listNamespaces(namespace: Namespace): Collection<Namespace> = emptyList()
```

alancai98 (approved):
some minor comments in the Catalog interface. approving, assuming those get addressed in this PR or a subsequent PR

PR #1504, opened 2024-07-10
Fix published jars for non-shadowed packages

alancai98:
## Relevant Issues
- None

## Description
- Fixes the publishing of jars for non-shadowed packages. Previously, all the packages that don't shadow any packages would fatjar all the dependencies (e.g. kotlin, ion-element). With this change, we now exclude those dependencies, so the jar contains what was there previously.
- Previous packages that shadowed certain dependencies (i.e. `partiql-parser` and `partiql-lang-kotlin`) did not have this issue
- Testing
  - Ran `publishToMavenLocal` and compared the jars between 0.14.5 and this version
  - Published jars for this version do not have the extra dependencies and retain the existing APIs

## Other Information
- Updated Unreleased Section in CHANGELOG: **[NO]** No api changes.

## License Information

### DIFF
--- a/partiql-ast/build.gradle.kts
+++ b/partiql-ast/build.gradle.kts
@@ -25,6 +25,17 @@ dependencies {
     api(project(":partiql-types"))
 }
 
+tasks.shadowJar {
+    configurations = listOf(project.configurations.shadow.get())
+}
+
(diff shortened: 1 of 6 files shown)


yliuuuu (approved):
I verified the patch removed the unshadowed dependencies in the fat jar. 

Thanks for the quick fix.

PR #1506, opened 2024-07-10
Add set op typing; fixes various behaviors in set op parsing and modeling

### LINKED ISSUE #1501: [main] Support `UNION` typing in the `main` branch
`UNION` is already in the `v1` plan and plan typer. For adding `UNION` support in Scribe https://github.com/partiql/partiql-scribe/issues/55, we may need to cut a release off of `main`/`v0.14.6` branch that includes `UNION` in the plan and plan typer (assuming we aren't releasing `v1` soon).

`UNION` in the `main` branch does not support set quantifiers. To align the `main` branch plan with `v1`, we would need to update the plan, which may necessitate a major version release.

Another complication is whether the removal of `null` and `missing` should be included in such a release, given the usage of these unknown types in Scribe.

alancai98:
## Relevant Issues
- Closes #1501 

## Description
- partiql-plan:
  - adds set quantifier to `Rel` set ops. Plan is slightly simpler than the set operator modeling used in `v1` branch. Primarily, no additional set class wrapping the set operators
  - adds `Rex` set ops to plan
- partiql-planner: adds typing support for set operator
  - Currently just performs the schema size validation; comparability of column types will not be part of this release to unblock https://github.com/partiql/partiql-scribe/issues/55
- partiql-parser:
  -  distinguishes between PartiQL bag and SQL set ops
  - updates and adds tests following this behavioral change
- partiql-ast:
  -  SqlDialect -- fixes missing parens set op printing
- Tested the updates in Scribe https://github.com/partiql/partiql-scribe/pull/56
- Prepares v0.14.6 release

## Other Information

  - Yes, from prior discussion, this is a backwards incompatible change due to adding an additional field (set quantifier) to set operators in the plan. But since the plan typer always gave an error when typing the plan, no customers should have been using the previous modeling.

## License Information


partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -241,16 +242,95 @@ internal class PlanTyper(
...
+            // TODO: [RFC-0007](https://github.com/partiql/partiql-lang/blob/main/RFCs/0007-rfc-bag-operators.md)
+            //  states that the types must be "comparable". The below code ONLY makes sure that types need to be
+            //  the same. In the future, we need to add support for checking comparable types.
+            for (i in 0..lhs.type.schema.lastIndex) {
+                val lhsBindingType = lhs.type.schema[i].type
+                val rhsBindingType = rhs.type.schema[i].type
+                if (lhsBindingType != rhsBindingType) {
+                    return false
+                }
+            }
```
alancai98, line 315: Since the current schema type comparison is only partially implemented, I'll delete this validation block for the 0.14.6 release to unblock https://github.com/partiql/partiql-scribe/issues/55. The different Scribe dialects may have their own notion of comparability and perform their own set operator checks.

alancai98:
Will mark this PR as a draft to account for some
- parsing changes (current parser parses every set op to `bag_op`)
  - we should distinguish between `bag_op` and `SFW.set_op`
    - `SFW.set_op` when `DISTINCT` is not specified and both arguments are `SFW`
    - `bag_op` otherwise
- plan changes
  - current plan modeling maps `bag_op`s and `SFW.set_op`s as a `Rel` node w/ two `Rel` arguments
  - we should have a separate node for `bag_op`s over `Rex`s and `SFW.set_op`s over `Rel`s

alancai98:
Above TODOs should be addressed. Still have a couple more before PR is ready for review:
- Check SqlDialect behavior for printing `BagOp` -- seems to be omitting parens for right-associative case 
- Test PR's changes in local maven release
- Update added set op parse tests to use the `AstBuilder` to remove some unneeded code

partiql-ast/src/main/kotlin/org/partiql/ast/sql/internal/InternalSqlDialect.kt
```diff
@@ -82,6 +82,13 @@ internal abstract class InternalSqlDialect : AstBaseVisitor<InternalSqlBlock, In
             t = t concat ")"
             t
         }
+        is Expr.BagOp -> {
```
alancai98, line 85: Parens were missing from output to indicate precedence of the arguments.

E.g. previously an ast generated from the following query:
```
SELECT * FROM foo UNION (SELECT * FROM bar UNION SELECT * FROM baz)
```
would get printed as
```
SELECT * FROM foo UNION SELECT * FROM bar UNION SELECT * FROM baz
```

partiql-ast/src/main/resources/partiql_ast.ion
```diff
@@ -500,7 +500,7 @@ expr::[
     where:    optional::expr,
     group_by: optional::group_by,
     having:   optional::expr,
-    set_op:   optional::{
+    set_op:   optional::{ // TODO modeling of `set_op` needs updated to support left-associative set ops https://github.com/partiql/partiql-lang-kotlin/issues/1507
```
alancai98, line 503: The current modeling of SQL set ops is not correct in the AST (see https://github.com/partiql/partiql-lang-kotlin/issues/1507). I chose to just reuse the `Expr.BagOp` node rather than use this `Expr.SFW.SetOp` node for this PR to preserve backwards compatibility for this release.

We should fix the modeling before the `v1` release.

partiql-lang/src/main/kotlin/org/partiql/lang/prettyprint/QueryPrettyPrinter.kt
```diff
@@ -320,7 +320,14 @@ class QueryPrettyPrinter {
             is PartiqlAst.Expr.Or -> writeNAryOperator("OR", node.operands, sb, level)
             is PartiqlAst.Expr.InCollection -> writeNAryOperator("IN", node.operands, sb, level)
             is PartiqlAst.Expr.BagOp -> {
-                var name = node.op.javaClass.simpleName.toUpperCase().replace("_", " ")
+                var name = when (node.op) {
```
alancai98, line 323: Buggy previous behavior would print `OUTER UNION` node as `OUTERUNION`.

Could have been more clever w/ printing but since this is legacy code that will be removed ahead of `v1`, just opted for a simple solution.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -1113,6 +1113,17 @@ internal class PartiQLParserDefault : PartiQLParser {
...
+        /**
+         * Verifies if all of the [args] are
+         * 1. [Expr.SFW] or
+         * 2. [Expr.BagOp] and is a SQL Set op (i.e. [Expr.BagOp.outer] != true)
+         */
+        private fun argsAreSFW(args: List<Expr>): Boolean {
+            return args.all { arg ->
+                arg is Expr.SFW || (arg is Expr.BagOp && arg.outer != true)
+            }
+        }
```
alancai98, line 1125: New parsing logic to distinguish between SQL set ops and PartiQL bag ops
- SQL set ops
  - require no `OUTER` before set operation and
  - require args are SQL set ops or SFW expression
- PartiQL bag ops
  - use `OUTER` keyword or
  - do not use `OUTER` keyword but one or more arguments is a PartiQL bag op or non-SFW expression

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -241,16 +242,88 @@ internal class PlanTyper(
...
+
+        /**
+         * @return whether each type of the [lhs] is comparable to its counterpart on the [rhs]
+         * @param lhs should be typed already
+         * @param rhs should be typed already
+         */
+        private fun setOpSchemaTypesMatch(lhs: Rel, rhs: Rel): Boolean {
+            // TODO: [RFC-0007](https://github.com/partiql/partiql-lang/blob/main/RFCs/0007-rfc-bag-operators.md)
+            //  states that the types must be "comparable". For now, we will always return true. In the future, we need
+            //  to add support for checking comparable types.
```
alancai98, line 308: For now, do not do comparability check of types. Only check same number of columns for both set ops.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -163,6 +160,29 @@ internal object RelConverter {
             return rel
         }
 
+        // Create a SQL set op
+        override fun visitExprBagOp(node: Expr.BagOp, ctx: Rel): Rel {
+            // Assumes parser correctly only allows Expr.SFW or other Expr.BagOps with Expr.SFW arguments when
+            // converting to the SQL set op
+            assert(node.lhs is Expr.SFW || node.lhs is Expr.BagOp)
+            assert(node.rhs is Expr.SFW || node.rhs is Expr.BagOp)
```
RCHowell, line 168: I suggest adding a message here.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -163,6 +160,29 @@ internal object RelConverter {
...
+            }
+            val lhsRel = visitExpr(node.lhs, ctx)
+            val rhsRel = visitExpr(node.rhs, ctx)
+            val op = when (node.type.type) {
+                SetOp.Type.UNION -> Rel.Op.Union(setq, lhsRel, rhsRel)
+                SetOp.Type.EXCEPT -> Rel.Op.Except(setq, lhsRel, rhsRel)
+                SetOp.Type.INTERSECT -> Rel.Op.Intersect(setq, lhsRel, rhsRel)
+            }
+            return Rel(
+                type = Rel.Type(listOf(Rel.Binding("_0", StaticType.ANY)), props = emptySet()),
```
RCHowell, line 181: May just set type to nil?

RCHowell (changes requested):
Still reviewing, minor comments

RCHowell (approved):

PR #1508, opened 2024-07-11
Updates rules for variable resolution

johnedquinn:
## Description
- This increases conformance by 184 tests (3.14%). See the [report](https://github.com/partiql/partiql-lang-kotlin/actions/runs/9898594883?pr=1508).
- This updates rules for variable resolution
- Adds support for casting from dynamic
- Updates tests to give greater visibility into errors

## Other Information

## License Information

### DIFF
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeEnv.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeEnv.kt
@@ -1,44 +1,16 @@
 package org.partiql.planner.internal.typer
 
 import org.partiql.planner.internal.Env
-import org.partiql.planner.internal.ir.Rel
 import org.partiql.planner.internal.ir.Rex
-import org.partiql.planner.internal.ir.rex
-import org.partiql.planner.internal.ir.rexOpLit
-import org.partiql.planner.internal.ir.rexOpPathKey
-import org.partiql.planner.internal.ir.rexOpPathSymbol
-import org.partiql.planner.internal.ir.rexOpVarLocal
-import org.partiql.spi.BindingCase
-import org.partiql.spi.BindingName
 import org.partiql.spi.BindingPath
-import org.partiql.types.PType
-import org.partiql.types.PType.Kind
-import org.partiql.types.StaticType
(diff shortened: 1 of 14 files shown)


partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeEnv.kt
```diff
@@ -22,10 +23,16 @@ import org.partiql.value.stringValue
  * @property outer refers to the outer variable scopes that we have access to.
  */
 internal data class TypeEnv(
+    private val globals: Env,
```
RCHowell, line 26: I think we should consider a "scope" class — [I experimented with it here.](https://github.com/partiql/partiql-lang-kotlin/tree/v1-planner/partiql-planner/src/main/kotlin/org/partiql/planner/intern/validate)

The "scope" would replace TypeEnv since TypeEnv is explicitly the name for locals rather than here we have globals+locals which makes sense.

Scope/SqlScope in this example also operates on its own internal classes rather than over stateless IR

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/TypeEnv.kt
```diff
@@ -22,10 +23,16 @@ import org.partiql.value.stringValue
...
     public val schema: List<Rel.Binding>,
     public val outer: List<TypeEnv>
 ) {
 
+    enum class LookupStrategy {
+        LOCALS_FIRST,
+        GLOBALS_FIRST
+    }
```
RCHowell, line 34: Perhaps we rename Scope.kt to Strategy.kt per https://github.com/partiql/partiql-lang-kotlin/blob/v1-planner/partiql-planner/src/main/kotlin/org/partiql/planner/intern/validate/Strategy.kt

RCHowell (changes requested):

johnedquinn:
> I think we should consider a "scope" class ...

I performed a minimal update by renaming TypeEnv to an internal Scope (representing locals). Renamed internal Scope to Strategy. And I created a new internal class called TypeEnv that wraps both locals and globals. Since everything is internal, we can change the name as we see fit.

RCHowell (approved):

PR #1511, opened 2024-07-12
Identifier mode

yliuuuu:
## Relevant Issues
- [Closes/Related To] Issue #1122 

## Note
- This is a draft PR that allows us to further experiment with the idea of adding modes for `lvalue` and `rvalue` behavior. 
- Some tests cases are still to-be-fixed, but the purpose of this PR is as a code reference point to further discuss the identifier behavior. 

## Semantics
- This draft PR enables adding planner modes to determine Case Preservation and Variable Look Up behavior. 
- Case Preservation: 
    -  A binary flag set on the planner: 
    -  This flag determines the lvalue behavior (As Binding, DDL, etc). 
    - If this flag is included: Lvalue (AS binding, DDL, etc) are case-sensitive and stored in the environment as-is. 
    - if this flag is not included: Lvalue are normalized using an implementation defined normalization rule.
         - In this PR we made a conscious decision to set this implementation defined normalization rule to be using the input string. 
         - This is: even without the case preservation flag, the lvalue will behave case sensitively. 
         - This allows us to reduce one moving element for the purpose of testing. 
         - While technically not breaking the semantics of "implementation-defined normalization rule", it does bug the question: do we also want to allow for different case preservation flag? If so, does it make sense to set case preservation to "FOLDING DOWN" and set Variable look up behavior to "FOLDING UP" ? 

- Variable Look Up (RValue): 
   - FOLDING UP: Using upper case text to look up
   - FOLDING DOWN: Using lower case text to look up
   - SENSITIVE: Using original text to look up
   - INSENSITIVE: "Match" behavior - text comparison with case ignored. This is the current behavior. 

## Modeling
- Behavior
   - Consider the follow expression `SELECT bAr .....` 
       - Folding UP -> the query should behave the same as `SELECT "BAR" ...` 
       - Folding DOWN -> the query should behave the same as `SELECT "bar" ...` 
       - SENSITIVE -> the query should behave the same as `SELECT "bAR" ...` 
   - Based on the above observation: we can keep the resolution logic as-is, with the only change to be normalize the identifier in "planning"
   - This PR does this during AST normalization steps, which is an internal process during planning. 

- AST Node: `Binder`
    - The Binder node differs from Identifier in two aspect: 
        - It is used for LValue. 
        - It is always present as unqualified.
        - i.e., `t.c.a AS A`, `t.c.a` is a qualified identifier and `t`, `c`, `a` are identifier symbol. `A` is a binder. 
    - The addition of binder node allow us to modifier the `lvalue` and `rvalue` on the AST level easily, without the need to go down each node and distinguish the type of identifier based on the node parameter. (check if the identifier symbol is an `as_alias`, etc). 

Removed to reduce noise in this PR.
~~- Decouple SelectNormalization with RelConverter.~~
~~- The previous invocation of SelectNormalization is invoked by RelConverter.~~
~~- This is due to the need of handling subquery coercion.~~
~~- I found this jumping between AST and Relational Operation introduces confusion, but to decouple the two, we are forced to add a boolean flag in the AST to indicate coercible. Not something I necessarily like either...~~
~~- We can revert this decouple per later discussion.~~
    
## Other Information
  - No. 

  -  Yes. 

  - No.

 - Yes.

## License Information

### DIFF
--- a/partiql-ast/src/main/resources/partiql_ast.ion
+++ b/partiql-ast/src/main/resources/partiql_ast.ion
@@ -18,7 +18,7 @@ statement::[
     insert::{
       target:       identifier,
       values:       expr,
-      as_alias:     optional::'.identifier.symbol',
+      as_alias:     optional::binder,
       on_conflict:  optional::on_conflict,
     },
 
@@ -34,14 +34,14 @@ statement::[
     upsert::{
       target:   identifier,
       values:   expr,
-      as_alias: optional::'.identifier.symbol',
+      as_alias: optional::binder,
     },
 
     // REPLACE INTO <target> [AS <alias>] <values>
     replace::{
       target:   identifier,
       values:   expr,
-      as_alias: optional::'.identifier.symbol',
+      as_alias: optional::binder,
     },
 
     // UPDATE <target> SET <set clause list> WHERE <expr>
@@ -65,8 +65,8 @@ statement::[
     delete::{
       target: {
         path:     path,
-        as_alias: optional::'.identifier.symbol',
-        at_alias: optional::'.identifier.symbol',
+        as_alias: optional::binder,
+        at_alias: optional::binder,
         by_alias: optional::'.identifier.symbol',
       },
       where:      optional::expr,
@@ -84,7 +84,7 @@ statement::[
           insert::{
             target:       identifier,
             values:       expr,
-            as_alias:     optional::'.identifier.symbol',
+            as_alias:     optional::binder,
             on_conflict:  optional::on_conflict,
           },
           insert_legacy::{
@@ -233,7 +233,7 @@ type::[
     fields: list::[field],
     _ : [
       field :: {
-        name: '.identifier.symbol',
+        name: binder,
         type: '.type',
         // This could be a boolean flag since we only support NOT NULL constraint
         // for struct subfield. But modeling this to be a list of constraints
@@ -273,6 +273,17 @@ identifier::[
     ],
   ],
 ]
+
+// identifier for lvalue
+// binder can not be qualified.
+// it exists for eaiser visitor implementation and maintainence
+binder::{
+  symbol: string,
+  case_sensitivity: [
+    SENSITIVE,
+    INSENSITIVE,
+  ],
+}
 
 // Path Literals
 //  - Much like qualified identifier but allowing bracket notation '[' <int> | <string> ']'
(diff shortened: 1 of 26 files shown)


partiql-planner/src/main/kotlin/org/partiql/planner/internal/astPasses/NormalizeSelect.kt
```diff
@@ -219,7 +212,8 @@ internal object NormalizeSelect {
                     args = tupleUnionArgs,
                     setq = null // setq = null for scalar fn
                 ),
-                setq = select.setq
+                setq = select.setq,
+                coercible = false
```
RCHowell, line 216: What was coercible added for exactly?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
```diff
@@ -120,20 +120,20 @@ internal object RexConverter {
...
+            val sfw = node as? Expr.SFW ?: return rex
+            val select = node.select as? Select.Value ?: return rex
+            return if (select.coercible) {
+                val selectRex = rex.op as Rex.Op.Select
+                rex(
+                    CompilerType(PType.typeDynamic()),
+                    rexOpSubquery(
+                        constructor = selectRex.constructor,
+                        rel = selectRex.rel,
+                        coercion = coercion
```
RCHowell, line 133: When do we coerce SELECT VALUE? I was under the impression we only coerce SQL SELECT

RCHowell (changes requested):

partiql-ast/src/main/resources/partiql_ast.ion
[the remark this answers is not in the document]
yliuuuu: Binder is different than Identifier. 

Identifier: used at place where we do variable reference. It can be optionally qualified. 

Binder: used at place where we do variable declaration. It can not be qualified (AFAIK).

partiql-planner/src/main/kotlin/org/partiql/planner/internal/astPasses/NormalizeSelect.kt
yliuuuu: I removed this part to reduce noise in this PR.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
yliuuuu: I removed this part to reduce noise in this PR.

yliuuuu:
Close in favor of PR #1519

PR #1514, opened 2024-07-16
Updates catalog definitions and tests to fix assemble.

RCHowell:
## Description

This PR stubs out all planner and tests so that I can begin implementing and executing planner tests.

CONTRIBUTING.md)

## License Information


partiql-planner/build.gradle.kts
```diff
@@ -24,17 +24,13 @@ plugins {
...
     api(project(":partiql-types"))
-    implementation(project(":partiql-ast"))
     implementation(Deps.dotlin)
     implementation(Deps.ionElement)
     // Test
     testImplementation(project(":partiql-parser"))
-    testImplementation(project(":plugins:partiql-local"))
-    testImplementation(project(":plugins:partiql-memory"))
-    // Test Fixtures
-    testFixturesImplementation(project(":partiql-spi"))
```
RCHowell, line 37: partiql-planner no long depends on spi which means we cannot use any of the plugins.

partiql-planner/src/main/kotlin/org/partiql/planner/PartiQLPlanner.kt
```diff
@@ -32,30 +34,44 @@ public interface PartiQLPlanner {
...
     public companion object {
 
         @JvmStatic
-        public fun builder(): PartiQLPlannerBuilder = PartiQLPlannerBuilder()
+        public fun builder(): Builder = Builder()
+    }
 
-        @JvmStatic
-        public fun default(): PartiQLPlanner = PartiQLPlannerBuilder().build()
+    public class Builder {
```
RCHowell, line 43: The recurring pattern is to namespace all builders under the interface class.

partiql-planner/src/main/kotlin/org/partiql/planner/PartiQLPlanner.kt
```diff
@@ -32,30 +34,44 @@ public interface PartiQLPlanner {
...
-     * @property instant
-     */
-    public class Session(
-        public val queryId: String,
-        public val userId: String,
-        public val currentCatalog: String,
-        public val currentDirectory: List<String> = emptyList(),
-        public val catalogs: Map<String, ConnectorMetadata> = emptyMap(),
-        public val instant: Instant = Instant.now(),
-    )
```
RCHowell, line 52: Session is moved to an interface.

partiql-planner/src/main/kotlin/org/partiql/planner/catalog/Catalog.kt
```diff
@@ -16,23 +18,19 @@ public interface Catalog {
...
+     *
+     * @param name  The case-sensitive [Table] name.
+     * @return The [Table] or null if not found.
      */
     public fun getTable(session: Session, name: Name): Table? = null
 
-    /**
-     * Get a table by identifier; note that identifiers may be case-insensitive.
-     */
-    public fun getTable(session: Session, identifier: Identifier): Table? = null
```
RCHowell, line 25: Consider `getTable(session: Session, name: Name)` and `searchTable(session, identifier: Identifier)` or something similar.

We have no way of knowing which is the catalog/namespace, the table name, and the path expression — so we must do searching.

partiql-planner/src/main/kotlin/org/partiql/planner/catalog/Catalog.kt
```diff
@@ -55,4 +53,97 @@ public interface Catalog {
...
+            }
+            this.tables[name.getName()] = Table.of(name.getName(), schema)
+            return this
+        }
+
+        public fun build(): Catalog {
+
+            val name = this.name ?: throw IllegalArgumentException("Catalog name must be provided")
+
+            return object : Catalog {
```
RCHowell, line 97: Replacement for partiql-memory plugin

partiql-planner/src/main/kotlin/org/partiql/planner/catalog/Routine.kt
```diff
@@ -79,12 +79,12 @@ public sealed interface Routine {
...
             name: String,
             parameters: Collection<Parameter>,
-            returnType: PType.Kind,
+            returnType: PType,
             properties: Properties = DEFAULT_PROPERTIES,
         ): Scalar = object : Scalar {
             override fun getName(): String = name
             override fun getParameters(): Array<Parameter> = parameters.toTypedArray()
-            override fun getReturnType(): PType.Kind = returnType
+            override fun getReturnType(): PType = returnType
```
RCHowell, line 87: For now use PType to minimize changes.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/FnMatch.kt
```diff
@@ -17,7 +15,7 @@ internal sealed class FnMatch {
      * @property mapping
      */
     data class Static(
-        val signature: FnSignature,
+        val signature: Routine,
```
RCHowell, line 18: You'll see Routine replaces FnSignature everywhere.

partiql-planner/build.gradle.kts
```diff
@@ -112,14 +107,14 @@ tasks.register<Copy>("copyUtils") {
...
 //
-// tasks.register<Copy>("copyNodes") {
-//     includeEmptyDirs = false
-//     dependsOn("codegen")
-//     filter { it.replace(Regex("public (?!(override|(fun visit)))"), "internal ") }
-//     from("$buildDir/tmp")
-//     include("**/Nodes.kt")
-//     into("src/main/kotlin")
-// }
+tasks.register<Copy>("copyNodes") {
```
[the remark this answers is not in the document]
alancai98: we can fix `Nodes.kt` after all the internal plan nodes are more finalized.

alancai98 (commented):
only reviewed a few files -- leaving just an initial comment about `Nodes.kt` getting overwritten.

partiql-planner/src/main/kotlin/org/partiql/planner/catalog/Routine.kt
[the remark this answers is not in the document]
RCHowell: I agree and can go back to `Function`. I was only avoiding it because most often _function_ is taken to mean _scalar function_.

But I do like,

- `aggregation function`
- `scalar function`
- `operator function`

Where an operator function is an _operator_ backed by a function a'la postgres.

partiql-planner/build.gradle.kts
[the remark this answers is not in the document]
RCHowell: Yes, this was my intention as it is being changes frequently.

PR #1518, opened 2024-07-18
[EXPERIMENTAL] Example of how we can source the dynamic dispatch logic to the planner

johnedquinn:
## Description
- DRAFT / EXPERIMENTAL. Not meant to merge.
- Example of how we can source the dynamic dispatch logic to the planner by exposing a FunctionResolver in the planner package (to reduce duplicated logic).

## Other Information

## License Information


johnedquinn:
Closing in favor of #1522. While this solves the problem, it has cons: exposing a public API in the planner to keep the typing logic intact. Rather, we should act like database implementers and write the logic in the evaluator as well.

PR #1519, opened 2024-07-18
identifier normalization

yliuuuu:
## Relevant Issues
- [Closes/Related To] Issue https://github.com/partiql/partiql-lang-kotlin/issues/1122

## Description
- This PR adds a planner flag that determines the case normalization behavior. 
    - UPPERCASE: normalize `Lvalue` and `Rvalue` by uppercasing the original text. 
    - LOWERCASE: normalize `Lvalue` and `Rvalue` by lowercasing the original text. 
    - EXACTCASE: normalize `Lvalue` and `Rvalue` by preserving the original text. 
    - If no flag is set: normalize `Lvalue` by preserving the original text, no Rvalue normalization. 

- To better separate `Lvalue` and `Rvalue`, this PR introduce a Binder Node in AST. 

- The Identifier/Binder normalization happens at the planner package, as the first step for AST normalization. 
    - Note that order matter for conducting AST normalization. 
    - Subsequence passes may generate binder, i.e., `AS` alias for from source. 
    - The binder generated should always be delimited. 
    - Example: 
```
UPPERCASE -> FROM bAr -> FROM "BAR" -> FROM "BAR" AS "BAR"
LOWERCASE -> FROM bAr -> FROM "bar" -> FROM "bar" AS "bar"
EXACTCASE -> FROM bAr -> FROM "bAr" -> FROM "bAr" AS "bAr"
NO FLAG -> FROM bAr -> FROM bAr -> FROM bAr AS "bAr"
```

## Other Information
  - No. Targeting v1

  -  No.

Yes. 

## License Information


partiql-planner/src/main/kotlin/org/partiql/planner/internal/astPasses/NormalizeIdentifier.kt
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Case Normalization behavior:
+ *   - UPPERCASE: normalize identifier using uppercase for both lvalue and rvalue
+ *   - LOWERCASE: normalize identifier using lowercase for both lvalue and rvalue
+ *   - EXACTCASE: normalize identifier by preserving case for both lvalue and rvalue
+ *   - NULL: Normalize lvalue by preserving case, no normalization for rvalue.
+ */
+internal class NormalizeIdentifier(
+    caseNormalization: CaseNormalization?
+) : AstPass {
```
RCHowell, line 20: - Merge this with IdentifierManager
- Use `init` to pick the normalization function ie `(string) -> string`


For example,

```kotlin
internal class NormalizeIdentifier(
    mode: CaseNormalization,
) {

   val normalize: (String) -> String

  init {
      normalize = when (mode) {
          UPPERCASE -> Strings.uppercase
          LOWERCASE -> Strings.lowercase
          EXACTCASE -> { it } 
      }
  }

}
```

Something like that so we aren't null checking and swithcing on mode every time.

RCHowell (changes requested):

partiql-planner/src/main/kotlin/org/partiql/planner/internal/astPasses/NormalizeIdentifier.kt
yliuuuu: I guess this suggestion is based off the assumption that we also case preserve (after normalization), which is indeed how this PR is implemented. 

But my thinking is to keep those two separate layer of indirection until at least the RFC is in draft.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/astPasses/Normalize.kt
RCHowell, line 1: nit, but please keep the same package name "normalize" rather than changing it. I'm ok with moving this to planner, in fact, I think that's a great idea!

partiql-planner/src/main/kotlin/org/partiql/planner/internal/IdentifierManager.kt
RCHowell, line 1: Please remove IdentifierManager per my previous comments. The identifier normalization pass should hold/take an enum for the normalization type — that is all. There is no need for the identifier manager.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/IdentifierManager.kt
```diff
@@ -0,0 +1,37 @@
+package org.partiql.planner.internal
+
+import org.partiql.ast.Binder
+import org.partiql.ast.Identifier
+import org.partiql.ast.binder
+import org.partiql.ast.identifierSymbol
+
+// The identifier manager class takes in two parameters,
+// in case we want to decouple case preservation and case normalization behavior in the future.
```
RCHowell, line 9: We have explicitly decided NOT to decouple these per our latest discussion. This is an internal class, so we can always change this later, but for now let's not create edge-cases that we've explicitly decided against.

RCHowell (changes requested):

yliuuuu:
Closing for now.

PR #1521, opened 2024-07-23
Merges v1-catalog into v1

RCHowell:
## Relevant Issues

#1497 
#1496 

## Description

This PR merges the catalog interfaces into the v1 branch.

## Other Information
- Updated Unreleased Section in CHANGELOG: **No**
- Any backward-incompatible changes? **No**
- Any new external dependencies? **No**
  and [Code Style Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CODE_STYLE.md)? **Yes**

## License Information

### DIFF
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/catalog/Catalog.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/catalog/Catalog.kt
@@ -1,0 +1,72 @@
+package org.partiql.planner.catalog
+
+/**
+ * Catalog interface for access to tables and routines.
+ *
+ * Related
+ *  - Iceberg — https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/catalog/Catalog.java
+ *  - Calcite — https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/schema/Schema.java
+ */
+public interface Catalog {
+
+    /**
+     * Returns the catalog name.
+     */
+    public fun getName(): String
+
+    /**
+     * Get a table by name.
+     */
+    public fun getTable(session: Session, name: Name): Table? = null
+
+    /**
+     * Given an [Identifier], returns a [Table.Handle] that corresponds to the longest-available requested path.
+     *
+     * For example, given a table named "Table" located within Catalog "AWS" and Namespace "a".b"."c", a user could
+     * call [getTableHandle] with the identifier "a"."b"."c"."Table". The returned [Table.Handle] will contain the table
+     * representation and the matching path: "a"."b"."c"."Table"
+     *
+     * As another example, consider a table within a [Namespace] that may be a struct with nested attributes.
+     * A user could call [getTableHandle] with the identifier "a"."b"."c"."Table"."x". In the Namespace, only table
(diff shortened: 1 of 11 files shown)


partiql-planner/src/main/kotlin/org/partiql/planner/catalog/Catalog.kt
```diff
@@ -0,0 +1,72 @@
...
+     * List all child namespaces from the namespace.
+     *
+     * @param namespace
+     */
+    public fun listNamespaces(session: Session, namespace: Namespace): Collection<Namespace> = emptyList()
+
+    /**
+     * Get a routine's variants by name; the default implementation is backed by the SQL-99 builtins.
+     */
+    public fun getFunctions(session: Session, name: Name): Collection<Function> {
```
johnedquinn, line 65: Not blocking -- but should this take in an `Identifier`? Similar to how `getTableHandle` also exposes a similar API.

partiql-planner/src/main/kotlin/org/partiql/planner/catalog/Function.kt
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * The function return type. Required.
+     */
+    public fun getReturnType(): PType.Kind
+
+    /**
+     * Compute a [PType] from the given arguments.
+     */
+    public fun computeReturnType(args: List<PType>): PType = SqlTypes.from(getReturnType())
```
johnedquinn, line 29: I've been thinking about this a bit, and I'm curious what you've been thinking about the following situation:
```
CREATE FUNCTION FOO(arg1 DECIMAL(15, 5)) RETURN DECIMAL(15, 5);
CREATE FUNCTION FOO(arg1 DECIMAL(10, 5)) RETURN DECIMAL(10, 5);
CREATE FUNCTION FOO(arg1 DECIMAL(5, 3)) RETURN DECIMAL(5, 3);

FOO(1); -- where the literal 1 is an INTEGER.
```

I imagine that, in the above function definitions, this would be represented by a single function signature (described with parameters of PType.Kind and a return type of PType.Kind). Now, my question is: when we have arguments that are coercible to a function's parameters' kinds, which specific variant of a kind will we choose? In this scenario, what would INT get coerced to?

johnedquinn (approved):
Everything looks good, but I do have 2 non-blocking questions.

partiql-planner/src/main/kotlin/org/partiql/planner/catalog/Function.kt
RCHowell: It would be coerced to the PType which is created by `SqlTypes.fromKind(kind)`.

For you specific example, it would be an arbitrary decimal.

`PType.fromKind(Kind.DECIMAL) -> PType.decimalArbitrary()`

johnedquinn: Interesting -- so coercions have some pre-determined mapping. So, in this model, now that we've coerced the literal integer to a decimal arbitrary, how do we know what the output type will be?
```
FOO(1::decimal_arbitrary) -> ?
```

I'm just mentally trying to understand how this works for coercions, and since I've recently done work on some of the arithmetic operators, I'm curious to see if this modeling is flexible enough to implement the operators with this logic.

PR #1522, opened 2024-07-24
Adds arithmetic operators as separate nodes & fixes casts

johnedquinn:
## Description
- Adds arithmetic operators as separate nodes. This allows for us to type the arithmetic operators how we please. I followed [SQL Server's rules for typing](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/arithmetic-operators-transact-sql?view=sql-server-ver16).
- Fixes casts to invalid types for conformance. The conformance tests allow them to pass -- so I made them warnings instead of errors.
- Fixes function resolution (static) for dynamic arguments.
  - Previously, `1 > some_dynamic_value` would result in `Fn.Gt(INT, INT) -> BOOL`. While this had the most number of static exact matches, it didn't necessarily have the most number of runtime exact matches -- causing behavior that would break the gradual guarantee. Fixed this by checking whether any argument is dynamic.
- Simplifies dynamic dispatch (runtime) by following the same rules as function resolution.

## Other Information
  - Added nodes specifically for the arithmetic operators.

## License Information


johnedquinn:
Closing in favor of #1533.

PR #1523, opened 2024-07-24
Fixes function invocation for missing arguments

johnedquinn:
## Description
- Fixes function invocation for missing arguments and fixes evaluation of `Rex.Op.Missing`.

## Other Information

## License Information

### DIFF
--- a/partiql-eval/src/main/java/org/partiql/eval/value/Datum.java
+++ b/partiql-eval/src/main/java/org/partiql/eval/value/Datum.java
@@ -345,6 +345,9 @@ public interface Datum extends Iterable<Datum> {
     @Deprecated
     default PartiQLValue toPartiQLValue() {
         PType type = this.getType();
+        if (this.isMissing()) {
+            return PartiQL.missingValue();
+        }
         switch (type.getKind()) {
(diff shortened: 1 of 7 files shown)


yliuuuu (approved):
I understand this PR is for evaluator only. 

Curious about your thoughts on whether we should also fix the planer. i.e, The Planner is operating under signal mode, will turn 1 + MISSING in to an rexOpErr.

johnedquinn:
> I understand this PR is for evaluator only.
> 
> Curious about your thoughts on whether we should also fix the planer. i.e, The Planner is operating under signal mode, will turn 1 + MISSING in to an rexOpErr.

Just checked to be sure, always missing arguments currently returns a `Rex.Op.Missing`.

yliuuuu:
The conversion happens when we convert the internal plan to public plan. 

i.e., 
https://github.com/partiql/partiql-lang-kotlin/blob/b5c596cd39ab59573bb4ec9959a3a37f57dba615/partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt#L259-L272

PR #1525, opened 2024-07-25
V1 roll back ddl

yliuuuu:
## Relevant Issues

## Description
- Roll back new CREATE Syntax Parsing support. 
- The plan is to add those back on either just before or right after v1 launch. 
- The goal is to use this as an opportunity to check if we can add a major feature addition without releasing a new major version(breaking customer). 
- The new DDL modeling lives under feature branch [v1-ddl](https://github.com/partiql/partiql-lang-kotlin/tree/v1-ddl). 
- The legacy DDL AST and parser implementation will be deprecated and removed in pre-v1 release, and will be deleted in V1 release. 

## Other Information
  - No. 

  -  No. 

No.

Yes. 

## License Information


alancai98 (commented):
Revert looks good. Was not able to add a comment directly to `test/partiql-tests` but I think it should still point to `v1`'s current commit, https://github.com/partiql/partiql-tests/tree/26ab08ec49b889550ecca4878e2a338a712dcf35.

Also in a comment or PR description, could be helpful to add:
- the branch where the new DDL changes will live before we merge it back to `v1`
- information that the legacy DDL modeling in the AST and parser will be deprecated in release pre-`v1` and deleted in `v1`
That way when we come back to merging the DDL changes back to `v1`, we're clear on the process we'll take.

alancai98 (approved):

PR #1527, opened 2024-07-26
Add more cast tests

am357:
## Description
```SQL
CREATE TABLE PETS (
    id INT4,
    breed STRING,
    tag INT2
)
```

Tests:
```SQL
-- #1 should succeed
CAST(true AS INT2) --

-- #2  should fail
CAST((s : STRING) AS LIST)

-- #3 should succeed
CAST((i : INT2) AS INT4)

-- #4 should succeed
CAST((i : INT4) AS INT2)
```

#### Background context
As part of a specification work, we want to make the behavior for the #4 more visible:

 From SQL-99 Section 6.22: <cast specification>:

```
 If SD (source data type) is exact numeric or approximate numeric, then
 Case:
   i) If there is a representation of SV (source value) in the data type TD (target data type) that does not lose any leading si-gnificant digits after rounding or truncating if necessary, then TV is that representation.
 The choice of whether to round or truncate is implementation-defined.
 ii) Otherwise, an exception condition is raised: data exception — numeric value out of
 range.
 ```

We can see this evaluation behavior in PostgresSQL v15:
Ref.: https://www.db-fiddle.com/f/vMY3Srd2bG6f3NshvTp3w1/0

```
-- succeeds
SELECT CAST((1::INT4) AS INT2);

-- fails: `Query Error: error: smallint out of range` SELECT CAST((327686::INT4) AS INT2);
```

## Other Information
  - No, adding tests only.

  - No.

  - No.

  Yes.

## License Information

### DIFF
--- a/partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt
+++ b/partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt
@@ -39,6 +39,7 @@ import org.partiql.types.StaticType
 import org.partiql.types.StaticType.Companion.ANY
 import org.partiql.types.StaticType.Companion.DECIMAL
 import org.partiql.types.StaticType.Companion.INT
+import org.partiql.types.StaticType.Companion.INT2
 import org.partiql.types.StaticType.Companion.INT4
 import org.partiql.types.StaticType.Companion.INT8
 import org.partiql.types.StaticType.Companion.unionOf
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt [hunk 1]
```diff
@@ -289,6 +293,37 @@ class PlanTyperTestsPorted {
...
+            ),
+            SuccessTestCase(
+                name = "CAST INT8 literal to INT2",
+                query = "CAST(343434534534 AS INT2)",
+                expected = INT2
+            ),
+            ErrorTestCase(
+                name = "CAST STRING to INT4",
+                query = "CAST((SELECT t.breed FROM ddb.pets AS t) AS INT2)",
+            ),
```
RCHowell, line 324: `CAST(<string> TO <int>)` should not be a type-check error, but it _may_ be a runtime error.

- I checked postgres and they allow `CAST('1' AS INT)`
- Existing [PLK behavior allows this](https://github.com/partiql/partiql-lang-kotlin/wiki/Functions#cast----since-v010) and per recent discussions we should do our best to maintain current functional behavior.


Additionally, the test name has `INT4` whereas the query uses `INT2`.

RCHowell (changes requested):

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt
am357: Makes sense, will fix.

RCHowell (approved):

PR #1532, opened 2024-07-30
Initializes plan interfaces for expressions and relations

RCHowell:
## Description

This PR begins defining the plan interfaces which we will use for defining a stable and evolve-able API for the logical algebra. This is a deliberate departure from generated IRs which we are not able to evolve without API breaking changes. Notably we avoid sealed classes/interfaces and only use java-style get/set methods.

I am open to using the kotlin interface values (aka properties) as these become Java-style get methods anyways. It would just be kotlin syntax sugar.

## Other Information

## License Information


partiql-plan/src/main/kotlin/org/partiql/plan/v1/rex/Rex.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import org.partiql.plan.v1.Node
+import org.partiql.types.PType
+
+/**
+ * TODO DOCUMENTATION
+ */
+public interface Rex : Node {
+
+    public fun getType(): PType
```
alancai98, line 11: nice, I like that the type is on the same level as the `Rex` implementations. Was pretty annoying with the current plan APIs when the `Rex.Op`s couldn't directly reference their `type` field.

partiql-plan/src/main/kotlin/org/partiql/plan/v1/Statement.kt
```diff
@@ -0,0 +1,20 @@
+package org.partiql.plan.v1
+
+import org.partiql.plan.v1.rex.Rex
+
+/**
+ * TODO DOCUMENTATION
+ */
+public interface Statement {
```
alancai98, line 8: should `Statement` also implement `Node`?

partiql-plan/src/main/kotlin/org/partiql/plan/v1/rex/RexCast.kt
```diff
@@ -0,0 +1,13 @@
+package org.partiql.plan.v1.rex
+
+import org.partiql.types.PType
+
+/**
```
alancai98, line 5: What's the sequencing for the PRs? Saw that this was marked as 1/2 in the title. Will `Rel`s be in a different PR? Similarly for the visitors + builders?

partiql-plan/src/main/kotlin/org/partiql/plan/v1/rex/Rex.kt
RCHowell: This is one of the primary goals, on top of being able to have internal implementations of these operators for the typers without having to transform across domains. I am making an effort to flatten things for ease-of-use.

partiql-plan/src/main/kotlin/org/partiql/plan/v1/rex/RexCast.kt
RCHowell: Correct, rels will be another **commit**.  Visitors/builders will be another **pull request**.

partiql-plan/src/main/kotlin/org/partiql/plan/v1/rel/RelExcludeStep.kt
```diff
@@ -0,0 +1,40 @@
+package org.partiql.plan.v1.rel
+
+/**
+ * TODO DOCUMENTATION
+ */
+public interface RelExcludeStep : Rel {
```
alancai98, line 6: don't think the `ExcludeStep` needs to implement `Rel`

partiql-plan/src/main/kotlin/org/partiql/plan/v1/rel/RelExcludeStep.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    public fun getSteps(): List<RelExcludeStep>
```
alancai98, line 8: nit: could be clearer to use `getSubsteps`

```suggestion
    public fun getSubsteps(): List<RelExcludeStep>
```

partiql-plan/src/main/kotlin/org/partiql/plan/v1/rel/Rel.kt
```diff
@@ -0,0 +1,15 @@
+package org.partiql.plan.v1.rel
+
+import org.partiql.plan.v1.Node
+import org.partiql.plan.v1.Schema
+
+public interface Rel : Node {
+
+    public fun getInputs(): List<Rel>
```
alancai98, line 8: So most of the `Rel`s have a `getInput(): Rel` function. Should they actually just use this `getInputs`s function?

partiql-plan/src/main/kotlin/org/partiql/plan/v1/rel/RelUnpivot.kt
```diff
@@ -0,0 +1,13 @@
+package org.partiql.plan.v1.rel
+
+import org.partiql.plan.v1.rex.Rex
+
+/**
+ * TODO DOCUMENTATION
+ */
+public interface RelUnpivot : Rel {
+
+    public fun getInput(): Rex
```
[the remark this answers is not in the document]
RCHowell: Unpivot does have an input, and the return type is rex.

```
FROM UNPIVOT T
```

Is

```
RelUnpivot(input = RexVar(T))
```

partiql-plan/src/main/kotlin/org/partiql/plan/v1/rex/RexCall.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Scalar function calls.
+ */
+public interface RexCall : Rex {
+
+    /**
+     * Returns the function to invoke.
+     *
+     * @return
+     */
+    public fun getFunction(): String
```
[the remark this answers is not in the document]
RCHowell: That is correct. The given methods are the **required** ones which we do not change.  Yes we will set defaults for **additional** methods we may introduce. Additionally I am working on abstract bases for extension.

partiql-plan/src/main/kotlin/org/partiql/plan/v1/rel/RelScan.kt
```diff
@@ -0,0 +1,11 @@
+package org.partiql.plan.v1.rel
+
+/**
+ * TODO DOCUMENTATION
+ */
+public interface RelScan : Rel {
+
+    public fun getInput(): Rel
```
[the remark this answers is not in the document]
RCHowell: Yes this should be a rex, nice catch

partiql-plan/src/main/kotlin/org/partiql/plan/v1/rel/RelExcludeStep.kt
RCHowell: Correct, bad paste

partiql-plan/src/main/kotlin/org/partiql/plan/v1/rel/RelExcludeStep.kt
RCHowell: Ack

partiql-plan/src/main/kotlin/org/partiql/plan/v1/rel/Rel.kt
RCHowell: Some of them have getInput() when it's a single input, and some have getLeft() and getRight() if it's a multi-set operator. This is a top-level getInputs() for tree-traversal

partiql-plan/src/main/kotlin/org/partiql/plan/v1/Statement.kt
```diff
@@ -0,0 +1,20 @@
...
+
+    /**
+     * PartiQL Query Statement — i.e. SELECT-FROM
+     */
+    public interface Query {
```
johnedquinn, line 13:
```suggestion
    public interface Query : Statement {
```

partiql-plan/src/main/kotlin/org/partiql/plan/v1/rel/RelUnion.kt
```diff
@@ -0,0 +1,67 @@
+package org.partiql.plan.v1.rel
+
+/**
+ * Logical `UNION [ALL|DISTINCT]` operator for set (or multiset) union.
+ */
+public interface RelUnion : Rel {
+
+    public fun isAll(): Boolean
```
alancai98, line 8: nit: wonder if it would be preferred to keep the set quantifier naming consistent. In `RelUnion`, `RelExcept` and `RelIntersect`, we use `isAll` while in `RelAggregateCall` we use `isDistinct`.

alancai98 (approved):

partiql-plan/src/main/kotlin/org/partiql/plan/v1/rel/RelUnion.kt
RCHowell: This has to do with which is the "default".

For union/intersect/except, DISTINCT is default — so "isAll()" is like asking "is this different than the default"?

For aggregate calls, ALL is default — so "isDistinct()" is like asking "is this different than the default"?

PR #1533, opened 2024-07-31
Optimizes dynamic dispatch & casts. Fixes arithmetic operations, casts, and more.

johnedquinn:
## Preface
- This PR improves conformance by 2.66%. We now sit at over 97% conformance (legacy is at 92.49%). See the [latest report](https://github.com/partiql/partiql-lang-kotlin/actions/runs/10202549600).

## Description
- The biggest change in this PR is that we are fixing dynamic dispatch (while also maintaining high performance). Before, when there were dynamic arguments, we sometimes didn't invoke dynamic dispatch. It depended on whether there was a single function that had the highest number of exact matches at compile time. This was wrong. Please see `FnResolver` to see how we pass along the invocable matches when we have any dynamic argument (so we can see which has the most exact matches at runtime). Also, see `ExprCallDynamic` -- it is now simplified to run through the sorted candidates. See the KDocs there.
- I've also simplified and optimized `ExprCast`. It now internally uses a `CastTable` which is a two-dimensional array (indexed by the `PType.Kind` ordinals) to receive a cast function. This speeds up the casts, and it's also much easier to read. It retains the same functionality as before while simplifying it and speeding it up.
- To make these changes, I've been using the profiler to keep track of the performance changes. This PR tackles several performance problems (casts, dynamic dispatch, and unnecessary conversions to/from PartiQLValue).

## What's Next?
- I'm going to work on converting the `Fn`s to use `Datum` instead. With this, our performance should be substantially better.

## Other Information

## License Information


partiql-eval/src/main/java/org/partiql/eval/value/Datum.java
```diff
@@ -547,6 +557,33 @@ static Datum int32Value(int value) {
         return new DatumInt(value);
     }
 
+    @Deprecated
+    @NotNull
+    static Datum intArbitrary(@NotNull BigInteger value) {
+        return new DatumBigInteger(value);
+    }
```
RCHowell, line 564: This is mixing C-style names with SQL names. I say go for all SQL names for symmetry with PType. So update update int64Value to bigint.

Or visa-versa! No strong preference, just that right now they're mixed

partiql-eval/src/main/java/org/partiql/eval/value/DatumCollection.java
```diff
@@ -36,4 +36,34 @@ public Iterator<Datum> iterator() {
...
+        }
+        if (_type.getKind() == PType.Kind.LIST) {
+            sb.append("]");
+        } else if (_type.getKind() == PType.Kind.BAG) {
+            sb.append(">>");
+        } else if (_type.getKind() == PType.Kind.SEXP) {
+            sb.append(")");
+        }
+        return sb.toString();
+    }
```
RCHowell, line 68: Interesting. Perhaps add a comment about being for debug?

RCHowell: This is outside the scope of the PR, but I would suggest using the Ion (and PartiQLValue) reader/writer style and have toString() just invoke a text writer btw.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
```diff
@@ -252,13 +251,13 @@ internal class Compiler(
                     error("Dynamic call candidate had different name than others; found $fnName but expected $name")
                 }
             }
-            ExprCallDynamic.Candidate(fn, coercions)
+            ExprCallDynamic.Candidate(fn)
```
RCHowell, line 254: Nice!

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/CastTable.kt
```diff
@@ -0,0 +1,701 @@
...
+        val cast = castTable[source.type.kind.ordinal][target.kind.ordinal]
+            ?: throw TypeCheckException("CAST(${source.type} AS $target) is not supported.")
+        return try {
+            cast.invoke(source, target)
+        } catch (t: Throwable) {
+            throw TypeCheckException("Failed to cast $source to $target")
+        }
+    }
+
+    private val castTable: Array<Array<((Datum, PType) -> Datum)?>> = Array(PType.Kind.values().size) { inputKind ->
```
RCHowell, line 36: nit: introduce two type aliases

1. alias for the cast function signature
2. alias for the 2d array / table

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/CastTable.kt
```diff
@@ -0,0 +1,701 @@
...
+     * CAST(<tinyint> AS <type>)
+     * TODO: CHAR, VARCHAR, SYMBOL
+     */
+    private fun registerTinyInt() {
+        register(PType.Kind.TINYINT, PType.Kind.BOOL) { x, _ -> Datum.boolValue(x.byte.toInt() != 0) }
+        register(PType.Kind.TINYINT, PType.Kind.TINYINT) { x, _ -> x }
+        register(PType.Kind.TINYINT, PType.Kind.SMALLINT) { x, _ -> Datum.smallInt(x.byte.toShort()) }
+        register(PType.Kind.TINYINT, PType.Kind.INT) { x, _ -> Datum.int32Value(x.byte.toInt()) }
+        register(PType.Kind.TINYINT, PType.Kind.BIGINT) { x, _ -> Datum.int64Value(x.byte.toLong()) }
+        register(PType.Kind.TINYINT, PType.Kind.INT_ARBITRARY) { x, _ ->
```
RCHowell, line 102: nit: clean up by importing all of the PType.Kind to reduce a lot of the noise here

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/CastTable.kt
```diff
@@ -0,0 +1,701 @@
...
+import org.partiql.errors.DataException
+import org.partiql.errors.TypeCheckException
+import org.partiql.eval.value.Datum
+import org.partiql.types.PType
+import java.math.BigDecimal
+import java.math.BigInteger
+import java.math.RoundingMode
+
+@Suppress("DEPRECATION")
+internal object CastTable {
```
RCHowell, line 15: **Could you add a toString() for generating documentation?**

This is great. Nice consolidation of the logic and putting it in a fast data structure.

_Outside the scope..._

I wonder if there are an opportunities to make this part of the datum package in some capacity (or in SPI).

What I'm getting at is for the planner and evaluator to use the same source of truth for casts, and seeing what abstraction we would need.

partiql-spi/src/main/kotlin/org/partiql/spi/fn/builtins/FnIsTuple.kt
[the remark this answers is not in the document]
RCHowell: You should change the RexConverter to have `IS TUPLE` lower to `is_struct` since these are the same function.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -769,7 +769,7 @@ internal class PlanTyper(private val env: Env) {
 
         override fun visitRexOpCastUnresolved(node: Rex.Op.Cast.Unresolved, ctx: CompilerType?): Rex {
             val arg = visitRex(node.arg, null)
-            val cast = env.resolveCast(arg, node.target) ?: return ProblemGenerator.errorRex(
+            val cast = env.resolveCast(arg, node.target) ?: return ProblemGenerator.missingRex(
```
RCHowell, line 772: Why MISSING if we _know_ the cast will never succeed?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt
```diff
@@ -138,11 +138,18 @@ internal class PlanTransform(
...
+            return when (signalMode) {
+                true -> {
+                    onProblem.invoke(problem)
+                    rexOpErr(problem.toString(), emptyList())
+                }
+                false -> {
+                    onProblem.invoke(ProblemGenerator.asWarning(problem))
+                    org.partiql.plan.rexOpMissing(problem.toString(), emptyList())
+                }
+            }
```
RCHowell, line 152: Maybe we can add a note/reference to spec as to why we don't fail here

partiql-planner/src/main/kotlin/org/partiql/planner/internal/PlanningProblemDetails.kt
```diff
@@ -166,6 +166,13 @@ internal open class PlanningProblemDetails(
         "Unknown function `$identifier($types)"
     })
 
+    data class UnknownCast(
```
RCHowell, line 169:
```suggestion
    public data class UnknownCast(
```

partiql-planner/src/main/kotlin/org/partiql/planner/internal/FnResolver.kt
```diff
@@ -46,34 +46,33 @@ internal object FnResolver {
...
-                }
-            }
-        }
-        if (isDynamic) {
-            return FnMatch.Dynamic(matches.map { it.match })
-        }
+        // 5. Run through all candidates and keep those that accept preferred types (of the input data type's type category) at the most positions where type conversion will be required.
 
         // 6. Find the highest precedence one. NOTE: This is a remnant of the previous implementation. Whether we want
         //  to keep this is up to us.
```
RCHowell, line 78: I'm sure we could factor this a dozen different ways.. I'm curious if you gave a one-loop solution a go?

Something along the lines of `var match: FnMatch? = null` and one-loop + break? I'm sure this isn't a priority, but I'm just curious.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
```diff
@@ -26,22 +36,114 @@ internal class ExprCallDynamic(
...
+    }
+
+    /**
+     * This represents SQL:1999 Section 4.1.2 "Type conversions and mixing of data types" and breaks down the different
+     * coercion groups.
+     *
+     * TODO: [UNKNOWN] should likely be removed in the future. However, it is needed due to literal nulls and missings.
+     * TODO: [DYNAMIC] should likely be removed in the future. This is currently only kept to map function signatures.
+     */
+    enum class CoercionFamily {
```
RCHowell, line 92:
```suggestion
    private enum class CoercionFamily {
```

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
```diff
@@ -26,22 +36,114 @@ internal class ExprCallDynamic(
...
+        STRUCTURE,
+        DATE,
+        TIME,
+        TIMESTAMP,
+        COLLECTION,
+        UNKNOWN,
+        DYNAMIC
+    }
+
+    companion object {
```
RCHowell, line 106:
```suggestion
    private companion object {
```

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
```diff
@@ -54,177 +156,29 @@ internal class ExprCallDynamic(
      */
     data class Candidate(
```
RCHowell, line 157:
```suggestion
    private data class Candidate(
```

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
```diff
@@ -4,20 +4,30 @@ import org.partiql.errors.TypeCheckException
...
+ * @property paramIndices the last index of the [args]
+ * @property paramTypes represents a two dimensional array, where the first dimension maps to the [candidates], and the
+ * second dimensions maps to the [org.partiql.spi.fn.FnSignature.parameters]'s [PType].
+ * @property paramFamilies represents a two dimensional array, where the first dimension maps to the [candidates], and the
+ * second dimensions maps to the [org.partiql.spi.fn.FnSignature.parameters]'s [CoercionFamily]. This allows the algorithm
+ * to know if an argument is coercible to the target parameter.
+ * @property cachedMatches a memoization cache for the [match] function.
  */
 @OptIn(PartiQLValueExperimental::class)
 internal class ExprCallDynamic(
```
RCHowell, line 33: In regards to all the private comments (knowing that this is internal).

Generally if expression operators need to share logic, then we can internalize that somewhere common rather than nested within a class.

However, I can see from partiql-eval.api that it doesn't matter — so the private comments are somewhat pointless.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
```diff
@@ -26,22 +36,114 @@ internal class ExprCallDynamic(
     private val args: Array<Operator.Expr>
 ) : Operator.Expr {
 
-    private val candidateIndex = CandidateIndex.All(candidates)
+    private val paramIndices: IntRange = args.indices
+    private val paramTypes: List<List<PType>> = this.candidates.map { candidate -> candidate.fn.signature.parameters.map { it.type } }
+    private val paramFamilies: List<List<CoercionFamily>> = this.candidates.map { candidate -> candidate.fn.signature.parameters.map { family(it.type.kind) } }
+    private val cachedMatches: MutableMap<List<PType>, Int> = mutableMapOf()
```
RCHowell, line 42: Interesting that this is a `MutableMap<List<PType>>` rather than something like `MutableMap<List<PType.Kind>>`.

RCHowell (changes requested):

partiql-eval/src/main/java/org/partiql/eval/value/Datum.java
johnedquinn: Yeah, I was tempted to change all of them to SQL as I ran through this PR, though I held back due to the diff. The intention is to use the SQL names as it's the public API. If you don't mind, my follow-up PRs aim to clean up some of our APIs.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
johnedquinn: There are many conformance tests that follow this behavior. Though, in this PR I leave it up to the signaling mode.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/PlanningProblemDetails.kt
johnedquinn: This is embedded within the internal `PlanningProblemDetails`.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/FnResolver.kt
johnedquinn: For now, this work is focused on making sure everything is right and setting up the baseline for tests. In the future, we can make these small optimizations.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
johnedquinn: It has to do with our function resolution. We follow PostgreSQL for exact matches -- which take into account parameters. So, in one dispatch (say argument of type `DECIMAL(2,0)`) could be matched to one candidate, while another dispatch (say argument `DECIMAL(3,0)`) could be matched to another candidate.

If we change our modelling of function signatures, then this can change as well.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/Compiler.kt
```diff
@@ -251,7 +251,7 @@ internal class Compiler(
                     error("Dynamic call candidate had different name than others; found $fnName but expected $name")
                 }
             }
-            ExprCallDynamic.Candidate(fn)
+            fn
```
johnedquinn, line 254: This is due to making `ExprCallDynamic.Candidate` private.

PR #1534, opened 2024-08-02
Fix antlr dep partiql-lang

alancai98:
## Relevant Issues
- None

## Description
- Tries fixing the `partiql-lang` dependency on `partiql-parser`
- Since the ANTLR shading change (https://github.com/partiql/partiql-lang-kotlin/pull/1439), any call to the `PartiQLParserBuilder.standard()` from the Maven release of 0.14.6 would result in a `NoSuchMethodError` when attempting to parse. This was due to a `partiql-lang` using its own ANTLR dependency rather than `partiql-parser`'s shaded ANTLR dependency.
  - It was not detected sooner in the normal gradle build because we would previously only generate the shaded ANTLR dependency when publishing the jars to Maven.
  - We should have been using this [configuration](https://gradleup.com/shadow/multi-project/#depending-on-the-shadow-jar-from-another-project) provided by the shadow plugin for multi-project builds.
- Simple reproduction case for previous failure w/ v0.14.6
```
// in build.gradle.kts
dependencies {
    implementation("org.partiql:partiql-lang-kotlin:0.14.6")
}
```

```
// in Main.kt
import org.partiql.lang.syntax.PartiQLParserBuilder

fun main(args: Array<String>) {
    val input = "SELECT 1 FROM <<>>"
    val parser = PartiQLParserBuilder.standard().build()
    println(parser.parseAstStatement(input))
}

```
above results in the following error:
```
Exception in thread "main" org.partiql.lang.syntax.ParserException: Unhandled exception.
	Evaluator Error: at line <UNKNOWN>, column <UNKNOWN>: internal error, <UNKNOWN> : <UNKNOWN>

	at org.partiql.lang.syntax.impl.PartiQLPigParser.parseAstStatement(PartiQLPigParser.kt:74)
	at MainKt.main(Main.kt:6)
Caused by: java.lang.NoSuchMethodError: 'void org.partiql.parser.antlr.PartiQLTokens.<init>(org.partiql.lang.thirdparty.antlr.v4.runtime.CharStream)'
	at org.partiql.lang.syntax.impl.PartiQLPigParser.createTokenStream$partiql_lang(PartiQLPigParser.kt:111)
	at org.partiql.lang.syntax.impl.PartiQLPigParser.parseQuery$partiql_lang(PartiQLPigParser.kt:95)
	at org.partiql.lang.syntax.impl.PartiQLPigParser.parseQuery(PartiQLPigParser.kt:85)
	at org.partiql.lang.syntax.impl.PartiQLPigParser.parseAstStatement(PartiQLPigParser.kt:58)
	... 1 more
```

Note that when using the `PartiQLParserBuilder.experimental...` rather than `PartiQLParserBuilder.standard...`, the query parses without errors.

## Other Information

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Thank you to all who have contributed!
 -->
 
+## [0.14.7]
+
+### Fixed
+- `partiql-lang`'s `PartiQLParserBuilder.standard()` will use the ANTLR dependency from `partiql-parser` to
+prevent `NoSuchMethodError`s
+
 ## [0.14.6]
 
 ### Added
@@ -1087,7 +1093,8 @@ breaking changes if migrating from v0.9.2. The breaking changes accidentally int
 ### Added
 Initial alpha release of PartiQL.
 
-[Unreleased]: https://github.com/partiql/partiql-lang-kotlin/compare/v0.14.6...HEAD
+[Unreleased]: https://github.com/partiql/partiql-lang-kotlin/compare/v0.14.7...HEAD
+[0.14.7]: https://github.com/partiql/partiql-lang-kotlin/compare/v0.14.6...v0.14.7
 [0.14.6]: https://github.com/partiql/partiql-lang-kotlin/compare/v0.14.5...v0.14.6
 [0.14.5]: https://github.com/partiql/partiql-lang-kotlin/compare/v0.14.4...v0.14.5
 [0.14.4]: https://github.com/partiql/partiql-lang-kotlin/compare/v0.14.3...v0.14.4
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 group=org.partiql
-version=0.14.6
+version=0.14.7
 
 ossrhUsername=EMPTY
 ossrhPassword=EMPTY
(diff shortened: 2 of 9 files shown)


alancai98:
Local testing within partiql-lang-kotlin
- gradle clean and build `partiql-lang-kotlin` -- all previous tests passed
- publishToMavenLocal

In a sample project
- Created a simple gradle project w/ `org.partiql:partiql-lang-kotlin:0.14.7` dependency
```
// In build.gradle.kts
repositories {
    mavenCentral()
    mavenLocal()
}

dependencies {
    implementation("org.partiql:partiql-lang-kotlin:0.14.7")
}
```
and in
```
// Main.kt
import org.partiql.lang.syntax.PartiQLParserBuilder

fun main(args: Array<String>) {
    val input = "SELECT 1 FROM <<>>"
    val parser = PartiQLParserBuilder.standard().build()
    println(parser.parseAstStatement(input))
}
```
Verified that the reproduction case from PR description now works without an error.

Other validation
- Inspected the locally-published `partiql-lang-kotlin-0.14.7.jar` and checked that the ANTLR dependencies used in `partiql-lang` come from `partiql-parser`'s shadowed ANTLR dependency under `org.partiql.parser.thirdparty.antlr`

johnedquinn (approved):
Tested locally. Thanks for tackling this issue!

PR #1535, opened 2024-08-02
Initializes logical plan builders

RCHowell:
## Description

This PR adds dataframe-like fluent builders to the logical plans. We have recurring customer asks of having calcite style algebra builders, and [we have had _some_ offering like this with the generated IRs](https://github.com/partiql/partiql-lang-kotlin/tree/main/partiql-ast), but these were more like convenience methods than actual dataflow builders.

For example, the builders were just POJO-style builders, and the DSL is just kotlin stuff for calling methods. Whereas these are consistent with a dataflow builder.

## Example

**Before**
```kotlin
// SELECT x FROM T

val query = rex(
    type = ..,
    op = RexOpSelect(
        constructor = rex(
            type = ...,
            op = rexOpStruct(fields = listOf(rex(type=..., op=rexopVar)
       input = rel(
           type = ...,
           op = relOpProject(
                projections = ...
               ..
)
```

This is _really_ bad and not user friendly. The highly-detailed and typed resolved plans enabled us to deliver statically resolved and typed plans, but were not conducive to ease-of-use.

This plan abstract is a work-in-progress to regain the simplicity of AST-like plan builders (based upon dataframe style construction) with the finer details hidden behind small interfaces.

**After**
```kotlin
val query = Rel.scan("T").select("x").build()

// all logic for planning/resolution is hidden behind build()
```

## Other Information

## License Information

### DIFF
--- a/partiql-plan/src/main/kotlin/org/partiql/plan/v1/rex/RexSelect.kt
+++ b/partiql-plan/src/main/kotlin/org/partiql/plan/v1/rex/RexSelect.kt
@@ -1,6 +1,7 @@
 package org.partiql.plan.v1.rex
 
 import org.partiql.plan.v1.rel.Rel
+import org.partiql.types.PType
 
 /**
  * TODO DOCUMENTATION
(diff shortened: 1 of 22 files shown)


partiql-plan/src/main/kotlin/org/partiql/plan/v1/rex/RexSelect.kt
```diff
@@ -11,6 +12,10 @@ public interface RexSelect : Rex {
 
     public fun getConstructor(): Rex
 
+    public override fun getType(): PType = PType.typeBag(getConstructor().getType())
```
johnedquinn, line 15: I'm sure you're planning on fixing everything soon -- but pointing out ordering here

johnedquinn (approved):

PR #1537, opened 2024-08-05
Updates functions and aggregation calls to use Datum

johnedquinn:
## Description
- Updates functions and aggregation calls to use Datum
- Renames the Datum static functions to use the SQL names (AKA `bigint` instead of `int64`)
- Looking at the conformance test, we pass 2 additional tests and do not fail any others.

## Other Information
- Any backward-incompatible changes? **YES**: Function signatures, invocations, and aggregations.

## License Information


partiql-spi/src/main/java/org/partiql/eval/value/Datum.java
```diff
@@ -548,13 +549,13 @@ static Datum smallInt(short value) {
     }
 
     @NotNull
-    static Datum int64Value(long value) {
-        return new DatumLong(value);
+    static Datum integer(int value) {
```
johnedquinn, line 552: `int` is a keyword in Java, hence the use of `integer`

johnedquinn:
Rebased off of V1

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprMissing.kt
```diff
@@ -10,6 +10,6 @@ internal class ExprMissing(
 ) : Operator.Expr {
 
     override fun eval(env: Environment): Datum {
-        return Datum.missingValue(type)
+        return Datum.missing(type)
```
RCHowell, line 13: missing takes a type? interesting..

partiql-spi/src/main/kotlin/org/partiql/spi/fn/builtins/AggAvg.kt
```diff
@@ -7,158 +7,139 @@ import org.partiql.spi.fn.Agg
...
-
-@OptIn(PartiQLValueExperimental::class)
+import org.partiql.types.PType
+
 internal object Agg_AVG__INT8__INT8 : Agg {
 
     override val signature: AggSignature = AggSignature(
         name = "avg",
-        returns = DECIMAL,
+        returns = PType.typeDecimalArbitrary(),
```
RCHowell, line 16: It may be worth adding a comment

> If AVG is specified and DT is exact numeric, then the declared type of the result is exact numeric with implementation-defined precision not less than the precision of DT and implementation-defined scale not less than the scale of DT. SQL-99 p.156

RCHowell (approved):
Approving with a fast-follow for PType cleanup I will PR

Issue #1507, opened 2024-07-11
[main/v1] partiql-ast: modeling of `s_f_w.set_op` does not allow for left associative set operations


alancai98:
## Description
- Currently in the `main` and `v1` branches, the ast's modeling of `set_op.s_f_w` does not allow for defining left associative set operations (UNION, EXCEPT, INTERSECT).

Current modeling on `main`/`v1`: https://github.com/partiql/partiql-lang-kotlin/blob/edef7fb8048a07fc3fef913ccb812d49434bf4c1/partiql-ast/src/main/resources/partiql_ast.ion#L494-L510

Assuming SFW_1, SFW_2, SFW_3 represent 3 different SFW queries. Then the following cannot be modeled in the current AST using `set_op.s_f_w`:

```
(SFW_1 EXCEPT SFW_2) EXCEPT SFW_3
```

Set operations are by default left-associative, so `SFW_1 <set_op> SFW_2 <set_op> SFW_3` is also not representable.

## To Reproduce
1. Attempt to create the above query using the AST

## Expected Behavior
- Allow users + library owner to create a left associative set operation

## Additional Context
- Java version: 17
- PartiQL version: 0.14.5 and later + `v1` branch
- Add any other context about the problem here.

alancai98:
From prior discussion, the modeling of SQL set ops needs to also support `LIMIT`, `OFFSET`, and `ORDER BY`, which is not currently supported in the parser.

For example the following does not currently parse:
```
PartiQL> (SELECT a FROM <<{'a': 1}>> LIMIT 1 UNION SELECT a FROM <<{'a': 2}>> LIMIT 1) LIMIT 1
   | !!
Parser Error: at line 1, column 79: unexpected token found, LIMIT : LIMIT
mismatched input 'LIMIT' expecting {<EOF>, ';'}
```

In the AST, we can consider a modeling more similar to what partiql-lang-rust uses:
https://github.com/partiql/partiql-lang-rust/blob/c91c5a4727f65523dea55fbdd9b9530d1df19d63/partiql-ast/src/ast.rs#L270-L278.

Alternatively, we can look at what another SQL parsing crate uses: https://docs.rs/sqlparser/0.48.0/sqlparser/ast/enum.SetExpr.html.

alancai98:
Closed by https://github.com/partiql/partiql-lang-kotlin/pull/1538.

PR #1538, opened 2024-08-05
[v1] Refactor representation of set ops

alancai98:
## Relevant Issues
- Resolves https://github.com/partiql/partiql-lang-kotlin/issues/1507

## Description
- Refactors modeling of SQL set ops in parser, ast, plan, and evaluator
- In the AST, new modeling supports
  - `ORDER BY`, `LIMIT`, `OFFSET` on the SQL set op and the SFW arguments
  - left-associative SQL set ops
- In the parser, 
  - fixes parsing when `ORDER BY`, `LIMIT`, `OFFSET` specified on SQL set op
  - updates parse tests outdated tests (updates taken from #1506)
- In the plan and planner, uses refactored plan modeling from #1506 
- In the evaluator, includes the implementations of PartiQL bag operators (expressions)

## Other Information

  - Changes to AST and plan but on `v1` branch

## License Information


partiql-eval/src/main/kotlin/org/partiql/eval/internal/helpers/ToBag.kt
```diff
@@ -12,7 +11,14 @@ import org.partiql.value.PartiQLValueExperimental
  *  - F(array_value)  -> bag_value          # discard ordering
  *  - F(bag_value)    -> bag_value          # identity
  */
-@OptIn(PartiQLValueExperimental::class)
-internal fun PartiQLValue.toBag(): BagValue<*> {
-    TODO("For OUTER set operators")
+internal fun Datum.asIterator(): Iterator<Datum> {
```
alancai98, line 14: Based on an offline discussion w/ @johnedquinn, mistyped cases in strict mode should result in an error.

For reference:
- partiql-lang issue https://github.com/partiql/partiql-lang/issues/76
- partiql-tests fix https://github.com/partiql/partiql-tests/pull/120/files

I will correct this in a new commit

partiql-ast/src/main/resources/partiql_ast.ion
```diff
@@ -489,22 +489,10 @@ expr::[
...
-    from:     from,
-    let:      optional::let,
-    where:    optional::expr,
-    group_by: optional::group_by,
-    having:   optional::expr,
-    set_op:   optional::{
-      type:    '.set_op',
-      operand: '.expr.s_f_w',
-    },
+  query_set::{
```
alancai98, line 494: (self-review) primary addition of this PR is adding `query_set` and `query_expr` to the AST. This modeling allows us to define `ORDER BY`, `LIMIT`, and `OFFSET` on SFW queries as well as SQL set ops.

Modeling is partially based off
- partiql-lang-rust -- https://github.com/partiql/partiql-lang-rust/blob/e5b8ce7497f515bc998a31cd8b45fb2f3e9287a2/partiql-ast/src/ast.rs#L238-L244
- similar modeling in the sqlparser Rust crate -- https://docs.rs/sqlparser/0.48.0/sqlparser/ast/struct.Query.html

partiql-lang/src/main/kotlin/org/partiql/lang/prettyprint/QueryPrettyPrinter.kt
```diff
@@ -320,7 +320,14 @@ class QueryPrettyPrinter {
             is PartiqlAst.Expr.Or -> writeNAryOperator("OR", node.operands, sb, level)
             is PartiqlAst.Expr.InCollection -> writeNAryOperator("IN", node.operands, sb, level)
             is PartiqlAst.Expr.BagOp -> {
-                var name = node.op.javaClass.simpleName.toUpperCase().replace("_", " ")
+                var name = when (node.op) {
```
alancai98, line 323: (self-review) changes to `partiql-lang` tests follow what I did for a similar PR targeting `v0.14.6` to fix some legacy tests -- https://github.com/partiql/partiql-lang-kotlin/pull/1506.

partiql-plan/src/main/resources/partiql_plan.ion
```diff
@@ -322,7 +342,7 @@ rel::{
       _: [
         call::{
           agg:  ref,
-          set_quantifier: [ ALL, DISTINCT ],
+          setq: set_quantifier,
```
alancai98, line 345: (self-review) moved `set_quantifier` out of this nested aggregation plan node. Also renamed to `setq` for consistency w/ the set ops.

test/partiql-tests
alancai98, line 1: (self-review) currently points to the source branch of https://github.com/partiql/partiql-tests/pull/121. Once that `partiql-tests` PR is merged in, I will update the submodule commit.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/helpers/ToBag.kt
```diff
@@ -1,18 +1,35 @@
...
  *  - F(tuple_value)  -> << tuple_value >>  # singleton bag, see future extensions
  *  - F(array_value)  -> bag_value          # discard ordering
  *  - F(bag_value)    -> bag_value          # identity
+ *
+ * Throws a [TypeCheckException] when coercing a non-collection to a bag.
  */
-@OptIn(PartiQLValueExperimental::class)
-internal fun PartiQLValue.toBag(): BagValue<*> {
-    TODO("For OUTER set operators")
+internal fun Datum.asIterator(coerce: Boolean): Iterator<Datum> {
```
alancai98, line 17: (self-review) coercion behavior is slightly different than other type check exceptions in permissive mode. For the bag operators, we coerce to a bag rather than return missing.

partiql-ast/src/main/resources/partiql_ast.ion
johnedquinn: Nice work. Would we want to add the naming followed by PLR and SQLParser? AKA `Query` is the top-level product type holding the expression, order-by, etc. And the union types `QuerySet` (PLR) or `SetExpr` (SQLParser) holding the different variants (expr, select, wrapped Query, values, etc).

Especially since the Java name will now be: `Expr.QuerySet` and `QueryExpr`. I found it a bit confusing in the rest of the code. Could just be: `Expr.Query` and `Expr.Query.Body`.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprUnionAll.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.partiql.eval.internal.Environment
+import org.partiql.eval.internal.helpers.IteratorChain
+import org.partiql.eval.internal.helpers.asIterator
+import org.partiql.eval.internal.operator.Operator
+import org.partiql.eval.value.Datum
+
+internal class ExprUnionAll(
+    private val lhs: Operator.Expr,
+    private val rhs: Operator.Expr,
+    private val coerce: Boolean,
```
johnedquinn, line 12: IMO, there should be as little logic in the execution nodes as possible, unless absolutely necessary. So, for `ExprUnionAll`, maybe there are two implementations: `ExprUnionAll.Permissive` and `ExprUnionAll.Strict`. That way, there is no additional planning logic in the actual execution.

Same goes for the other nodes.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprUnionDistinct.kt
```diff
@@ -0,0 +1,42 @@
...
+    private val lhs: Operator.Expr,
+    private val rhs: Operator.Expr,
+    private val coerce: Boolean,
+) : Operator.Expr {
+    override fun eval(env: Environment): Datum {
+        val lDatum = lhs.eval(env)
+        val rDatum = rhs.eval(env)
+        val iter = Iterable {
+            // TODO: Add support for equals/hashcode in Datum
+            val seen: MutableSet<PartiQLValue> = mutableSetOf()
```
johnedquinn, line 23:
```suggestion
            val seen = TreeSet(PartiQLValue.comparator())
```

johnedquinn (changes requested):
Still need to go through the evaluation implementations for the bag ops -- but overall looks good. I've left some comments on the modelling.

alancai98:
Thanks for the review @johnedquinn. Based on the offline discussion, we decided on the following changes
- allow `LIMIT`, `OFFSET`, and `ORDER BY` on the outer bag ops
- coercion function for `NULL` and `MISSING` will the same as the spec's `FROM` source coercion (basically wrap the absent type in a bag rather than empty bag).

Will mark this PR as a draft till I address those concerns and the other PR comments.

partiql-parser/src/main/antlr/PartiQLParser.g4
```diff
@@ -541,9 +542,9 @@ exprSelect
         where=whereClauseSelect?
         group=groupClause?
         having=havingClause?
-        order=orderByClause?
-        limit=limitClause?
-        offset=offsetByClause? # SfwQuery
+        order=orderByClause??
+        limit=limitClause??
+        offset=offsetByClause?? # SfwQuery
```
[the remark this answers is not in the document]
alancai98: > Since the AST nodes don't allow for this, this may result in ambiguous results.

Could you clarify on this? Are you referring to if a set op query is defined without parens and it includes a `LIMIT`, `OFFSET`, or `ORDER BY`? For example,

```
SELECT * FROM a UNION SELECT * FROM b LIMIT 5
```

In this case, I believe the parsing behavior we want is to apply the `LIMIT` to the `UNION` set operator. This behavior follows what PostgreSQL's parsing does - https://www.postgresql.org/docs/current/queries-union.html.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprUnionAll.kt
alancai98: Latest commit removes these `Rex` exprs and defers to use the existing `Rel` scan logic.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprUnionDistinct.kt
alancai98: Similarly, these `Rex`s were removed in latest commit to use the existing `Rel` scan logic.

partiql-ast/src/main/resources/partiql_ast.ion
alancai98: I added some more context on PLR's AST definition in the partiql-tests PR https://github.com/partiql/partiql-tests/pull/121#discussion_r1719123490. The AST definition I used in this PR allows for us to be more strict in what expressions we want to support with `LIMIT`, `OFFSET`, and `ORDER BY`. For now, I'm thinking we only permit those clauses on
- SFW
- SQL set ops
- PartiQL bag ops

PLR also allows those clauses on general expressions. Since it's not in the PartiQL spec, I haven't chosen to allow those clauses on general expressions.

I agree that the previous naming was confusing, so updated to use `query_body` rather than `query_expr`.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -441,24 +471,34 @@ internal object RelConverter {
...
-            if (setOp == null) {
-                return input
-            }
-            val type = input.type.copy(props = emptySet())
-            val lhs = input
-            val rhs = visitExprSFW(setOp.operand, nil)
-            val quantifier = when (setOp.type.setq) {
-                SetQuantifier.ALL -> Rel.Op.Set.Quantifier.ALL
-                null, SetQuantifier.DISTINCT -> Rel.Op.Set.Quantifier.DISTINCT
+        private fun convertSetOp(setExpr: QueryBody.SetOp): Rel {
```
alancai98, line 489: (self-review) one benefit of reusing the SQL set op plan nodes for the PartiQL bag ops is that we can reuse the `Rel` scan behavior for non-collections (i.e. coercing to a bag in permissive and erroring in strict mode).

johnedquinn (commented):
Looks good to me! Once the tests PR is merged and the submodule commit is updated, I can approve.

alancai98:
Since the last review,
- rebased the PR off of `v1` to include the latest datum comparator commit
- [`53e96c2` (#1538)](https://github.com/partiql/partiql-lang-kotlin/pull/1538/commits/53e96c25d821636211efbd0c7423380e799330be) uses the latest tests from https://github.com/partiql/partiql-tests/pull/121 and updates the missing to null coercion for the set op evaluation

partiql-eval/src/main/kotlin/org/partiql/eval/internal/helpers/RecordUtility.kt
```diff
@@ -1,12 +1,17 @@
...
+    /**
+     * Converts the [Record] into an array of Datum, while coercing any missing values into null values.
+     */
+    fun Record.toDatumArrayCoerceMissing(): Array<Datum> = Array(this.values.size) {
+        val d = this@toDatumArrayCoerceMissing.values[it]
+        when (d.isMissing) {
+            true -> Datum.nullValue()
+            else -> d
+        }
     }
```
johnedquinn, line 16: Just clarifying why you're doing this: even though the comparator treats null/missing as equal, it isn't deterministic which value will be returned in the TreeSet -- so you need to coerce them ahead of time.

Also, I'm leaving a suggestion here because I don't think you need to create a new Array and move all previous pointers to the new one.

```suggestion
    fun Array<Datum>.coerceMissing() {
        for (i in 0 until this.size) {
            val d = this[i]
            if (d.isMissing) {
                this[i] = Datum.nullValue()
            }
        }
    }
```

alancai98: > Just clarifying why you're doing this: even though the comparator treats null/missing as equal, it isn't deterministic which value will be returned in the TreeSet -- so you need to coerce them ahead of time.

Yes, that's correct. I added a brief comment in the latest commit to give some more context on the behavior.

---
> Also, I'm leaving a suggestion here because I don't think you need to create a new Array and move all previous pointers to the new one.

Thanks for the suggestion. You're right -- it can be simplified quite a bit. I applied your suggestion in the latest commit.

johnedquinn (approved):

PR #1539, opened 2024-08-07
Addresses some conformance edge cases

johnedquinn:
## Description
- This PR addresses some small bugs that are fairly self explanatory.
- The only meaningful change has to do with [Section 6.1.1.1 Tuple Constructors (Mistyping Cases)](file:///Users/johqunn/Development/partiql-lang-jvm/test/partiql-tests-runner/build/reports/tests/generateTestReport/index.html):
> In the permissive mode, the query:
>
> SELECT VALUE {v.a: v.b}
> FROM [{'a':'legit', 'b':1}, {'a':400, 'b':2}] AS v
>
> results into
>
> <<{'legit':1}, {}>>
>
> Notice that the attempt to create an attribute named 400 failed, thus leading to a tuple with no attributes.

## Other Information

## License Information

### DIFF
--- a/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprStructPermissive.kt
+++ b/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprStructPermissive.kt
@@ -1,0 +1,37 @@
+package org.partiql.eval.internal.operator.rex
+
+import org.partiql.eval.internal.Environment
+import org.partiql.eval.internal.operator.Operator
+import org.partiql.eval.value.Datum
+import org.partiql.types.PType
+
+internal class ExprStructPermissive(private val fields: List<ExprStructField>) : Operator.Expr {
+    override fun eval(env: Environment): Datum {
+        val fields = fields.mapNotNull {
+            val key = it.key.eval(env)
+            if (key.isNull) {
+                return Datum.nullValue()
+            }
+            val keyString = key.getTextOrNull() ?: return Datum.struct(emptyList())
+            val value = it.value.eval(env)
+            when (value.isMissing) {
+                true -> null
+                false -> org.partiql.eval.value.Field.of(keyString, value)
+            }
+        }
+        return Datum.struct(fields)
+    }
+
+    companion object {
+        /**
+         * @throws NullPointerException if the value itself is null
+         * @return the underlying string value of a textual value; null if the type is not a textual value.
+         */
+        fun Datum.getTextOrNull(): String? {
+            return when (this.type.kind) {
+                PType.Kind.STRING, PType.Kind.SYMBOL, PType.Kind.CHAR -> this.string
+                else -> null
+            }
+        }
+    }
+}
--- a/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprStructPermissive.kt
+++ b/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprStructPermissive.kt
@@ -9,9 +9,6 @@ internal class ExprStructPermissive(private val fields: List<ExprStructField>) :
     override fun eval(env: Environment): Datum {
         val fields = fields.mapNotNull {
             val key = it.key.eval(env)
-            if (key.isNull) {
-                return Datum.nullValue()
-            }
             val keyString = key.getTextOrNull() ?: return@mapNotNull null
             val value = it.value.eval(env)
             when (value.isMissing) {
@@ -24,11 +21,14 @@ internal class ExprStructPermissive(private val fields: List<ExprStructField>) :
 
     companion object {
         /**
-         * @throws NullPointerException if the value itself is null
-         * @return the underlying string value of a textual value; null if the type is not a textual value.
+         * @return the underlying string value of a textual value; null if the type is not a textual value or if the
+         * value itself is absent.
          */
         @JvmStatic
         fun Datum.getTextOrNull(): String? {
+            if (this.isNull || this.isMissing) {
+                return null
+            }
             return when (this.type.kind) {
                 PType.Kind.STRING, PType.Kind.SYMBOL, PType.Kind.CHAR -> this.string
                 else -> null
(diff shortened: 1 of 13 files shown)


partiql-spi/src/main/kotlin/org/partiql/spi/fn/builtins/FnAnd.kt
```diff
@@ -27,11 +27,14 @@ internal object Fn_AND__BOOL_BOOL__BOOL : Fn {
     override fun invoke(args: Array<Datum>): Datum {
         val lhs = args[0]
         val rhs = args[1]
+        val lhsIsUnknown = lhs.isNull || lhs.isMissing
```
alancai98, line 30: I happened to be checking out the other boolean ops (`OR` and `NOT`). Seems like this logic should also be applied to `OR` and `NOT`

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -1089,6 +1089,9 @@ internal class PlanTyper(private val env: Env) {
          * Calculate output type of a scalar subquery.
          */
         private fun visitRexOpSubqueryScalar(subquery: Rex.Op.Subquery, cons: CompilerType): Rex {
+            if (cons.kind == Kind.DYNAMIC) {
```
alancai98, line 1092: Why was this check needed?

johnedquinn: In the following lines, `cons.fields` is called -- which is not relevant to a dynamic type, causing an exception to be thrown.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprStructPermissive.kt
```diff
@@ -0,0 +1,37 @@
...
+            val value = it.value.eval(env)
+            when (value.isMissing) {
+                true -> null
+                false -> org.partiql.eval.value.Field.of(keyString, value)
+            }
+        }
+        return Datum.struct(fields)
+    }
+
+    companion object {
```
alancai98, line 25: nit: any reason this is within a companion object?

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprStructPermissive.kt
```diff
@@ -0,0 +1,37 @@
...
+import org.partiql.types.PType
+
+internal class ExprStructPermissive(private val fields: List<ExprStructField>) : Operator.Expr {
+    override fun eval(env: Environment): Datum {
+        val fields = fields.mapNotNull {
+            val key = it.key.eval(env)
+            if (key.isNull) {
+                return Datum.nullValue()
+            }
+            val keyString = key.getTextOrNull() ?: return Datum.struct(emptyList())
```
alancai98, line 15: Not sure this logic is correct. Consider the struct constructor expression call in permissive mode: 
```
{100: 1, 'b': 2, 'c': 3}
```

As it's currently written, this would result in an empty struct rather than omitting that single invalid struct field.

partiql-spi/src/main/kotlin/org/partiql/spi/fn/builtins/FnOr.kt
```diff
@@ -25,9 +25,18 @@ internal object Fn_OR__BOOL_BOOL__BOOL : Fn {
...
 
     override fun invoke(args: Array<Datum>): Datum {
-        if (args[0].isNull || args[1].isNull) return Datum.nullValue(PType.typeBool())
-        val lhs = args[0].boolean
-        val rhs = args[1].boolean
-        return Datum.bool(lhs || rhs)
+        val lhs = args[0]
+        val rhs = args[1]
+        val lhsIsUnknown = lhs.isNull || lhs.isMissing
+        val rhsIsUnknown = rhs.isNull || rhs.isMissing
```
alancai98, line 31: I think we should apply this change for `FnNot`?

johnedquinn: Added. Not exactly the same logic since the NULL CALL status handles it for `FnNot`, but I've added support for the missing check.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprStructPermissive.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.partiql.eval.internal.operator.Operator
+import org.partiql.eval.value.Datum
+import org.partiql.types.PType
+
+internal class ExprStructPermissive(private val fields: List<ExprStructField>) : Operator.Expr {
+    override fun eval(env: Environment): Datum {
+        val fields = fields.mapNotNull {
+            val key = it.key.eval(env)
+            if (key.isNull) {
+                return Datum.nullValue()
```
alancai98, line 13: Why is `null` returned for the overall struct constructor expression if one key is `null`? For the following query,

```
SELECT VALUE {v.a: v.b} FROM [{'a':'legit', 'b':1}, {'a':null, 'b':2}] AS v
```

`<< { legit: 1 }, null >>` will currently be the output.

My read of the "Treatment of mistyped attribute names" section of the PartiQL spec is that `null` struct key values (alike non-strings and `missing` values) would just omit that struct field from the struct in permissive mode. Following my read, the above query would result in `<< { 'legit': 1 }, {} >>`.

alancai98 (commented):
one followup comment for a struct constructor edge case. otherwise looks good

alancai98 (commented):
one followup comment for a struct constructor edge case. otherwise looks good

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprStructPermissive.kt
johnedquinn: Ok, nice! From our offline discussion, I've created an issue with the team's consensus on behavior as well as the need for potential clarification in the PartiQL Spec: https://github.com/partiql/partiql-lang/issues/88

I've just updated this PR accordingly. I also needed to rebase off of the type-renaming PR. The last two commits represent what has been updated.

alancai98 (approved):

PR #1542, opened 2024-08-07
[v1] Fix try-catch for conformance tests expecting an error

alancai98:
## Relevant Issues
- None

## Description
- Seems like the conformance test runner was not reporting back test failures that expected an `EvaluationFail` but did not error

## Other Information

## License Information

### DIFF
--- a/test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/test/TestRunner.kt
+++ b/test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/test/TestRunner.kt
@@ -26,6 +26,8 @@ class TestRunner<T, V>(private val factory: TestExecutor.Factory<T, V>) {
         case.statements.forEach { run(it, case, executor) }
     }
 
+    private class TestRunnerException(message: String) : RuntimeException(message)
+
     private fun run(input: String, case: TestCase, executor: TestExecutor<T, V>) {
         val assertion = case.assertion
         try {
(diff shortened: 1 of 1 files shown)


alancai98:
Some additional test failures due to this change -- https://github.com/partiql/partiql-lang-kotlin/actions/runs/10292632996.

Some of these errors are properly caught by the planner but not detected by the evaluator. Regarding the planner errors, we should probably throw on these cases -- https://github.com/partiql/partiql-lang-kotlin/issues/1543.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/test/TestRunner.kt
```diff
@@ -36,14 +38,16 @@ class TestRunner<T, V>(private val factory: TestExecutor.Factory<T, V>) {
                     val expect = executor.fromIon(assertion.expectedResult)
                     if (!executor.compare(actual, expect)) {
                         val ion = executor.toIon(actual)
-                        error("Expected: ${assertion.expectedResult}\nActual: $ion\nMode: ${case.compileOptions.typingMode}")
+                        throw TestRunnerException("Expected error to be thrown but none was thrown.\n${case.name}\nActual result: $ion")
```
johnedquinn, line 41: This error message isn't the intention behind Assertion.EvaluationSuccess.

johnedquinn (changes requested):
This PR makes me think, why are we even catching these exceptions? Using exceptions for business logic is generally advised against. IMO, this whole method can be simplified by creating the `statement`/`actual` variables in each assertion block. Then, we can use a try-catch just for the `Assertion.EvaluationFailure` where it's required.

Though, it would also be nice to have each variant of `TestCase` implement their own assertion logic.

alancai98:
> This PR makes me think, why are we even catching these exceptions? Using exceptions for business logic is generally advised against. IMO, this whole method can be simplified by creating the `statement`/`actual` variables in each assertion block. Then, we can use a try-catch just for the `Assertion.EvaluationFailure` where it's required.
> Though, it would also be nice to have each variant of `TestCase` implement their own assertion logic.

Yeah the current `TestRunner` is somewhat confusing. This PR was initially meant to unblock some testing as part of #1542, but I'll take another look at refactoring the current `TestRunner`.

johnedquinn (approved):
Discussed offline -- to unblock Alan's work on the bag ops, I'm approving this PR. A follow-up will ensue.

PR #1545, opened 2024-08-12
Adds Datum comparator

### LINKED ISSUE #1541: [V1] Add Datum comparator/hashcode/equals for grouping
(from offline discussion) yeah, seems reasonable to tackle this TODO later since it will affect a lot of other parts of the evaluator (e.g. bag ops).

might be worthwhile to cut an issue to track

_Originally posted by @alancai98 in https://github.com/partiql/partiql-lang-kotlin/pull/1513#discussion_r1707546985_

johnedquinn:
## Relevant Issues
- Closes #1541

## Description
- Adds a performant comparator for Datum. Please read the Javadocs associated with the comparator.
- Replaced all usages of the `PartiQLValue.comparator()` in favor of `Datum.comparator()`.
  - This should significantly speed up evaluation of aggregations, ordering, etc.
  - The only caveat is for the conformance tests -- I believe Robert was working on replacing the use of PartiQLValue with Datum with the Ion reader, so I didn't switch the comparator there.
- Passes all tests previously passing.

## Other Information

## License Information


partiql-spi/src/test/kotlin/org/partiql/eval/value/DatumComparatorTest.kt
johnedquinn, line 1: This is essentially a copy-paste of PartiQLValueComparatorTest. The only difference is that annotations tests have been commented out.

partiql-spi/src/main/java/org/partiql/eval/value/DatumComparator.java
```diff
@@ -0,0 +1,720 @@
...
+import java.util.Map;
+import java.util.function.Supplier;
+
+/**
+ * This class allows for the comparison between two {@link Datum}s. This is internally implemented by constructing
+ * a comparison table, where each cell contains a reference to a {@link DatumComparison} to compute the comparison.
+ * The table's rows and columns are indexed by the {@link PType.Kind#ordinal()}. The first dimension matches the
+ * left-hand-side's type of {@link #compare(Datum lhs, Datum rhs)}. The second dimension matches the right-hand-side's
+ * type of {@link #compare(Datum lhs, Datum rhs)}. As such, this implementation allows for O(1) comparison of scalars.
+ */
```
alancai98, line 22: Wonder if you had ran any benchmarks comparing the performance between `PartiQLValueComparator` and the `DatumComparator`. If `DatumComparator` is significantly more performant than the previous `PartiQLValueComparator`, it could be a good callout for our upcoming `v1` release.

alancai98 (commented):
Nice work! The new factoring of the datum comparator looks to be more efficient than the previous versions of our value comparator. Left some questions and other minor comments.

partiql-spi/src/main/java/org/partiql/eval/value/DatumComparator.java
johnedquinn: Just finished running a set of benchmarks that shadow this PR and the V1.0.0-perf.1 release. The benchmarks compare evaluation of aggregations, distinct, order by, and comparison operators for the latest engine (this PR), the `v1.0.0-perf.1` engine, and the legacy engine.

See the benchmarks at this [tag](https://github.com/partiql/partiql-lang-kotlin/releases/tag/v1-datum-comparator-benchmark-results.1). See the results [here](https://github.com/partiql/partiql-lang-kotlin/blob/b9d7def1ec4aeab1a93bad69a577ac54efd6dcc7/test/benchmarks/src/jmh/kotlin/org/partiql/benchmarks/tests/EvaluationBenchmarks.kt#L36).

johnedquinn: And, to hint: it's much faster. In 2 of the 3 benchmarks, the new implementation is over 2x faster than `v1.0.0-perf.1` and over 5x* faster than the legacy implementation.

To get a better overall view of how much the comparator itself is contributing to the performance, it could be a good idea to benchmark AND profile the comparator in action. For now though, we have a good glimpse as for how Datum, PType, and DatumComparator in coordination are improving performance.

The `*` above is discussed in the results link above. The legacy comparator is hard to quantify due to its immediate materialization of aggregations, but preliminary analysis indicates a huge performance improvement nonetheless.

partiql-spi/src/main/kotlin/org/partiql/spi/fn/builtins/internal/AccumulatorDistinct.kt
```diff
@@ -1,24 +1,17 @@
...
 
-@OptIn(PartiQLValueExperimental::class)
 internal class AccumulatorDistinct(
     private val _delegate: Accumulator,
 ) : Accumulator() {
 
-    // TODO: Add support for a datum comparator once the accumulator passes datums instead of PartiQL values.
-    @OptIn(PartiQLValueExperimental::class)
-    private val seen = TreeSet(PartiQLValue.comparator())
+    private val seen = TreeSet(Datum.comparator())
```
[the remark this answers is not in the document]
johnedquinn: Yeah, I'll add a comment. But, it comes down to the lack of support for hashCode and equals. The reason we can't support them outright for aggregations/grouping is because of how the PartiQL Specification dictates that the grouping should act similarly to the `<` operator. This operator compares the arithmetic values of INT, BIGINT, etc -- so a simple hashCode/equals is non-trivial. It *may* be fast to coerce values into some larger type specifically for aggregations -- but this requires further research.

So, a TreeSet is necessary since we can pass in the comparator.

partiql-spi/src/main/java/org/partiql/eval/value/DatumDouble.java
```diff
@@ -25,4 +29,9 @@ public double getDouble() {
     public PType getType() {
         return _type;
     }
+
+    @Override
+    public String toString() {
+        return "dp::" + _value;
+    }
```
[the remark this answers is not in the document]
johnedquinn: https://github.com/partiql/partiql-lang-kotlin/issues/1553

partiql-spi/src/main/java/org/partiql/eval/value/DatumComparator.java
```diff
@@ -0,0 +1,720 @@
...
+
+    @NotNull
+    private static final PType.Kind[] TYPE_KINDS = PType.Kind.values();
+
+    private static final int TYPE_KINDS_LENGTH = TYPE_KINDS.length;
+
+    /**
+     * This array defines the precedence of types when comparing values of different types.
+     * The lower the index, the higher the precedence. However, this is used for generating the BASELINE precedence.
+     * This may be overridden by families of types. For example, all number values are compared by their mathematical
```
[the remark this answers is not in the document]
johnedquinn: Ah, yeah. I was having trouble finding the right wording for this.

In this implementation, I initialize the two-dimensional array (see `initializeComparatorArray()`) by only performing comparisons between any two type-family's precedence.. So, for the cell representing the comparison between INT and STRING, I initialize the cell to return LESS (due to the numbers coming before text types). Similarly, for INT and BIGINT, I initialize the cell to return EQUALS (due to the shared type precedence). Note that this is only on the initialization of the array.

Now, in order to handle the comparisons *within* types of the same precedence, I need to manually rewrite some of those comparisons (the ones where there originally returned an EQUALS). So, for example, `fillBigIntComparator` is invoked before the static 2D table is returned, and it specifically handles cases such as the comparison of BIGINT and INT. It'll place within the table a function that compares the underlying JVM values of the BIGINT and INT.

So, with both of these passes, we can ensure that we have memoized the comparisons of values by handling both the overarching type family comparisons and the value comparisons (where necessary).

I'll update this specific Javadoc since it probably causes more confusion than necessary and will include comments in other private methods for better clarification.

partiql-spi/src/main/java/org/partiql/eval/value/DatumComparator.java
alancai98: Awesome, thanks for creating the benchmarks and tests! The preliminary analysis looks good. Those statements seem like a good starting point for any future performance benchmarking test suite. Ahead of the `v1` release, we should definitely benchmark more queries to show off the performance benefit of the new evaluator.

partiql-spi/src/main/java/org/partiql/eval/value/DatumComparator.java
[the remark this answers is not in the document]
alancai98: Thanks for the explanation. The updated javadoc is clearer for what's going on

alancai98 (approved):
Nice work!

PR #1549, opened 2024-08-15
Preserve constraint for decimal, string in dynamic typer

yliuuuu:
## Relevant Issues

## Description

1. Preserve constraint in dynamic typer

Previous, 
```
CASE  WHEN <condition> THEN <DECIMAL(10,5)>
    ELSE null
END;
```
return Decimal Type with arbitrary precision and scale
Similarly: 
```
CASE  WHEN <condition> THEN <VARCHAR(10)>
    ELSE null
END;
```
return String Type with arbitrary length constraint. 
With this PR: 
```
CASE  WHEN <condition> THEN <DECIMAL(10,5)>
    ELSE null
END;
```
returns DECIMAL(10,5).
```
CASE  WHEN <condition> THEN <VARCHAR(10)>
    ELSE null
END;
```
returns VARCHAR(10)

2. Type Constraint Promotion
If the common super type is of single type and require no coercion, the dynamic typer should perform promotion if applicable. 

Consider 
```
CASE  WHEN <condition> THEN <DECIMAL(10,5)>
    ELSE <DECIMAL(5,3)>
END;
```
With type promotion, the return type will be : Decimal(10,5)

Similarly: 
```
CASE  WHEN <condition> THEN <VARCHAR(10)>
    ELSE <VARCHAR(5)>
END;
```

The return type will be VARCHAR(10). 

In addition, because STRING and CHAR share the same runtime type: 

```
CASE  WHEN <condition> THEN <VARCHAR(10)>
    ELSE <CHAR(5)>
END;
```

returns VARCHAR(10) without the need of coercion. 

## Other Information
  - No. 

  - < If YES, why? >
No. 

 No. 

Yes. 

## License Information

### DIFF
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
@@ -3,7 +3,10 @@
 package org.partiql.planner.internal.typer
 
 import org.partiql.planner.internal.ir.Rex
+import org.partiql.types.DecimalType
+import org.partiql.types.NumberConstraint
 import org.partiql.types.StaticType
+import org.partiql.types.StringType
 import org.partiql.value.MissingValue
 import org.partiql.value.NullValue
 import org.partiql.value.PartiQLValueExperimental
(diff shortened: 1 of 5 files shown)


yliuuuu:
Closing in favor of #1551

PR #1551, opened 2024-08-16
[v0.14.8-SNAPSHOT] Preserve constraint for decimal, string in dynamic typer

yliuuuu:
## Relevant Issues

## Description
- Base #1549 on tip of v0.14.7. 
- Change version to v0.14.8-SNAPSHOT. (The target branch is not a release branch). 

1. Preserve constraint in dynamic typer

Previous, 
```
CASE  WHEN <condition> THEN <DECIMAL(10,5)>
    ELSE null
END;
```
return Decimal Type with arbitrary precision and scale
Similarly: 
```
CASE  WHEN <condition> THEN <VARCHAR(10)>
    ELSE null
END;
```
return String Type with arbitrary length constraint. 
With this PR: 
```
CASE  WHEN <condition> THEN <DECIMAL(10,5)>
    ELSE null
END;
```
returns DECIMAL(10,5).
```
CASE  WHEN <condition> THEN <VARCHAR(10)>
    ELSE null
END;
```
returns VARCHAR(10)

2. Type Constraint Promotion
If the common super type is of single type and require no coercion, the dynamic typer should perform promotion if applicable. 

Consider 
```
CASE  WHEN <condition> THEN <DECIMAL(10,5)>
    ELSE <DECIMAL(5,3)>
END;
```
With type promotion, the return type will be : Decimal(10,5)

Similarly: 
```
CASE  WHEN <condition> THEN <VARCHAR(10)>
    ELSE <VARCHAR(5)>
END;
```

The return type will be VARCHAR(10). 

In addition, because STRING and CHAR share the same runtime type: 

```
CASE  WHEN <condition> THEN <VARCHAR(10)>
    ELSE <CHAR(5)>
END;
```

returns VARCHAR(10) without the need of coercion. 

## Other Information
  - Yes. 

  - < If YES, why? >
No. 

No.

Yes.

## License Information

### DIFF
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
@@ -2,9 +2,12 @@
 
 package org.partiql.planner.internal.typer
 
+import org.partiql.types.DecimalType
 import org.partiql.types.MissingType
 import org.partiql.types.NullType
+import org.partiql.types.NumberConstraint
 import org.partiql.types.StaticType
+import org.partiql.types.StringType
 import org.partiql.value.PartiQLValueExperimental
 import org.partiql.value.PartiQLValueType
 import org.partiql.value.PartiQLValueType.ANY
@@ -35,6 +38,7 @@ import org.partiql.value.PartiQLValueType.STRUCT
 import org.partiql.value.PartiQLValueType.SYMBOL
 import org.partiql.value.PartiQLValueType.TIME
 import org.partiql.value.PartiQLValueType.TIMESTAMP
+import kotlin.math.max
 
 /**
  * Graph of super types for quick lookup because we don't have a tree.
@@ -127,19 +131,27 @@ internal class DynamicTyper {
         val modifiers = mutableSetOf<StaticType>()
         if (nullable) modifiers.add(StaticType.NULL)
         if (missable) modifiers.add(StaticType.MISSING)
-        // If at top supertype, then return union of all accumulated types
-        if (supertype == ANY) {
-            return StaticType.unionOf(types + modifiers).flatten() to null
-        }
-        // If a collection, then return union of all accumulated types as these coercion rules are not defined by SQL.
-        if (supertype == STRUCT || supertype == BAG || supertype == LIST || supertype == SEXP) {
-            return StaticType.unionOf(types + modifiers) to null
-        }
-        // If not initialized, then return null, missing, or null|missing.
-        val s = supertype
-        if (s == null) {
-            val t = if (modifiers.isEmpty()) StaticType.MISSING else StaticType.unionOf(modifiers).flatten()
-            return t to null
+        val s = when (val superT = supertype) {
+            // If not initialized, then return null, missing, or null|missing.
+            null -> {
+                val t = if (modifiers.isEmpty()) StaticType.MISSING else StaticType.unionOf(modifiers).flatten()
+                return t to null
+            }
+            // If at top supertype, then return union of all accumulated types
+            ANY -> return StaticType.unionOf(types + modifiers).flatten() to null
+            // If a collection, then return union of all accumulated types as these coercion rules are not defined by SQL.
+            STRUCT, BAG, LIST, SEXP -> return StaticType.unionOf(types + modifiers) to null
+            DECIMAL -> {
+                val type = computeDecimal()
+                // coercion required. fall back
+                if (type == null) superT else return StaticType.unionOf(setOf(type) + modifiers).flatten() to null
+            }
+            STRING -> {
+                val type = computeString()
+                // coercion required. fall back
+                if (type == null) superT else return StaticType.unionOf(setOf(type) + modifiers).flatten() to null
+            }
+            else -> superT
         }
         // Otherwise, return the supertype along with the coercion mapping
         val type = s.toNonNullStaticType()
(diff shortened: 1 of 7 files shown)


plugins/partiql-local/src/main/kotlin/org/partiql/plugins/local/LocalSchema.kt
```diff
@@ -149,6 +153,27 @@ public fun StructElement.toStructType(): StaticType {
...
 }
 
+public fun StructElement.toDecimalType(): StaticType {
+    val precision = get("precision").bigIntegerValue.intValueExact()
+    val scale = get("scale").bigIntegerValue.intValueExact()
+    val precisionScaleConstraint = DecimalType.PrecisionScaleConstraint.Constrained(precision, scale)
+    return DecimalType(precisionScaleConstraint)
+}
+
+public fun StructElement.toStringType(): StaticType {
```
alancai98, line 163: (minor) btw, for Scribe we had added constraints to the LocalSchema Ion reader -- https://github.com/partiql/partiql-scribe/pull/19. Could consider reusing that logic or making things the same in the future.

One difference is default values for decimal scale if unspecified. Also separating out string into string (i.e. `VARCHAR`) and char (i.e. `CHAR`).

test/partiql-tests
alancai98, line 1: Any reason the test submodule was updated? Perhaps we could revert it back to what's on v0.14.7?

partiql-planner/src/testFixtures/resources/inputs/basics/case.sql
```diff
@@ -292,6 +292,56 @@ CASE t_item.t_string
...
+
+--#[case-when-40]
+-- type: char(10)
+CASE t_item.t_string
+    WHEN 'a' THEN t_item.t_char_10
+    ELSE t_item.t_char_5
+END;
+
+--#[case-when-41]
+-- type: varchar(10)
```
alancai98, line 338: I'm curious what would happen if you had a `CHAR(x)` and `VARCHAR(y)` where `x` > `y`?

partiql-planner/src/testFixtures/resources/inputs/basics/case.sql
```diff
@@ -292,6 +292,56 @@ CASE t_item.t_string
...
+CASE t_item.t_string
```
alancai98, line 339: Could add a few more tests
- Mixing arbitrary decimal with parameterized decimals
- Mixing parameterized decimal with other numeric types
- Using a decimal with a specified `precision` and unspecified `scale` 
- Mixing arbitrary varchar / char with parameterized varchar / char

alancai98: Also, perhaps it would be helpful to add a `toString()` to static type's `DecimalType` since it currently does not output the precision/scale constraint.

partiql-planner/src/testFixtures/resources/inputs/basics/case.sql
yliuuuu: VARCHAR(x). 

It seems to be common practice and sane approach to envision varchar having higher type precedence than char. SQL Server [example](https://learn.microsoft.com/en-us/sql/t-sql/data-types/data-type-precedence-transact-sql?view=sql-server-ver16). 

It also makes sense to preserve the length constraint to be able to hold the largest possible data.

partiql-planner/src/testFixtures/resources/inputs/basics/case.sql
yliuuuu: Great suggestion on test addition. 

> Using a decimal with a specified precision and unspecified scale
Unspecified scale essentially just default to 0. In my opinion this is perhaps better covered by syntax to IR. 

>it would be helpful to add a toString() to static type's DecimalType...
Yes. I agree. But perhaps leaving this out of this PR?

plugins/partiql-local/src/main/kotlin/org/partiql/plugins/local/LocalSchema.kt
yliuuuu: Thanks for the context. 

Perhaps we can do this in a follow up PR, the reason being split string and char causes potentially a lot diff in our tests.

test/partiql-tests
yliuuuu: reverted.

partiql-planner/src/testFixtures/resources/inputs/basics/case.sql
alancai98: Ok cool. That lines up w/ my expectation for the behavior.

partiql-planner/src/testFixtures/resources/inputs/basics/case.sql
alancai98: > Yes. I agree. But perhaps leaving this out of this PR?

I'm fine leaving it out or adding it. Just may be easier to debug anywhere we print types in the error message. Also applies to the string constraints.

partiql-planner/src/testFixtures/resources/inputs/basics/case.sql
```diff
@@ -292,6 +292,100 @@ CASE t_item.t_string
...
+
+--#[case-when-44]
+-- type: decimal
+-- coercion required. Common super type is arbitrary decimal
+CASE t_item.t_string
+    WHEN 'a' THEN t_item.t_decimal_10_5
+    ELSE t_item.t_int32
+END;
+
+--#[case-when-45]
```
alancai98, line 366: test case is repeated. think we can delete?

partiql-planner/src/testFixtures/resources/inputs/basics/case.sql
```diff
@@ -292,6 +292,100 @@ CASE t_item.t_string
...
+    WHEN 'a' THEN t_item.t_decimal_10_5
+    ELSE t_item.t_decimal
+END;
+
+--#[case-when-44]
+-- type: decimal
+-- coercion required. Common super type is arbitrary decimal
+CASE t_item.t_string
+    WHEN 'a' THEN t_item.t_decimal_10_5
+    ELSE t_item.t_int32
```
alancai98, line 363: (observation) seems somewhat interesting that the common supertype here goes all the way to arbitrary decimals. It's correct but wondering if we could return a non-arbitrary precision type?

Perhaps we should only add that logic if there's a particular use case.

alancai98 (approved):

partiql-planner/src/testFixtures/resources/inputs/basics/case.sql
yliuuuu: Yeah I agree. 

Spent some time thinking about this actually. 

So INT-32 -> 2,147,483,647 -> which all can be hold under decimal(10,0).

So in case of `<int32> || <decimal(10,5)>`, we can return `decimal(10,5)` , 

and  in case of `<int32> || <decimal(5,3)>` we can return `decimal(10,3)`

to ensure no loss of precision. 

But upon second thought this appears to be "too smart" and a decision we might need spending more time on.

alancai98: Yeah, agree. Let's just keep as-is. If a use-case arises for the more precise logic, we can decide then.

alancai98 (approved):

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
```diff
@@ -151,6 +163,34 @@ internal class DynamicTyper {
         }
     }
 
+    private fun computeDecimal(): DecimalType? {
+        val (precision, scale) = types.fold((0 to 0)) { acc, staticType ->
+            val decimalType = staticType as? DecimalType ?: return null
+            val constr = decimalType.precisionScaleConstraint as DecimalType.PrecisionScaleConstraint.Constrained
```
johnedquinn, line 169: Is there a need to check for `Unconstrained` in this situation? You do it further below in `computeString()`. This might throw a `ClassCastException` leading to unexpected runtime errors.

Especially since we know that CASE WHENs with multiple output types might lead to an unconstrained decimal. So, a nested CASE WHEN may very well end up in this situation.

alancai98: `computeDecimal` is only called when the `supertype` is a constrained `DECIMAL` (i.e. `PartiQLValueType.DECIMAL`) and not an arbitrary decimal (i.e. `PartiQLValueType.DECIMAL_ARBITRARY`). So a constrained decimal should never be in the list of `types` when we enter this block.

I can add the same logic from string constraints to prevent any accidental class cast exception in case there's a bug.

alancai98: Added in the v0.14.8 release PR

PR #1544, opened 2024-08-12
[v1] Remove `CAN_CAST` and `CAN_LOSSLESS_CAST` from partiql-ast

alancai98:
## Relevant Issues
- Related to https://github.com/partiql/partiql-lang-kotlin/issues/1477

## Description
- `partiql-ast` -- removes `CAN_CAST` and `CAN_LOSSLESS_CAST` nodes
- Adds TODOs in `partiql-parser` grammar and `partiql-lang` tests for parts we want to remove ahead of the `v1` release

## Other Information
- Updated Unreleased Section in CHANGELOG: **[NO]**
  - No, on `v1` branch.

- Any backward-incompatible changes? **[YES]**
  - Yes, deletes `CAN_CAST` and `CAN_LOSSLESS_CAST` nodes from the `partiql-ast`

- Any new external dependencies? **[NO]**

- Do your changes comply with the [Contributing Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CONTRIBUTING.md)
  and [Code Style Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CODE_STYLE.md)? **[YES]**

## License Information

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1477, opened 2024-05-30
[v1] Remove unsupported nodes from the AST


alancai98:
Currently the `v1` AST (https://github.com/partiql/partiql-lang-kotlin/blob/22bb46125b4f2271d0e4ad6e206e76cc0ac25175/partiql-ast/src/main/resources/partiql_ast.ion) supports some nodes that are not in SQL99/SQL++ and do not intend to be part of the `v1` release. E.g.
- `can_cast`
- `can_lossless_cast`
- some legacy DML nodes

This issue tracks the removal of such nodes.

alancai98:
Also should look at `set_op` defined within the `s_f_w` node of Ast's `Expr`:
https://github.com/partiql/partiql-lang-kotlin/blob/5b86afc7a3e1a98d379645b709de03caf6436695/partiql-ast/src/main/resources/partiql_ast.ion#L544-L547

It seems redundant with `bag_op`:
https://github.com/partiql/partiql-lang-kotlin/blob/5b86afc7a3e1a98d379645b709de03caf6436695/partiql-ast/src/main/resources/partiql_ast.ion#L527-L533

Currently, the default partiql-parser also creates `bag_op` rather than `set_op`:
https://github.com/partiql/partiql-lang-kotlin/blob/5b86afc7a3e1a98d379645b709de03caf6436695/partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt#L1221-L1258

PR #1552, opened 2024-08-16
[v1] Remove DML and EXEC ast nodes

alancai98:
## Relevant Issues
- Related to #1477
- Similar to #1544

## Description
Removes the DML and EXEC nodes from the partiql-ast. 

Eventually we will want to add back the DML nodes that are spec'd out in the RFC and in SQL. The previous modeling in the `partiql-ast` may need revised since there were
- some missing parts (e.g. `DO UPDATE` with assignment)
- redundant components
- TODOs throughout the parser and ast

So rather than committing to the potentially problematic AST modelings in the `v1` release, I removed the nodes from the AST. We can always add those nodes back into the AST as they are needed.

Also worth noting that the previous ANTLR parser supported both the spec'd features as well as some legacy DML features that we don't want to support going forward. Removing the legacy DML features simplify the parsing rules and resolve some of the existing TODOs related to parsing DML.

`EXEC` is not fully spec'd out and was only added for a legacy customer. If it's ever needed after `v1`, we can add it back later.

## Other Information
  - No, `v1` branch

  - Yes, but on `v1` branch

## License Information


partiql-lang/src/test/kotlin/org/partiql/lang/syntax/PartiQLParserDMLTest.kt
```diff
@@ -0,0 +1,2229 @@
...
+ * Tests parsing of DML using just the PIG parser.
+ */
+class PartiQLParserDMLTest : PartiQLParserTestBase() {
+    override val targets: Array<ParserTarget> = arrayOf(ParserTarget.DEFAULT)
+
+    // ****************************************
+    // DML
+    // ****************************************
+
+    @Test
```
alancai98, line 26: (self-review) DML parsing tests were copied directly from `PartiQLParserTest` and runs on just the PIG parser.

partiql-lang/src/test/kotlin/org/partiql/lang/syntax/PartiQLParserExecTest.kt
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Tests parsing of EXEC using just the PIG parser.
+ */
+class PartiQLParserExecTest : PartiQLParserTestBase() {
+    override val targets: Array<ParserTarget> = arrayOf(ParserTarget.DEFAULT)
+
+    // ****************************************
+    // EXEC clause parsing
+    // ****************************************
+    @Test
```
alancai98, line 19: (self-review) `EXEC` parsing tests were copied directly from `PartiQLParserTest` and runs on just the PIG parser.

johnedquinn (approved):

PR #1554, opened 2024-08-19
Prepare 0.14.8 release

alancai98:
## Relevant Issues
- None

## Description
- Prepares v0.14.8 release

## Other Information
- Updated Unreleased Section in CHANGELOG: **[YES]**

- Any backward-incompatible changes? **[NO]**

- Any new external dependencies? **[NO]**

- Do your changes comply with the [Contributing Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CONTRIBUTING.md)
  and [Code Style Guidelines](https://github.com/partiql/partiql-lang-kotlin/blob/main/CODE_STYLE.md)? **[YES]**

## License Information

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1565, opened 2024-08-22
Merges tag v0.14.8 into v1

johnedquinn:
## Description
- Merges tag `v0.14.8` into `v1`

## License Information


CHANGELOG.md
```diff
@@ -23,7 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Thank you to all who have contributed!
 -->
 
-## [Unreleased]
+## [0.14.8]
```
alancai98, line 26: The v0.14.8 CHANGELOG entry does not match w/ the v0.14.8 branch's CHANGELOG -- https://github.com/partiql/partiql-lang-kotlin/blob/v0.14.8/CHANGELOG.md. Perhaps it's because of a prior merge of `main` to `v1`?

partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEngineDefaultTest.kt
```diff
@@ -1389,22 +1389,18 @@ class PartiQLEngineDefaultTest {
...
+                    CASE
+                        WHEN NULL THEN 'isNull'
+                        WHEN MISSING THEN 'isMissing'
+                        WHEN FALSE THEN 'isFalse'
                     END
-                FROM << i, f, d, null, missing >> AS x
+                    ;
             """,
-            expected = boolValue(true),
+            expected = stringValue(null),
```
alancai98, line 1403: (minor) is this just a local test? perhaps we could just delete it from `v1` to prevent future rebase/merge conflicts?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
```diff
@@ -132,6 +146,47 @@ internal class DynamicTyper {
         return s to mapping
     }
 
+    private fun computeDecimal(): CompilerType? {
```
alancai98, line 149: Wonder if we should preserve the change from Yingtao's [comment](https://github.com/partiql/partiql-lang-kotlin/pull/1554#discussion_r1722419851) in commit [`873b6c4` (#1554)](https://github.com/partiql/partiql-lang-kotlin/pull/1554/commits/873b6c4f4ce835ded2dcc91e091cf1895c8b0f52). Basically,
- Make it clearer that this function is only for computing constrained decimals. So rename to `computeConstrainedDecimal`.
- Error when we receive an arbitrary decimal rather than returning `null`, since that would indicate a bug in the dynamic typer. This function shouldn't be called when we have previously encountered an arbitrary decimal.

alancai98 (commented):
Left a few minor comments. Merge looks good otherwise

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
johnedquinn: The merged change is still wrong. The supertype of INT, SMALLINT, and DECIMAL(5, 0), according to #1554 is DECIMAL. Though, when `computeDecimal` is invoked to compute the precision and scale, it throws an error when folding over the types of INT and SMALLINT.

This PR merges the not completely correct code while mitigating the thrown exception. I tried to apply as minimal changes as possible to the merge, but this is one area known to have a bug.

CHANGELOG.md
johnedquinn: Nice catch. Somehow that merged cleanly.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
alancai98: Ok -- then if the merged behavior is misaligned w/ the change in 0.14.8, then we should either
1. track it in a TODO and tracking issue
2. or fix it in this PR if it's not too much work

CHANGELOG.md
alancai98: Seems like we're also missing some of the CHANGELOG release links for v0.14.6-v0.14.8 -- https://github.com/johnedquinn/partiql-lang-jvm/blob/v1-merge-v0_14_8/CHANGELOG.md?plain=1#L1124-L1125.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/DynamicTyper.kt
johnedquinn: https://github.com/partiql/partiql-lang-kotlin/issues/1566

alancai98 (approved):

PR #1579, opened 2024-09-13
[v1] Add simplified AST classes and visitors

alancai98:
## Relevant Issues
- None

## Description
- Adds simplified AST classes and visitors written in Java
- Also unnests existing AST classes to 1 level of nesting maximum (e.g. will have `foo.bar`)
- Subsequent PRs will
  - Add docs for new classes
  - Add equals/hashcode impls
  - Integrate w/ existing `v1` code
  - Add factory functions + builders
- Note: DDL has also been removed for now. DDL will be added back at some point before the v1 release to show we can add new features without breaking existing code.
 
## Other Information
  - No -- v1 branch

## License Information


alancai98:
Converting to a draft. From offline discussion, will look into
- rewriting using simple Java classes
- see the viability of using Lombok w/ the Java classes
- unnest some AST nodes

partiql-ast/src/main/java/org/partiql/ast/v1/DDL.java
```diff
@@ -0,0 +1,23 @@
+package org.partiql.ast.v1;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public abstract class DDL extends Statement {
```
alancai98, line 8: TODO: delete DDL stuff from PR (for now)

partiql-ast/src/main/java/org/partiql/ast/v1/AstVisitor.java
```diff
@@ -0,0 +1,367 @@
...
+
+    R visitTypeStruct(TypeStruct node, C ctx);
+
+    R visitTypeAny(TypeAny node, C ctx);
+
+    R visitTypeCustom(TypeCustom node, C ctx);
+
+    R visitIdentifier(Identifier node, C ctx);
+
+    R visitIdentifierSymbol(Identifier.Symbol node, C ctx);
```
alancai98, line 173: (self-review) I tried to get rid of excessive nesting of classes from the original sprout-generated AST code. Commit [`1a4e2de` (#1579)](https://github.com/partiql/partiql-lang-kotlin/pull/1579/commits/1a4e2de73ce9a6d3f329472af0d8fcf618eaf0ee) shows the change to use at most 1 nested class in the Kotlin code before translating the classes and interfaces to Java.

partiql-ast/src/main/java/org/partiql/ast/v1/CaseSensitivity.java
johnedquinn, line 1: While I've already brought this up in person, I'm adding some context and links here.

Thinking about the backwards/forwards compatibility, I think it would be a better approach to use integers rather than enums. Some Java libraries have done this, and I've also seen a similar approach from protobuf. Here's an [interesting issue](https://github.com/open-telemetry/opentelemetry-proto/issues/424) that discusses the downsides of using closed enums.

We could potentially have a model that looks something like this:
```java
public interface IdentifierUnqualified {
    ...
    CaseSensitivity getCase();
}

public interface Enum {
    int getValue();
}

public interface CaseSensitivity implements Enum {
    static const int UNKNOWN = 0;
    static const int SENSITIVE = 1;
    static const int INSENSITIVE = 2;
}
```

OR, we could just use the integers directly without the Enum. Though, the enum does provide the possibility for better readability (`toString` overriding):
```java
public interface IdentifierUnqualified {
    ...
    /**
     * Valid returns are specified as constants in: `CaseSensitivity.CASE_*`
     */
    int getCase();
}

public class CaseSensitivity {
    static const int CASE_UNKNOWN = 0;
    static const int CASE_SENSITIVE = 1;
    static const int CASE_INSENSITIVE = 2;
}
```

This modeling enforces that Java users use the `default` branch for `switch` statements, protecting us from backward compatibility issues. Also, the `UNKNOWN` case helps us in forward compatibility scenarios for deserialization. This is an approach that [protobuf takes](https://protobuf.dev/programming-guides/proto3/#enum).

Our internal implementations, following the first model from above, could look like:
```kotlin
// Internal impl file
val id = idUnqualified("some_table", CaseSensitivity.SENSITIVE)
when (id.getCase().getValue()) {
    CaseSensitivity.SENSITIVE -> TODO()
    CaseSensitivity.INSENSITIVE -> TODO()
    else -> error("Unsupported. Or we can also convert to the UNKNOWN variant if need be.")
}

fun idUnqualified(name: String, case: Int): IdentifierUnqualified {
    return IdentifierUnqualifiedImpl(name, case)
}
```

partiql-ast/src/main/java/org/partiql/ast/v1/Identifier.java
```diff
@@ -0,0 +1,80 @@
...
+            return visitor.visitIdentifierQualified((Qualified) this, ctx);
+        } else {
+            throw new IllegalStateException("Unexpected value: " + this);
+        }
+    }
+
+    /**
+     * TODO docs, equals, hashcode
+     */
+    public static class Symbol extends Identifier {
```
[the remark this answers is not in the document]
johnedquinn: SQL:1999 also has `<local or schema qualified name>`, though, it doesn't fit our use-case particularly well, since we don't have the same two-level catalog structure. Could be of use to adopt this name. `Identifier` and `IdentifierChain`.

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprVar.java
```diff
@@ -0,0 +1,48 @@
...
+import org.partiql.ast.v1.Identifier;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public class ExprVar extends Expr {
```
johnedquinn, line 15:
```suggestion
// or VarRef, VarReference, Ref, BindingRef, etc.
public class Reference extends Expr {
```

I like this modeling as it provides a distinction between `AS <identifier>` vs referencing fields/globals/etc. In SQL:1999, this would be a `<column reference>` or `<table reference>`, though, PartiQL is a bit nuanced, since we allow for both everywhere an expression is allowed.

Anyways, the use of `reference` aligns with SQL.

```
<select list> ::=
  <asterisk>
  | <select sublist> [ { <comma> <select sublist> }... ]

<select sublist> ::=
  <derived column>
  | ...

<derived column> ::=
  <value expression> [ <as clause> ]

<as clause> ::= [ AS ] <column name>

<nonparenthesized value expression primary> ::=
  <unsigned value specification>
  | <column reference>
  | ...

<column reference> ::=
  <basic identifier chain>
  | MODULE <period> <qualified identifier> <period> <column name>
```

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprVar.java
```diff
@@ -0,0 +1,48 @@
...
+    @NotNull
+    public Identifier identifier;
```
johnedquinn, line 17:
```suggestion
    @NotNull
    public IdentifierChain identifier;
```

This would more closely align with the EBNF. Essentially all areas (except for projection aliases, `AS <identifier>`) uses `<identifier chain>` or `<local or schema qualified name>`. That being said, the chain may only have the root.

Up to you on this. It could help our modeling since we would never have to switch on the variants. You could have two distinct classes that don't derive from the same abstract class.

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprPathStep.java
johnedquinn, line 1: If you want to follow SQL more closely.

```java
/**
 * <element reference> ::= <array value expression> <left bracket or trigraph> <numeric value expression> <right bracket or trigraph>
 * Example: some_array_ref[2 + 2], some_ref['hello'], some_ref[CAST('' || 'a' AS STRING)]
 */
class ElementReference extends Expr {
    @NotNull
    Expr root;

    @NotNull
    Expr element;
}

/**
 * <field reference> ::= <value expression primary> <period> <field name>
 */
class FieldReference {
    @NotNull
    Expr root;

    @NotNull
    Identifier field;
}

/**
 * This is used as part of the projection list in SQL, but it directly uses <qualified asterisk>.
 * <all fields reference> ::= <value expression primary> <period> <asterisk>
 */
class AllFieldsReference {
  @NotNull
  Expr root;
}


/**
 * Below is not specified by SQL:1999.
 * <all elements reference> ::= <value expression primary> <left bracket or trigraph> <asterisk> <right bracket or trigraph>
 */
class AllElementsReference {
  @NotNull
  Expr root;
}
```

Anyway, you *could* scrap the use of `ExprPath` and `ExprPathStep`, and builders would help with code-readability.

```kotlin
// SELECT * FROM t WHERE t['a'].b[0] > 2
val query = QueryBuilder()
    .selectStar()
    .from("t")
    .where(ExprBuilder().id("t").elementRef("a").fieldRef("b").elementRef(0).gt(2))
    .build()
```

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprBetween.java
```diff
@@ -0,0 +1,48 @@
...
+    public final Expr value;
+
+    @NotNull
+    public final Expr from;
+
+    @NotNull
+    public final Expr to;
+
+    @NotNull
+    public final Boolean not;
```
johnedquinn, line 25:
```suggestion
    public final boolean not;
```

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprIon.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.Collections;
+import java.util.Collection;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public class ExprIon extends Expr {
+    @NotNull
+    public IonElement value;
```
johnedquinn, line 16: This might actually be a good opportunity to just pass this as a string since this is just the AST. This could remove the only 3rd party dependency we have. Since it wouldn't be in the public API for the modeling, we could shadow IonReader in the evaluator to allow our customers to use whatever version of Ion they want in their own libraries.

partiql-ast/src/main/java/org/partiql/ast/v1/type/TypeCustom.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public class TypeCustom extends Type {
+    @NotNull
+    public String name;
```
johnedquinn, line 15:
```suggestion
/**
 * TODO docs, equals, hashcode
 * <data type> ::=
 *    <predefined type>
 *    | <row type>
 *    | <user-defined type>
 *    | ...
 * <user-defined type> ::= <user-defined type name>
 * <user-defined type name> ::= <schema qualified type name>
 */
public class TypeUserDefined extends Type {
    @NotNull
    public IdentifierChain name;
```

partiql-ast/src/main/java/org/partiql/ast/v1/type/TypeFloat32.java
```diff
@@ -0,0 +1,24 @@
...
+import org.partiql.ast.v1.AstNode;
+import org.partiql.ast.v1.AstVisitor;
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public class TypeFloat32 extends Type {
```
johnedquinn, line 13:
```suggestion
public class TypeReal extends Type {
```

partiql-ast/src/main/java/org/partiql/ast/v1/type/TypeFloat64.java
```diff
@@ -0,0 +1,24 @@
...
+import org.partiql.ast.v1.AstNode;
+import org.partiql.ast.v1.AstVisitor;
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public class TypeFloat64 extends Type {
```
johnedquinn, line 13:
```suggestion
public class TypeDoublePrecision extends Type {
```

partiql-ast/src/main/java/org/partiql/ast/v1/type/TypeVarchar.java
```diff
@@ -0,0 +1,32 @@
...
+import org.partiql.ast.v1.AstNode;
+import org.partiql.ast.v1.AstVisitor;
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public class TypeVarchar extends Type {
```
johnedquinn, line 14: Just pointing out the `VARCHAR(l)` and `CHARACTER VARYING(l)` are represented by `TypeVarchar` -- but `INT4`, `INTEGER`, and `INT` are represented by two different types: `TypeInt` and `TypeInt4`.

partiql-ast/src/main/java/org/partiql/ast/v1/type/TypeByteString.java
```diff
@@ -0,0 +1,32 @@
...
+import org.partiql.ast.v1.AstNode;
+import org.partiql.ast.v1.AstVisitor;
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public class TypeByteString extends Type {
```
johnedquinn, line 14: Neither SQL nor Ion has `BYTE`. We can remove this for now.

partiql-ast/src/main/java/org/partiql/ast/v1/type/TypeAny.java
```diff
@@ -0,0 +1,24 @@
...
+import org.partiql.ast.v1.AstNode;
+import org.partiql.ast.v1.AstVisitor;
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public class TypeAny extends Type {
```
johnedquinn, line 13: I'd defer to @jpschorr, but this may be a good opportunity for us to use the `TypeDynamic` naming since these changes are breaking anyways.

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprOverlay.java
```diff
@@ -0,0 +1,49 @@
...
+    @NotNull
+    public Expr value;
+
+    @NotNull
+    public Expr placing;
+
+    @NotNull
+    public Expr from;
+
+    @NotNull
```
johnedquinn, line 24:
```suggestion
    @Nullable
```

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprNullIf.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public class ExprNullIf extends Expr {
+    @NotNull
+    public Expr value;
+
+    @NotNull
+    public Expr nullifier;
```
johnedquinn, line 19: You don't need to update this, but do you know the history of the name `nullifier`? I've always been confused by it:
> NULLIF (V1 , V2 ) is equivalent to the following <case specification>:
> CASE WHEN V1=V2 THEN NULL ELSE V1 END

I would personally just use `lhs` and `rhs`. I might just be missing something.

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprLit.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public class ExprLit extends Expr {
+    @NotNull
+    public PartiQLValue value;
```
johnedquinn, line 16: We've discussed this earlier, but would you like to create the literal AST representations as part of this PR? AKA:
```java
public class LiteralNumericExact extends Expr { ... }
public class LiteralNumericExactPeriod extends Expr { ... }
public class LiteralNumericApprox extends Expr { ... }
```

For:
```ebnf
<unsigned numeric literal> ::=
  <exact numeric literal>
  | <approximate numeric literal>;

<exact numeric literal> ::=
  <unsigned integer> [ <period> [ <unsigned integer> ] ]
  | <period> <unsigned integer>;

<approximate numeric literal> ::= <mantissa> E <exponent>
```

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprIsType.java
```diff
@@ -0,0 +1,45 @@
...
+import org.partiql.ast.v1.type.Type;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public class ExprIsType extends Expr {
```
johnedquinn, line 16: Do we need this?

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprDateDiff.java
```diff
@@ -0,0 +1,44 @@
...
+import org.partiql.ast.v1.DatetimeField;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public class ExprDateDiff extends Expr {
```
johnedquinn, line 15: Really hope we don't *need* to add this to the AST. Same goes for `DATE_ADD`. I wonder if there's a way we can rewrite this to be `+/-` for datetimes and intervals.

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprCollection.java
```diff
@@ -0,0 +1,48 @@
...
+import org.partiql.ast.v1.AstVisitor;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public class ExprCollection extends Expr {
```
johnedquinn, line 14: So that we don't have to switch on the type, we could just create a `BagConstructor` and `ArrayConstructor`?

johnedquinn (commented):
Thank you for this! I've gone through most of the files, though several remain. I can re-review once some modifications have been made.

partiql-ast/src/main/java/org/partiql/ast/v1/CaseSensitivity.java
alancai98: Thanks for the dive-deep into enum compatibility. Agree w/ the points that Java's closed enums are an issue and shouldn't be relied upon in our public APIs.

I'm open to using integers in some form. Probably will discuss w/ you offline further on the general API we should use:
- interfaces?
- classes w/ constants?
- do we still need an unknown/other variant?

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprVar.java
alancai98: I'll change to `ExprVarRef` to be a bit more descriptive

partiql-ast/src/main/java/org/partiql/ast/v1/Identifier.java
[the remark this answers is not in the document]
alancai98: Thanks for the suggestion. I've been confused at times by the previous names, `Symbol` and `Qualified`, so open to changing them.

I might need some clarity on the suggestion. Would `Identifier` and `IdentifierChain` be within some abstract class or be separate `AstNode`s entirely? I believe you're referring to separate `AstNode`s without some overarching abstract class but just checking before I refactor further.

partiql-ast/src/main/java/org/partiql/ast/v1/Identifier.java
```diff
@@ -0,0 +1,80 @@
...
+        @NotNull
+        public String symbol;
+
+        @NotNull
+        public CaseSensitivity caseSensitivity;
```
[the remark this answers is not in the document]
alancai98: Changed.

alancai98 (commented):
Made some changes based off John's initial PR review. Still some points for discussion. After we decide those points, I'll make further changes.

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprDateDiff.java
alancai98: From offline discussion: we'll remove `DATE_ADD` and `DATE_DIFF` from the AST with the eventual plan to support the `+/-` syntax with intervals.

To support customers dependent on these existing functions and special keyword syntax, we will still support parsing the special syntax but
- convert the datetime keyword (e.g. `month`, `day`) into a string or
- map to different functions for each variant (e.g. `date_add_month`, `date_diff_day`)

`DATE_ADD` and `DATE_DIFF` will still have support in the plan and evaluator.

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprCollection.java
alancai98: From offline discussion, these other variants will be removed, leaving just bag and array constructors. We will split these into separate classes.

partiql-ast/src/main/java/org/partiql/ast/v1/CaseSensitivity.java
alancai98: From offline discussion, chose to define an internal `Enum` interface within the `partiql-ast` package. Enums we which to add will be classes that implement this interface with a default constant of `UNKNOWN` set to 0.

There will be a private constructor with public functions to create.

E.g. for `Order`:

```Java
public class Order implements Enum {
    public static final int UNKNOWN = 0;
    public static final int ASC = 1;
    public static final int DESC = 2;

    public static Order UNKNOWN() {
        return new Order(UNKNOWN);
    }

    public static Order ASC() {
        return new Order(ASC);
    }

    public static Order DESC() {
        return new Order(DESC);
    }

    private final int code;

    private Order(int code) {
        this.code = code;
    }

    @Override
    public int code() {
        return code;
    }
}
```

And example client code calling and using this class:
```Java
Order order = Order.ASC();
switch (order.code()) {
    case Order.ASC:
        System.out.println("Ascending order");
        break;
    case Order.DESC:
        System.out.println("Descending order");
        break;
    default:
        System.out.println("Unknown order");
}
```

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprPathStep.java
alancai98: From offline discussion, will model w/ a root and steps but will model the steps using a linked list rather than a java collection list. Also I'm coming around to the naming you had suggested for the path steps. I made some slight adjustments to be a less verbose (i.e. removing `Reference`).

alancai98 (commented):
Latest revision
- Models the enums as classes implementing the `Enum` partiql-ast interface
- Refactors `Type` to be `DataType`, which will also implement `Enum`
- Refactors a few other classes -- `Identifier`, `PathStep`, `NullIf`, `ExprCollection`

partiql-ast/src/main/java/org/partiql/ast/v1/Identifier.java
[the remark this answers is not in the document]
alancai98: Changed to use `Identifier` and `IdentifierChain` as separate nodes.

partiql-ast/src/main/java/org/partiql/ast/v1/AstNode.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.Collection;
+import java.util.Random;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public abstract class AstNode {
+    @NotNull
+    public String tag = "Ast-" + String.format("%06x", new Random().nextInt());
+
```
RCHowell, line 14: I've created a follow-up issue for including source locations. https://github.com/partiql/partiql-lang-kotlin/issues/1608

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprLit.java
RCHowell: +1 to as a follow-up for 1589

partiql-ast/src/main/java/org/partiql/ast/v1/AstVisitor.java
```diff
@@ -0,0 +1,240 @@
...
+
+    R visitStatement(Statement node, C ctx);
+
+    R visitQuery(Query node, C ctx);
+
+    R visitExplain(Explain node, C ctx);
+
+    R visitTarget(Target node, C ctx);
+
+    R visitTargetDomain(Target.Domain node, C ctx);
```
RCHowell, line 54: I advise keeping the adjectives on target/domain because these are common terms in the SQL spec (>119 occurrences of "target" in different contexts).

And after looking more closely into the EXPLAIN grammar, I don't think target and domain are even necessary. Consider simplifying to.

```
public class Explain extends Statement {
    @NotNull
    public Map<String, PartiQLValue> options;

    @NotNull
    public Statement statement;

    // ....
}
```

Because the EXPLAIN grammar is _very_ simple and doesn't need three separate nodes. Yes, the PartiQLValue will be replaced in the follow-up. The grammar currently only supports identifier/string literals in the explain options, so this AST representation would cover things like numeric literals etc.

partiql-ast/src/main/java/org/partiql/ast/v1/AstVisitor.java
```diff
@@ -0,0 +1,240 @@
...
+
+    R visitSelectStar(SelectStar node, C ctx);
+
+    R visitSelectProject(SelectProject node, C ctx);
+
+    R visitProjectItem(ProjectItem node, C ctx);
+
+    R visitProjectItemAll(ProjectItem.All node, C ctx);
+
+    R visitProjectItemExpression(ProjectItem.Expression node, C ctx);
```
RCHowell, line 158: nit: ProjectItemExpr

only because we don't use the full "Expression" elsewhere afaik

partiql-ast/src/main/java/org/partiql/ast/v1/AstVisitor.java
```diff
@@ -0,0 +1,240 @@
...
+
+    R visitExcludeStepCollIndex(ExcludeStep.CollIndex node, C ctx);
+
+    R visitExcludeStepStructWildcard(ExcludeStep.StructWildcard node, C ctx);
+
+    R visitExcludeStepCollWildcard(ExcludeStep.CollWildcard node, C ctx);
+
+    R visitFrom(From node, C ctx);
+
+    R visitFromValue(FromValue node, C ctx);
```
RCHowell, line 180: nit: FromExpr

I am to blame that this is not FromExpr, but FromValue does not _really_ make sense .. so just a nit.

partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
```diff
@@ -0,0 +1,318 @@
...
+    public static final int FLOAT = 18;
+    public static final int DECIMAL = 19;
+    public static final int DEC = 20;
+    public static final int NUMERIC = 21;
+    public static final int BIT = 22; // TODO not defined in parser yet
+    public static final int BIT_VARYING = 23; // TODO not defined in parser yet
+    public static final int CHAR = 24;
+    public static final int CHARACTER = 25;
+    public static final int VARCHAR = 26;
+    public static final int CHARACTER_VARYING = 27;
```
RCHowell, line 31: Let's group things together based on SQL standard definitions. 

For example,
```java
// <character string type>
public static final int CHARACTER = 24;
public static final int CHAR = 25;
public static final int CHARACTER_VARYING = 26;
public static final int CHAR_VARYING = 27;
public static final int VARCHAR = 28;
public static final int CHARACTER_LARGE_OBJECT = 29;
public static final int CHAR_LARGE_OBJECT = 30;
public static final int CLOB = 31; 
```

Which would make implementations of some function `isCharacterString()` easy/efficient to answer aka code in range.

This will also help us cover all SQL things first, then we can have a separate codespace for extended type syntax. I would skip `<national character string type>`..

partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
```diff
@@ -0,0 +1,318 @@
...
+        return new DataType(TIMESTAMP_WITH_TIME_ZONE, precision, null, null);
+    }
+
+    public static DataType INTERVAL() {
+        return new DataType(INTERVAL);
+    }
+
+    public static DataType USER_DEFINED() {
+        return new DataType(USER_DEFINED);
+    }
```
RCHowell, line 291: We need the identifier of a user-defined type. The code is perfect, but there should be a `private final Identifier name` which is only NON-null for the UDT

partiql-ast/src/main/java/org/partiql/ast/v1/AstNode.java
alancai98: Good point on the source locations -- I'll link the issue.

partiql-ast/src/main/java/org/partiql/ast/v1/AstVisitor.java
alancai98: Yeah, I was just going off the old partiql-ast.ion definition. This can be simplified a bit. I'll follow your suggestion and add a TODO about getting rid of PartiQLValue.

partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
alancai98: Good idea. I wasn't too sure on the best way to group the constants. Will try to follow SQL's grouping/ordering.

Will also skip `<national character string type>`

partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
alancai98: Ah I missed that. I'll add a new private constructor for just the user-defined type `name`.

partiql-ast/src/main/java/org/partiql/ast/v1/SelectProject.java
```diff
@@ -0,0 +1,35 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public class SelectProject extends Select {
```
RCHowell, line 13: SelectList

partiql-ast/src/main/java/org/partiql/ast/v1/ProjectItem.java
```diff
@@ -0,0 +1,81 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public abstract class ProjectItem extends AstNode {
```
RCHowell, line 13: SelectItem.Expr
SelectItem.Star

RCHowell (changes requested):

partiql-ast/src/main/java/org/partiql/ast/v1/ProjectItem.java
alancai98: ack -- applied this and the other `SELECT` node renaming.

partiql-ast/src/main/java/org/partiql/ast/v1/AstVisitor.java
```diff
@@ -0,0 +1,236 @@
...
+
+    R visitSelectStar(SelectStar node, C ctx);
+
+    R visitSelectProject(SelectList node, C ctx);
+
+    R visitProjectItem(SelectItem node, C ctx);
+
+    R visitProjectItemAll(SelectItem.Star node, C ctx);
+
+    R visitProjectItemExpr(SelectItem.Expr node, C ctx);
```
RCHowell, line 154: update visitor methods to match the types

RCHowell (approved):

partiql-ast/src/main/java/org/partiql/ast/v1/expr/ExprCall.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public class ExprCall extends Expr {
+    public IdentifierChain function;
+    public List<Expr> args;
+    public SetQuantifier setq;
```
johnedquinn, line 19:
```suggestion
    @NotNull
    public IdentifierChain function;
    @NotNull
    public List<Expr> args;
    @Nullable
    public SetQuantifier setq;
```

johnedquinn (commented):
If we want to enforce immutable structures, we could write getters (no setters). Or use final. Or, in the javadocs state that properties should not be modified.

partiql-ast/src/main/java/org/partiql/ast/v1/FromValue.java
```diff
@@ -0,0 +1,57 @@
...
+import org.partiql.ast.v1.expr.Expr;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public class FromValue extends From {
```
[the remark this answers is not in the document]
alancai98: I see what you're saying. The existing modeling for `FROM` clause would "eagerly" convert the comma-separated table references into a join, when it should really be a different table reference. Following the SQL:1999 modeling may let us add more SQL:1999 features without breaking the AST, so I'll revise.

partiql-ast/src/main/java/org/partiql/ast/v1/FromValue.java
```diff
@@ -0,0 +1,57 @@
...
+    @Nullable
+    public Identifier asAlias;
+
+    @Nullable
+    public Identifier atAlias;
+
+    /**
+     * TODO get rid of `BY`
+     */
+    @Nullable
```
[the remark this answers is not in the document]
alancai98: I'll just delete `BY`

partiql-ast/src/main/java/org/partiql/ast/v1/JoinType.java
```diff
@@ -0,0 +1,68 @@
...
+    public static final int UNKNOWN = 0;
+    public static final int INNER = 1;
+    public static final int LEFT = 2;
+    public static final int LEFT_OUTER = 3;
+    public static final int RIGHT = 4;
+    public static final int RIGHT_OUTER = 5;
+    public static final int FULL = 6;
+    public static final int FULL_OUTER = 7;
+    public static final int CROSS = 8;
+    public static final int COMMA = 9;
```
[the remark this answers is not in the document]
alancai98: Yeah, removed.

partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
```diff
@@ -0,0 +1,382 @@
...
+        this.name = name;
+        this.precision = null;
+        this.scale = null;
+        this.length = null;
+    }
+
+    @Override
+    public int code() {
+        return code;
+    }
```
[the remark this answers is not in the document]
alancai98: Ah, my bad. Forgot to add those getters in this revision. Will add.

alancai98:
> If we want to enforce immutable structures, we could write getters (no setters). Or use final. Or, in the javadocs state that properties should not be modified.

I'll save the discussion for mutable vs immutable fields for another PR/discussion. I like immutable structures for most purposes. However I've dealt with some annoying patterns when I needed to modify parts of the AST which is a common use case for tree rewriting.

alancai98:
Hm not sure why the conformance test commenting is publishing a new comment for every commit rather than editing an existing comment. Deleting the stale comments for now. I'll investigate further if this happens for other PRs.

PR #1580, opened 2024-09-14
Add dml update to partiql_logical pig domain and improve DML modeling

dlurton:
## Description

Heretofore the `partiql_logical` and later domains didn't include support for DML `UPDATE`.  Also, I felt that the way DML was modeled in that domain was still a bit awkward.   My team is looking to support a limited form of `UPDATE/SET/WHERE` in the very near future so I've decided to just do it myself rather than waiting on the usual channels.  When this is merged, we'll need it released at the earliest convenience.

## Other Information
   - No--this will have to be done in a future commit to this same PR.

  - Yes, but AFAIK my team is the only customer of the APIs I changed.  If there are issues with this we'll have to resolve them during review.
  - In particular, the changes in `partiql.ion` and to `PartiQLResult` are not backward compatible.

  - no
 
## License Information


dlurton:
Yes, `ktlint` is not happy.  On my machine: `./gradlew ktlintCheck` and `./gradlew ktlintMainSourceSetCheck`  succeed so I have no idea why it's failing here.  `./gradlew ktlintFormat` fails but doesn't seem to provide any reason why.  Of course this will have to be fixed before the merge.

partiql-lang/src/main/kotlin/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransform.kt
```diff
@@ -368,37 +369,23 @@ internal class AstToLogicalVisitorTransform(
...
+                    dmlInsert(
+                        target = target,
+                        targetAlias = alias,
+                        rowsToInsert = transformExpr(dmlOp.values),
+                        metas = node.metas,
+                        onConflict = transformConflictAction(dmlOp.conflictAction)
+                    )
+                }
             }
+            // DL TODO: transformDmlUpdate
```
sadderchris, line 387: DL TODO, does this still need to be fleshed out here?

sadderchris: Looks like this is handled by the DmlOp.Set branch below if I'm reading this right, I think this comment can just be removed

partiql-lang/src/test/kotlin/org/partiql/lang/planner/transforms/AstToLogicalVisitorTransformTests.kt
```diff
@@ -1117,8 +1310,6 @@ class AstToLogicalVisitorTransformTests {
 
             // Unimplemented parts of DML
             ProblemTestCase(200, "FROM x AS xx INSERT INTO foo VALUES (1, 2)", unimplementedProblem("UPDATE / INSERT", 1, 14, 6)),
-            ProblemTestCase(201, "FROM x AS xx SET k = 5", unimplementedProblem("SET", 1, 14, 3)),
```
sadderchris, line 1120: I don't see any tests for this syntax.  I don't think we're planning on relying on it, but it should probably be covered if it's no longer unimplemented

partiql-ast/src/main/pig/partiql.ion
```diff
@@ -802,55 +802,87 @@ may then be further optimized by selecting better implementations of each operat
...
-                //  since dml_update is currently shared by UPSERT (which does not allow the aliasing of rows) and
-                //  INSERT ... ON CONFLICT DO UPDATE ... (which aliases the rows as "EXCLUDED" for use within the [condition]).
-                (dml_update target_alias::var_decl condition::(? expr) row_alias::(? var_decl))
-            )
-        )
-
         // Redefine statement.dml to be a simpler subset of the full DML functionality expressed with PartiQL's DML
         // syntax.  Full functionality is out of scope for now.  This is factored minimally to support
         // `INSERT INTO` and `DELETE FROM ... [WHERE <predicate>]` but this may need refactoring when
         // `FROM ... UPDATE` and `UPDATE` is supported later.
```
sadderchris, line 808: Comments probably need to be refreshed here

PR #1584, opened 2024-09-16
[1/2] Resolve functions from the catalog

RCHowell:
## Relevant Issues

#1496 

## Description

This Part 1 of 2 – the second part will remove SqlFnProvider entirely.

This PR removes direct calls to the SqlFnProvider for lookup ONLY as a piecewise step towards implementing the PartiQL-PATH in which we search the session PATH for the function names. The implementations are still provided via lookup by specific.

**Names**

* _Routines_ are the what SQL describes as "callable" objects.
* _Function_ is an interface for scalar routines.
* _Aggregation_ is an interface for aggregation routines.
* _Procedure_ does NOT exist yet, but could be added later, as it is a type of routine defined by SQL.

The following changes are incremental,

* The Env now searches for functions/aggregations via the Catalog.
* Resolution/matching is still dependent upon FnSignature
* Removing FnSignature will be a BIG diff, so I am setting everything up beforehand.

I have a follow-up issue for recent discussion points, https://github.com/partiql/partiql-lang-kotlin/issues/1585.

## Other Information

## License Information


partiql-spi/src/main/kotlin/org/partiql/spi/fn/Routine.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * The function return type. Required.
+     */
+    public fun getReturnType(): PType
+
+    /**
+     * TODO REMOVE ME ??
+     */
+    public fun getSpecific(): String
```
alancai98, line 28: Will this be kept for `v1` or is it just used as part of the refactoring process?

alancai98 (approved):
Just a small nit and question. Refactor looks good otherwise

partiql-spi/src/main/kotlin/org/partiql/spi/fn/Routine.kt
RCHowell: Part of refactoring, removed in next PR

PR #1586, opened 2024-09-16
[0.14.9-SNAPSHOT] Adds warnings for `EXCLUDE` paths that don't exclude any value

alancai98:
## Relevant Issues
- Related to https://github.com/partiql/partiql-lang/issues/91

## Description
Gives a plan-time **warning** when, in the presence of closed schema, and `EXCLUDE` path does not exclude a value.

---

#### Examples:
Attribute not in struct
```
SELECT * EXCLUDE t.attr_does_not_exist FROM << { 'a': 1 } >> AS t;
```

^ will result in the error message 

```
Exclude path 't.attr_does_not_exist' does not exclude any values
```

---
Exclude collection wildcard step on a struct

```
SELECT * EXCLUDE t.a[*]
FROM <<{ 'a': { 'b': { 'c': 0, 'd': 'foo' } } } >> AS t;
```

^ will result in the error message

```
Exclude path 't.a[*]' does not exclude any values
```
---
## Other Information

## License Information

### DIFF
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/ExcludeUtils.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/ExcludeUtils.kt
@@ -13,105 +13,92 @@ import org.partiql.types.CollectionType
 import org.partiql.types.StaticType
 import org.partiql.types.StructType
 
-/**
- * Checks for every exclude path in [paths], that a value in the [bindings] is excluded. If there is no value excluded,
- * a warning is added to the [onProblem] callback.
- */
-internal fun checkForInvalidExcludePaths(bindings: List<Rel.Binding>, paths: List<Rel.Op.Exclude.Item>, onProblem: ProblemCallback) {
-    paths.forEach { excludePath ->
-        val root = excludePath.root
-        val steps = excludePath.steps
-        val excludesSomething = bindings.any { binding ->
-            when (root) {
-                is Rex.Op.Var.Unresolved -> {
-                    when (val id = root.identifier) {
-                        is Identifier.Symbol -> {
-                            if (id.isEquivalentTo(binding.name)) {
-                                binding.type.checkExclude(steps)
-                            } else {
-                                false
+internal object ExcludeUtils {
+    /**
+     * Checks for every exclude path in [paths], that a value in the [bindings] is excluded. If there is no value excluded,
+     * a warning is added to the [onProblem] callback.
+     */
+    internal fun checkForInvalidExcludePaths(bindings: List<Rel.Binding>, paths: List<Rel.Op.Exclude.Item>, onProblem: ProblemCallback) {
+        paths.forEach { excludePath ->
+            val root = excludePath.root
+            val steps = excludePath.steps
+            val excludesSomething = bindings.any { binding ->
+                when (root) {
+                    is Rex.Op.Var.Unresolved -> {
+                        when (val id = root.identifier) {
+                            is Identifier.Symbol -> {
+                                if (id.isEquivalentTo(binding.name)) {
+                                    binding.type.checkExclude(steps)
+                                } else {
+                                    false
+                                }
                             }
-                        }
-                        is Identifier.Qualified -> {
-                            if (id.root.isEquivalentTo(binding.name)) {
-                                binding.type.checkExclude(steps)
-                            } else {
-                                false
+                            is Identifier.Qualified -> {
+                                if (id.root.isEquivalentTo(binding.name)) {
+                                    binding.type.checkExclude(steps)
+                                } else {
+                                    false
+                                }
                             }
                         }
                     }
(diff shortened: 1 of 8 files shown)


partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/ExcludeUtils.kt
```diff
@@ -0,0 +1,159 @@
...
+import org.partiql.types.AnyOfType
+import org.partiql.types.CollectionType
+import org.partiql.types.StaticType
+import org.partiql.types.StructType
+
+/**
+ * Checks for every exclude path in [paths], that a value in the [bindings] is excluded. If there is no value excluded,
+ * a warning is added to the [onProblem] callback.
+ */
+internal fun checkForInvalidExcludePaths(bindings: List<Rel.Binding>, paths: List<Rel.Op.Exclude.Item>, onProblem: ProblemCallback) {
```
RCHowell, line 20: Could you please move these inside of an internal object "ExcludeUtils"

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/ExcludeUtils.kt
```diff
@@ -0,0 +1,159 @@
...
+                        is Identifier.Qualified -> {
+                            if (id.root.isEquivalentTo(binding.name)) {
+                                binding.type.checkExclude(steps)
+                            } else {
+                                false
+                            }
+                        }
+                    }
+                }
+                is Rex.Op.Var.Resolved -> false
```
RCHowell, line 44: Why do we return false here?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/ExcludeUtils.kt
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * Checks whether [steps] will exclude a value from [this] [StructType].
+ */
+internal fun StructType.checkExclude(steps: List<Rel.Op.Exclude.Step>): Boolean {
+    // Ignore open structs
+    if (!this.contentClosed) {
+        return true
+    }
+    val step = steps.first()
```
RCHowell, line 81: Check step size before calling first. We don't want this to blow up if the steps list is empty (just in case) even though it _should_ never be empty.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/ExcludeUtils.kt
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+}
+
+/**
+ * Checks whether [steps] will exclude a value from [this] [CollectionType].
+ */
+internal fun CollectionType.checkExclude(steps: List<Rel.Op.Exclude.Step>): Boolean {
+    val step = steps.first()
+    return if (steps.size == 1) {
```
RCHowell, line 108: Same as before, check size so we don't call .first() on a empty list and get an exception.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/ExcludeUtils.kt
```diff
@@ -0,0 +1,159 @@
...
+    } else {
+        when (step) {
+            is Rel.Op.Exclude.Step.CollIndex, is Rel.Op.Exclude.Step.CollWildcard -> {
+                val e = this.elementType
+                e.checkExclude(steps.drop(1))
+            }
+            else -> false
+        }
+    }
+}
```
RCHowell, line 122: You could actually simplify these to something like

```kotlin
if (steps.isEmpty()) {
    return false
}
when (steps.first()) {
            is Rel.Op.Exclude.Step.CollIndex, is Rel.Op.Exclude.Step.CollWildcard -> {
                val e = this.elementType
                e.checkExclude(steps.drop(1))
            }
            else -> false
        }
}
```

I don't believe you need the if size thing because steps.drop(1) will return an empty list and then your recursive base case hits.

RCHowell (changes requested):

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/ExcludeUtils.kt
alancai98: Thanks for the suggestion. I simplified the `checkExclude` for structs and collection types based on your suggestion in the latest commit.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/ExcludeUtils.kt
alancai98: When this check is run in `PlanTyper`, the root for each exclude path should be an unresolved var, so this case should not occur. The value returned here shouldn't matter.

RCHowell (approved):

PR #1588, opened 2024-09-19
Removes unused SPI APIs and makes connector instantiate catalogs

RCHowell:
## Relevant Issues

#1496 

## Description

This PR is a cleanup of extra and unused SPI APIs, and it is the final modifications to the connector/catalog components of the SPI. We have effectively moved from Trino's model of plugins+connector to a "catalog" interface like Iceberg and Calcite.

## Other Information

## License Information


partiql-spi/src/main/kotlin/org/partiql/spi/Connector.kt
```diff
@@ -0,0 +1,48 @@
...
+     */
+    public interface Context
+
+    /**
+     * Get (or instantiate) a [Catalog] with the given name.
+     *
+     * @param name      Catalog name
+     * @return
+     */
+    public fun getCatalog(name: String): Catalog
```
johnedquinn, line 23: Will this ever be used?

partiql-spi/src/main/kotlin/org/partiql/spi/Connector.kt
```diff
@@ -0,0 +1,48 @@
...
+     * private fun getCatalog(context: MyContext): Catalog {
+     *      // ... get catalog from context
+     * }
+     * ```
+     *
+     * @param name      Catalog name
+     * @param context   Context is an arbitrary object
+     * @return
+     */
+    public fun getCatalog(name: String, context: Context): Catalog
```
johnedquinn, line 47: Given that a name is carried by the `Catalog` itself, it might not be necessary to pass it here (or not necessary to keep it in the `Catalog`). It may present the opportunity for mismatching names.

RCHowell: Think of this being looking up a catalog implementation (or instantiating if it does not exist). I see this as no different than the Catalog's `getTable(name: Name): Table` which also presents the opportunity for mismatching names.

Now .. that would be a bad implementation which is kind of like returning the wrong table given the name. We can't stop a customer from implementing something incorrect.

johnedquinn (approved):

Issue #1589, opened 2024-09-20
[V1] Remove PartiQLValue from the AST


RCHowell:
Some AST cleanup work,

* The AST should have its own representation of values independent of PartiQLValue (to be deprecated) and Datum.
* Similarly, the AST should not necessarily be dependent on partiql-types/partiql-spi packages (but ok if more practical).

To the first point, notice how AST types are independent of PType as there are many syntaxes for a single type e.g. BOOL|BOOLEAN.

This PR would require defining AST expr literal variants for the parser rules

```antlr4
literal
    : NULL                                                                                # LiteralNull
    | MISSING                                                                             # LiteralMissing
    | TRUE                                                                                # LiteralTrue
    | FALSE                                                                               # LiteralFalse
    | LITERAL_STRING                                                                      # LiteralString
    | LITERAL_INTEGER                                                                     # LiteralInteger
    | LITERAL_DECIMAL                                                                     # LiteralDecimal
    | ION_CLOSURE                                                                         # LiteralIon
    | DATE LITERAL_STRING                                                                 # LiteralDate
    | TIME ( PAREN_LEFT LITERAL_INTEGER PAREN_RIGHT )? (WITH TIME ZONE)? LITERAL_STRING   # LiteralTime
    | TIMESTAMP ( PAREN_LEFT LITERAL_INTEGER PAREN_RIGHT )? (WITH TIME ZONE)? LITERAL_STRING   # LiteralTimestamp
    ;
``` 

Along with updating the pretty-printer in `org.partiql.ast.sql`

alancai98:
Closed by #1650

PR #1591, opened 2024-09-20
Generalizes back-tick Ion to a variant pair in the AST

RCHowell:
## Relevant Issues

- https://github.com/orgs/partiql/discussions/92
- https://github.com/partiql/partiql-lang-kotlin/tree/v1-no-pv
- #1589 

## Description

This PR removes eager-parsing of Ion in the AST (ONLY) and introduces the parameterized VARIANT type. I am working on a subsequent[ PR to remove PartiQLValue from the planner ](https://github.com/partiql/partiql-lang-kotlin/tree/v1-no-pv)which will eliminate eager-parsing of Ion entirely.

This PR is limited to only the necessary AST changes and does not modify the grammar or other components.

## Other Information

## License Information

### DIFF
--- a/partiql-ast/src/main/kotlin/org/partiql/ast/sql/SqlDialect.kt
+++ b/partiql-ast/src/main/kotlin/org/partiql/ast/sql/SqlDialect.kt
@@ -224,9 +224,11 @@ public abstract class SqlDialect : AstBaseVisitor<SqlBlock, SqlBlock>() {
         return head concat r(value)
     }
 
-    override fun visitExprIon(node: Expr.Ion, head: SqlBlock): SqlBlock {
-        // simplified Ion value writing, as this intentionally omits formatting
-        val value = node.value.toString()
+    override fun visitExprVariant(node: Expr.Variant, head: SqlBlock): SqlBlock {
+        if (node.encoding != "ion") {
+            error("Unsupported encoding ${node.encoding}")
+        }
+        val value = node.value
         return head concat r("`$value`")
     }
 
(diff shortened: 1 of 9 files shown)


partiql-ast/src/main/kotlin/org/partiql/ast/sql/SqlDialect.kt
```diff
@@ -224,9 +224,11 @@ public abstract class SqlDialect : AstBaseVisitor<SqlBlock, SqlBlock>() {
         return head concat r(value)
     }
 
-    override fun visitExprIon(node: Expr.Ion, head: SqlBlock): SqlBlock {
-        // simplified Ion value writing, as this intentionally omits formatting
-        val value = node.value.toString()
+    override fun visitExprVariant(node: Expr.Variant, head: SqlBlock): SqlBlock {
+        if (node.encoding != "ion") {
```
am357, line 228: Maybe having the encoding as an `enum` is a better option.

am357 (approved):

partiql-ast/src/main/kotlin/org/partiql/ast/sql/SqlDialect.kt
RCHowell: Agreed, however I'm not sure the best way to make it extensible/extendable

RCHowell: I was thinking of it like the "content-type" header

am357: I see thinking of it same as "content-type" make sense if we are directly working on the serialized payload. But if the node is an in-memory (which can be deserialized from the payload), we may benefit from Enum.

Issue #1585, opened 2024-09-16
[V1] Function and aggregation improvements for typing and implementations.


RCHowell:
A couple of function things..

* Functions (and aggregations) should be typed on their families rather than specific PTypes.
* FnSignature (and AggSignature) should be removed in favor of the interface methods.
* Functions (and aggregations) should have an _implementation provider_ rather than a single invoke method.

```kotlin
    /**
     * TODO REPLACE ME WITH `getInstance(args: Array<PType>)` which returns an invocable instance of this function.
     *
     * Invoke the function with the given arguments. Required.
     *
     * @param args the arguments to the function
     * @return the result of the function
     */
    public fun invoke(args: Array<Datum>): Datum
```

PR #1598, opened 2024-09-26
[1/2] Function modeling as instance providers and remove signatures

RCHowell:
## Relevant Issues

#1585 

## Description

This PR is Part 1 of 2. It moves away from single static functions to having the Function interface hold the typing logical along with being an instance provider.

Why are we doing this? Accurate SQL typing requires knowing the _exact_ types of input arguments i.e. what is the precision and scale of the argument decimals. SQL defines this by "enumerating" all permutations and defining a static function for each permutation. In practice, this would mean defining at least 38^2 functions for each function that has a decimal argument. 

This implementation is how we get around that. The function interface is where we implement the computation of a return type. We intend to follow[ SQL Server as it is well documented](https://learn.microsoft.com/en-us/sql/t-sql/data-types/precision-scale-and-length-transact-sql?view=sql-server-ver16). This implementation is inspired by Calcite which types operators on families rather than static types.

This is Part 1 of 2 because I've simply transition all functions to a "static" implementation of the Function interface. I'm not sure if "static" is the right word, but it effectively a singleton-implementation. The second part will introduce type families for the operators.

## Other Information

## License Information


partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
```diff
@@ -148,33 +168,32 @@ internal class ExprCallDynamic(
...
-
-        fun eval(originalArgs: Array<Datum>): Datum {
-            val args = originalArgs.mapIndexed { i, arg ->
-                if (arg.isNull && fn.signature.isNullCall) {
-                    return nil.invoke()
-                }
+        fun eval(args: Array<Datum>): Datum {
+            val coerced = Array(args.size) { i ->
+                val arg = args[i]
+                if (function.isNullCall && arg.isNull) nil.invoke()
```
johnedquinn, line 188: This needs a return

partiql-eval/src/main/kotlin/org/partiql/eval/internal/SqlCompiler.kt
```diff
@@ -348,46 +348,45 @@ internal class SqlCompiler(
...
-            return when (fnTakesInMissing) {
-                true -> ExprCallStatic(fn, args.map { it.catch() }.toTypedArray())
-                else -> ExprCallStatic(fn, args.toTypedArray())
-            }
+        override fun visitCall(rex: RexCall, ctx: Unit): Operator.Expr {
+            val function = rex.getFunction()
+            val args = rex.getArgs()
+            // TODO HANDLE CALLS WITH DYNAMIC INPUT
+            // val fnTakesInMissing = fn.getParameters().any {
+            //     it.getType().kind == PType.Kind.DYNAMIC // TODO: Is this needed?
```
johnedquinn, line 383: Yes, I believe this is needed. If a function allows missing-valued arguments, then, in permissive mode, we need to be able to appropriately handle type exceptions that result in MISSING.

A scenario to consider is the equals function. In its function signature, it accepts dynamic. If run in permissive mode, we should be able to execute:
```partiql
{ 'a': 1 }.b = MISSING
```

I believe the PartiQL Spec says something along the lines of: "in permissive mode, equals never returns missing". Similar scenarios are encountered when we look at the logical operators.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
```diff
@@ -3,53 +3,73 @@ package org.partiql.eval.internal.operator.rex
...
 internal class ExprCallDynamic(
     private val name: String,
-    candidateFns: Array<Function>,
+    private val functions: Array<Function.Instance>,
     private val args: Array<Operator.Expr>
 ) : Operator.Expr {
 
-    private val candidates = Array(candidateFns.size) { Candidate(candidateFns[it]) }
+    /**
+     * @property paramIndices the last index of the [args]
```
johnedquinn, line 35:
```suggestion
     * @property paramIndices the indices of the [args]
```

johnedquinn (approved):

PR #1603, opened 2024-10-01
[2/2] Parameter match on SqlTypeFamily

RCHowell:
## Relevant Issues

#1585
#1598 

## Description

This PR is Part 2 of 2. It introduces the internal concept of a "SqlTypeFamily" which is internal-only and aids us in accurate typing.

Why are we doing this? Accurate SQL typing requires knowing the exact types of input arguments i.e. what is the precision and scale of the argument decimals. SQL defines this by "enumerating" all permutations and defining a static function for each permutation. In practice, this would mean defining at least 38^2 functions for each function that has a decimal argument.

This implementation is how we get around that. The SqlType family enables defining operators on "acceptable" types and checking if we can coerce to the preferred type when there are multiple choices (postgres style).

## Other Information

## License Information


RCHowell:
Merging for 1.0.0-rc1 with follow-up tracked in https://github.com/partiql/partiql-lang-kotlin/issues/1604

PR #1611, opened 2024-10-07
Fix V1 exclude implementation

rchowell:
## Description

This PR fixes the V1 implementation of the EXCLUDE operator, along with introducing the finalized V1 modeling of exclusion trees in the query plans.

The language slight diverges from roots/paths used by the AST modeling of exclusion paths. The subsumption algorithm merges several paths into a tree of exclusions which is modeled as a var+items where each item is its own subtree of exclusion items.

The implementation works by keep track of which exclusions to apply at the current level, and memoizes what should be applied at a deeper level. The previous implementation could memoize on the "step" because it was a generated data class with no children, however this implementation no longer uses those data classes so must memoize differently. 

Otherwise, the function is the same as before .. just now on the simplified tree representation.

## Other Information

## License Information

### DIFF
--- a/partiql-plan/src/main/kotlin/org/partiql/plan/rex/RexVar.kt
+++ b/partiql-plan/src/main/kotlin/org/partiql/plan/rex/RexVar.kt
@@ -24,18 +24,18 @@ public interface RexVar : Rex {
 /**
  * Default [RexVar] implementation intended for extension.
  */
-internal class RexVarImpl(depth: Int, offset: Int) : RexVar {
+internal class RexVarImpl(depth: Int, offset: Int, type: RexType) : RexVar {
 
     // DO NOT USE FINAL
     private var _depth = depth
     private var _offset = offset
-    // private var _type = type
-    //
+    private var _type = type
+
     override fun getDepth(): Int = _depth
 
     override fun getOffset(): Int = _offset
 
-    override fun getType(): RexType = TODO("getType not yet implemented")
+    override fun getType(): RexType = _type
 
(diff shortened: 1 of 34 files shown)


partiql-plan/src/main/kotlin/org/partiql/plan/rex/RexVar.kt
```diff
@@ -42,15 +42,13 @@ internal class RexVarImpl(depth: Int, offset: Int, type: RexType) : RexVar {
...
         if (_offset != other.getOffset()) return false
-        if (_type != other.getType()) return false
         return true
     }
 
     override fun hashCode(): Int {
         var result = 1
         result = 31 * result + _depth
         result = 31 * result + _offset
-        result = 31 * result + _type.hashCode()
```
alancai98, line 53: At some point, need to consider whether `type` should or shouldn't be included in the `equals` and `hashcode` functions for the plan node classes.

partiql-plan/src/main/kotlin/org/partiql/plan/Exclusion.kt
```diff
@@ -0,0 +1,274 @@
+package org.partiql.plan
+
+import org.partiql.plan.rex.RexVar
+
+/**
```
alancai98, line 5: nice set of kdoc for the public apis!

alancai98 (commented):
Sorry for the delayed review. Fixed implementation of subsumption looks good. I had one comment about the equals/hashcode function changes. Perhaps we need to look at whether a node's PType should be part of the equals/hashcode computation.

PR #1615, opened 2024-10-08
Initializes PartiQL's error reporting mechanism

### LINKED ISSUE #1597: Add uniform error reporting (codes) across parser, planner, and evaluator
## Requested Solution/Feature
- This issue is to track the consolidation of error reporting to provide a consistent experience for end-customers.
- Due to recent requests, it may be in the best interest for the PartiQL library to standardize a small set of error codes that may be used across implementations.

## DoD (Definition of Done)
- [x] Consistent error reporting mechanism across all components of the library
- [x] Standardized error codes
- [x] Exceptions should (likely) never be thrown on the component level

johnedquinn:
## Relevant Issues
- Closes #1597

## Description
- Creates V1's error reporting mechanism, highlighted by the introduction of `PError`, `PErrorListener`, and `PErrorListenerException`.
  - Other notable additions: `PEnum`, `Severity`, and `Classification`.
- Now, ALL errors/warnings will be reported via the registered error listener (for each component).
- I've added an implementation of this in the CLI.
- I've also added a usage guide (that may be updated prior to the release of V1). This is just a start.
- I've completely removed the legacy `Error` and `Problem` APIs.

## Follow-ups
- Need to remove `PartiQLSyntaxException` -- which was made internal by this PR. But we no longer need to convert to a syntax exception and then to an error code.

## License Information


partiql-eval/src/main/kotlin/org/partiql/eval/PartiQLEngine.kt
```diff
@@ -24,7 +24,7 @@ import org.partiql.spi.catalog.Session
  */
 public interface PartiQLEngine {
 
-    public fun prepare(plan: Plan, mode: Mode, session: Session): PartiQLStatement
+    public fun prepare(plan: Plan, session: Session, config: CompilerConfig): PartiQLStatement
```
RCHowell, line 27: I suggest changing the "Config" to something more like [Context](https://calcite.apache.org/javadocAggregate/org/apache/calcite/plan/Context.html) which may or may not be part of the session, but perhaps these are slightly different?

The Config reads like what should be part of a PartiQLEngine/PartiQLParser/PartiQLPlanner builders. Why not have `PartiQLParser.builder().addListener(...)` because the "config" is what the constructor arguments are for.

 I am curious the use-case of having a changing config for each invocation, and where a session context wouldn't achieve this.

partiql-parser/src/main/java/org/partiql/parser/ParserConfig.java
```diff
@@ -0,0 +1,18 @@
...
+ * @see PartiQLParser
+ * @see ParserConfigBuilder
+ */
+public interface ParserConfig {
+    /**
+     * @return The error listener to be used by the compiler.
+     */
+    @NotNull
+    PErrorListener getErrorListener();
+}
```
RCHowell, line 18: What are some reasons not to add listeners via constructors (or even mutable class methods) like ANTLR has. These configs add a several more classes to juggle and thread through all calls, which seems like additional complexity but I may be missing how these are more useful than just adding listeners via the builders.

partiql-parser/src/main/kotlin/org/partiql/parser/PartiQLParser.kt
```diff
@@ -14,19 +14,39 @@
...
-    )
+    ) {
+        public companion object {
+            @OptIn(PartiQLValueExperimental::class)
+            internal fun empty(source: String): Result {
+                val locations = SourceLocations.Mutable().toMap()
+                return Result(source, Statement.Query(Expr.Lit(nullValue())), locations)
+            }
+        }
+    }
```
RCHowell, line 49: We can remove the data class. I'm also curious your thoughts on these "Result" classes and if we should be pushing for better consistency across the components.

RCHowell (changes requested):

partiql-spi/src/main/java/org/partiql/spi/errors/PError.java
```diff
@@ -0,0 +1,398 @@
...
+ * <br><br>
+ * All error codes are defined as static final integers in {@link PError}.
+ * <br><br>
+ * <b>WARNING</b>: The available properties are subject to change without prior notice. The property's values are also
+ * subject to change without prior notice. Your application <b>MUST</b> be able to handle these scenarios.
+ * @see PError#code()
+ * @see PError#getOrNull(String, Class)
+ * @see PError#getListOrNull(String, Class)
+ */
+public class PError extends PEnum {
```
RCHowell, line 33: consider a public final class unless we want it to be open for extension

partiql-planner/src/main/kotlin/org/partiql/planner/internal/problems/AlwaysMissing.kt
```diff
@@ -0,0 +1,20 @@
...
+ */
+internal class AlwaysMissing(
+    location: SourceLocation?
+) : PError(
+    ALWAYS_MISSING,
+    Severity.WARNING(),
+    Classification.SEMANTIC(),
+    location,
+    null
+)
```
RCHowell, line 20: Rather than a bunch of extra classes, how about simplifying to just static methods? We already have codes so the additionally classes don't help us and they do not extend functionality of the PError class. The constructors here are effectively being used as named factory methods, so let's do just that. This would allow PError to be final as well.

_Errors.kt_
```kotlin

internal object Errors {

    @JvmStatic
    fun alwaysMissing(location: SourceLocation) = PError(
       ALWAYS_MISSING,
       Severity.WARNING(),
       Classification.SEMANTIC(),
       location,
       null,
    )

   // a place to instantiate errors
}

```

RCHowell (changes requested):

partiql-spi/src/main/java/org/partiql/spi/errors/Classification.java
```diff
@@ -0,0 +1,107 @@
...
+     */
+    @NotNull
+    public static Classification COMPILATION() {
+        return new Classification(COMPILATION);
+    }
+
+    /**
+     * TODO
+     */
+    public static final int EXECUTION = 4;
```
RCHowell, line 74: What are some examples of compilation vs execution errors? 

In a recent message we were talking about how lazy datums hide execution problems. I'm not sure I understand how we might get an execution PError rather than something akin to a "runtime exception" when materializing a lazy datum.

RCHowell: Perhaps some driver in partiql-lang would take a listener an it might emit execution errors? I somewhat think partiql-eval should have an evaluator driver that can emit them. By driver, I just mean something that materializes Datums. So a customer may choose to use the lazy datum, or just invoke an eval() which actually performs the work and accepts a listener. Does that make sense?

docs/wiki/v1/ErrorHandling.md
```diff
@@ -0,0 +1,233 @@
...
+        System.out.println(message);
+    }
+
+    static String getMessage(@NotNull Error error, @NotNull String prefix) {
+        switch (error.getCode()) {
+            case Error.ALWAYS_MISSING:
+                Integer line = (Integer) error.getProperty(Property.LINE_NO);
+                Integer column = (Integer) error.getProperty(Property.COL_NO);
+                String location = getNullSafeLocation(line, column);
+                return prefix + location + " Expression always evaluates to missing.";
```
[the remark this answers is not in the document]
johnedquinn: Yeah, that document is outdated since we now do expose a `location` field in `PError`. I'll update.

partiql-eval/src/main/kotlin/org/partiql/eval/PartiQLEngine.kt
johnedquinn: Having the error listener as part of a context/config allows for end-users to create singleton planners (if they want to). They're given the freedom to always pass the same object, or to pass in new error listeners.

That being said, I'm up for adding it to the session context -- but I wasn't entirely sure the _right_ place to put it.

partiql-parser/src/main/java/org/partiql/parser/ParserConfig.java
johnedquinn: See https://github.com/partiql/partiql-lang-kotlin/pull/1615#discussion_r1805257316.

partiql-parser/src/main/kotlin/org/partiql/parser/PartiQLParser.kt
johnedquinn: Discussed. I'm in favor, though outside of this PR.

partiql-spi/src/main/java/org/partiql/spi/errors/Classification.java
johnedquinn: > What are some examples of compilation vs execution errors?

Compilation errors might happen if the plan is referencing a custom physical operator that doesn't exist, or if a feature is not supported for execution (yet). It describes taking the process of taking the input plan and outputting an executable. There may be many types of errors at this stage.

Execution errors may be emitted during actual execution, such as data exceptions. I actually discuss this briefly in the usage guide, but the idea is that all components will standardize the fact the the only JVM exception that these components will see is the `PErrorListenerException`. Though, for exceptions that arise during the execution of the prepared statement, I detailed an `EvaluationException` that essentially wraps a `PError`.

RCHowell (changes requested):

partiql-eval/src/main/kotlin/org/partiql/eval/PartiQLEngine.kt
```diff
@@ -24,10 +25,18 @@ import org.partiql.spi.catalog.Session
...
+        return prepare(plan, session, Context.standard(), Mode.STRICT)
+    }
+
+    public fun prepare(plan: Plan, session: Session, ctx: Context): PartiQLStatement {
+        return prepare(plan, session, ctx, Mode.STRICT)
+    }
+
+    public fun prepare(plan: Plan, session: Session, mode: Mode): PartiQLStatement {
+        return prepare(plan, session, Context.standard(), mode)
     }
```
RCHowell: no default mode and the ctx with/without are the overloads so something like

```suggestion
    public fun prepare(plan: Plan, mode: Mode): PartiQLStatement {
        return prepare(plan, mode, Context.NONE)
    }

    public fun prepare(plan: Plan, mode: Mode, ctx: Context): PartiQLStatement
```

I can follow up since this will go on the java PartiQLCompiler interface so no more kt engine thing.

partiql-spi/src/main/java/org/partiql/spi/Context.java
```diff
@@ -0,0 +1,44 @@
...
+public interface Context {
+
+    /**
+     * The default utilizes {@link PErrorListener#abortOnError()}.
+     * @return the registered {@link PErrorListener}.
+     */
+    @NotNull
+    default PErrorListener getErrorListener() {
+        return PErrorListener.abortOnError();
+    }
```
RCHowell, line 20: I'm not sure we want to instantiate a new error listener every time, this is somewhat like the argument changing everytime. We can have a TODO to make AbortOnError a static variable

RCHowell (approved):

RCHowell (approved):

Issue #1610, opened 2024-10-07
[v1] Replace sprout-generated AST with hand-written AST


alancai98:
Tracking issue for replacing the sprout-generated AST with the hand-written AST. The following will need to be completed to mark this issue as completed:

- [x] https://github.com/partiql/partiql-lang-kotlin/pull/1579
- [x] Add builders - https://github.com/partiql/partiql-lang-kotlin/pull/1616
- [x] Make fields/properties final - https://github.com/partiql/partiql-lang-kotlin/pull/1616
- [x] Add factory methods for AST classes - https://github.com/partiql/partiql-lang-kotlin/pull/1622
  - [x] Rewrite factory methods in Kotlin for better usability from Kotlin - https://github.com/partiql/partiql-lang-kotlin/pull/1623
- [x] Create AstBaseVisitor and AstRewriter equivalents - https://github.com/partiql/partiql-lang-kotlin/pull/1629
- Switchover existing sprout-generated AST code w/ the new Java AST definitions
  - [x] partiql-ast: `SqlDialect` - https://github.com/partiql/partiql-lang-kotlin/pull/1638
  - [x] partiql-parser: ANTLR visitor -> partiql-ast - https://github.com/partiql/partiql-lang-kotlin/pull/1626
  - [x] partiql-planner: visitor to convert AST -> partiql-plan - https://github.com/partiql/partiql-lang-kotlin/pull/1631
  - [x] migrate v1 names to existing names (parser) - https://github.com/partiql/partiql-lang-kotlin/pull/1637
- [ ] Add docs
- [x] Delete sprout-generated AST code - https://github.com/partiql/partiql-lang-kotlin/pull/1639
- Misc other tasks
  - [x] Add equals and hashcode for AST nodes (does not use the AstNode tag) - https://github.com/partiql/partiql-lang-kotlin/pull/1627
  - [x] Add `.valueOf` and `.values` methods for AST enums - https://github.com/partiql/partiql-lang-kotlin/pull/1628
  - [x] Add `.name` method to AST enum interface - https://github.com/partiql/partiql-lang-kotlin/pull/1630
  - [ ] Create `.codes() -> int[]`, `.names() -> strings[]`, `code(string) -> int`, `name(int) -> string` methods for `AstEnum`s
[...]

PR #1616, opened 2024-10-08
[v1] Add Lombok Builders to classes; make every field final

alancai98:
## Relevant Issues

## Description
- Add Lombok builders to classes
- Makes every field final

## Other Information
  - No, v1

  - Yes but on v1

  - Yes, add Lombok Kotlin plugin (to reference Lombok classes/interfaces from Kotlin) and Lombok API dependency

## License Information

### DIFF
--- a/partiql-ast/build.gradle.kts
+++ b/partiql-ast/build.gradle.kts
@@ -16,7 +16,6 @@
 plugins {
     id(Plugins.conventions)
     id(Plugins.publish)
-    id(Plugins.kotlinLombok) version Versions.kotlinLombok
 }
 
 dependencies {
--- a/partiql-ast/build.gradle.kts
+++ b/partiql-ast/build.gradle.kts
@@ -16,6 +16,7 @@
 plugins {
     id(Plugins.conventions)
     id(Plugins.publish)
+    id(Plugins.kotlinLombok) version Versions.kotlinLombok
 }
 
 dependencies {
@@ -24,6 +25,8 @@ dependencies {
     // TODO REMOVE ME ONCE PartiQLValue IS REMOVED
     // THE AST NEEDS ITS OWN "VALUE" REPRESENTATION
     api(project(":partiql-spi"))
+    compileOnly(Deps.lombok)
+    annotationProcessor(Deps.lombok)
 }
 
 tasks.shadowJar {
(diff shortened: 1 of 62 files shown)


alancai98:
seems like we still are getting a new conformance comparison report for every commit. I'll investigate further

partiql-ast/build.gradle.kts
```diff
@@ -16,6 +16,7 @@
 plugins {
     id(Plugins.conventions)
     id(Plugins.publish)
+    id(Plugins.kotlinLombok) version Versions.kotlinLombok
```
johnedquinn, line 19: What's the purpose of this plugin? From what I can tell online, it's not required for using the annotations.

partiql-ast/src/main/java/org/partiql/ast/v1/Exclude.java
```diff
@@ -9,9 +10,10 @@
 /**
  * TODO docs, equals, hashcode
  */
+@Builder
```
johnedquinn, line 13: Discussed offline, but the generated builder lies at: `Exclude.ExcludeBuilder`. A bit redundant -- this is non-blocking, but if you'd want to look into it, I'm just pointing it out.

partiql-ast/build.gradle.kts
alancai98: I was following the Kotlin Lombok sample code, which had included the plugin -- https://github.com/kotlin-hands-on/kotlin-lombok-examples/blob/master/kotlin_lombok_gradle/nokapt/build.gradle.kts.

From what I can tell, it's just used to enable using Java Lombok code from Kotlin -- https://kotlinlang.org/docs/lombok.html:

> The Kotlin Lombok compiler plugin allows the generation and use of Java's Lombok declarations by Kotlin code in the same mixed Java/Kotlin module. If you call such declarations from another module, then you don't need to use this plugin for the compilation of that module.

Since I'm not sure we have such usages in the Kotlin AST code, I'll just remove for now. We can add it back if we ever need it.

partiql-ast/src/main/java/org/partiql/ast/v1/Exclude.java
alancai98: Yeah, seems like I can override the builder name to just `Exclude.Builder` -- `@Builder(builderClassName = "Builder")`. Will revise

johnedquinn (approved):

PR #1621, opened 2024-10-16
Adds parsing and modeling for `INSERT INTO <table name>`

### LINKED ISSUE #1595: Support for `INSERT INTO <qualified identifier>`
## Requested Solution/Feature
- Several of the DDL and DML statements don't allow for qualified identifiers. This ticket is specifically asking for support for insert statements with qualified identifiers.
- The following should parse:
```
INSERT INTO some_catalog."some_SCHEMA".some_table VALUES (1, 2, 3);
```
- Also, the experimental planner implementation should add planning support for this.

## Additional Context
- The AST and plan modeling on `main` currently do not have support for qualified identifiers. This should be addressed as part of this issue.
- There is likely significant reworking of the definition of bindings. This is already handled on the `v1` branch, however, this request is for `main`'s experimental planner implementation.
- This originally came as a request relevant to: #1592

## DoD (Definition of Done)
- [ ] AST and plan modeling adding support for qualified identifiers
- [ ] Parser support for `INSERT INTO <qualified identifier>`
- [ ] Planner support for qualified identifiers

johnedquinn:
## Relevant Issues
- Closes #1595 

## Description
- Adds parsing and modeling for `INSERT INTO <table name>`
- Updates release version to `0.14.10` and updates README/CHANGELOG.

## Low Level Details
- ANTLR
  - Adds `<table name>` and `<identifier chain>` rules to `PartiQLParser.g4`, which largely matches the SQL:1999 EBNF.
- PIG AST
  - Adds `table_name` and `identifier_chain` rules to `partiql.ion`.
  - `table_name` really just wraps an `identifier_chain`.
  - `insert`'s `target` is now a `table_name` -- not an expression (which realistically was always `expr.id`, which isn't right because it held a scope qualifier (AKA `@`).
  - I didn't update `insert_value`, `remove`, or `set` -- since it's not part of the feature request. However, we might want to consider updating these in the future.
- PIG LOGICAL
  - Modifies `dml_target` to hold a `table_name` -- no longer holds an unqualified `identifier`.
  - This has direct implications on `dml_insert` and `dml_update` and indirect implications on `UPSERT`/`REPLACE` since these are modeled using insert.
- LOGICAL TO LOGICAL RESOLVED
  - I had to update the `GlobalVariableResolver` to allow for the resolution of qualified identifiers, also resulting in a `BindingId`, which is essentially a qualified `BindingName`.
  - I also needed to add a new `GlobalResolutionResult` that allows for namespaced names. See the Javadocs there.

## Why in draft?
- I'm waiting for some preliminary feedback before I write more comprehensive tests for this.

## Other Information

  - < If YES, why? >

## License Information

### DIFF
--- a/partiql-lang/src/main/kotlin/org/partiql/lang/eval/Bindings.kt
+++ b/partiql-lang/src/main/kotlin/org/partiql/lang/eval/Bindings.kt
@@ -57,63 +57,6 @@ fun PartiqlAst.CaseSensitivity.toBindingCase(): BindingCase = when (this) {
 }
 
 /**
- * Represents a namespaced global binding.
- */
-class BindingId(
-    private val qualifier: List<BindingName>,
-    private val id: BindingName
-) : Iterable<BindingName> {
-    /**
-     * For input such as `SELECT a FROM catalog1."schema1".table1`, the identifier is `table1`.
-     */
-    fun getIdentifier(): BindingName = id
-
-    /**
-     * For input such as `SELECT a FROM catalog1."schema1".table1`, the qualifier is `catalog1."schema1"`
-     */
-    fun getQualifier(): List<BindingName> = qualifier
-
-    /**
-     * Returns whether the id is qualified.
-     */
-    fun hasQualifier(): Boolean = qualifier.isNotEmpty()
-
-    /**
-     * Returns a collection of the parts of the identifier.
-     */
-    fun getParts(): List<BindingName> {
-        return qualifier + id
-    }
-
-    override fun iterator(): Iterator<BindingName> {
-        return getParts().iterator()
-    }
-
-    override fun toString(): String {
-        return when (hasQualifier()) {
-            true -> "${qualifier.joinToString(".")}.$id"
-            false -> id.toString()
-        }
-    }
-
-    override fun equals(other: Any?): Boolean {
-        if (this === other) return true
-        if (other !is BindingId) return false
-
-        if (qualifier != other.qualifier) return false
-        if (id != other.id) return false
-
-        return true
-    }
-
-    override fun hashCode(): Int {
-        var result = qualifier.hashCode()
-        result = 31 * result + id.hashCode()
-        return result
-    }
-}
-
-/**
  * Encapsulates the data necessary to perform a binding lookup.
  */
 data class BindingName(val name: String, val bindingCase: BindingCase) {
--- a/partiql-lang/src/main/kotlin/org/partiql/lang/eval/Bindings.kt
+++ b/partiql-lang/src/main/kotlin/org/partiql/lang/eval/Bindings.kt
@@ -57,6 +57,63 @@ fun PartiqlAst.CaseSensitivity.toBindingCase(): BindingCase = when (this) {
 }
 
 /**
+ * Represents a namespaced global binding.
+ */
+class BindingId(
+    private val qualifier: List<BindingName>,
+    private val id: BindingName
+) : Iterable<BindingName> {
+    /**
+     * For input such as `SELECT a FROM catalog1."schema1".table1`, the identifier is `table1`.
+     */
+    fun getIdentifier(): BindingName = id
+
+    /**
+     * For input such as `SELECT a FROM catalog1."schema1".table1`, the qualifier is `catalog1."schema1"`
+     */
+    fun getQualifier(): List<BindingName> = qualifier
+
+    /**
+     * Returns whether the id is qualified.
+     */
+    fun hasQualifier(): Boolean = qualifier.isNotEmpty()
+
+    /**
+     * Returns a collection of the parts of the identifier.
+     */
+    fun getParts(): List<BindingName> {
+        return qualifier + id
+    }
+
+    override fun iterator(): Iterator<BindingName> {
+        return getParts().iterator()
+    }
+
+    override fun toString(): String {
+        return when (hasQualifier()) {
+            true -> "${qualifier.joinToString(".")}.$id"
+            false -> id.toString()
+        }
+    }
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other !is BindingId) return false
+
+        if (qualifier != other.qualifier) return false
+        if (id != other.id) return false
+
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = qualifier.hashCode()
+        result = 31 * result + id.hashCode()
+        return result
+    }
+}
+
+/**
  * Encapsulates the data necessary to perform a binding lookup.
  */
 data class BindingName(val name: String, val bindingCase: BindingCase) {
@@ -65,6 +122,13 @@ data class BindingName(val name: String, val bindingCase: BindingCase) {
      * Compares [name] to [otherName] using the rules specified by [bindingCase].
      */
     fun isEquivalentTo(otherName: String?) = otherName != null && name.isBindingNameEquivalent(otherName, bindingCase)
+
+    override fun toString(): String {
+        return when (bindingCase) {
+            BindingCase.SENSITIVE -> "\"$name\""
+            BindingCase.INSENSITIVE -> "$name"
+        }
+    }
 }
 
 /**
--- a/partiql-lang/src/main/kotlin/org/partiql/lang/planner/GlobalVariableResolver.kt
+++ b/partiql-lang/src/main/kotlin/org/partiql/lang/planner/GlobalVariableResolver.kt
@@ -1,6 +1,7 @@
 package org.partiql.lang.planner
 
 import org.partiql.lang.eval.BindingCase
+import org.partiql.lang.eval.BindingId
 import org.partiql.lang.eval.BindingName
 
 /**
@@ -16,6 +17,21 @@ sealed class GlobalResolutionResult {
      * information about global variables.
      */
     data class GlobalVariable(val uniqueId: String) : GlobalResolutionResult()
+
+    /**
+     * A success case. Refers to a variable that is contained within a namespace (AKA catalog/schema).
+     */
+    abstract class NamespacedVariable(
+        val uniqueId: String
+    ) : GlobalResolutionResult() {
+        /**
+         * When attempting to resolve a qualified identifier, say `FROM cat1.schema1.table1.attr1.attr2`, the [GlobalVariableResolver]
+         * _may_ resolve only the first few steps of the identifier, say `cat1.schema1.table1`. Therefore, it must
+         * return the resolved [uniqueId] (say `cat1:::schema1:::table1`) and the remaining steps `attr1.attr2` in order for the planner
+         * to convert these to path expressions.
+         */
+        abstract fun getRemainingSteps(): List<BindingName>
+    }
 
     /** A failure case, indicates that resolution did not match any variable. */
     object Undefined : GlobalResolutionResult()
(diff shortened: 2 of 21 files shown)


partiql-lang/src/main/kotlin/org/partiql/lang/eval/Bindings.kt
```diff
@@ -56,6 +56,63 @@ fun PartiqlAst.CaseSensitivity.toBindingCase(): BindingCase = when (this) {
     is PartiqlAst.CaseSensitivity.CaseSensitive -> BindingCase.SENSITIVE
 }
 
+/**
+ * Represents a namespaced global binding.
+ */
+class BindingId(
+    private val qualifier: List<BindingName>,
+    private val id: BindingName
```
dlurton, line 64: As I mentioned in a comment in the PIG model--the corollary here is that there should just be one list of `BindingName`, but if for some reason you don't end up going that route:

- `id` -> `head` or `root` (pick one, remain consistent)
- `qualifier` should be the 2nd argument, not the first

partiql-lang/src/main/kotlin/org/partiql/lang/eval/Bindings.kt
```diff
@@ -56,6 +56,63 @@ fun PartiqlAst.CaseSensitivity.toBindingCase(): BindingCase = when (this) {
...
+        if (id != other.id) return false
+
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = qualifier.hashCode()
+        result = 31 * result + id.hashCode()
+        return result
+    }
```
dlurton, line 113: Why can't this be a `data` class to avoid the  `equals`, `hashCode` boilerplate?   AFAICT with a quick glance the only one you really need here is `toString()`.

...

On closer inspection does this class even need custom implementations of `equals` and `hashCode`?  I would consider not implementing those things at all since there are hard questions like what is the correct definition of equality is between instances of `BindingId` and `BindingName`, etc.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/Bindings.kt
```diff
@@ -56,6 +56,63 @@ fun PartiqlAst.CaseSensitivity.toBindingCase(): BindingCase = when (this) {
...
+     * Returns whether the id is qualified.
+     */
+    fun hasQualifier(): Boolean = qualifier.isNotEmpty()
+
+    /**
+     * Returns a collection of the parts of the identifier.
+     */
+    fun getParts(): List<BindingName> {
+        return qualifier + id
+    }
```
dlurton, line 86: If you just have a single list of `BindingName` as I suggested above, the first three functions above aren't needed.

Also, why not use kotlin's read-only property syntax here?

```Kotlin
val parts get() = qualifier + id
```

It's quite identical to your `getParts`, Java even should still see a `getParts` , AFAIK.

partiql-lang/src/main/kotlin/org/partiql/lang/planner/transforms/LogicalToLogicalResolvedVisitorTransform.kt
```diff
@@ -352,25 +354,48 @@ internal data class LogicalToLogicalResolvedVisitorTransform(
...
+                if (resolvedVariable.getRemainingSteps().isNotEmpty()) {
+                    problemHandler.handleProblem(
+                        Problem(
+                            node.metas.sourceLocationMetaOrUnknown.toProblemLocation(),
+                            PlanningProblemDetails.UndefinedDmlTarget(
+                                bindingId,
+                            )
+                        )
+                    )
+                    "DML Target could not be completely resolved - do not run"
```
dlurton, line 370: This message seems geared to the developers applications that embed partiql--not to query authors.  It should be comprehensible to the query author.  i.e. as a query author I don't know what to do with the "do not run" instruction.

partiql-lang/src/main/kotlin/org/partiql/lang/planner/transforms/LogicalToLogicalResolvedVisitorTransform.kt
```diff
@@ -352,25 +354,48 @@ internal data class LogicalToLogicalResolvedVisitorTransform(
 
     override fun transformDmlTarget(node: PartiqlLogical.DmlTarget): PartiqlLogicalResolved.DmlTarget {
         // We only support DML targets that are global variables.
-        val bindingName = BindingName(node.identifier.name.text, node.identifier.case.toBindingCase())
-        val tableUniqueId = when (val resolvedVariable = globals.resolveGlobal(bindingName)) {
+        val bindingId = bindingId(node.identifier)
+        val tableUniqueId = when (val resolvedVariable = globals.resolveGlobal(bindingId)) {
             is GlobalResolutionResult.GlobalVariable -> resolvedVariable.uniqueId
+            is GlobalResolutionResult.NamespacedVariable -> {
+                if (resolvedVariable.getRemainingSteps().isNotEmpty()) {
```
dlurton, line 361: Nit: `isNotEmpty()` is a Java-ism.  Use `.any()` instead.

partiql-lang/src/main/kotlin/org/partiql/lang/planner/GlobalVariableResolver.kt
```diff
@@ -17,6 +18,21 @@ sealed class GlobalResolutionResult {
...
+    abstract class NamespacedVariable(
+        val uniqueId: String
+    ) : GlobalResolutionResult() {
+        /**
+         * When attempting to resolve a qualified identifier, say `FROM cat1.schema1.table1.attr1.attr2`, the [GlobalVariableResolver]
+         * _may_ resolve only the first few steps of the identifier, say `cat1.schema1.table1`. Therefore, it must
+         * return the resolved [uniqueId] (say `cat1:::schema1:::table1`) and the remaining steps `attr1.attr2` in order for the planner
+         * to convert these to path expressions.
+         */
+        abstract fun getRemainingSteps(): List<BindingName>
```
dlurton, line 33: IMHO this might be a little too forward thinking... i.e. trying to design for a future we know little about.  Or is there more to this than I am aware of?  

In my case, all parts of a table identifier will be completely resolved or we will abort w/undefined table error.  This *might* make sense in the event that we have only part of the schema... but that is not in my foreseeable future. I suggest dropping this.

If this is dropped, I don't think we need a distinction between a `GlobalVariable` and a `NamespacedVariable`  either... `GlobalVariable` will still be fine for our case...  Even though we're kind of using it in a way not entirely consistent with the name.  Maybe, "ResolvedVariable" is a better name?

partiql-lang/src/main/kotlin/org/partiql/lang/planner/GlobalVariableResolver.kt
```diff
@@ -56,6 +72,18 @@ fun interface GlobalVariableResolver {
      */
     fun resolveGlobal(bindingName: BindingName): GlobalResolutionResult
```
dlurton, line 73: I think you can drop this--it is redundant when you have `resolveGlobal(BindingId)`.  Or, more accurately, drop this function and move its documentation (with modification) to the other overload.  Then, remove the default implementation of `resolveGlobal(BindingId)` so we're forced to implement it.

partiql-lang/src/main/kotlin/org/partiql/lang/planner/GlobalVariableResolver.kt
```diff
@@ -17,6 +18,21 @@ sealed class GlobalResolutionResult {
      */
     data class GlobalVariable(val uniqueId: String) : GlobalResolutionResult()
```
dlurton, line 19: A change I've been wanting to make forever is to make `uniqueId` here an `IonElement` so we can put arbitrary Ion data here instead of having to "abuse" this field by putting Ion-text here instead.  I'll leave a note in the main PIG domain where we need to change this as well.... EDIT: github won't let me comment there--to do this we would also need to change `partiql_logical_resolved.expr.global_id.unique_id` to the `ion` type.

Since we're the API anyway, now might be a good time to do that as well...

partiql-lang/src/main/kotlin/org/partiql/lang/eval/Bindings.kt
```diff
@@ -56,6 +56,63 @@ fun PartiqlAst.CaseSensitivity.toBindingCase(): BindingCase = when (this) {
     is PartiqlAst.CaseSensitivity.CaseSensitive -> BindingCase.SENSITIVE
 }
 
+/**
+ * Represents a namespaced global binding.
+ */
+class BindingId(
```
dlurton, line 62: This class isn't used by anything else in this file--consider moving it to its own file.

dlurton (changes requested):
Oh my I do

partiql-lang/src/main/kotlin/org/partiql/lang/planner/GlobalVariableResolver.kt
dlurton: Thinking again about this... was the intention to handle the scenario when some of the parts of the qualified identifier could be resolved but others not?  I guess I'm not 100% sure I understand the intent here.

partiql-lang/src/main/kotlin/org/partiql/lang/eval/Bindings.kt
johnedquinn: I've obviously been writing a lot of Java recently 😅

partiql-lang/src/main/kotlin/org/partiql/lang/planner/transforms/LogicalToLogicalResolvedVisitorTransform.kt
johnedquinn: Would it be presented to the user? Line 384 has very similar wording, but it's used to return a table's "unique id", even when a problem has been emitted to the problem handler. That being said, I'll update the type of problem here.

partiql-lang/src/main/kotlin/org/partiql/lang/planner/GlobalVariableResolver.kt
johnedquinn: For backwards compatibility purposes, when adding new APIs, we've been adding default implementations that "may" utilize existing implemented methods, so that users who don't want to update their code don't have to. That being said, you're the only customer here, so it's up to you.

partiql-parser/src/main/antlr/PartiQLParser.g4
```diff
@@ -165,9 +202,26 @@ removeCommand
 insertCommandReturning
     : INSERT INTO pathSimple VALUE value=expr ( AT pos=expr )? onConflictLegacy? returningClause?;
```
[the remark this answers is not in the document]
johnedquinn: Limited scope, since this change is for your use-case.

partiql-lang/src/main/kotlin/org/partiql/lang/planner/GlobalVariableResolver.kt
johnedquinn: > Thinking again about this... was the intention to handle the scenario when some of the parts of the qualified identifier could be resolved but others not?

Yes, that is the intention. In the `INSERT INTO <table name>` use-case, you wouldn't allow `remainingSteps` to have anything. But, in a query such as the one mentioned in the Javadocs, it is useful to know what is actually a global vs what is a path expression. This allows for the engine-implementer to enforce the accuracy of the path expression and for the global-variable-resolver implementer to only take care of returning "global" variables.

This is something that has been added to `v1`.

johnedquinn:
Closing until there is an imminent need for this.

alancai98:
Tagging @johnedquinn and @dlurton. This may need re-opened or a new PR.

PR #1622, opened 2024-10-17
[v1] Add AST factory methods; some fixes to AST classes

alancai98:
## Relevant Issues

## Description
- Adds factory methods for every hand-written AST class
- Fixes some AST classes/functions
  - Add constructors for some classes
  - Fix naming of `visitSelectItemAll` -> `visitSelectItemStar`
  - Change `@NotNull Long` -> `long`

## Other Information
  - No, targeting v1 branch

  - Yes but on v1 branch

## License Information

### DIFF
--- a/partiql-ast/src/main/java/org/partiql/ast/v1/Ast.java
+++ b/partiql-ast/src/main/java/org/partiql/ast/v1/Ast.java
@@ -268,13 +268,13 @@ public class Ast {
     }
 
     @NotNull
-    public static GraphLabel.Conj graphLabelConj(@NotNull List<GraphLabel> args) {
-        return new GraphLabel.Conj(args);
+    public static GraphLabel.Conj graphLabelConj(@NotNull GraphLabel lhs, @NotNull GraphLabel rhs) {
+        return new GraphLabel.Conj(lhs, rhs);
     }
 
     @NotNull
-    public static GraphLabel.Disj graphLabelDisj(@NotNull List<GraphLabel> args) {
-        return new GraphLabel.Disj(args);
+    public static GraphLabel.Disj graphLabelDisj(@NotNull GraphLabel lhs, @NotNull GraphLabel rhs) {
+        return new GraphLabel.Disj(lhs, rhs);
     }
 
     @NotNull
--- a/partiql-ast/src/main/java/org/partiql/ast/v1/Ast.java
+++ b/partiql-ast/src/main/java/org/partiql/ast/v1/Ast.java
@@ -1,0 +1,481 @@
+package org.partiql.ast.v1;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.partiql.ast.v1.expr.Expr;
+import org.partiql.ast.v1.expr.ExprAnd;
+import org.partiql.ast.v1.expr.ExprArray;
+import org.partiql.ast.v1.expr.ExprBag;
+import org.partiql.ast.v1.expr.ExprBetween;
+import org.partiql.ast.v1.expr.ExprCall;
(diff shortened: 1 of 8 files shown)


partiql-ast/src/main/java/org/partiql/ast/v1/graph/GraphSelector.java
```diff
@@ -36,6 +36,8 @@ public <R, C> R accept(@NotNull AstVisitor<R, C> visitor, C ctx) {
      */
     @Builder(builderClassName = "Builder")
     public static class AnyShortest extends GraphSelector {
+        public AnyShortest() {}
```
alancai98, line 39: (self-review) some missing constructors in the graph/ directory.

partiql-ast/src/main/java/org/partiql/ast/v1/AstVisitor.java
```diff
@@ -149,7 +149,7 @@ public interface AstVisitor<R, C> {
 
     R visitSelectItem(SelectItem node, C ctx);
 
-    R visitSelectItemAll(SelectItem.Star node, C ctx);
```
alancai98, line 152: (self-review) I forgot to change naming of this visitor function after the class was renamed.

partiql-ast/src/main/java/org/partiql/ast/v1/graph/GraphQuantifier.java
```diff
@@ -14,13 +14,12 @@
  */
 @Builder(builderClassName = "Builder")
 public class GraphQuantifier extends AstNode {
-    @NotNull
-    public final Long lower;
```
alancai98, line 18: (self-review) type could just be primitive `long`.

partiql-ast/src/main/java/org/partiql/ast/v1/Ast.java
```diff
@@ -0,0 +1,481 @@
...
+
+    @NotNull
+    public static ExprMatch exprMatch(@NotNull Expr expr, @NotNull GraphMatch pattern) {
+        return new ExprMatch(expr, pattern);
+    }
+
+    @NotNull
+    public static ExprNot exprNot(@NotNull Expr value) {
+        return new ExprNot(value);
+    }
```
alancai98, line 137: (self-review) for PartiQL expression `NOT false`,

example usage from Kotlin
```Kotlin
exprNot(exprLit(boolValue(false)))
```

example usage from Java
```Java
exprNot(exprLit(boolValue(false)));
```

partiql-ast/src/main/java/org/partiql/ast/v1/Ast.java
```diff
@@ -0,0 +1,481 @@
...
+import java.util.List;
+import java.util.Map;
+
+// TODO docs for all factory methods
+public class Ast {
+    // Expr
+    @NotNull
+    public static ExprAnd exprAnd(Expr lhs, Expr rhs) {
+        return new ExprAnd(lhs, rhs);
+    }
```
alancai98, line 61: One potentially annoying thing I've noticed about these factory methods being written in Java is that they cannot be called from Kotlin with named parameters.

E.g. below results in an error "Named arguments are not allowed for non-Kotlin functions"
```
identifier(
    symbol = ctx.IDENTIFIER_QUOTED().getStringValue(),
    isDelimited = true
)
```

Wondering if we should actually have these factory methods written in Kotlin instead?

RCHowell (approved):

partiql-ast/src/main/java/org/partiql/ast/v1/Ast.java
alancai98: I'll bring up next week w/ the team if these should be written in Kotlin.

Issue #1013, opened 2023-03-20
Explore SQL-consistent syntax for joins


vgapeyev:
PartiQL grammar for join expressions currently allows syntax that is not in SQL: 
*  `OUTER JOIN` -- OUTER must come with LEFT, RIGHT, or FULL, not by itself; 
* `LEFT OUTER CROSS JOIN `, `RIGHT CROSS JOIN`, and other variations of `joinType CROSS JOIN` -- in SQL, `CROSS JOIN` is a thing on its own, without further qualification. 

While PartiQL sometimes extends SQL syntax in order to anchor constructs with PartiQL-specific semantics distinct from SQL (for example, `SELECT VALUE`), this does not appear to be the case here.  Also, since this involves only the syntactic tokens from the domain of joins -- just in new combinations -- this is particularly confusing/misleading.

Some of this syntax appears in the specification, while there is lore suggesting that `LEFT CROSS JOIN` is there to work around grammar ambiguity in the presence on the `ON condition` clause in a join.  

The purpose of this ticket it to experiment with the grammar to understand whether there is indeed an issue with SQL-consistent syntax and explore whether it can be resolved. Upon resolution, this will entail adjustments to the specification as well, but those should be purely narrative.

PR #1626, opened 2024-10-22
[v1] Migrate parser to new AST

alancai98:
## Relevant Issues

## Description
- Migrates the partiql-parser to use the new AST

## Other Information
  - On v1 branch

  - Some changes but on v1 and hasn't been release yet

## License Information

### DIFF
--- a/partiql-ast/src/main/java/org/partiql/ast/v1/FromJoin.java
+++ b/partiql-ast/src/main/java/org/partiql/ast/v1/FromJoin.java
@@ -17,10 +17,10 @@ import java.util.List;
 @EqualsAndHashCode(callSuper = false)
 public class FromJoin extends FromTableRef {
     @NotNull
-    public final From lhs;
+    public final FromTableRef lhs;
 
     @NotNull
-    public final From rhs;
+    public final FromTableRef rhs;
 
     @Nullable
     public final JoinType joinType;
@@ -28,7 +28,7 @@ public class FromJoin extends FromTableRef {
     @Nullable
     public final Expr condition;
 
-    public FromJoin(@NotNull From lhs, @NotNull From rhs, @Nullable JoinType joinType, @Nullable Expr condition) {
+    public FromJoin(@NotNull FromTableRef lhs, @NotNull FromTableRef rhs, @Nullable JoinType joinType, @Nullable Expr condition) {
         this.lhs = lhs;
         this.rhs = rhs;
         this.joinType = joinType;
--- a/partiql-parser/src/main/antlr/PartiQLParser.g4
+++ b/partiql-parser/src/main/antlr/PartiQLParser.g4
@@ -358,7 +358,7 @@ excludeExprSteps
     ;
 
 fromClause
-    : FROM tableReference;
+    : FROM ( tableReference ( COMMA tableReference)* );
 
 whereClauseSelect
     : WHERE arg=exprSelect;
@@ -468,16 +468,16 @@ edgeAbbrev
  */
 
 tableReference
-    : lhs=tableReference joinType? CROSS JOIN rhs=joinRhs     # TableCrossJoin
-    | lhs=tableReference COMMA rhs=joinRhs                    # TableCrossJoin
-    | lhs=tableReference joinType? JOIN rhs=joinRhs joinSpec  # TableQualifiedJoin
-    | tableNonJoin                                            # TableRefBase
-    | PAREN_LEFT tableReference PAREN_RIGHT                   # TableWrapped
+    : tablePrimary # TableRefPrimary
+    | lhs=tableReference LEFT CROSS JOIN rhs=tablePrimary # TableLeftCrossJoin // PartiQL spec defines LEFT CROSS JOIN; other variants are not defined yet
+    | lhs=tableReference CROSS JOIN rhs=tablePrimary # TableCrossJoin // SQL99 defines just CROSS JOIN
+    | lhs=tableReference joinType? JOIN rhs=tableReference joinSpec  # TableQualifiedJoin
     ;
 
-tableNonJoin
+tablePrimary
     : tableBaseReference
     | tableUnpivot
+    | tableWrapped
     ;
 
 tableBaseReference
@@ -487,15 +487,11 @@ tableBaseReference
     ;
 
 tableUnpivot
-    : UNPIVOT expr asIdent? atIdent? byIdent?;
-
-joinRhs
-    : tableNonJoin                           # JoinRhsBase
-    | PAREN_LEFT tableReference PAREN_RIGHT  # JoinRhsTableJoined
+    : UNPIVOT expr asIdent? atIdent? byIdent?
     ;
 
-joinSpec
-    : ON expr;
+tableWrapped
+    : PAREN_LEFT tableReference PAREN_RIGHT;
 
 joinType
     : mod=INNER
@@ -504,6 +500,9 @@ joinType
     | mod=FULL OUTER?
     | mod=OUTER
     ;
+
+joinSpec
+    : ON expr;
 
 /**
  *
--- a/partiql-parser/src/main/kotlin/org/partiql/parser/V1PartiQLParser.kt
+++ b/partiql-parser/src/main/kotlin/org/partiql/parser/V1PartiQLParser.kt
@@ -1,0 +1,39 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at:
+ *
+ *      http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ */
+
+package org.partiql.parser
+
+import org.partiql.ast.v1.Statement
+import org.partiql.parser.internal.V1PartiQLParserDefault
+
+public interface V1PartiQLParser {
+
+    @Throws(PartiQLSyntaxException::class, InterruptedException::class)
+    public fun parse(source: String): Result
+
+    public data class Result(
+        val source: String,
+        val root: Statement,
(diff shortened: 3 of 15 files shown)


partiql-parser/src/main/antlr/PartiQLParser.g4
```diff
@@ -358,7 +358,7 @@ excludeExprSteps
     ;
 
 fromClause
-    : FROM tableReference;
+    : FROM ( tableReference ( COMMA tableReference)* );
```
alancai98, line 361: Made some ANTLR grammar changes to account for the new modeling of `FROM` clauses to be better aligned with the SQL1999 BNF definition. See https://github.com/partiql/partiql-lang-kotlin/pull/1579#discussion_r1788110562 for some further context.

partiql-ast/src/main/java/org/partiql/ast/v1/FromJoin.java
```diff
@@ -17,18 +17,18 @@
 @EqualsAndHashCode(callSuper = false)
 public class FromJoin extends FromTableRef {
     @NotNull
-    public final From lhs;
+    public final FromTableRef lhs;
 
     @NotNull
-    public final From rhs;
+    public final FromTableRef rhs;
```
alancai98, line 23: This was an error in the original PR adding the hand-written ASTs. The `lhs` and `rhs` should have been `FromTableRef`s.

partiql-parser/src/main/kotlin/org/partiql/parser/V1PartiQLParser.kt
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ */
+
+package org.partiql.parser
+
+import org.partiql.ast.v1.Statement
+import org.partiql.parser.internal.V1PartiQLParserDefault
+
+public interface V1PartiQLParser {
```
alancai98, line 20: Duplicated all of the existing `PartiQLParser` APIs for now to keep the other tests passing (e.g. planner, eval). After those other packages are migrated to use the parser outputting the new v1 AST, then we can delete the existing sprout AST `PartiQLParser` APIs.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/V1PartiQLParserDefault.kt
```diff
@@ -0,0 +1,2197 @@
...
+        }
+
+        override fun visitTableWrapped(ctx: PartiQLParser.TableWrappedContext): FromTableRef = translate(ctx) {
+            visitAs<FromTableRef>(ctx.tableReference())
+        }
+
+        override fun visitTableCrossJoin(ctx: PartiQLParser.TableCrossJoinContext): FromTableRef = translate(ctx) {
+            val lhs = visitAs<FromTableRef>(ctx.lhs)
+            val rhs = visitAs<FromTableRef>(ctx.rhs)
+            // TODO support other CROSS JOIN variants (e.g. LEFT CROSS JOIN)
```
alancai98, line 1254: The previous ANTLR parser visitor would parse `l LEFT CROSS JOIN r` to the AST equivalent of `l LEFT JOIN r ON TRUE`. This was mentioned in the PartiQL spec in section 5.4 (https://partiql.org/dql/joins.html#left-join), but I couldn't find it mentioned in the SQL1999 spec. It's not yet clear to me what other JOIN types (e.g. `RIGHT`, `LEFT OUTER`, `INNER`, `FULL`, etc.) should support this behavior so I've left it out of the new ANTLR parser visitor for now.

alancai98: From offline discussion, will change the ANTLR parsing rules slightly:
1. Remove the optional `joinType` before the `CROSS JOIN` since that's what is defined in SQL99 spec
2. Add an explicit rule for `LEFT CROSS JOIN` since that's what is defined in the PartiQL spec
3. Add TODO linking Vladimir's issue for the other cross join variant investigation -- https://github.com/partiql/partiql-lang-kotlin/issues/1013

partiql-parser/src/main/kotlin/org/partiql/parser/internal/V1PartiQLParserDefault.kt
```diff
@@ -0,0 +1,2195 @@
...
+                if (it.type == GeneratedLexer.QUESTION_MARK && parameterIndexes.containsKey(token.tokenIndex).not()) {
+                    parameterIndexes[token.tokenIndex] = ++parametersFound
+                }
+            }
+            return token
+        }
+    }
+
+    /**
+     * Translate an ANTLR ParseTree to a PartiQL
```
RCHowell, line 356:
```suggestion
     * Translate an ANTLR ParseTree to a PartiQL AST
```

partiql-parser/src/main/kotlin/org/partiql/parser/internal/V1PartiQLParserDefault.kt
```diff
@@ -0,0 +1,2195 @@
...
+     */
+    @OptIn(PartiQLValueExperimental::class)
+    private class Visitor(
+        private val tokens: CommonTokenStream,
+        private val locations: SourceLocations.Mutable,
```
RCHowell, line 361: not for this PR, but I would like to put source locations class under the microscope..

partiql-parser/src/main/kotlin/org/partiql/parser/internal/V1PartiQLParserDefault.kt
```diff
@@ -0,0 +1,2195 @@
...
+            graphMatchNode(prefilter, variable, label)
+        }
+
+        private fun visitRestrictor(ctx: GeneratedParser.PatternRestrictorContext?): GraphRestrictor? {
+            if (ctx == null) return null
+            return when (ctx.restrictor.text.lowercase()) {
+                "trail" -> GraphRestrictor.TRAIL()
+                "acyclic" -> GraphRestrictor.ACYCLIC()
+                "simple" -> GraphRestrictor.SIMPLE()
+                else -> throw error(ctx, "Unrecognized pattern restrictor")
```
RCHowell, line 1194: Curious if the message ever propagates beyond here

RCHowell (changes requested):
Small comments, otherwise let's get it merged.

partiql-parser/src/main/kotlin/org/partiql/parser/V1PartiQLParser.kt
alancai98: Yeah good catch and thanks for cutting the issue. Parser APIs should also be migrated to Java (especially the nested `Result` data class).

I'll link the issue in the kdoc.

partiql-parser/src/main/kotlin/org/partiql/parser/V1PartiQLParserBuilder.kt
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * A builder class to instantiate a [V1PartiQLParser].
+ */
+public class V1PartiQLParserBuilder {
+
+    public fun build(): V1PartiQLParser {
+        return V1PartiQLParserDefault()
+    }
+}
```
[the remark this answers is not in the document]
alancai98: Yeah true. I will link the issue in the file.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/V1PartiQLParserDefault.kt
```diff
@@ -0,0 +1,2195 @@
...
+        } catch (e: PErrorListenerException) {
+            throw e
+        } catch (throwable: Throwable) {
+            val error = PError.INTERNAL_ERROR(PErrorKind.SYNTAX(), null, throwable)
+            ctx.errorListener.report(error)
+            return V1PartiQLParser.Result.empty(source)
+        }
+    }
+
+    companion object {
```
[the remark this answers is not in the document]
alancai98: These were just copied from the existing `PartiQLParserDefault` for context. Since the static methods within the companion object are all private, I will just move them out of the companion object.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/V1PartiQLParserDefault.kt
```diff
@@ -0,0 +1,2195 @@
...
+        ) {
+            if (offendingSymbol is Token) {
+                val rule = e?.ctx?.toString(rules) ?: "UNKNOWN" // TODO: Do we want to display the offending rule?
+                val token = offendingSymbol.text
+                val tokenType = GeneratedParser.VOCABULARY.getSymbolicName(offendingSymbol.type)
+                val location = org.partiql.spi.SourceLocation(line.toLong(), charPositionInLine + 1L, token.length.toLong())
+                val error = PErrors.unexpectedToken(location, tokenType, null)
+                listener.report(error)
+            } else {
+                throw IllegalArgumentException("Offending symbol is not a Token.")
```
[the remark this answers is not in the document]
alancai98: Ack. Changed this and the `TokenizeErrorListener` logic to cast directly.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/V1PartiQLParserDefault.kt
alancai98: I tested locally with a misspelt graph restrictor (i.e. `TRAILL` rather than `TRAIL`). The message will be part of a syntax `PError` with a message captured in the "CAUSE" field -- "org.partiql.parser.PartiQLParserException: Unrecognized pattern restrictor".

> ErrorException{error=PError{code=1, severity=ERROR, kind=SYNTAX, location=null, properties={CAUSE=org.partiql.parser.PartiQLParserException: Unrecognized pattern restrictor}}}

alancai98: I haven't been following the latest `PError` developments. Is this the expected behavior?

RCHowell: Yes it's ok, we explicitly state that properties are not semvered

RCHowell (approved):

PR #1627, opened 2024-10-22
[v1] Add equals and hashcode for AST classes

alancai98:
## Relevant Issues

## Description
- Adds equals and hashcode for AST classes using Lombok annotations. Notice the use of `@EqualsAndHashCode(callSuper = false)`, which will not take into account the super class. This will prevent the `AstNode` tag from being used by the equals and hashcode functions.

## Other Information
  - No, on v1

## License Information

### DIFF
--- a/partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
+++ b/partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
@@ -1,5 +1,8 @@
 package org.partiql.ast.v1;
 
+import lombok.EqualsAndHashCode;
+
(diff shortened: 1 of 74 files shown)


partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
```diff
@@ -158,6 +161,14 @@ public static DataType DEC() {
         return new DataType(DEC);
     }
 
+    public static DataType DEC(int precision) {
```
alancai98, line 164: these were missing from the origin commit adding the simplified AST

RCHowell (approved):

PR #1628, opened 2024-10-22
[v1] Add .valueOf and .values methods for enums

alancai98:
## Relevant Issues

## Description
- The previous sprout-generated code used Kotlin enum classes which provided some helper methods when working with enum constants -- https://kotlinlang.org/docs/enum-classes.html#working-with-enum-constants. I found I needed these in the partiql-parser migration, particularly `.valueOf(<string>)` and `.values`. This PR adds those two functions to the interface and implements them for existing enum impls.

## Other Information
  - No, on v1.

  - Adds to the enum interface but on v1

## License Information

### DIFF
--- a/partiql-ast/src/main/java/org/partiql/ast/v1/AstEnum.java
+++ b/partiql-ast/src/main/java/org/partiql/ast/v1/AstEnum.java
@@ -1,0 +1,8 @@
+package org.partiql.ast.v1;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public abstract class AstEnum extends AstNode {
+    public abstract int code();
+}
--- a/partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
+++ b/partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
@@ -1,6 +1,7 @@
 package org.partiql.ast.v1;
 
 import lombok.EqualsAndHashCode;
+import org.jetbrains.annotations.NotNull;
 
 @EqualsAndHashCode(callSuper = false)
 public class DataType implements Enum {
@@ -349,7 +350,11 @@ public class DataType implements Enum {
         return new DataType(INTERVAL);
     }
 
-    public static DataType USER_DEFINED(IdentifierChain name) {
+    public static DataType USER_DEFINED() {
+        return new DataType(USER_DEFINED);
+    }
+
+    public static DataType USER_DEFINED(@NotNull IdentifierChain name) {
         return new DataType(USER_DEFINED, name);
     }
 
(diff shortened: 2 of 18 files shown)


partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
```diff
@@ -1,6 +1,7 @@
 package org.partiql.ast.v1;
 
 import lombok.EqualsAndHashCode;
+import org.jetbrains.annotations.NotNull;
 
 @EqualsAndHashCode(callSuper = false)
 public class DataType implements Enum {
```
alancai98, line 7: for every AST class implementing the `enum` interface, added the `valueOf(String value)` and `values()` static methods

partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
```diff
@@ -391,6 +396,109 @@ public int code() {
...
+            case "LIST": return LIST();
+            case "SEXP": return SEXP();
+            case "BAG": return BAG();
+            case "TIME": return TIME();
+            case "TIME_WITH_TIME_ZONE": return TIME_WITH_TIME_ZONE();
+            case "TIMESTAMP": return TIMESTAMP();
+            case "TIMESTAMP_WITH_TIME_ZONE": return TIMESTAMP_WITH_TIME_ZONE();
+            case "INTERVAL": return INTERVAL();
+            case "USER_DEFINED": return USER_DEFINED();
+            default: return UNKNOWN();
```
alancai98, line 447: unidentified string values will map to `UNKNOWN`

partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
```diff
@@ -391,6 +396,109 @@ public int code() {
...
+        }
+    }
+
+    public static DataType[] values() {
```
alancai98, line 451: `.values()` does not include `UNKNOWN`

partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
```diff
@@ -391,6 +396,109 @@ public int code() {
         return code;
     }
 
+    public static DataType valueOf(String value) {
```
RCHowell, line 399: Maybe annotate both return and parameter as NotNull?

partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
RCHowell, line 451: I suggest these are all singletons, we want `enum.FOO() === enum.FOO()` – aka they return the same reference.

partiql-ast/src/main/java/org/partiql/ast/v1/graph/GraphRestrictor.java
```diff
@@ -39,4 +39,21 @@ private GraphRestrictor(int code) {
...
+    }
+
+    public static GraphRestrictor[] values() {
+        return new GraphRestrictor[] {
+            TRAIL(),
+            ACYCLIC(),
+            SIMPLE()
+        };
+    }
 }
```
RCHowell, line 59: Comments from DataType apply to all enums.

RCHowell (changes requested):

partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
alancai98: Yeah. I will add.

alancai98:
Other changes I'll make from offline discussion:
1. Change naming of `Enum` to `AstEnum`
2. Change `AstEnum` to be an abstract class that extends `AstNode`. `AstNode` will still be an abstract class since it has state (i.e. the tag)
3. Change static method `values()` to be named `codes()` and return an array of ints
4. Change static method `valueOf(String value)` to be named `parse(String value)`. Still will return the enum class.

partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
alancai98: From offline discussion, we want separate instances since some enum nodes will extend `AstNode` in subsequent PRs. Those nodes like `DataType` may need source locations which require a tag in the parser's source location map.

partiql-ast/src/main/java/org/partiql/ast/v1/AstEnum.java
```diff
@@ -0,0 +1,8 @@
+package org.partiql.ast.v1;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public abstract class AstEnum extends AstNode {
+    public abstract int code();
```
RCHowell, line 7:
```suggestion
    public abstract int code();
    public abstract String name();
```

partiql-ast/src/main/java/org/partiql/ast/v1/DataType.java
RCHowell: this is a good example of the backwards compatibility

partiql-ast/src/main/java/org/partiql/ast/v1/AstEnum.java
RCHowell: This could be a follow-up

RCHowell (approved):

partiql-ast/src/main/java/org/partiql/ast/v1/AstEnum.java
alancai98: Yeah -- I'll cut a followup PR since this PR is getting somewhat large. Tracked in https://github.com/partiql/partiql-lang-kotlin/issues/1610.

alancai98: Added `name` method in https://github.com/partiql/partiql-lang-kotlin/pull/1630.

PR #1629, opened 2024-10-23
[v1] Adds AstVisitor and AstRewriter; port partiql-ast normalization passes to partiql-planner

alancai98:
## Relevant Issues

## Description
- Makes `AstVisitor` an abstract class with defaults
- Adds an `AstRewriter`
- Port the partiql-ast normalization passes to partiql-planner package and makes them `internal`
- Other partiql-planner `AstVisitor` impls are to be added in a subsequent PR

## Other Information
  - No, on v1 branch.

## License Information

### DIFF
--- a/partiql-ast/src/main/java/org/partiql/ast/v1/AstRewriter.kt
+++ b/partiql-ast/src/main/java/org/partiql/ast/v1/AstRewriter.kt
@@ -43,6 +43,7 @@ import org.partiql.ast.v1.graph.GraphQuantifier
 import org.partiql.ast.v1.graph.GraphSelector
 import org.partiql.value.PartiQLValueExperimental
 
+// TODO docs and move to Kotlin sources
 public abstract class AstRewriter<C> : AstVisitor<AstNode, C>() {
     override fun defaultReturn(node: AstNode, context: C): AstNode = node
 
--- a/partiql-ast/src/main/java/org/partiql/ast/v1/AstRewriter.kt
+++ b/partiql-ast/src/main/java/org/partiql/ast/v1/AstRewriter.kt
@@ -1,0 +1,752 @@
+package org.partiql.ast.v1
+
+import org.partiql.ast.v1.Ast.explain
+import org.partiql.ast.v1.Ast.exprQuerySet
+import org.partiql.ast.v1.Ast.identifier
+import org.partiql.ast.v1.Ast.query
+import org.partiql.ast.v1.expr.Expr
+import org.partiql.ast.v1.expr.ExprAnd
+import org.partiql.ast.v1.expr.ExprArray
+import org.partiql.ast.v1.expr.ExprBag
+import org.partiql.ast.v1.expr.ExprBetween
+import org.partiql.ast.v1.expr.ExprCall
+import org.partiql.ast.v1.expr.ExprCase
+import org.partiql.ast.v1.expr.ExprCast
+import org.partiql.ast.v1.expr.ExprCoalesce
+import org.partiql.ast.v1.expr.ExprExtract
+import org.partiql.ast.v1.expr.ExprInCollection
+import org.partiql.ast.v1.expr.ExprIsType
+import org.partiql.ast.v1.expr.ExprLike
+import org.partiql.ast.v1.expr.ExprLit
+import org.partiql.ast.v1.expr.ExprMatch
+import org.partiql.ast.v1.expr.ExprNot
+import org.partiql.ast.v1.expr.ExprNullIf
+import org.partiql.ast.v1.expr.ExprOperator
+import org.partiql.ast.v1.expr.ExprOr
+import org.partiql.ast.v1.expr.ExprOverlay
+import org.partiql.ast.v1.expr.ExprParameter
+import org.partiql.ast.v1.expr.ExprPath
+import org.partiql.ast.v1.expr.ExprPosition
+import org.partiql.ast.v1.expr.ExprQuerySet
+import org.partiql.ast.v1.expr.ExprSessionAttribute
+import org.partiql.ast.v1.expr.ExprStruct
+import org.partiql.ast.v1.expr.ExprSubstring
+import org.partiql.ast.v1.expr.ExprTrim
+import org.partiql.ast.v1.expr.ExprValues
+import org.partiql.ast.v1.expr.ExprVarRef
+import org.partiql.ast.v1.expr.ExprVariant
+import org.partiql.ast.v1.expr.ExprWindow
+import org.partiql.ast.v1.expr.PathStep
+import org.partiql.ast.v1.graph.GraphMatch
+import org.partiql.ast.v1.graph.GraphPattern
+import org.partiql.ast.v1.graph.GraphQuantifier
+import org.partiql.ast.v1.graph.GraphSelector
+import org.partiql.value.PartiQLValueExperimental
+
+public abstract class AstRewriter<C> : AstVisitor<AstNode, C>() {
+    override fun defaultReturn(node: AstNode, context: C): AstNode = node
+
+    private inline fun <reified T> _visitList(
+        nodes: List<T>,
+        ctx: C,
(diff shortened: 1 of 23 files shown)


partiql-ast/src/main/java/org/partiql/ast/v1/expr/SessionAttribute.java
```diff
@@ -49,4 +49,12 @@ public static SessionAttribute[] values() {
             CURRENT_DATE()
         };
     }
+
+    public String name() {
```
alancai98, line 53: Something else I noticed was missing from the AST enum classes, was the `name` property (https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-enum/name.html). Added since it was used for one of the ast visitors. Perhaps we should add it for all of the AST enums?

alancai98: TODO: add `name` to `enum` methods to Java interface.

alancai98: Added to the `AstEnum` interface in https://github.com/partiql/partiql-lang-kotlin/pull/1630.

alancai98:
Going to mark this as a draft till I fix the visitor modeling a bit.

partiql-ast/src/main/java/org/partiql/ast/v1/ExcludeStep.java
```diff
@@ -15,6 +16,7 @@ public abstract class ExcludeStep extends AstNode {
      * TODO docs, equals, hashcode
      */
     @Builder(builderClassName = "Builder")
+    @EqualsAndHashCode(callSuper = false)
```
alancai98, line 19: (self-review) these were missing from PR adding equals and hashcode

partiql-ast/src/main/java/org/partiql/ast/v1/FromTableRef.java
```diff
@@ -1,19 +1,6 @@
 package org.partiql.ast.v1;
 
-import org.jetbrains.annotations.NotNull;
-
 /**
  * TODO docs, equals, hashcode
  */
-public abstract class FromTableRef extends AstNode {
-    @Override
-    public <R, C> R accept(@NotNull AstVisitor<R, C> visitor, C ctx) {
```
alancai98, line 10: (self-review) `accept` over the variants were moved to the `AstVisitor`

partiql-ast/src/main/java/org/partiql/ast/v1/AstRewriter.kt
```diff
@@ -0,0 +1,758 @@
...
+import org.partiql.ast.v1.expr.ExprVariant
+import org.partiql.ast.v1.expr.ExprWindow
+import org.partiql.ast.v1.expr.PathStep
+import org.partiql.ast.v1.graph.GraphMatch
+import org.partiql.ast.v1.graph.GraphPattern
+import org.partiql.ast.v1.graph.GraphQuantifier
+import org.partiql.ast.v1.graph.GraphSelector
+import org.partiql.value.PartiQLValueExperimental
+
+public abstract class AstRewriter<C> : AstVisitor<AstNode, C> {
```
alancai98, line 46: (self-review) `AstRewriter` written in Kotlin since it was slightly easier to create based on the sprout-generated `AstRewriter`. I believe it should have similar public API as if it was in Java but if not, I can migrate it to Java.

partiql-ast/src/main/java/org/partiql/ast/v1/AstVisitor.java
```diff
@@ -33,206 +32,557 @@
...
+    }
+
+    public R visitStatement(Statement node, C ctx) {
+        if (node instanceof Query) {
+            return visitQuery((Query) node, ctx);
+        } else if (node instanceof Explain) {
+            return visitExplain((Explain) node, ctx);
+        } else {
+            throw new IllegalStateException("Unexpected value: " + node);
+        }
```
johnedquinn, line 64: I'm trying to think about how we can make it easier for users to add their own custom nodes without breaking their visitors completely. Is there something wrong with just:
```suggestion
        node.accept(this, ctx)
```

partiql-planner/src/main/kotlin/org/partiql/planner/internal/normalize/Normalize.kt
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package org.partiql.planner.internal.normalize
+
+import org.partiql.ast.v1.Statement
+
+/**
+ * AST normalization
+ */
+internal fun Statement.normalize(): Statement {
```
johnedquinn, line 23: Thank you for putting this in the planner! Long overdue.

partiql-ast/src/main/java/org/partiql/ast/v1/AstRewriter.kt
johnedquinn: I can check the decompiled bytecode, but this resides in the `java` source set btw.

partiql-ast/src/main/java/org/partiql/ast/v1/AstVisitor.java
johnedquinn: Theoretically, each variant of a sum type would look like:
```java
public class Query : AstNode() {
    @Override
    public <R, C> R accept(@NotNull AstVisitor<R, C> visitor, C ctx) {
        return visitor.visitQuery(this, ctx);
    }
}
```

So, it would definitely work for the built-in variants. For the other variants, I'm not entirely sure how they'd be able to invoke a custom visit (AKA `visitor.visitCustomNode(this, ctx)`).

partiql-ast/src/main/java/org/partiql/ast/v1/AstRewriter.kt
alancai98: > but this resides in the java source set btw.

Yeah, I'll fix this in a later PR (probably the PR that deletes the sprout AST code). There's some other Kotlin sources in the `java` directory that I'll also move in that PR.

partiql-ast/src/main/java/org/partiql/ast/v1/AstVisitor.java
alancai98: You're right about the simplification. Those sum type visit functions can be simplified to `node.accept(this, ctx)`.

alancai98: From offline discussion, there's pros/cons to both approaches. There are some drawbacks to the `node.accept` default for the sum types, primarily a user of this library would need to be careful with overriding the proper methods when adding their own variants. This can hopefully be resolved by including some good usage guides.

I do like your suggested approach a bit more since it will simplify what we (the PartiQL team) need to add when adding a new variant.

partiql-ast/src/main/java/org/partiql/ast/v1/AstVisitor.java
```diff
@@ -33,206 +32,557 @@
...
-    R visitGraphLabelName(GraphLabel.Name node, C ctx);
-
-    R visitGraphLabelWildcard(GraphLabel.Wildcard node, C ctx);
-
-    R visitGraphLabelNegation(GraphLabel.Negation node, C ctx);
-
-    R visitGraphLabelConj(GraphLabel.Conj node, C ctx);
-
-    R visitGraphLabelDisj(GraphLabel.Disj node, C ctx);
+public abstract class AstVisitor<R, C> {
```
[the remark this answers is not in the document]
alancai98: Replacing the `instanceOf` checking with the `node.accept(...)` calls for the sum type visit methods (from [your comment](https://github.com/partiql/partiql-lang-kotlin/pull/1629#discussion_r1825161053)) would get rid of the `IllegalStateException`s.

I'll add docs and a usage guide in a subsequent PR once the APIs are more finalized.

johnedquinn (approved):

PR #1634, opened 2024-10-30
Adds Parser APIs in Java and allows for parsing multiple statements

### LINKED ISSUE #1632: [V1] Replace PartiQLParser interface with Java interface
## Requested Solution/Feature

The current parser interface is written in kotlin which introduces unnecessary companion object classes and doesn't have our with/without context overloads.  We will also nest the builder so that we have a consistent builder style following our configured lombok.

This interface should support parsing multiple statements.

johnedquinn:
## Relevant Issues
- Closes #1632

## Description
- Adds Parser APIs in Java
- Allows for parsing multiple statements
- Updates SourceLocations to be in Java. Utilizes the SourceLocation class from SPI.
- PartiQLParser.Result no longer contains the input string -- this didn't make any sense.
- PartiQLParser now contains the Builder as a nested class, Lombok style

## License Information


partiql-cli/src/main/kotlin/org/partiql/cli/shell/ShellHighlighter.kt
```diff
@@ -89,7 +89,7 @@ internal object ShellHighlighter : Highlighter {
 
         // Parse and Replace Token Style if Failures
         try {
-            parser.root()
+            parser.file()
```
RCHowell, line 92:
```suggestion
            parser.statements()
```

partiql-cli/src/main/kotlin/org/partiql/cli/pipeline/Pipeline.kt
```diff
@@ -38,7 +38,10 @@ internal class Pipeline private constructor(
         val result = listen(ctx.errorListener as AppPErrorListener) {
             parser.parse(source, ctx)
         }
-        return result.root
+        if (result.statements.size != 1) {
+            throw PipelineException("Expected exactly one statement, got: ${result.statements.size}")
+        }
+        return result.statements[0]
```
RCHowell, line 44: Can you add a `parse` and `parseAll` so that customers do not need to always handle this. The parse can either call parseAll and do the assertion and throw IllegalArgument, or it can invoke the single-statement ANTLR rule resulting in an "expected EOF" type error. This would be _slightly_ more efficient.

partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEvaluatorTest.kt
```diff
@@ -1320,7 +1321,9 @@ class PartiQLEvaluatorTest {
         )
 
         internal fun assert() {
-            val statement = parser.parse(input).root
+            val parseResult = parser.parse(input)
+            assertEquals(1, parseResult.statements.size)
+            val statement = parseResult.statements[0]
```
RCHowell, line 1326: Here is an example where the _parse_ and _parseAll_ would introduce functionality without introducing complexity for the consumer.

partiql-parser/src/main/antlr/PartiQLParser.g4
```diff
@@ -11,14 +11,16 @@ options {
  *
  */
 
-root
-    : (EXPLAIN (PAREN_LEFT explainOption (COMMA explainOption)* PAREN_RIGHT)? )? statement;
+file
+    : statement (COLON_SEMI statement)* COLON_SEMI? EOF
+    ;
```
RCHowell, line 16:
```suggestion
statements
    : statement (COLON_SEMI statement)* COLON_SEMI? EOF
    ;
```

RCHowell: Interesting that the last statement doesn't require a semi-colon. I see no semi-colon making sense for a single statement, but something is off about this succeeding..

```
select 1 + 1;
select 2 + 2;
select 3 + 3
```

Maybe multiple statements require each to have a semi-colon? The other way to look at it is like an implicit list that does not require trailing "commas" aka semi-colons

```
[ statement1; statement2; statement3 ]   -- reasonable
[ statement1; statement2; statement3; ]  -- reasonable
```

I'm not sure, but something seems off about not requiring a semi-colon on the last statement. Maybe that's just my experience with languages whose omission of a semi-colon on the last statement _means_ something different than when a semi-colon is present (rust implicit return)!

partiql-parser/src/main/java/org/partiql/parser/PartiQLParserBuilder.java
```diff
@@ -12,16 +12,22 @@
...
-        return PartiQLParserDefault()
+    /**
+     * TODO
+     * @return TODO
+     */
+    @NotNull
+    public PartiQLParser build() {
+        return new PartiQLParserDefault();
     }
 }
```
RCHowell, line 33: Lombok style, let's nest this as "Builder" under the PartiQLParser interface.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PFile.kt
```diff
@@ -0,0 +1,16 @@
...
+internal class PFile(
+    internal val statements: List<Statement>,
+) : AstNode() {
+
+    override val children: List<AstNode> = this.statements
+
+    override fun <R, C> accept(visitor: AstVisitor<R, C>, ctx: C): R {
+        error("This is an internal ast node and cannot be passed to a visitor.")
+    }
+}
```
RCHowell, line 16: Discussed offline, not need to make these extra classes (pfile and pfilev1)

partiql-spi/src/main/java/org/partiql/spi/SourceLocations.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * This class maps a set of identifiers to their corresponding source locations.
+ * <br>
+ * <b>Note!</b>: This class is immutable and does not support {@link Map#put(Object, Object)}, amongst others. Please
+ * handle the runtime exceptions indicated by {@link Map}'s Javadocs.
+ */
+public class SourceLocations implements Map<String, SourceLocation> {
```
RCHowell, line 16: nice

RCHowell (changes requested):

partiql-parser/src/main/antlr/PartiQLParser.g4
johnedquinn: Updated to enforce semi-colon for the last statement within a multi-statement block. We can always add support for it later.

partiql-parser/src/main/java/org/partiql/parser/PartiQLParserBuilder.java
RCHowell, line 33: Remove this class?

partiql-parser/src/main/java/org/partiql/parser/PartiQLParserV1.java
```diff
@@ -0,0 +1,151 @@
...
+            Map<String, Object> properties = new HashMap<String, Object>() {{
+                put("EXPECTED_TOKENS", new ArrayList<String>() {{
+                    add("EOF");
+                }});
+            }};
+            PError pError = new PError(PError.UNEXPECTED_TOKEN, Severity.ERROR(), PErrorKind.SYNTAX(), location, properties);
+            ctx.getErrorListener().report(pError);
+        }
+        return result;
+    }
```
RCHowell, line 76: Parse single doesn't make sense actually with the result. That was a bad suggestion on my part (: 

Let me know your thoughts, but removing this seems fine.

RCHowell (approved):

partiql-parser/src/main/java/org/partiql/parser/PartiQLParserBuilder.java
johnedquinn: I believe this is part of the follow-up that @alancai98 will be undertaking for removing the older parser.

PR #1635, opened 2024-10-31
Define datum reader/writer interfaces with variant

RCHowell:
## Description

This PR introduces the DatumReader and DatumWriter interfaces along with the IonDatumReader implementation. In the interest of time, I've opened the PR now to get the interfaces in – and will continue to flesh out the implementations when I have time.

Once completed, and RFC will be raised with the encoding scheme.

## Other Information

## License Information

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.is pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


partiql-spi/src/main/java/org/partiql/spi/Enum.java
```diff
@@ -3,16 +3,20 @@
 import java.util.Objects;
 
 /**
- * Represents an enum in the PartiQL Library. This is used for backward/forward compatibility purposes.
+ * All enumerated types should extend this class for backward/forward compatibility.
  */
-public abstract class PEnum {
+public abstract class Enum {
```
johnedquinn, line 8: Originally named `PEnum` due to the ubiquity of `java.lang.Enum`. I don't mind it either way since most people don't use `Enum` directly, but bringing it up for awareness.

partiql-spi/src/main/java/org/partiql/spi/value/Variant.java
```diff
@@ -0,0 +1,43 @@
...
+ * @param <T>
+ */
+public interface Variant<T> extends Datum {
+
+    /**
+     * Unpack the inner variant value.
+     *
+     * @return T
+     */
+    T unpack();
```
johnedquinn, line 19: It was my assumption that PType.Kind would have another variant: `VARIANT` with some method to expose the parameterization. And, a `Datum` holding type `VARIANT` would allow for the invocation of two methods: a new `Datum unpack()` and the existing `byte[] getBytes()`.

Doing it like this, we could still unpack the variant to a PartiQL value (Datum), and _if_ we expose `IonVariant` as a public class, then Ion's API of `IonElement` wouldn't be exposed either. I _think_ it would provide enough APIs for variant consumers (ourselves included) to create a DatumReader/Writer for the variant type. Also, it would allow for us to internalize `Variant`, `IonDatumReader`, `IonDatumWriter`, and `IonVariant`.

partiql-spi/src/main/java/org/partiql/spi/value/DatumReader.java
```diff
@@ -0,0 +1,21 @@
...
+ *  - public void reset();
+ *  - public void skip(long n);
+ */
+public interface DatumReader extends AutoCloseable {
+
+    /**
+     * @return next Datum or null.
+     */
+    @Nullable
+    public Datum next();
```
johnedquinn, line 20: Why wouldn't this look a bit more like the `IonReader` with `stepIn()` and `stepOut()`?

Kind of similar to `IonReader`, the constructors for `IonTextReader` takes in an IonValue or stream of data , and it doesn't return `IonValues` -- it returns primitives as you walk the data. I assumed this would be similar. They do expose an `IonElementLoader` which seemingly wraps a reader and creates an `IonElement`, but it's different than the reader itself.

As a side, this came into my mind because I saw the JSON directory placeholder. JSON doesn't allow for multiple top-level values, so the `DatumReader` would always just return a single `Datum`. As a consumer of this, you'd still need to know the in's and out's of the Datum API.

partiql-spi/src/main/java/org/partiql/spi/value/DatumWriter.java
```diff
@@ -0,0 +1,16 @@
...
+ * {@see java.io.Writer}
+ */
+public interface DatumWriter extends AutoCloseable {
+
+    /**
+     * Like java.io.Reader with combined `append` and `write` since this does not implement Appendable.
+     *
+     * @param datum to write.
+     */
+    public DatumWriter write(Datum datum);
```
johnedquinn, line 15: I've got similar thoughts as the `DatumReader` comments.

partiql-spi/src/main/java/org/partiql/spi/Enum.java
RCHowell: Yeah I figured since not used directly, parameterized, and requires explicit import then there's little to no ambiguity.

partiql-spi/src/main/java/org/partiql/spi/value/Variant.java
RCHowell: Let me know if this clarifies the model.

* A variant _value_ is a datum (partiql _value_) but holds an inner type (hence unpack)
* You do not unpack the variant to a value because it already is a value – variant extends datum
* The VARIANT PType is parameterized by a _string_ aka `variant("ion"), variant("json"), variant("cbor")`
* Unpack is to provide direct access (like old .ionValue)
* Pack is to encode as bytes for the writer.

Hopefully this model is a bit more clear? To address the concerns

* no need to unpack, a _variant_ is a _datum_ implementation
* DatumReader/DatumWriter are the interfaces to read some encoding of PartiQL values into the Datum interface
* Variant, IonDatumReader, IonDatumWriter, and IonVariant are explicitly public.

partiql-spi/src/main/java/org/partiql/spi/value/DatumReader.java
RCHowell: The reader/write are really the two functions "encode" and "decode" you could conceive of backing them by a cursor-like API, but that is not the intention here. Notice how the methods return datums and not primitives. These encode a Datum into some bytes, and can decode back into the Datum interface.

The consumer of a reader is not responsible for how the value is traversed, that's what the Datum interface allows for them. This is simple about getting bytes into and outof the Datum interface.

I don't believe you would need to know the in's and out's of Datum to use this. It just wraps a stream of one-or-more values. Yes, json is one top-level value, but there's others like

* jsonlines
* yaml
* csv
* etc.

which allow for multiple values in a stream

partiql-spi/src/main/java/org/partiql/spi/value/DatumWriter.java
RCHowell: Yes, this is intentionally not a cursor. The functionality is really just

```
write(value) -> bytes
read(bytes) -> value
```

partiql-spi/src/main/java/org/partiql/spi/value/Variant.java
RCHowell: I see you are describing adding variant to the datum fat interface, I thought about that but variant is parameterized on some type. I'm not sure of a better solution without parameterizing Datum.

johnedquinn: > no need to unpack, a variant is a datum implementation

Yes, exactly. We've established that variant is a runtime type and should be reflected in our runtime values.

In the `IonVariant` impl, there is a TODO to return `PType.variant("ion")` on `getType()`. That's where it's a bit odd to me, because now, we've broken the contract between `getType` and any of the other getters. It seems like the contract would be better between `PType.Kind.VARIANT` and `Datum.unpack(): Datum`.

We're back to a similar `.check<Variant>()` as before, where we can't fully trust the relationship between `Datum#getType()` and the getters. For example:
```kotlin
// This is all pseudo-code, for brevity
class Udf : Function {
    override fun signature() = Signature("udf", params = listOf(PType.variantIon()))

    override fun eval(args: Array<Datum>) {
        val datum = args[0]
        val variant: IonVariant = datum.check<IonVariant>() // which will throw a TypeCheckException and perform the cast
        val underlyingValue: IonElement = datum.unpack()
        return // whatever we want
    }
}
```

Rather than:
```kotlin
// This is all pseudo-code, for brevity
class Udf : Function {
    override fun signature() = Signature("udf", params = listOf(PType.variantIon()))

    override fun eval(args: Array<Datum>) {
        val underlyingValue: Datum = datum.unpack() // Or, we can call datum.getBytes()
        return // whatever we want
    }
}
```

This is especially apparent when we want to be in charge of casting. How would we know which Datum getter to invoke when `getType()` returns `PType.variant("ion")`?

johnedquinn (approved):

PR #1650, opened 2024-11-19
[v1] Remove PartiQLValue from AST; refactor AST literals

alancai98:
## Relevant Issues
- Closes https://github.com/partiql/partiql-lang-kotlin/issues/1589

## Description
- Removes `PartiQLValue` from AST
- Refactors AST literals to own representation

## Other Information
  - No on v1 branch.

  - Yes but on v1 branch

## License Information


partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
```diff
@@ -126,26 +144,92 @@ internal object RexConverter {
...
+        private fun Literal.toPType(): PType {
+            return when (this) {
+                is LiteralNull, is LiteralMissing -> PType.unknown()
+                is LiteralString -> PType.string()
+                is LiteralBool -> PType.bool()
+                is LiteralDecimal -> PType.decimal(this.value.precision(), this.value.scale())
+                is LiteralInt -> PType.integer()
+                is LiteralLong -> PType.bigint()
+                is LiteralDouble -> PType.real()
+                is LiteralTypedString -> {
```
alancai98, line 165: (self-review) move the datetime literal validation to the ast->plan conversion.

alancai98:
Marking as a draft to look at the conformance test failures.

partiql-ast/src/main/java/org/partiql/ast/Explain.java
```diff
@@ -16,14 +16,13 @@
 @Builder(builderClassName = "Builder")
 @EqualsAndHashCode(callSuper = false)
 public class Explain extends Statement {
-    // TODO get rid of PartiQLValue once https://github.com/partiql/partiql-lang-kotlin/issues/1589 is resolved
     @NotNull
-    public final Map<String, PartiQLValue> options;
+    public final Map<String, Literal> options;
```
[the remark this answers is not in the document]
RCHowell: Literal makes sense

partiql-ast/src/main/java/org/partiql/ast/literal/LiteralInt.java
```diff
@@ -0,0 +1,16 @@
...
+ */
+@EqualsAndHashCode(callSuper = false)
+public class LiteralInt extends Literal {
+    public int value;
+
+    public LiteralInt(int value) {
+        super(String.format("%d", value));
+        this.value = value;
+    }
+}
```
RCHowell, line 16: I don't believe there should be both a LiteralInt and LiteralLong, there should be one "integral literal" which may just be called LiteralInteger.

RCHowell: To be honest, it would be nice to just have LiteralNumber which has the necessary APIs / details to model floating point, decimals, and integers. But also there are places in the BNF where we might want to enforce exact numeric or integral.

RCHowell (changes requested):

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
```diff
@@ -127,26 +145,104 @@ internal object RexConverter {
...
         }
 
+        private fun Literal.toPType(): PType {
+            return when (this) {
+                is LiteralNull, is LiteralMissing -> PType.unknown()
+                is LiteralString -> PType.string()
+                is LiteralBool -> PType.bool()
+                is LiteralDecimal -> {
+                    if (this.value.scale() == 0) {
+                        PType.numeric()
```
alancai98, line 164: TODO Alan. since we're keeping SQL's `numeric`, need to
- look into modeling of `PType.numeric`, `Datum.numeric()`, and `PartiQLValue.intValue` further
- depending on the AST modeling of literal numbers, may also add a numeric

partiql-ast/src/main/java/org/partiql/ast/literal/LiteralInt.java
alancai98: Yeah -- the literals, especially the numerics, had a lot of different ways to model. I played around w/ some different approaches but wasn't yet sure of which stage (i.e. parser, ast, planner) should perform the validation and conversion of the literal text.

I'll sync offline w/ you and @johnedquinn on where the text -> value conversion should take place and where to perform any validation.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
alancai98: Checked w/ John -- the behavior of PType.numeric will get changed in some upcoming changes next week. Added a todo for now.

partiql-ast/src/main/java/org/partiql/ast/literal/LiteralApprox.java
```diff
@@ -0,0 +1,52 @@
+package org.partiql.ast.literal;
```
alancai98, line 1: I discussed a different modeling with @RCHowell using just one class rather than three for the numerics (`LiteralApprox`, `LiteralExact`, `LiteralInt`) that used a enum/kind to distinguish between the three numeric types.

```Java
public class LiteralNumber extends Literal {
    @Nullable
    private final Long p1;

    @Nullable
    private final BigDecimal p2;

    @NotNull
    public final ParseContext kind;

    private LiteralNumber(@Nullable Long p1, @Nullable BigDecimal p2, @NotNull ParseContext kind) {
        this.p1 = p1;
        this.p2 = p2;
        this.kind = kind;
    }

    // Factory methods
    public static LiteralNumber integer(long value) {
        return new LiteralNumber(value, null, ParseContext.INTEGER);
    }

    public static LiteralNumber integer(int value) {
        return new LiteralNumber((long) value, null, ParseContext.INTEGER);
    }

    public static LiteralNumber exact(BigDecimal value) {
        return new LiteralNumber(null, value, ParseContext.EXACT);
    }

    public static LiteralNumber approx(BigDecimal value, long exponent) {
        return new LiteralNumber(exponent, value, ParseContext.APPROX);
    }

    public static LiteralNumber approx(BigDecimal value) {
        return new LiteralNumber(null, value, ParseContext.APPROX);
    }

    public static LiteralNumber approx(float value) {
        return approx(BigDecimal.valueOf(value));
    }

    public static LiteralNumber approx(double value) {
        return approx(BigDecimal.valueOf(value));
    }

    // Getting the value out
    @NotNull
    public BigDecimal getDecimal() {
        if (kind == ParseContext.INTEGER) {
            return p2;
        } else {
            throw new IllegalStateException("Unknown context: " + kind);
        }
    }

    public long getInteger() {
        if (kind == ParseContext.INTEGER) {
            return p1;
        } else {
            throw new IllegalStateException("Unknown context: " + kind);
        }
    }

    // similarly for getDouble

    // TODO if we keep this representation, change to extend `AstEnum`
    public enum ParseContext {
        INTEGER,
        EXACT,
        APPROX
    }

    @NotNull
    @Override
    public String getText() {
        switch (kind) {
            // Since they're nullable, can be slightly annoying to extract the value but it's internal code
            case INTEGER:
                assert p1 != null;
                return p1.toString();
            case EXACT:
                assert p2 != null;
                return p2.toString();
            case APPROX:
                assert p1 != null;
                assert p2 != null;
                return p2 + "e" + p1;
            default:
                throw new IllegalStateException("Unknown context: " + kind);
        }
    }
}
```

and if we wanted to internalize the enum, could add some additional methods
```Java
    public boolean isInteger() {
        return kind == ParseContext.INTEGER;
    }

    public boolean isExact() {
        return kind == ParseContext.EXACT;
    }

    public boolean isApprox() {
        return kind == ParseContext.APPROX;
    }
```

alancai98: imo, the customer code using these different modelings is pretty similar. But I feel like the internal implementation code is simpler if we break apart the `LiteralNumeric` class into separate classes since we don't have to do the `null` checks or casing on the enum/kind.

Using the single class with an enum
```Kotlin
val v = when (lit) {
    is LiteralNumber -> {
        val kind = lit.kind
        when (kind) {
            LiteralNumber.ParseContext.EXACT -> {
                lit.decimal
            }
            LiteralNumber.ParseContext.INTEGER -> {
                lit.integer
            }
            LiteralNumber.ParseContext.APPROX -> {
                lit.double
            }
            else -> error("Unexpected numeric literal: $lit")
        }
    }
    else -> error("Unexpected literal: $lit")
}
```

Not using the enum directly but using helper methods
```Kotlin
val v2 = when (lit) {
    is LiteralNumber -> {
        if (lit.isExact) {
            lit.decimal
        }
        else if (lit.isInteger) {
            lit.integer
        }
        else if (lit.isApprox) {
            lit.double
        } else {
            error("Unexpected numeric literal: $lit")
        }
    }
    else -> error("Unexpected literal: $lit")
}
```

Using current PR's three classes
```Kotlin
val v3 = when (lit) {
    is LiteralInteger -> lit.integer
    is LiteralExact -> lit.decimal
    is LiteralApprox -> lit.double
    else -> error("Unexpected literal: $lit")
}
```

partiql-ast/src/main/java/org/partiql/ast/literal/LiteralExact.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @NotNull
+    public static LiteralExact litExact(BigInteger value) {
+        return new LiteralExact(new BigDecimal(value));
+    }
+
+    @NotNull
+    public BigDecimal getDecimal() {
+        return value;
+    }
```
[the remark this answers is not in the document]
alancai98: Discussed offline. I had not considered those interface methods in the customer code I was testing out. Agree with your comment that the typical OO approach (i.e. effective java item 23 to prefer class hierarchies to tagged classes) may not be what we want here. Something more akin to PType, Datum, and AnyElement could offer a more cohesive interface.

As mentioned by @RCHowell, the Calcite follows the tagged class approach using their `SqlTypeName` property -- https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/SqlLiteral.java. I'll try out a similar approach for literal modeling.

johnedquinn (changes requested):

partiql-ast/src/main/java/org/partiql/ast/literal/LiteralApprox.java
```diff
@@ -0,0 +1,52 @@
...
+import lombok.EqualsAndHashCode;
+import org.jetbrains.annotations.NotNull;
+
+import java.math.BigDecimal;
+
+/**
+ * TODO docs
+ */
+@EqualsAndHashCode(callSuper = false)
+public class LiteralApprox extends Literal {
```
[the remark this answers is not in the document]
alancai98: I had originally thought this as well. There are some edge cases here. For example, `123` and `123.` cannot both be represented by a BigDecimal alone since BigDecimal doesn't preserve whether the `123` mantissa is specified with or without a decimal point.

We could provide some boolean field to distinguish between the two (e.g. `hasPeriod`); however I came to release that approximate numeric also has this same issue and may need that field as well. Example: `123e0` vs `123.e0`. To exactly preserve the text, we would even need to distinguish whether the approximate numeric's exponent had an implicit plus (`123e0`), explicit plus (`123e+0`), or explicit minus (`123e-0`).

This led me to think perhaps we should just be passing around strings for the AST representation of numerics, leaving the parser to perform simple validation and type tagging.

partiql-parser/src/main/antlr/PartiQLTokens.g4
```diff
@@ -388,12 +388,16 @@ LITERAL_STRING
...
-    | '.' DIGIT DIGIT* ([e] [+-]? DIGIT+)?
-    | DIGIT DIGIT* ([e] [+-]? DIGIT+)?
+LITERAL_DECIMAL
+    : DIGIT+ '.' DIGIT*
+    | '.' DIGIT+
+    ;
+
+LITERAL_FLOAT
+    : DIGIT+ ('.' DIGIT*)? 'E' [+-]? DIGIT+
+    | '.' DIGIT+ 'E' [+-]? DIGIT+
```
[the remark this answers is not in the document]
alancai98: I played around with some different ANTLR parsing modelings and it led to some other issues. I think the functionality is fine to parse as unsigned numerics. We could always change the modeling in the future or add some AstVisitor pass to unify unary +/- with AST numeric literals.

Changing literal parsing to be in the parser rather than lexer -- cannot use the single-quoted `'E'`, so `E` would need to be in the tokens.

Changing literal parsing in the tokens may bind too eagerly -- e.g. `1 - 23` would tokenize `-23` together.

partiql-ast/src/main/java/org/partiql/ast/literal/LiteralTypedString.java
```diff
@@ -0,0 +1,34 @@
...
+
+    private LiteralTypedString(@NotNull DataType type, @NotNull String value) {
+        this.type = type;
+        this.value = value;
+    }
+
+    @NotNull
+    public static LiteralTypedString litTypedString(@NotNull DataType type, @NotNull String value) {
+        return new LiteralTypedString(type, value);
+    }
```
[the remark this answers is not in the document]
alancai98: > To make sure people don't put in bad types, I'd consider making this a final class and then exchanging this method for litTime(int precision, @NotNull String value), litTimeZ(int precision, @NotNull String value), and so on and so forth.

I like the idea to create date/time/timestamp-specific factory methods. Perhaps I could add those after I get the general design approved first? We could always add these in a subsequent PR or release.

---
> Or, just making them their own classes (LiteralTime, LiteralTimestamp, LiteralDate). I imagine it would have the same benefits that you described in [#1650 (files)](https://github.com/partiql/partiql-lang-kotlin/pull/1650/files#r1851189538), and it would more closely match the EBNF and more straightforward.

For now, I liked the modeling of these `<data type > '<string repr in quotes>'` to be under one class and reuse the data type enum. The parser does perform some validation to confirm that the strings match the EBNF rules.

partiql-ast/src/main/java/org/partiql/ast/literal/LiteralExact.java
```diff
@@ -0,0 +1,41 @@
...
+        this.value = value;
+    }
+
+    @NotNull
+    public static LiteralExact litExact(BigDecimal value) {
+        return new LiteralExact(value);
+    }
+
+    @NotNull
+    public static LiteralExact litExact(BigInteger value) {
```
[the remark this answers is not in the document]
alancai98: We could always add them later. With the current design I didn't actually need a factory method or constructor to create an exact numeric literal using a long or int.

partiql-ast/src/main/java/org/partiql/ast/literal/LiteralApprox.java
```diff
@@ -0,0 +1,52 @@
...
+    @NotNull
+    private final BigDecimal mantissa;
+
+    private final int exponent;
```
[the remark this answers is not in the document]
alancai98: TBH I'm not quite sure what **needs** to be public atm. When looking at an approximate numeric literal, are there use cases where an API user would actually need just the mantissa or just the exponent? The use cases we currently have use both the arguments (e.g. pretty-printing, AST->plan conversion).

partiql-ast/src/main/java/org/partiql/ast/literal/Literal.java
```diff
@@ -0,0 +1,125 @@
...
+import org.partiql.ast.AstNode;
+import org.partiql.ast.AstVisitor;
+import org.partiql.ast.DataType;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.Collection;
+import java.util.Collections;
+
+public abstract class Literal extends AstNode {
```
alancai98, line 13: Latest design currently models things a bit differently than before. Key summary is
- fat interface w/ a `LiteralKind` enum. Includes relevant constructors and property accessor methods. much like PType, Datum, AnyElement, and Calcite's [SqlLiteral](https://calcite.apache.org/javadocAggregate/org/apache/calcite/sql/SqlLiteral.html)
- Three numeric variants (approx, int, exact) with each containing just a string
  - The parser will still perform some type ascription
  - Modeling certain numerics with a string proved less cumbersome. E.g.
    - `123` vs `123.` -- BigDecimal does not actually preserve whether there was a period or not
    - `123e0` vs `123.e0` -- BigDecimal for the mantissa of an approximate does not preserve whether there was a period or not
    - `123e0` vs `123.e+0` -- a long/integer for the exponent does not preserve whether there was an explicit `+`
  - Simplifying to just a string allows us to hold off on any eager computation
- Date/time/timestamp represented as a typed string
  - The parser already performs some validations and basic type ascription. 
  - One potential issue with eagerly converting the fields into java is all of the additional properties that would need added. E.g. date would need year, month, day. Time would need hour, minute, second, precision, timezone. Timestamp would need all of date and time's fields. We would need to at some point convert these AST fields to the plan's representation.

RCHowell: > One potential issue with eagerly converting the fields into java is all of the additional properties that would need added. E.g. date would need year, month, day. Time would need hour, minute, second, precision, timezone. Timestamp would need all of date and time's fields. We would need to at some point convert these AST fields to the plan's representation.


That's a good point, and I didn't think of that. We wouldn't be able to zero-pad and would need to insert nulls for those fields. Seems like a lot of extra work (on read aka scribe) when a (rule,string) is sufficient for capturing syntax which ultimately is the goal.

partiql-ast/src/main/java/org/partiql/ast/literal/Literal.java
```diff
@@ -0,0 +1,125 @@
...
+        return new LiteralMissing();
+    }
+
+    public static Literal litString(@NotNull String value) {
+        return new LiteralString(value);
+    }
+
+    public static Literal litTypedString(@NotNull DataType type, @NotNull String value) {
+        return new LiteralTypedString(type, value);
+    }
```
RCHowell, line 79: Drop the "lit" prefix from these as we are already namespaced under Literal. This way it's not redundant `Literal.lit*`. So `Literal.integer()` to avoid keyword-conflict and something like `nil()` or `nul()` or `absent(bool missing)` to avoid null keyword. Datum has to do something similar.

partiql-ast/src/main/java/org/partiql/ast/literal/Literal.java
```diff
@@ -0,0 +1,125 @@
...
+
+public abstract class Literal extends AstNode {
+    @NotNull
+    public abstract LiteralKind kind();
+
+    @Override
+    @NotNull
+    public Collection<AstNode> children() {
+        return Collections.emptyList();
+    }
```
RCHowell, line 21: In a quick follow-up PR, could you please change this to

```
public List<AstNode> getChildren() {
    ...
}
```

This way it has ordered semantics and the `getFoo` while clunky (imo) plays better with kotlin accessors.

partiql-ast/src/main/java/org/partiql/ast/literal/Literal.java
```diff
@@ -0,0 +1,125 @@
...
+import org.partiql.ast.DataType;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.Collection;
+import java.util.Collections;
+
+public abstract class Literal extends AstNode {
+    @NotNull
+    public abstract LiteralKind kind();
```
RCHowell, line 15: Not exactly a fat-interface but similar pattern! We are doing a tagged-union, and we don't actually want a LiteralKind enum and the code (aka tag) should go directly on the Literal (like we did with DataType).

RCHowell (changes requested):

partiql-ast/src/main/java/org/partiql/ast/literal/Literal.java
alancai98: I'll try using the code directly on the Literal. I had ran into some slightly repetitive + confusing code pattern when using a LiteralKind:

```
val litText = when (lit.kind().code()) {
    LiteralKind.NULL -> "NULL"
    LiteralKind.MISSING -> "MISSING"
    LiteralKind.BOOLEAN -> lit.booleanValue().toString()
    ...
    else -> error("Unsupported literal kind ${lit.kind()}")
}
```
^ `lit.kind().code()` rather than just `lit.code()`.

partiql-ast/src/main/java/org/partiql/ast/literal/Literal.java
alancai98: > capturing syntax which ultimately is the goal

Yeah that was my mindset w/ this latest refactor -- AST at the end of the day just captures syntax with some slight validation w/ the parser. I think most users are not actually performing computation directly on the AST. Computation (like in our case) will come later in the plan or even pushed all the way to the evaluator.

partiql-ast/src/main/java/org/partiql/ast/literal/Literal.java
alancai98: I had tried dropping the `lit` prefix but ran into the keyword-conflict with bool and null. I'll change to some other spelling as per your suggestion.

partiql-ast/src/main/java/org/partiql/ast/literal/Literal.java
alancai98: Sure. I'll track this as a followup in https://github.com/partiql/partiql-lang-kotlin/issues/1610.

partiql-ast/src/main/java/org/partiql/ast/literal/Literal.java
johnedquinn, line 79: If you want to be as restrictive as possible, you can have explicit `time(boolean withTimeZone, int precision, @NotNull String value)` for the datetime values, rather than the typed string allowing all types.

partiql-ast/src/main/java/org/partiql/ast/literal/Literal.java
johnedquinn, line 13: Because of the flexibility PType provides with the fat interface, I explicitly wrote in the Javadocs that consumers of this should not author their own implementations. See https://github.com/partiql/partiql-lang-kotlin/blob/f5c6efff22543c8bb9da3fba0cb0c3cd21ea8f42/partiql-types/src/main/java/org/partiql/types/PType.java#L23-L24.

Could add that to this API as well.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
```diff
@@ -127,26 +139,101 @@ internal object RexConverter {
...
+                    val typedString = this.stringValue()
+                    when (type.code()) {
+                        DataType.DATE -> {
+                            val value = LocalDate.parse(typedString, DateTimeFormatter.ISO_LOCAL_DATE)
+                            val date = DateTimeValue.date(value.year, value.monthValue, value.dayOfMonth)
+                            dateValue(date)
+                        }
+                        DataType.TIME, DataType.TIME_WITH_TIME_ZONE -> {
+                            val time = DateTimeUtils.parseTimeLiteral(typedString)
+                            val precision = type.precision ?: 6
```
johnedquinn, line 221: I believe these are safe to `!!`.

partiql-ast/src/main/java/org/partiql/ast/literal/Literal.java
alancai98: I ended up changing this to not be an abstract class and a concrete class extending `AstEnum` (like `DataType`) -- https://github.com/partiql/partiql-lang-kotlin/pull/1650#discussion_r1873778924. So adding another implementation shouldn't be an issue anymore.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
alancai98: `DataType`'s `precision` method returns a nullable integer (null indicates no precision was explicitly specified like `TIME WITH TIME ZONE '12:34:56'`, so I think we still need to check for null w/ the elvis operator.

alancai98: Hmm actually seems like the parser is eagerly inferring the precision from the time/timestamp string when the type precision argument is unspecified -- https://github.com/partiql/partiql-lang-kotlin/blob/v1-ast-remove-partiqlvalue/partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt#L2201-L2208. I don't think we should be doing this eager inference in the parser so I'll change.

partiql-ast/src/main/java/org/partiql/ast/literal/Literal.java
alancai98: I still think there's some value in having a public factory method to construct typed string literals w/ arguments `DataType` and `String`. For instance, if an API user had a `DataType` already (date, time, timetz, timestamp, timestamptz), they could simply call this factory method. If we only make the specific typed string methods public (e.g. `time(boolean withTimeZone, int precision, @NotNull String value)`, then they would need to perform additional conversion logic.

We could always add such methods in the future.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -2202,41 +2192,6 @@ internal class PartiQLParserDefault : PartiQLParser {
...
-        }
-
-        private fun getPrecisionFromTimeString(timeString: String): Int {
-            val matcher = GENERIC_TIME_REGEX.toPattern().matcher(timeString)
-            if (!matcher.find()) {
-                throw IllegalArgumentException("Time string does not match the format 'HH:MM:SS[.ddd....][+|-HH:MM]'")
-            }
-            val fraction = matcher.group(1)?.removePrefix(".")
-            return fraction?.length ?: 0
-        }
```
alancai98, line 2238: (self-review): this prior parsing logic was deleted for the following reasons
- was eagerly inferring the seconds precision for TIME and TIMESTAMP when the precision argument was unspecified. Thus it was classifying `TIME '12:34:56'` as the same as `TIME (0) '12:34:56'"`
- applying the time string's regex check for timestamp
- only performing the string regex check when the precision was unspecified

partiql-ast/src/main/java/org/partiql/ast/literal/Literal.java
alancai98: Changed to be more like `DataType` in the lateset commit.

partiql-ast/src/main/java/org/partiql/ast/Literal.java
```diff
@@ -0,0 +1,254 @@
...
+import java.util.Collection;
+import java.util.Collections;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * TODO docs
+ */
+@EqualsAndHashCode(callSuper = false)
+public class Literal extends AstEnum {
```
RCHowell, line 17: Modeling looks good to me; very concise union and easy to extend if needed.

RCHowell (approved):
Approved on the modeling, no comment/preferences on changes to partiql-parser since the proof is in the tests and the details are internal. I have some nits but I'll spare you those considering the cycles of this PR. Thanks for closing on this.

Issue #1572, opened 2024-08-28
[V1] Remove symbol and sexp from PType and Datum


RCHowell:
## Description 

Symbol and S-Expression types are to be removed from PType (and Datum) until some definitive decision is made on their place in the type system.

We still ready Ion in backticks like so,

```
-- sexp example
`(1 2 3)` -> [ 1, 2, 3 ] -> type: ARRAY

-- symbol example
`foo` -> 'foo' -> type: STRING
```

RCHowell:
Closed by #1651

PR #1651, opened 2024-11-19
Fixes conformance failures

johnedquinn:
## Description
- Fixes conformance failures stemming from the merge of #1647 and #1572.
- To do this:
  - Added precision/scale to AVG
  - Rewrote more operators (MULT, DIV, MOD, IN, etc)
  - Adds the comparison operators (GTE, GTE, etc)
  - Added a DatumReader. The PartiQLValue to Datum conversion didn't maintain precision/scale.
  - Fixed the null-call property of the EQ function.
- For more in-depth info, see each commit's expanded message.

## License Information


partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnGt.kt
```diff
@@ -4,201 +4,138 @@
...
+
+    override fun getBigIntInstance(bigIntLhs: PType, bigIntRhs: PType): Function.Instance {
+        return basic(PType.bool(), PType.bigint()) { args ->
+            val lhs = args[0].long
+            val rhs = args[1].long
+            Datum.bool(lhs > rhs)
+        }
+    }
+
+    // TODO: Update
```
alancai98, line 53: Guessing this TODO is related to the upcoming change for updating `numeric` representation? If so, could add a bit more context in the TODO comments in this and the other comparison functions.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/io/DatumIonReader.kt
```diff
@@ -0,0 +1,369 @@
...
+            return null
+        }
+    }
+}
+
+public class DatumIonReaderBuilder private constructor(
+    private var sourceDataFormat: SourceDataFormat = SourceDataFormat.IonForPartiQL,
+    private var ionReaderBuilder: IonReaderBuilder,
+) {
+    public enum class SourceDataFormat {
```
alancai98, line 321: nit: could indicate that `IonForPartiQL` is used for the standard builder

alancai98 (commented):
Nice -- looks like we're back to 50 conformance failures (prev run before the drop - https://github.com/partiql/partiql-lang-kotlin/actions/runs/11899775807). Left some minor comments. Looks good otherwise.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/AggAvg.kt
```diff
@@ -10,22 +10,22 @@ import org.partiql.types.PType
 
 internal val Agg_AVG__INT8__INT8 = Aggregation.static(
     name = "avg",
-    returns = PType.decimal(),
+    returns = PType.decimal(38, 19),
```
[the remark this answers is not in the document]
johnedquinn: Ah, yes. I discussed with @jpschorr about what the appropriate precision/scale of average should be. It comes down to the fact that `AVG(x)` = `SUM(x) / COUNT(x)`. Since `COUNT` returns a BIGINT (JVM long) (which has a decimal precision of 19). We were originally failing conformance tests with AVG since the declared output type was a decimal with scale 0, so the average of 2 and 3 was returning 2 rather than 2.5. So, we opted for a max precision (38) followed by the precision of BIGINT for the division. It needs to be verified via RFC, but for now, we need to retain the fractional parts to some degree.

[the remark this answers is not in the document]
alancai98: I see. Perhaps we could include a TODO in the code and/or cut an issue?

Code-style wise, I think you should add some private top-level variable that can be reused by the AVG function instances that return `PType.decimal(38, 19)`.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnModulo.kt
```diff
@@ -4,126 +4,78 @@
...
+
+    override fun getDecimalInstance(decimalLhs: PType, decimalRhs: PType): Function.Instance {
+        val p = decimalLhs.precision - decimalLhs.scale + decimalRhs.scale + Math.max(6, decimalLhs.scale + decimalRhs.precision + 1)
+        val s = Math.max(6, decimalLhs.scale + decimalRhs.precision + 1)
+        return basic(PType.decimal()) { args ->
+            val arg0 = args[0].bigDecimal
+            val arg1 = args[1].bigDecimal
+            Datum.decimal(arg0 % arg1, p, s)
+        }
+    }
```
alancai98, line 64: The computation for precision and scale is still following SQL Server's division rather than modulo. Should be 

p = min(p1 - s1, p2 - s2) + max(s1, s2)
s = max(s1, s2)

alancai98: Also would be helpful for future maintenance to include a comment on the precision/scale computation

alancai98 (changes requested):

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/AggAvg.kt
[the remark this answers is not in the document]
johnedquinn: Added!

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnModulo.kt
johnedquinn: Added!

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnBitwiseAnd.kt
```diff
@@ -8,9 +8,10 @@ import org.partiql.spi.value.Datum
...
 import kotlin.experimental.and
 
-internal object FnBitwiseAnd : ArithmeticDiadicOperator() {
-    override fun getName(): String {
-        return "bitwise_and"
+internal object FnBitwiseAnd : DiadicArithmeticOperator("bitwise_and") {
+
+    init {
+        fillTable()
     }
```
RCHowell, line 15: nit and don't change anything, but I think the pattern here is "must call super" which is fairly common and might have some annotations to compile-time enforce this, but I wouldn't change it

RCHowell (approved):

alancai98 (approved):

PR #1652, opened 2024-11-21
Added parsing support for Create Table Statement 

yliuuuu:
## Relevant Issues
- [Closes/Related To] N/A

## Description
- Scope: 
    - Table Definition (i.e., CREATE TABLE .... )
    - Table Element
    - Table Constraint (PRIMARY KEY, UNIQUE)
    - Column Declaration with OPTIONAL and COMMENT. 
    - Column Constraint (PRIMARY KEY, UNIQUE, NULL, NOT NULL, CHECK)
    - Parameterized STRUCT type and LIST type in type declaration. 
    - PARTITION BY
    - TBLPROPERTIES. 

### TABLE DEFINITION
It is worth to mention that in PartiQL, the legacy implementation of CREATE TABLE can hold no attribute/column declaration. This is: 
```
CREATE TABLE foo
```
should parse successfully. 
### Table Element
At the moment: we support two types of table elements, Table Constraint and Column Definition. 
It is worth to note that: based on the SQL grammar, table constraint and column definition are at the same level. 
```
<table element>  ::=
         <column definition>
     |   <table constraint definition>
     |   <like clause>
     |   <self-referencing column specification>
     |   <column options>
```
This means: 
```
CREATE TABLE foo (
     <table_constraint>
)
```
will pass parsing. 

### Table Constraint
SQL Constraints to be placed on Table. 

At the moment: we support Primary Key and Unique Constraint at table level. 

### Column Declaration
```
<column definition>  ::=
         <columnName> [OPTIONAL] <data type> [<column constraint def>....] [<comment>]
```
The two addition we add are: 
1. OPTIONAL keyword -> used to indicate that field may not exist in the underlying data. 
2. COMMENT clause -> used to associated an arbitrary text with the column. 

### Column Constraint
At the moment: we support 5 attribute level constraints. 
PRIMARY KEY, UNIQUE, NOT NULL, NULL, CHECK. 

The NULL Constraint is not in the SQL specification. The semantics of such is  is logically equivalent to omitting the NOT NULL constraint from the column definition.

i.e., `foo INT2 NULL` === `foo INT2`

The NULL Constraint is supported for two reasons: 
1. It is supported in the legacy parser. 
2. It is useful to provide a way for user to explicitly statement their intension regarding NULL Value. 

### Parameterized STRUCT type and LIST type in type declaration. 
<struct_type> ::= STRUCT [ANGLE_LEFT structField [(, structField )....] ANGLE_RIGHT]
<struct_field> := <column Name> [OPTIONAL] COLON <type> [<columnConstraintDef>...] [<comment>]
<array type> := ARRAY [ANGLE_LEFT type ANGLE_RIGHT]

### PARTITION BY And TBLProperties
Useful extensions that are not in the SQL spec. 

Example: 
```
CREATE TABLE ... 
PARTITION BY (foo)
TBLPROPERTIES ('k1' = 'v1')
```

### Implementation
- Antlr Grammar: 
   - The Antlr grammar is constructed based on [SQL99 EBNF grammar](https://ronsavage.github.io/SQL/sql-99.bnf.html). 
   - whenever applicable, the naming was consistent with SQL Spec. The intension was to allow cross referencing between PartiQL Grammar and SQL specification. 

- AST
   - AST implementation was inspired by [SqlParser](https://docs.rs/sqlparser/latest/sqlparser/ast/struct.CreateTable.html). 
   - Occasionally, there might be conflict noun between grammar rule and ast node. (i.e., column constraint in grammar rule get renamed to attribute constraint in AST node), for using a more generic name to support nested data type. 

## Other Information
  - N/A, Release candidate. 

  - No. 

  - No. 

Yes. 

## License Information


partiql-ast/src/main/java/org/partiql/ast/AttributeConstraint.java
```diff
@@ -0,0 +1,105 @@
...
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public abstract class AttributeConstraint extends AstNode {
+
+    // NULL & NOT NULL
```
alancai98, line 16: Could we actually just split this into two classes? `Null` and `NotNull`

partiql-ast/src/main/java/org/partiql/ast/AttributeConstraint.java
```diff
@@ -0,0 +1,105 @@
...
+            return Collections.emptyList();
+        }
+
+        @Override
+        public <R, C> R accept(@NotNull AstVisitor<R, C> visitor, C ctx) {
+            return visitor.visitNullable(this, ctx);
+        }
+    }
+
+    // Unique and primary
```
alancai98, line 46: Could we split this into `Unique` and `PrimaryKey`?

partiql-ast/src/main/java/org/partiql/ast/AttributeConstraint.java
```diff
@@ -0,0 +1,105 @@
...
+
+    // NULL & NOT NULL
+    /**
+     * TODO docs, equals, hashcode
+     */
+    @Builder(builderClassName = "Builder")
+    @EqualsAndHashCode(callSuper = false)
+    public static class Nullable extends AttributeConstraint {
+
+        public final String name;
```
alancai98, line 24: If `name` is common to every variant, then it should go under the abstract class rather than be a field of every variant.

partiql-ast/src/main/java/org/partiql/ast/TableConstraint.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+public abstract class TableConstraint extends AstNode {
+
+    @Builder(builderClassName = "Builder")
+    @EqualsAndHashCode(callSuper = false)
+    public static class Unique extends TableConstraint {
+        // TODO: Should this be an identifier
```
alancai98, line 17: SQL BNF seems to use more than just a string name -- https://ronsavage.github.io/SQL/sql-99.bnf.html#constraint%20name.

Also `name` could be a field of `TableConstraint` rather than each of the variants.

partiql-ast/src/main/java/org/partiql/ast/AstVisitor.java
```diff
@@ -61,6 +61,58 @@ public R visitStatement(Statement node, C ctx) {
         return node.accept(this, ctx);
     }
 
+    //
+    // DDL
+    //
+    public R visitDdl(Ddl node, C ctx) {
+        return node.accept(this, ctx);
```
alancai98, line 68: nit: could just use the default visit

```suggestion
        return defaultVisit(node, ctx);
```

alancai98 (commented):
Thanks for adding the parsing support back and using the new AST modeling. I still need to look further into the tests but left some representation comments for the AST.

partiql-ast/src/main/java/org/partiql/ast/DataType.java
```diff
@@ -456,115 +546,6 @@ public String name() {
         }
     }
 
-    @NotNull
```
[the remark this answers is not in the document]
yliuuuu: This is weird. I will revert.

partiql-ast/src/main/java/org/partiql/ast/AttributeConstraint.java
```diff
@@ -0,0 +1,105 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.partiql.ast.expr.Expr;
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public abstract class AttributeConstraint extends AstNode {
```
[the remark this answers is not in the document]
yliuuuu: I was thinking about the same thing.

partiql-ast/src/main/java/org/partiql/ast/Options.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public abstract class Options extends AstNode {
```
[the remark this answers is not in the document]
yliuuuu: Agree we do not need the hierarchy at the moment.

partiql-ast/src/main/java/org/partiql/ast/AttributeConstraint.java
yliuuuu: I think the boolean makes the constant folding a bit easier. 

Consider: 
```
foo INT2 NOT NULL NULL NULL NOT NULL
```

you would just grab all AttributeConstraint with Type `NULL` and folding as you wish.

partiql-ast/src/main/java/org/partiql/ast/AttributeConstraint.java
yliuuuu: Similar to above.

partiql-ast/src/main/java/org/partiql/ast/AttributeConstraint.java
yliuuuu: Lombok has some limitations on builder annotation with inheritance.

There seems to be work around with the annotation [`SuperBuilder`](https://projectlombok.org/features/experimental/SuperBuilder), but is not really familiar with this...

From their doc: 
> all superclasses also have the @SuperBuilder annotation

Not sure how this would imply with multiple level of hierarchy. 

So if Ok with you, may be we should leave this out of the scope for this CR.

partiql-ast/src/main/java/org/partiql/ast/TableConstraint.java
yliuuuu: SQL has that as identifier since it requires constraint name to be unique within the same schema... 

The constraint name semantics is quite hand wavy now in PartiQL... 

Let's discuss a little more ...

partiql-ast/src/main/java/org/partiql/ast/ColumnDefinition.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @NotNull
+    private final DataType dataType;
+
+    @NotNull
+    private final Boolean isOptional;
+
+    private final List<AttributeConstraint> constraints;
+
+    private final String comment;
```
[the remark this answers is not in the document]
yliuuuu: Yeah sorry missed this... 

Side note: Why we set all fields to be public?

partiql-ast/src/main/java/org/partiql/ast/AttributeConstraint.java
alancai98: I see. I guess I'm fine leaving it as one class since we follow a similar pattern for other negatable exprs like `BETWEEN` and `LIKE`.

Out of curiosity, for your above example what would be the constant-folded equivalent?

partiql-ast/src/main/java/org/partiql/ast/ddl/CreateTable.java
```diff
@@ -0,0 +1,66 @@
...
+ * TODO docs, equals, hashcode
+ */
+@Builder(builderClassName = "Builder")
+@EqualsAndHashCode(callSuper = false)
+public class CreateTable extends Ddl {
+
+    @NotNull
+    public final IdentifierChain name;
+
+    @Nullable
```
alancai98, line 25: Should this be nullable? In your factory method, the `columns` argument is non-null.

partiql-ast/src/main/kotlin/org/partiql/ast/Ast.kt
```diff
@@ -506,4 +512,64 @@ public object Ast {
...
+    public fun columnConstraintUnique(name: String?, isPrimaryKey: Boolean): AttributeConstraint.Unique {
+        return AttributeConstraint.Unique(name, isPrimaryKey)
+    }
+
+    @JvmStatic
+    public fun columnConstraintCheck(name: String?, searchCondition: Expr): AttributeConstraint.Check {
+        return AttributeConstraint.Check(name, searchCondition)
+    }
+
+    @OptIn(PartiQLValueExperimental::class)
```
alancai98, line 565: Redundant annotation.

partiql-ast/src/main/kotlin/org/partiql/ast/Ast.kt
```diff
@@ -506,4 +512,64 @@ public object Ast {
...
+
+    @JvmStatic
+    public fun columnConstraintUnique(name: String?, isPrimaryKey: Boolean): AttributeConstraint.Unique {
+        return AttributeConstraint.Unique(name, isPrimaryKey)
+    }
+
+    @JvmStatic
+    public fun columnConstraintCheck(name: String?, searchCondition: Expr): AttributeConstraint.Check {
+        return AttributeConstraint.Check(name, searchCondition)
+    }
```
alancai98, line 563: I haven't checked too closely, but we need to confirm which arguments should be nullable vs non-null in the factory methods compared with the constructor.

Some of the args like `createTable`'s `columns` and `constraints` are nullable in the constructor but marked as non-null in the factory method.

partiql-ast/api/partiql-ast.api
```diff
@@ -1,5 +1,10 @@
 public final class org/partiql/ast/Ast {
 	public static final field INSTANCE Lorg/partiql/ast/Ast;
+	public static final fun columnConstraintCheck (Ljava/lang/String;Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/ddl/AttributeConstraint$Check;
```
alancai98, line 3: Taking a look at the api dump, seems like we're also missing DDL support in `AstRewriter` and `SqlDialect`. If those are tackled later, can we add TODOs in those classes?

partiql-ast/src/main/java/org/partiql/ast/TableConstraint.java
alancai98: To followup on this comment, will the PartiQL RFC for CREATE TABLE opt for a string name here or an identifier?

alancai98 (commented):
I think we're almost ready to merge in. Mostly left some minor comments/suggestions. Perhaps need to follow up on the constraint `name` for whether it should be a simple string or identifier.

partiql-ast/src/main/java/org/partiql/ast/ddl/AttributeConstraint.java
```diff
@@ -0,0 +1,109 @@
...
+import org.partiql.ast.AstVisitor;
+import org.partiql.ast.expr.Expr;
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * TODO docs, equals, hashcode
+ */
+public abstract class AttributeConstraint extends AstNode {
```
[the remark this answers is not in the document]
yliuuuu: Named attribute constraint because we use the same object in Struct Field construction. i.e., `Struct<foo: INT2 NOT NULL>`. 

The mental model is to have the grammar file follows SQL as close as possible, but rename certain things in AST to better fits our use case.

partiql-ast/src/main/java/org/partiql/ast/ddl/AttributeConstraint.java
```diff
@@ -0,0 +1,109 @@
...
+     * TODO docs, equals, hashcode
+     */
+    @Builder(builderClassName = "Builder")
+    @EqualsAndHashCode(callSuper = false)
+    public static class Unique extends AttributeConstraint {
+        @Nullable
+        public final String name;
+
+        @NotNull
+        public final Boolean isPrimary;
```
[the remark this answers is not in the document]
yliuuuu: The table one is separated because the possibility of having different target (hence not so straight forward to do constant folding). 

i.e., 

```
CREATE TABLE foo (
   attr1 INT2, 
   attr2 INT2, 
   PRIMARY KEY (attr1),
   UNIQUE (attr2)
)
```

For inline ones: 
Consider :

1. 
```
CREATE TABLE foo (
   attr1 INT2 PRIMARY KEY UNIQUE
   attr2 INT2
)
```

The same as
```
CREATE TABLE foo (
   attr1 INT2 PRIMARY KEY
   attr2 INT2
)
```

2. 
```
CREATE TABLE foo (
   attr1 INT2 PRIMARY KEY UNIQUE
   attr2 INT2
)
```

The same as
```
CREATE TABLE foo (
   attr1 INT2 PRIMARY KEY
   attr2 INT2
)
```

3. 
```
CREATE TABLE foo (
   attr1 INT2 PRIMARY KEY UNIQUE PRIMARY KEY
   attr2 INT2
)
```

Potentially illegal **or** the same as
```
CREATE TABLE foo (
   attr1 INT2 PRIMARY KEY
   attr2 INT2
)
```

So in the constraint analysis: 

1. for each node: constraint folding
2. for across column declaration: select all inline constraint that are unique -> fold to collection unique column and note inline PK. If multiple throw: 
3. for table analysis: append unique constraint columns to the unique set, make decision on valid primary key declaration based on if inline PK exists.

partiql-ast/src/main/java/org/partiql/ast/ddl/CreateTable.java
yliuuuu: Good callout: 

In case of 
```
CREATE TABLE foo
```

do we prefer List<ColumnDefinition> and List<TableConstraint> constraints to be null, or do we prefer those to be emptyList()

partiql-ast/src/main/java/org/partiql/ast/TableConstraint.java
yliuuuu: Actually: let's make the constraint name a identifier (simple identifier) for now...

Not sure what would end up being in the RFC. Practically: we probably just going to leave the semantics of identifier name out of scope. Something like: 

> The PartiQL parser supports parsing of constraint name (as simple identifier, regular or delimited). However, the operation semantics for constraint name is out of scope. 

In implementation, practically speaking, as PartiQL does not have a persistence metadata layer, it is difficult for it to support things like assertion constraint name conflict across schema, or operations such as drop constraint via name. (Potentially possible via SPI, but not in the road map as for today).

yliuuuu: Discussed offline: In the sprit of following SQL, will make the constraint name a IdentifierChain.

partiql-ast/src/main/java/org/partiql/ast/AttributeConstraint.java
yliuuuu: Discussed offline. 

The constant-folded equivalent maybe latest wins or invalid semantics.

partiql-ast/src/main/java/org/partiql/ast/ddl/AttributeConstraint.java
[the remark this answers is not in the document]
yliuuuu: Offline discussion: We want to make the Unique/Primary key constraint at attribute level and at table table level modeled more consistently. 

I will remodel the table level PK /Unique Constraint to use the boolean flag.

partiql-ast/src/main/java/org/partiql/ast/AstVisitor.java
```diff
@@ -61,6 +68,50 @@ public R visitStatement(Statement node, C ctx) {
...
+        return defaultVisit(node, ctx);
+    }
+
+    public R visitUnique(AttributeConstraint.Unique node, C ctx) {
+        return defaultVisit(node, ctx);
+    }
+
+    public R visitCheck(AttributeConstraint.Check node, C ctx) {
+        return defaultVisit(node, ctx);
+    }
```
alancai98, line 108: nit: could consider using the full name in the `visit*`. E.g. `visitUnique` -> `visitAttributeConstraintUnique` or if you can think of an appropriate way to shorten.

It might be confusing if some of the function names clash.

partiql-ast/src/main/java/org/partiql/ast/ddl/AttributeConstraint.java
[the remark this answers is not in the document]
alancai98: From offline discussion, column here can be confusing since these constraints can be applied to structs and not just top-level table columns. Field is another consideration but applies to row types from the SQL1999 spec. And attribute refers to structured types. So leaving it as attribute seems reasonable so long as we add the appropriate docs in the future.

alancai98 (approved):
One last minor suggestion for the visitor function names. Otherwise looks good to me.

Issue #1654, opened 2024-11-25
[V1] Tracking issue for sink and source APIs for PartiQL data streams.


RCHowell:
## Description

This issue is a tracking issue for several PRs in a sequence to finalize value representations along with the reader/writer implementations.

## Tasks

* [ ] Move datetime to values spi
* [ ] Package datum into SPI datum
* [ ] Push sink/source APIs for PartiQL streams
* [ ] Add default sink/source implementations.
* [ ] Split ion/json/csv implementations from SPI package.
* [ ] Define type provider interface for different variant forms.

PR #1655, opened 2024-11-25
[V1] Defines source and sink APIs for PartiQL data streams.

RCHowell:
## Description

This PR adds a source and sink API to the SPI package which will become the backing interfaces for various PartiQL encoding schemes (firstly Ion). These APIs are just like the reader/writer APIs from IonJava, but are intended to be extended to get values into a PartiQL system. The primary/default application here will be to encode PartiQL data in typed Ion.

See #1654 

## Other Information

## License Information

### DIFF
--- a/partiql-spi/src/main/java/org/partiql/spi/stream/PSource.java
+++ b/partiql-spi/src/main/java/org/partiql/spi/stream/PSource.java
@@ -1,0 +1,86 @@
+package org.partiql.spi.stream;
+
+import org.jetbrains.annotations.NotNull;
+import org.partiql.types.PType;
+import org.partiql.value.datetime.Date;
+import org.partiql.value.datetime.Time;
+import org.partiql.value.datetime.Timestamp;
+
+import java.math.BigDecimal;
+
+/**
+ * This is a PartiQL value stream source.
+ * <br>
+ * Developer Note:
+ * - There should be a method for every Datum *java* value and all PType arguments.
+ * - Method names are derived from PType.Kind as pascal case.
+ */
+public interface PSource {
+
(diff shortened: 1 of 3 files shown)


partiql-spi/src/main/java/org/partiql/spi/stream/PSource.java
```diff
@@ -0,0 +1,86 @@
...
+
+    short readSmallint();
+
+    int readInt();
+
+    long readBigint();
+
+    @NotNull
+    BigDecimal readDecimal();
+
```
RCHowell, line 45: missing

```
@NotNull
BigDecimal readNumeric();
```

partiql-spi/src/main/java/org/partiql/spi/stream/PSource.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /**
+     * Positions the source to the next value, return its type.
+     */
+    PType next();
+
+    void readNull();
+
+    void readMissing();
```
RCHowell, line 31: likely not necessary as these have no effect

partiql-spi/src/main/java/org/partiql/spi/stream/PSource.java
```diff
@@ -0,0 +1,86 @@
...
+public interface PSource {
+
+    default void close() {
+        // no-op
+    }
+
+    /**
+     * Positions the source to the next value, return its type.
+     */
+    PType next();
```
RCHowell, line 27: Annotate as nullable

RCHowell:
Replaced by #1687

Issue #1664, opened 2024-11-27
[V1] Post rc.3 plan cleanup follow-up issue


rchowell:
## Description

This issue holds a list of follow-up actions for post rc.3 plan development.

## Tasks

- [ ] Collation getColumn should return a column index rather than an expression.
- [ ] RelAggregate group strategy
- [ ] Document all arg ordering (import because of efficient serde and rewriter rules/strategies!)
- [ ] #1575 
- [ ] Planner subquery support (proper)
- [ ] Consider non-lateral cross join and equijoins (both not in spec)
- [ ] Add back stack-style builders
- [ ] Consider "replace" methods for mutability and "copy" for immutable patterns
- [ ] Document children with vararg children support (coalesce, array, bag, etc)
- [ ] Fix cross join in planner to use correlate
- [ ] Make error return UNKNOWN rather than dynamic.

Issue #1661, opened 2024-11-27
[V1] Plan cleanup and finalization


RCHowell:
## Description

This is a tracking issue for plan cleanup efforts and finalization, see #1665 

## Tasks

- [x] Convert interfaces to Java
- [x] Make full tree since we have the Operator interface now.
- [x] Remove default methods
- [x] Update RelScan input name
- [x] Rewriter
- [x] Factory updates
- [x] Builder updates (punted to #1664)
- [x] Fix the build! integration hell 😄

RCHowell:
need rebase on #1658

PR #1658, opened 2024-11-26
Simplifies PType's APIs

johnedquinn:
## Draft
Waiting on #1651 to be merged.

## Description
- Sets up the backwards-compatibility story for PType by removing `PType.Kind` and converting all variants to static final integers.
- Removed the deprecation notices on PType.
- Updated `NUMERIC` to use precision/scale. Note that the change of NUMERIC to use BigDecimal will come later when updating Datum.
- Removed the no-arg `PType#decimal()` and `PType#numeric()` static methods.

## License Information

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1665, opened 2024-11-27
[V1] Address 1.0 partiql-plan feedback and cleanup

RCHowell:
## Issue

#1661 

## Description

This PR address a number of issues to ([mostly](https://github.com/partiql/partiql-lang-kotlin/issues/1664)) complete the 1.0 plan implementation. It includes,

- interfaces/classes/enums converted to java
- removes all default methods (uses abcs)
- adds rewriter
- cleanup factory methods
- remove builders until reader to be finalized

## Design

> This is also included in part of the partiql-plan README

For the rule and strategy patterns to work, we need to model classes whose operands have a stable ordering;
so we have defined an abstract base class for all operators which holds operands and controls the access to them.
We use base implementations for state management and enforcing operands ordering; however we use interfaces for the
top-level of each domain. The abstract bases can be extended, and the operator/rel/rex interfaces can be implemented
directly.

What are operands?

* An operand is an input to some operator.
* An operator may have more than one operand e.g. join (left and right).
* Rel operands are typically called "inputs"
* Operands unify inputs since PartiQL bridges rel/rex domains.
* Not all operators are operands e.g. the limit of RelLimit is a rex, but not an operand - it is an "arg"

Why interfaces for top-level domain entities and not abstract base classes?

* We don’t want to force materialization of operands (consider wrapping a serde class)
* We don’t want to force holding state (aka having to call super constructors)
* The operator/rel/rex should be flexible for extension and the interface is the most open-ended / non-prescriptive.

Why abstract base classes for individual operators and not interfaces?

* Enforce operands ordering is the primary reason; also can memoize operands.
* Hold state such as mutable tags and computed types.
* We want the standard operators to be prescriptive but not limiting.
 

## License Information


RCHowell:
Rel ABCs done, moving on to Rex. Have lots of cleanup for builders and compiler, still draft.

RCHowell:
need rebase on #1658

partiql-plan/src/main/java/org/partiql/plan/rel/RelSort.java
```diff
@@ -0,0 +1,87 @@
...
+    @Override
+    protected final RelType type() {
+        return getInput().getType();
+    }
+
+    @NotNull
+    @Override
+    protected final List<Operator> operands() {
+        Rel c0 = getInput();
+        return List.of(c0);
```
johnedquinn, line 39: Need to pass the collations

partiql-plan/src/main/java/org/partiql/plan/rel/RelType.java
```diff
@@ -0,0 +1,65 @@
+package org.partiql.plan.rel;
+
+import org.jetbrains.annotations.NotNull;
+import org.partiql.types.Field;
+
+/**
+ * Analogous to a ROW type, consider cardinality estimates or other hint mechanisms.
+ */
+public class RelType {
```
johnedquinn, line 9:
```suggestion
public final class RelType {
```

partiql-plan/src/main/java/org/partiql/plan/rex/RexCast.java
```diff
@@ -0,0 +1,71 @@
...
+    public abstract PType getTarget();
+
+    @NotNull
+    protected final RexType type() {
+        return RexType.of(getTarget());
+    }
+
+    @Override
+    protected List<Operator> operands() {
+        return List.of();
```
johnedquinn, line 42: Needs operand.

Also, this is interesting regarding the type. If we wanted to add a wrapper PlanNode around PType (for whatever reason), you could add the type to the children. Though, probably not necessary.

partiql-plan/src/main/java/org/partiql/plan/rex/RexBag.java
```diff
@@ -0,0 +1,62 @@
...
+    @Override
+    protected final RexType type() {
+        return RexType.of(PType.bag());
+    }
+
+    @NotNull
+    @Override
+    protected final List<Operator> operands() {
+        Collection<? extends Operator> varargs = getValues();
+        return List.copyOf(varargs);
```
johnedquinn, line 40: While it probably doesn't pertain to this specific class (since I doubt additional fields would be added), I'm curious how we could mitigate breaking changes if we were to add another field to a class which has variable arguments. One way would be to create an OperatorList which extends Operator. That way, for the `operands()` method, you can still add elements (potentially variable as well) to the end of the list without ambiguity.

partiql-plan/src/main/java/org/partiql/plan/OperatorTransform.java
```diff
@@ -0,0 +1,695 @@
...
+        Rel input = rel.getInput();
+        Rel input_new = visit(input, ctx, Rel.class);
+        // rewrite measures
+        List<Measure> measures = rel.getMeasures();
+        List<Measure> measures_new = visitAll(measures, ctx, this::visitAggregateMeasure);
+        // rewrite groups
+        List<Rex> groups = rel.getGroups();
+        List<Rex> groups_new = visitAll(groups, ctx, this::visitAggregateGroup);
+        // rewrite aggregate
+        if (input != input_new || measures != measures_new || groups != groups_new) {
```
johnedquinn, line 62: One thing to highlight is that Java's `!=` will use referential equality, which wouldn't take into consideration the methods that now make plan nodes mutable such as `setTag()` and `setType()`.

johnedquinn (changes requested):
I still have some files left to review, but getting the review out quicker to allow you to address feedback sooner.

partiql-plan/src/main/java/org/partiql/plan/Action.java
```diff
@@ -0,0 +1,22 @@
+package org.partiql.plan;
+
+import org.jetbrains.annotations.NotNull;
+import org.partiql.plan.rex.Rex;
+
+/**
+ * A PartiQL statement action within a plan.
+ */
+public interface Action {
```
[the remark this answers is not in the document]
RCHowell: No, only the operators are part of the tree. Plans and actions are not in this.

partiql-plan/src/main/java/org/partiql/plan/rel/RelAggregate.java
```diff
@@ -0,0 +1,165 @@
...
+    protected final RelType type() {
+        throw new UnsupportedOperationException("Derive type is not implemented");
+    }
+
+    @NotNull
+    @Override
+    protected final List<Operator> operands() {
+        Rel c0 = getInput();
+        return List.of(c0);
+    }
```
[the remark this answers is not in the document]
RCHowell: I am trying to make the distinction between operands and arguments clear. For this example (RelAgggregate) the input relation is the only *operand* and the measures are an *argument*. A pattern matches operand trees, but a particular node match can define custom predicate logic on any arguments.

partiql-plan/src/main/java/org/partiql/plan/rel/RelBase.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+        return operands;
+    }
+
+    /**
+     * PROTECTED (could also be package private atm).
+     *
+     * @return computed type.
+     */
+    protected abstract RelType type();
```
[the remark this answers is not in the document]
RCHowell: RelBase memoizes calls to getType() so concrete RelBase classes implement the internal "type()" method which RelBase holds the state for; same pattern for operands(). It's just a lateinit pattern which the abstract classs can handle.

partiql-plan/src/main/java/org/partiql/plan/rel/RelSort.java
RCHowell: Collations are not operators, so they're not operands. For RelSort, the only operand is the "input" and the "collations" list is an argument.

partiql-plan/src/main/java/org/partiql/plan/rex/RexCast.java
RCHowell: There's no PlanNode, just operator tree and types aren't operators.

partiql-plan/src/main/java/org/partiql/plan/rex/RexBag.java
RCHowell: Yes I have a note in #1664 about handling varargs and just using a holder works, but would require a larger refactor for what an "operand" is. I think having operands mirror patterns would be easiest here, which would mean having a special vararg "operand".

partiql-plan/src/main/java/org/partiql/plan/OperatorTransform.java
RCHowell: We want referential equality for a transform. You'll see all the pig rewriters and sprout rewriters use kotlin referential equality. 

You don't need a rewriter for mutable things because mutable.

partiql-plan/src/main/java/org/partiql/plan/rel/RelProject.java
```diff
@@ -0,0 +1,93 @@
...
+    @Override
+    protected final RelType type() {
+        throw new UnsupportedOperationException("Derive type is not implemented");
+    }
+
+    @NotNull
+    @Override
+    protected final List<Operator> operands() {
+        Rel c0 = getInput();
+        return List.of(c0);
```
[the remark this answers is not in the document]
RCHowell: RelProject

operands: input
arguments: projections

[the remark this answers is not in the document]
johnedquinn: How would someone write a strategy for RelProject if the projections themselves aren't being compiled?

partiql-plan/src/main/java/org/partiql/plan/rel/RelAggregate.java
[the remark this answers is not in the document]
johnedquinn: Yeah, the main reason I bring it up has to do with where `getOperands()` is used, namely within the whole compilation strategies approach. If someone were to write a strategy for this RelAggregate, they'd have access to the compiled input in the operand tree. But, the Measures are arguments in this scenario, and thus aren't compiled, even though they contain plan Rex's (not compiled). Hence, they can't be used.

partiql-plan/src/main/java/org/partiql/plan/rel/RelSort.java
johnedquinn: Yeah, this comment has the same reasoning as https://github.com/partiql/partiql-lang-kotlin/pull/1665#discussion_r1870363550, but this one could potentially be disregarded since I see that the Rex in the collation will be replaced by an `int`.

partiql-plan/src/main/java/org/partiql/plan/rel/RelAggregate.java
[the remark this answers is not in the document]
RCHowell: Passing a compiler reference is easy enough

docs/wiki/documentation/Numeric Data Type & Arithmetic Operations.md
```diff
@@ -17,13 +17,13 @@ PartiQL also support `Positive Infinity, Negative Infinity, and NaN` as `FLOAT`
 
 Unless otherwise specified, a mathematical operator: 
 1) takes a field name or expression of a numeric data type as operand
-2) For Unary operation, the result type will be the same as the operand
-3) For Binary operation, and the two operands are of the same type, the result will be the same as the type of operands.
-4) For Binary operation, and the two operands are not of the same type, PartiQL will attempt to automatically coerce the operand.
+2) For Unary action, the result type will be the same as the operand
```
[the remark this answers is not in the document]
RCHowell: That was an accident and I'll have to revert all of that.

partiql-plan/src/main/java/org/partiql/plan/Action.java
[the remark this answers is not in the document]
johnedquinn: I see. Should the PartiQLCompiler take in the operator tree and not the plan? For a customer who wants to perform a bunch of rewrites on the plan, the lack of a PlanTransform (only an OperatorTransform) and generic PlanNode  makes the user experience a bit odd.

partiql-eval/src/main/java/org/partiql/eval/compiler/Strategy.java
```diff
@@ -34,8 +34,20 @@ public Pattern getPattern() {
      * Applies the strategy to a logical plan operator and returns the physical operation (expr).
      *
      * @param match holds the matched operators
+     * @param callback for compiling arguments of matched operators
      * @return the physical operation
      */
     @NotNull
-    public abstract Expr apply(@NotNull Match match);
+    public abstract Expr apply(@NotNull Match match, @NotNull Callback callback);
```
johnedquinn, line 41: Mentioned this in another comment, but if we hope to provide default strategies for some nodes (such as `RelScan`), we need to also pass the execution mode.

partiql-plan/src/main/java/org/partiql/plan/rex/RexBase.java
```diff
@@ -0,0 +1,72 @@
...
+     * @return computed type.
+     */
+    protected abstract RexType type();
+
+    /**
+     * PROTECTED (could also be package private atm).
+     *
+     * @return computed operands.
+     */
+    protected abstract List<Operand> operands();
```
johnedquinn, line 71: Both `@NotNull`. Same with the param for `setType()`.

partiql-plan/src/main/kotlin/org/partiql/plan/Operators.kt
```diff
@@ -0,0 +1,444 @@
...
+     * Create a [RelLimit] instance.
+     *
+     * @param input
+     * @param limit
+     * @return
+     */
+    public fun limit(input: Rel, limit: Rex): RelLimit = RelLimit.create(input, limit)
+
+    /**
+     * Create a [RelLimit] instance.
```
johnedquinn, line 184:
```suggestion
     * Create a [RelOffset] instance.
```

partiql-plan/src/main/java/org/partiql/plan/Plan.java
```diff
@@ -0,0 +1,25 @@
...
+    @NotNull
+    default public Version getVersion() {
+        return Version.UNKNOWN();
+    }
+
+    /**
+     * @return statement actions to execute.
+     */
+    @NotNull
+    public List<Action> getActions();
```
johnedquinn, line 24: The list here. Is this supposed to represent essentially sub-plans?

johnedquinn (changes requested):
Minor comments/questions remaining. Once addressed, I can approve.

partiql-plan/src/main/java/org/partiql/plan/Plan.java
RCHowell: Initially for a transaction, but I believe now that "transaction" should be an "action" which holds more actions. I will update this

partiql-plan/src/main/java/org/partiql/plan/rex/RexBase.java
RCHowell: Doesn't matter because these are internal java-only apis, but doesn't hurt to add just superfluous

johnedquinn: These would still be public, since anyone can extend the public RexBase.

johnedquinn (approved):

Issue #1660, opened 2024-11-27
[V1] SPI Feedback Cleanup


RCHowell:
## Description

This is a tracking issue for finalization/cleanup efforts for the SPI classes.

## Tasks

- [ ] Split getTable(identifier) functionality to leverage getTable(name)
- [ ] Remove some default impls so its clear what to implement
- [ ] Make function instances easier to use
- [ ] Have hidden PartiQL catalog (env calls it directly)

PR #1670, opened 2024-12-03
Adds a built-in PartiQL System catalog, support for SQL-Path, and SPI cleanup

johnedquinn:
## Relevant Issues
- Related to #1660

## Description
- Adds a built-in PartiQL System catalog, which is installed via the Session builder.
- Adds support for SQL Path by plumbing the path through scalar and aggregate function resolution.
- Changes some APIs from using Collection to List, where order matters.
- Splits `Catalog#getTable()` and `Catalog#resolveTable()`
- The remaining action items in #1660 will be taken in a follow-up PR.

## License Information


partiql-system/build.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+tasks.withType<Jar> {
+    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
+}
+
+publish {
+    artifactId = "partiql-catalog"
+    name = "PartiQL Catalog"
+    description = "Contains all of the PartiQL System's functions and operators."
```
RCHowell, line 42: probably meant to be partiql-system here.

Why split this vs. adding as part of the partiql-spi? I don't think consuming one without the other makes sense at the moment imo.

partiql-system/src/main/java/org/partiql/system/PartiQLSession.java
```diff
@@ -0,0 +1,101 @@
...
+import org.partiql.spi.catalog.Session;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * This is a package-private implementation of {@link Session}.
+ * @see PartiQLSessionBuilder
+ */
+public final class PartiQLSession implements Session {
```
RCHowell, line 17: Session already has a builder and standard implementation, so I'm not sure we need to make these things public. You could internalize these by using the named builder pattern like `Session.builder()..build()` which has the PartiQL system installed, or use a `Session.bare()...build()` or something like that.

partiql-spi/src/main/kotlin/org/partiql/spi/catalog/Session.kt
```diff
@@ -43,7 +43,8 @@ public interface Session {
     public companion object {
 
         /**
-         * Returns a [Session] with only the "empty" catalog implementation.
+         * Returns a [Session] with only the "empty" catalog implementation. Note that this does NOT add the
+         * PartiQL System Catalog.
          */
         @JvmStatic
         public fun empty(): Session = object : Session {
```
RCHowell, line 50: Session.empty() is used in several places so would need the psystem available. I don't believe we need bare()  factory method either. I was referring to a builder method in the previous PR, not a factory.

Think about it this, you have a named method for which builder either adds the system or not.

```
Session.builder()....build()   // standard builder comes with standard system
Session.system(catalog)...builder()  // .system() is a named *builder* factory which replaces the standard system.
```

johnedquinn: Yeah, the latest revision has this. I can just revert the last commit (the addition of the bare method).

johnedquinn: I just reverted and added the PartiQL system catalog to `Session.empty()`. See the last three commits.

partiql-spi/src/main/java/org/partiql/spi/catalog/PartiQLSystemCatalog.java
```diff
@@ -0,0 +1,66 @@
...
+    @NotNull
+    private final String name;
+
+    /**
+     * Creates a new PartiQL System Catalog with the given name.
+     * @param name the name of the PartiQL System Catalog
+     */
+    PartiQLSystemCatalog(@NotNull String name) {
+        this.name = name;
+    }
```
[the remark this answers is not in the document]
johnedquinn: Can't make a protected top-level class. Updated the Javadocs to make the visibility more apparent, since there is no package-private modifier. And added the singleton. See latest commit: https://github.com/partiql/partiql-lang-kotlin/pull/1670/commits/d1b33056d3405fdd4ee1f8dd23d20aa1144e3dd9

If we want to expose this class in the future, sure. But we don't need to right now.

partiql-spi/src/main/kotlin/org/partiql/spi/catalog/Session.kt
```diff
@@ -46,11 +46,9 @@ public interface Session {
...
          */
         @JvmStatic
-        public fun empty(): Session = object : Session {
-            override fun getIdentity(): String = "unknown"
-            override fun getCatalog(): String = "empty"
-            override fun getCatalogs(): Catalogs = Catalogs.empty()
-            override fun getNamespace(): Namespace = Namespace.empty()
+        public fun empty(): Session {
+            val catalog = Catalog.empty("empty")
+            return builder().catalog(catalog.getName()).catalogs(catalog).build()
```
RCHowell, line 51: Since we now have a system catalog, then I don't think we need to add _any_ catalogs with the empty session. But this means you cannot create tables or anything until you create a catalog. Also this means `getCatalog()` for the session will be "$system". 

This was necessary before because we needed the super-chain, but adding an empty catalog to an empty is no longer needed with $system at the end of the path-chain.

partiql-spi/src/main/kotlin/org/partiql/spi/catalog/Session.kt
```diff
@@ -64,6 +62,7 @@ public interface Session {
 
         private var identity: String = "unknown"
         private var catalog: String? = null
+        private var systemCatalog: Catalog = PartiQLSystemCatalog.INSTANCE
```
RCHowell, line 65:
```suggestion
        private var system: Catalog = System.INSTANCE
```

RCHowell (changes requested):

RCHowell (approved):

Issue #1642, opened 2024-11-04
[V1] Use int primitive for AST tag


RCHowell:
Each AST node gets a random tag. Rather than inserting a random value which becomes a string, we could improve parser perf and AST instantiation with a simple int primitive.

PR #1674, opened 2024-12-09
[v1] Refactor AstNode getChildren() method; change tag to be mutable int

alancai98:
## Relevant Issues
- https://github.com/partiql/partiql-lang-kotlin/issues/1642

## Description
- From recent PR comment https://github.com/partiql/partiql-lang-kotlin/pull/1650#discussion_r1873771424, changing AstNode's `children` method in a couple ways
1. use `List` in signature to retain order property
2. rename to `getChildren()` to allow for Kotlin accessor syntax sugar:
```Kotlin
// Previous modeling w/ `children()`
node.children().filter { it is Expr }
// New modeling w/ `getChildren()` allows for slight syntax improvements from Kotlin
node.children.filter { it is Expr }
```
- Also changes the `AstNode` tag to be a mutable int

## Other Information
  - No, `v1` not yet released.

  - Yes, but `v1` not yet released.

## License Information

### DIFF
--- a/partiql-ast/src/main/java/org/partiql/ast/AstNode.java
+++ b/partiql-ast/src/main/java/org/partiql/ast/AstNode.java
@@ -1,7 +1,5 @@
 package org.partiql.ast;
 
-import lombok.Getter;
-import lombok.Setter;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.List;
@@ -10,10 +8,16 @@ import java.util.List;
  * TODO docs, equals, hashcode
  * TODO support source location -- https://github.com/partiql/partiql-lang-kotlin/issues/1608
  */
-@Setter
-@Getter
 public abstract class AstNode {
     private int tag = 0;
+
+    public int getTag() {
+        return tag;
+    }
+
+    public void setTag(int tag) {
+        this.tag = tag;
+    }
 
     @NotNull
     public abstract List<AstNode> getChildren();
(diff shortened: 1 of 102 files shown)


RCHowell:
Can we also update the tag to mutable int like the plans? 

https://github.com/partiql/partiql-lang-kotlin/issues/1642

alancai98:
> Can we also update the tag to mutable int like the plans?
> 
> #1642

Let me sync w/ you on a few things for that issue.

alancai98:
> Can we also update the tag to mutable int like the plans?
> 
> #1642

Added mutable int tag for the `AstNode` -- https://github.com/partiql/partiql-lang-kotlin/pull/1674/commits/796f04eaad1334b18dd6edca6c669d70ab969cf7.

From offline discussion
- set `AstNode` tags using default ANTLR parser visitor
- use a private counter to ensure unique tags (since tag could have collisions using `Random.nextInt()`
- TBD if the parsed AST tree should have the highest tag node be the top-level node (for the purposes of AstVisitor pass that adds/modifies nodes)

partiql-ast/src/main/java/org/partiql/ast/AstNode.java
```diff
@@ -1,20 +1,22 @@
...
 /**
  * TODO docs, equals, hashcode
  * TODO support source location -- https://github.com/partiql/partiql-lang-kotlin/issues/1608
  */
+@Setter
+@Getter
 public abstract class AstNode {
-    @NotNull
-    public String tag = "Ast-" + String.format("%06x", new Random().nextInt());
+    private int tag = 0;
```
RCHowell, line 16: Let's add an explicit getter/setter like in the plans.

alancai98: Updated

RCHowell (approved):

Issue #1640, opened 2024-11-04
[V1] Add nullable default fields to AST node factories


RCHowell:
## Description

> when converting the SqlDialectTest over to use the v1 AST, it was somewhat annoying to not have defaults for all of the nullable fields. Deferring adding defaults to a different PR since we'll need to examine backwards compatibility of defaults a bit further.

https://github.com/partiql/partiql-lang-kotlin/pull/1638

PR #1675, opened 2024-12-10
[v1] Add null defaults to AST factory methods

alancai98:
## Relevant Issues
- https://github.com/partiql/partiql-lang-kotlin/issues/1640 as part of https://github.com/partiql/partiql-lang-kotlin/issues/1610

## Description
- Adds null defaults for nullable AST factory methods

## Other Information
  - No `v1` not yet released.

  - Yes but on `v1`

## License Information


partiql-ast/api/partiql-ast.api
```diff
@@ -1,17 +1,25 @@
 public final class org/partiql/ast/Ast {
 	public static final field INSTANCE Lorg/partiql/ast/Ast;
 	public static final fun columnConstraintCheck (Lorg/partiql/ast/IdentifierChain;Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/ddl/AttributeConstraint$Check;
+	public static synthetic fun columnConstraintCheck$default (Lorg/partiql/ast/IdentifierChain;Lorg/partiql/ast/expr/Expr;ILjava/lang/Object;)Lorg/partiql/ast/ddl/AttributeConstraint$Check;
```
alancai98, line 4: (self-review) adding a default to an existing field is an additive Kotlin API change

partiql-ast/src/main/kotlin/org/partiql/ast/Ast.kt
```diff
@@ -94,12 +94,12 @@ public object Ast {
     }
 
     @JvmStatic
-    public fun exprCall(function: IdentifierChain, args: List<Expr>, setq: SetQuantifier?): ExprCall {
+    public fun exprCall(function: IdentifierChain, args: List<Expr>, setq: SetQuantifier? = null): ExprCall {
```
alancai98, line 97: (self-review) this will allow for defaults from Kotlin code. What is our stance w/ providing Java defaults using `@JvmOverloads` annotation?

Alternative modeling could have not used defaults but provide two functions, though it can get unwieldy if there were multiple nullable arguments.

partiql-ast/src/main/java/org/partiql/ast/expr/ExprWindow.java
```diff
@@ -67,13 +67,13 @@ public <R, C> R accept(@NotNull AstVisitor<R, C> visitor, C ctx) {
...
-        @Nullable
+        @NotNull
         public final List<Expr> partitions;
 
-        @Nullable
+        @NotNull
         public final List<Sort> sorts;
 
-        public Over(@Nullable List<Expr> partitions, @Nullable List<Sort> sorts) {
+        public Over(@NotNull List<Expr> partitions, @NotNull List<Sort> sorts) {
```
alancai98, line 76: (self-review) prior modeling was slightly overcomplicated. An empty `partitions` or `sorts` list should represent no `PARTITION BY` or `ORDER BY` specifications.

partiql-ast/src/main/kotlin/org/partiql/ast/Ast.kt
```diff
@@ -569,15 +608,28 @@ public object Ast {
...
+    @JvmStatic
+    public fun queryBodySFW(select: Select, from: From): QueryBody.SFW {
+        return QueryBody.SFW(select, null, from, null, null, null, null)
+    }
+
+    // SFW variant with `WHERE`; optional fields are set to null
+    @JvmStatic
+    public fun queryBodySFW(select: Select, from: From, where: Expr): QueryBody.SFW {
+        return QueryBody.SFW(select, null, from, null, where, null, null)
+    }
```
alancai98, line 621: (self-review) defined the SFW required, SFW required with `WHERE`, and SFW all fields variants since many fields are nullable.

Also the SFW all fields variant is purposely not using `@JvmOverloads`.

partiql-ast/src/main/kotlin/org/partiql/ast/Ast.kt
```diff
@@ -188,8 +194,15 @@ public object Ast {
...
+    // query set variant with only required field; optional fields set to null
     @JvmStatic
-    public fun exprQuerySet(body: QueryBody, orderBy: OrderBy?, limit: Expr?, offset: Expr?): ExprQuerySet {
+    public fun exprQuerySet(body: QueryBody): ExprQuerySet {
+        return ExprQuerySet(body, null, null, null)
+    }
+
+    // query set variant with all fields and nullable defaults set to null
+    @JvmStatic
+    public fun exprQuerySet(body: QueryBody, orderBy: OrderBy? = null, limit: Expr? = null, offset: Expr? = null): ExprQuerySet {
```
alancai98, line 205: (self-review) similarly for query set, create the most common use case of required field only since there are a few nullable arguments. All of the other (besides SFW) will specify null defaults and will use `@JvmOverloads`

partiql-ast/src/main/kotlin/org/partiql/ast/Ast.kt
alancai98: From offline discussion,
- use `@JvmOverloads` when factory method does not have many nullable fields with defaults of `null`
- when factory method has many nullable fields (e.g. query set, SFW), specify common factory methods of required fields
  - also create the default nullable variant without the `@JvmOverloads`

JamesMcGaa:
thank you @alancai98

partiql-ast/src/main/java/org/partiql/ast/expr/ExprWindow.java
RCHowell: This would make a good doc comment!

RCHowell (approved):

partiql-ast/src/main/java/org/partiql/ast/expr/ExprWindow.java
alancai98: True. I'll add

RCHowell (approved):

PR #1678, opened 2024-12-11
Removes PartiQLValue from public API

johnedquinn:
## Description
This PR can be broken up into TWO main commits: the removal of PartiQLValue and the true onboarding of VARIANT as a first-class type. I couldn't merge just the first commit, since the conformance tests failed an additional 1000+ tests. So, the second one naturally followed.

### Removal of PartiQLValue (1st commit)
- This PR removes PartiQLValue from the public API. It also removes the public conversions between PartiQLValue, Datum, PartiQLValueType, and PType.
- It does this by moving its definition (and other relevant classes such as impls and conversions) into `testFixtures`.
  - Test sources will temporarily rely on PartiQLValue, until it is fully deleted in the future.
  - The only main source set relying on the `testFixtures` is `partiql-cli`
- I also updated the internal Sprout model to use Datum instead of PartiQLValue.
- This also removes `PartiQLValueExperimental`

### Support for VARIANT (2nd commit)
- VARIANT needed to be supported due to the lack of parity between the PartiQLValueReader and the IonDatumReader -- which is for decoding PartiQL values that are encoded in Ion. Rather than what VARIANT is, which is decoding arbitrary Ion data into PartiQL structures.
- So, this PR was born. This essentially just uses `lower()` whenever we find a variant value at runtime. It also adds the static type checking required for variant.

## Reviewer's Notices

- I'd look at the two commits individually so that it is easier to understand why something is changing. They're both relatively straightforward.
- We now **pass additional conformance tests** 🎊  without failing any previous ones, bringing us from 54 failing tests to 47.

## License Information


partiql-eval/src/main/kotlin/org/partiql/eval/internal/helpers/ValueUtility.kt
```diff
@@ -51,29 +52,42 @@ internal object ValueUtility {
      */
     fun Datum.getText(): String {
         return when (this.type.code()) {
+            PType.VARIANT -> this.lower().getText()
```
RCHowell, line 55: No need to be changed here, but this is interesting and makes me think Variant needs to be modified slightly so callers aren't required to call `.lower()` even though the same logic will be applied.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelOpUnpivot.kt
```diff
@@ -58,7 +58,13 @@ internal sealed class RelOpUnpivot : ExprRelation {
     class Strict(private val expr: ExprValue) : RelOpUnpivot() {
 
         override fun struct(): Datum {
-            val v = expr.eval(env.push(Row()))
+            val v = expr.eval(env.push(Row())).let {
+                if (it.type.code() == PType.VARIANT) {
+                    it.lower()
+                } else {
+                    it
+                }
```
RCHowell, line 66: Could you please leave a comment for why we need to lower here? Thanks, I'm just guessing it's to ensure we have Datum struct rather than a datum variant|struct

partiql-eval/src/test/kotlin/org/partiql/eval/internal/SuccessTestCase.kt
```diff
@@ -26,13 +26,12 @@ public class SuccessTestCase(
...
 ) : PTestCase {
 
-    @OptIn(PartiQLValueExperimental::class)
     constructor(
         input: String,
         expected: PartiQLValue,
         mode: Mode = Mode.PERMISSIVE(),
         globals: List<Global> = emptyList(),
-    ) : this(input, Datum.of(expected), mode, globals)
+    ) : this(input, DatumUtils.toDatum(expected), mode, globals)
```
RCHowell, line 34: nit, but a `PartiQLValue.toDatum()` would properly invert since then no Datum related classes like DatumUtils would reference PartiQLValue.

partiql-planner/build.gradle.kts
```diff
@@ -115,7 +115,6 @@ tasks.register<Exec>("codegen") {
         "--poems", "visitor",
         "--poems", "builder",
         "--poems", "util",
-        "--opt-in", "org.partiql.value.PartiQLValueExperimental",
```
RCHowell, line 118: Love this

partiql-planner/src/main/kotlin/org/partiql/planner/internal/casts/CastTable.kt
```diff
@@ -63,6 +63,11 @@ internal class CastTable private constructor(
                     cast(it)
                 }
             }
+            graph[PType.VARIANT] = relationships {
+                PType.codes().map {
+                    cast(it)
+                }
+            }
```
RCHowell, line 70: Interesting, may be worth a one-line note like "a variant can be cast to any type"

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
```diff
@@ -148,73 +132,57 @@ internal object RexConverter {
...
                     } catch (ex: NumberFormatException) {
                         // ignore
                     }
 
                     // 3rd, try parse as BigInteger
                     try {
                         val v = BigInteger(n)
-                        return intValue(v)
+                        val vDecimal = BigDecimal(v)
+                        return Datum.decimal(vDecimal, vDecimal.precision(), vDecimal.scale())
```
RCHowell, line 179: Farewell arbitrary integers. This makes me think creating a decimal datum from a BigInteger is a reasonable API.

partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
```diff
@@ -774,6 +577,13 @@ static Datum struct(@NotNull Iterable<Field> values) {
         return new DatumStruct(values);
     }
 
+    @NotNull
+    static Datum ion(@NotNull String value) {
+        IonElementLoader loader = ElementLoader.createIonElementLoader();
+        AnyElement element = loader.loadSingleElement(value);
+        return new IonVariant(element);
+    }
```
RCHowell, line 585: I went back-and-forth on this quite a bit, but stopped myself from setting up a whole VariantLoader which is where this eventually leads. I would suggest leaving any variant instance (i.e. "ion" or "json) out of the datum public API, rather it makes sense to have either

1.  IonVariant be public (which I know you would not like) with private constructor
2. Have the API be `Datum.variant(String encoding, String payload)` or something equivalent.

This way we don't couple any variant name/instance to Datum's public API.

RCHowell (dismissed):

partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
RCHowell: Seems I've contradicted myself (: 

https://github.com/partiql/partiql-lang-kotlin/blob/main/partiql-spi/src/main/java/org/partiql/spi/value/DatumReader.java#L31-L37

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
```diff
@@ -148,73 +132,57 @@ internal object RexConverter {
...
                     } catch (ex: NumberFormatException) {
                         throw ex
                     }
                 }
                 Literal.APPROX_NUM -> {
-                    float64Value(lit.numberValue().toDouble())
+                    return Datum.doublePrecision(lit.numberValue().toDouble())
```
alancai98, line 185: More so a question for the plan representation: are we committing to using `Datum` as the literal representation in the plan? I haven't looked much into serialization of plans but not sure if eagerly converting some of the types into Java can lead to some later problems.

At least for `VARIANT`, we don't appear to be any `lower`ing till evaluation time so wondering if the same can apply for literals in general.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/helpers/ValueUtility.kt
jpschorr: FWIW, this is pretty similar to the approach I'm developing in partiql-lang-rust.

There's kind of a parallel with `Dynamic` here:
`Dynamic` is a compile-time type that says introspect the type at runtime, because we don't know statically.
`Variant` is a runtime-value-type that says introspect and/or convert  the value first, because it is not yet computable.

By 'lower'ing the Variant value, you allow it to take part in the rest of the already existing PartiQL computation engine. There are probably other approaches that might have some efficiency gains. But this approach is probably the least invasive/chatty.

johnedquinn: If you look at the PlanTyper, we skip over type checks when we see dynamic (and variant). The downside to this is that variant is a type that exists in our runtime, and we need to introspect the value (by using `lower()`) before passing it to other areas of our engine. We've done a ton of work to make sure that we can safely receive values (via `PType`) in functions and expressions, and `lower()` allows us to continue to do so. Also leverages our cast tables, comparator, etc. without breaking the typing logic.

So, yes, we need to use `lower()` here and at other places before passing it anywhere else.

That being said, if we _know_ that variant will never be encountered due to static type analysis, we can then decide to forgo `lower()`. But right now, we don't have this.

partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
johnedquinn: 😄 Yeah, haha, that's actually where I got the naming from. But, LMK whatever your preference is! Since the default encoding is Ion when a user writes back-ticks, I'm not opposed to keeping the method as `ion()`.

partiql-spi/src/main/kotlin/org/partiql/spi/value/ion/IonDatumReader.kt
```diff
@@ -66,36 +67,64 @@ internal class IonDatumReader internal constructor(
...
+            IonType.INT -> bigint()
+            IonType.FLOAT -> double()
+            IonType.DECIMAL -> decimal0()
+            IonType.TIMESTAMP -> TODO("timestamp without annotation")
+            IonType.STRING -> varchar0()
+            IonType.CLOB -> clob0()
+            IonType.BLOB -> clob0()
+            IonType.LIST -> array()
+            IonType.STRUCT -> struct()
+            IonType.SYMBOL -> missing()
```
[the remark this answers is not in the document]
jpschorr: I think Ion symbols should be considered strings when reading/converting to PartiQL values.

> symbol: provides efficient encoding for frequently occurring strings 

from: https://amazon-ion.github.io/ion-docs/

> In binary Ion, common text tokens such as struct field names are automatically stored in a symbol table. This allows these tokens to be efficiently encoded as table offsets instead of repeated copies of the same text. As a further space optimization, symbol tables can be pre-shared between producer and consumer so that only the table name and version are included in the payload, eliminating the overhead involved with repeatedly defining the same symbols across multiple pieces of Ion data.

from: https://amazon-ion.github.io/ion-docs/guides/why.html#rich-type-system

Cf. https://amazon-ion.github.io/ion-docs/docs/symbols.html

partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
```diff
@@ -245,7 +203,7 @@ default short getShort() {
      *                                       {@link #isNull()} returns false before attempting to invoke this method.
      */
     default int getInt() {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException("Has type: " + getType() + " and class " + this.getClass().getName());
```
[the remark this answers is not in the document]
johnedquinn: Removed for now, but yes, we should have something. Future PR.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
johnedquinn: The plan representation of literals is out of scope of this PR -- so I'm not sure.

And, what do you mean by the second observation?

RCHowell: This PR is (internal) plan representation of literals since the definition of literals in plans has been changed by this,

```
    - partiql_value::'org.partiql.value.PartiQLValue',
    + partiql_value::'org.partiql.spi.value.Datum',
```

But I think you two might be talking about _serialization_ of values which is covered by "PartiQL encoding in Ion" in #1655

partiql-eval/src/main/kotlin/org/partiql/eval/internal/helpers/DatumUtils.kt
```diff
@@ -0,0 +1,20 @@
...
+import org.partiql.spi.value.Datum
+import org.partiql.types.PType
+
+internal object DatumUtils {
+
+    /**
+     * Calls [Datum.lower] if the datum is a variant, otherwise returns the datum. If you don't know whether the value
+     * is of type [PType.VARIANT], you should use [Datum.lowerSafe] before invoking whatever methods you intend to use.
+     * This is essentially a workaround for the fact that we currently don't know whether a particular expression will be
+     * [PType.VARIANT] or not. This can eventually be optimized to accommodate this.
```
alancai98, line 12: nit: could we reword the last sentence to not refer to "this" and be more specific? Might be confusing as to what "this" is referring to

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/internal/Accumulator.kt
```diff
@@ -97,12 +97,16 @@ private fun Long.checkOverflowPlus(other: Long): Number {
 }
 
 internal fun checkIsBooleanType(funcName: String, value: Datum) {
+    if (value.type.code() == PType.VARIANT) {
```
alancai98, line 100: Is there a reason we add this variant type check for booleans but not for numbers in `checkIsNumberType`?

partiql-spi/src/main/kotlin/org/partiql/spi/value/ion/IonVariant.kt
```diff
@@ -54,127 +51,63 @@ internal class IonVariant(private var value: AnyElement) : Datum {
...
+            TIMESTAMP -> Datum.timestamp(DateTimeValue.timestamp(value.timestampValue))
+            INT -> {
+                val bigInteger = value.bigIntegerValue
+                when {
+                    bigInteger < Int.MAX_VALUE.toBigInteger() && bigInteger > Int.MIN_VALUE.toBigInteger() -> Datum.integer(bigInteger.toInt())
+                    bigInteger < Long.MAX_VALUE.toBigInteger() && bigInteger > Long.MIN_VALUE.toBigInteger() -> Datum.bigint(bigInteger.toLong())
+                    else -> {
+                        val dec = bigInteger.toBigDecimal()
+                        Datum.decimal(dec, dec.precision(), 0)
+                    }
```
alancai98, line 73: nit: could use `IntElementSize` to simplify some of the logic here -- https://javadoc.io/doc/com.amazon.ion/ion-element/1.2.0/com/amazon/ionelement/api/IntElementSize.html. It only distinguishes between long and big_integer (not sure if that's intentional) but it could help eliminate some of the `.toBigInteger` conversions.

alancai98 (approved):

PR #1679, opened 2024-12-13
[v1] Add IS [NOT] TRUE|FALSE|UNKNOWN; make IS NULL and IS MISSING separate AST nodes

alancai98:
## Relevant Issues
- Related PR -- https://github.com/partiql/partiql-lang-kotlin/pull/1520
- partiql-lang discussion on MISSING propagation and predicates -- https://github.com/orgs/partiql/discussions/93

## Description
- Adds AST, parser, evaluation support for SQL99's `IS [NOT] TRUE|FALSE|UNKNOWN` boolean test predicate
- AST: refactors `NULL` and `MISSING` out of `DataType` and creates dedicated null and missing predicate nodes

## Other Information
  - No `v1` yet to be released.

  - Removal of some `DataType` AST values and functions (for `NULL` and `MISSING`); `v1` not yet released so not an issue.

## License Information

### DIFF
--- a/partiql-ast/src/main/java/org/partiql/ast/DataType.java
+++ b/partiql-ast/src/main/java/org/partiql/ast/DataType.java
@@ -10,7 +10,6 @@ import java.util.List;
 
 @EqualsAndHashCode(callSuper = false)
 public class DataType extends AstEnum {
-
     /**
      * A field definition with in a Struct Type Definition
      */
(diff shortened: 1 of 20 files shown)


partiql-ast/src/main/java/org/partiql/ast/DataType.java
```diff
@@ -70,80 +69,68 @@ public <R, C> R accept(@NotNull AstVisitor<R, C> visitor, C ctx) {
     }
 
     public static final int UNKNOWN = 0;
-    // TODO remove `NULL` and `MISSING` variants from DataType
-    // <absent types>
-    public static final int NULL = 1;
-    public static final int MISSING = 2;
```
alancai98, line 76: (self-review) removed `NULL` and `MISSING` from `DataType`. `IS [NOT] NULL|MISSING` handled by **different** node than `ExprIsType`.

Rest of changes in file are to fix the numbering.

partiql-ast/src/main/java/org/partiql/ast/expr/ExprBoolTest.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * TODO docs
+ * Corresponds to SQL99's boolean test (6.30).
+ */
+@Builder(builderClassName = "Builder")
+@EqualsAndHashCode(callSuper = false)
+public class ExprBoolTest extends Expr {
```
alancai98, line 18: (self-review) there are a few different ways to model bool test predicate and null/missing predicate. Open to some other suggestions on modeling if it could be improved. Some alternatives I considered
- separate nodes for all (i.e. ExprIsTrue, ExprIsFalse, ExprIsUnknown, ExprIsNull, ExprIsMissing)? <- seemed a bit redundant w/ all the visitors + rewriter methods that would get added
- modeling `IS [NOT] NULL` and `IS [NOT] MISSING` together as an absent node? <- creates the same number of classes (i.e. ExprIsAbsent and AbsentEnum) as just having them separated out. I feel like we won't be adding more absent predicates? But if we will, then an enum might be better.

partiql-parser/src/main/antlr/PartiQLParser.g4
```diff
@@ -934,8 +941,8 @@ literal
 
 type
     : datatype=(
-        NULL | BOOL | BOOLEAN | SMALLINT | INTEGER2 | INT2 | INTEGER | INT | INTEGER4 | INT4
-        | INTEGER8 | INT8 | BIGINT | REAL | CHAR | CHARACTER | MISSING
```
alancai98, line 938: (self-review): `NULL` and `MISSING` are no longer passed as data types.

partiql-ast/src/main/java/org/partiql/ast/expr/ExprIsType.java
```diff
@@ -11,8 +11,7 @@
 import java.util.List;
 
 /**
- * TODO docs, equals, hashcode
- * TODO also support IS NULL, IS MISSING, IS UNKNOWN, IS TRUE, IS FALSE
+ * TODO docs + further specification on supported types
```
RCHowell, line 14: Possibly mark as experimental for now.

partiql-ast/src/main/java/org/partiql/ast/expr/TruthValue.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+
+/**
+ * TODO docs
+ */
+public class TruthValue extends AstEnum {
+    public static final int UNKNOWN = 0;
+    public static final int TRUE = 1;
+    public static final int FALSE = 2;
+    public static final int UNK = 3;
```
RCHowell, line 17: Interesting! Maybe OTHER, or maybe we don't need to define the `UNK` variant for this. It would still force non-exhaustive and we have -1 reserved. Maybe you make it

```
// implicit uknown is 0
TRUE = 1
FALSE = 2
UNKNOWN = 3
```

partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEvaluatorTest.kt
```diff
@@ -1024,13 +1025,146 @@ class PartiQLEvaluatorTest {
...
+                expected = boolValue(false)
+            ),
+            SuccessTestCase(
+                input = "MISSING IS TRUE;",
+                expected = boolValue(false)
+            ),
+            SuccessTestCase(
+                input = "'foo' IS TRUE",
+                expected = Datum.missing(),
+                mode = Mode.PERMISSIVE()
```
RCHowell, line 1049: Something interesting is Postgres does

```
'1' IS TRUE  -- true
'foo' IS TRUE -- err!
```

partiql-parser/src/main/antlr/PartiQLParser.g4
RCHowell: Nice!

RCHowell (changes requested):

partiql-ast/src/main/java/org/partiql/ast/expr/ExprIsType.java
alancai98: Yeah, I'll do that

partiql-ast/src/main/java/org/partiql/ast/expr/TruthValue.java
alancai98: I wasn't too sure what to do w/ the name collision. Ideally for the sake of consistency, I would like to keep the unknown/other variant the same across `AstEnum`'s.

Perhaps we could rename all of the unknown/other variants something that won't have a collision like `_UNKNOWN` or `_OTHER`? The static functions could then be named `_UNKNOWN()`/`_OTHER()`.

partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEvaluatorTest.kt
alancai98: So in strict mode, `'foo' IS TRUE` should error (assuming https://github.com/partiql/partiql-lang-kotlin/issues/1680 gets fixed) since the expression for `IS <truth value>` expects a boolean value.

This is also consistent w/ what we do for the other boolean constructs like `AND`, `OR`, `NOT` (i.e. data type mismatch error in strict, missing in permissive).

Postgres allows certain strings as input to functions that expect booleans. From their [docs](https://www.postgresql.org/docs/current/datatype-boolean.html):
> The datatype input function for type `boolean` accepts these string representations for the “true” state:
> - true 
> - yes 
> - on 
> - 1 
> 
> and these representations for the “false” state: 
> 
> - false 
> - no 
> - off 
> - 0 
> 
> Unique prefixes of these strings are also accepted, for example `t` or `n`. Leading or trailing whitespace is ignored, and case does not matter.

```SQL
-- in Postgres
-- IS TRUE
SELECT 'no' IS TRUE -- false
SELECT 'foo' IS TRUE -- Query Error: invalid input syntax for type boolean: "foo"
-- IS FALSE
SELECT 'no' IS FALSE -- true
SELECT 'foo' IS FALSE -- Query Error: invalid input syntax for type boolean: "foo"
-- IS UNKNOWN
SELECT 'no' IS UNKNOWN -- false
SELECT 'foo' IS UNKNOWN -- Query Error: invalid input syntax for type boolean: "foo"
-- other boolean exprs
SELECT NOT 'no' -- true
SELECT NOT 'foo' -- Query Error: invalid input syntax for type boolean: "foo"
SELECT true AND 'no' -- false
SELECT true AND 'foo' -- Query Error: invalid input syntax for type boolean: "foo"
SELECT false OR 'no' -- false
SELECT false OR 'foo' -- Query Error: invalid input syntax for type boolean: "foo"
```

partiql-ast/src/main/java/org/partiql/ast/expr/TruthValue.java
RCHowell: I don't think I have a preference on the exact name so let's check in with John on Tuesday's meeting. I think the important thing is avoiding current and future collisions with what you have suggested.

johnedquinn: Yeah, I've been thinking about this. If there is good separation between our library APIs and serialization, then I _believe_ we can remove the UNKNOWN variants of the other enums.

RCHowell (approved):
I'm approving AS-IS to keep you moving forward. I've created a follow-up for tues. https://github.com/partiql/partiql-lang-kotlin/issues/1682

partiql-ast/src/main/java/org/partiql/ast/expr/TruthValue.java
alancai98: Yeah, let's discuss this tomorrow. I'll need to check w/ the original rationale for including an `UNKNOWN` variant for all of the enums. Perhaps it's required for serde? But we could always add the variant and relevant functions back in the future.

PR #1685, opened 2024-12-18
[v1] Change public fields to Lombok getters; make concrete classes final

alancai98:
## Relevant Issues

## Description
- Makes the public AST fields private and adds Lombok getters
- Makes existing concrete classes final

## Other Information
  - No, v1 release not out yet.

  - Yes, but v1 release not out yet.

  - Yes, Kotlin lombok plugin to allow for using Java lombok annotations from Kotlin (required for AstRewriter and SqlDialect)

## License Information

### DIFF
--- a/partiql-ast/build.gradle.kts
+++ b/partiql-ast/build.gradle.kts
@@ -21,10 +21,6 @@ plugins {
     kotlin("plugin.lombok") version "2.1.0"
 }
 
-kotlinLombok {
-    lombokConfigurationFile(file("lombok.config"))
-}
-
 dependencies {
     api(Deps.ionElement)
     compileOnly(Deps.lombok)
--- a/partiql-ast/build.gradle.kts
+++ b/partiql-ast/build.gradle.kts
@@ -18,7 +18,7 @@ plugins {
     id(Plugins.publish)
     // To allow for Kotlin code in partiql-ast to understand Java Lombok annotations, need the following plugin.
     // https://kotlinlang.org/docs/lombok.html
-    kotlin("plugin.lombok") version "2.1.0"
+    id(Plugins.kotlinLombok) version Versions.kotlinLombok
 }
 
 dependencies {
--- a/partiql-ast/build.gradle.kts
+++ b/partiql-ast/build.gradle.kts
@@ -16,7 +16,7 @@
 plugins {
     id(Plugins.conventions)
     id(Plugins.publish)
-    // To allow for Kotlin code in partiql-ast to understand Java Lombok annotations, need the following plugin.
+    // Need the Kotlin lombok plugin to allow for Kotlin code in partiql-ast to understand Java Lombok annotations.
     // https://kotlinlang.org/docs/lombok.html
     id(Plugins.kotlinLombok) version Versions.kotlinLombok
 }
--- a/partiql-ast/build.gradle.kts
+++ b/partiql-ast/build.gradle.kts
@@ -16,6 +16,13 @@
 plugins {
     id(Plugins.conventions)
     id(Plugins.publish)
+    // To allow for Kotlin code in partiql-ast to understand Java Lombok annotations, need the following plugin.
+    // https://kotlinlang.org/docs/lombok.html
+    kotlin("plugin.lombok") version "2.1.0"
+}
+
+kotlinLombok {
+    lombokConfigurationFile(file("lombok.config"))
 }
 
 dependencies {
--- a/partiql-ast/src/main/java/org/partiql/ast/DataType.java
+++ b/partiql-ast/src/main/java/org/partiql/ast/DataType.java
@@ -40,8 +40,11 @@ public final class DataType extends AstEnum {
         private final boolean optional;
 
         @Nullable
+        @Getter
         private final List<AttributeConstraint> constraints;
+
         @Nullable
+        @Getter
         private final String comment;
 
         public StructField(
--- a/partiql-ast/src/main/java/org/partiql/ast/DataType.java
+++ b/partiql-ast/src/main/java/org/partiql/ast/DataType.java
@@ -1,6 +1,7 @@
 package org.partiql.ast;
 
 import lombok.EqualsAndHashCode;
+import lombok.Getter;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.partiql.ast.ddl.AttributeConstraint;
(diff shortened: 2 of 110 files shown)


partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
```diff
@@ -757,23 +775,25 @@ public abstract class SqlDialect : AstVisitor<SqlBlock, SqlBlock>() {
             else -> throw UnsupportedOperationException("Cannot print $node")
         }
         t = t concat list("", "") { node.keys }
-        t = if (node.asAlias != null) t concat " GROUP AS ${node.asAlias.sql()}" else t
+        val asAlias = node.asAlias
+        t = if (asAlias != null) t concat " GROUP AS ${asAlias.sql()}" else t
```
alancai98, line 779: (self-review): previously, when we used public fields, we could use the smart-casting of these nullable fields to non-null values (e.g. call `node.asAlias.sql()` directly). Since we switch to use `getters`, will need to define a variable first that we perform null-checks on.

partiql-ast/src/main/java/org/partiql/ast/Explain.java
```diff
@@ -13,12 +14,14 @@
  */
 @Builder(builderClassName = "Builder")
 @EqualsAndHashCode(callSuper = false)
-public class Explain extends Statement {
+public final class Explain extends Statement {
```
alancai98, line 17: (self-review): change all of the concrete classes to be `final`.

partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
```diff
@@ -394,19 +396,20 @@ public abstract class SqlDialect : AstVisitor<SqlBlock, SqlBlock>() {
     override fun visitExprLike(node: ExprLike, tail: SqlBlock): SqlBlock {
         var t = tail
         t = visitExprWrapped(node.value, t)
-        t = t concat if (node.not) " NOT LIKE " else " LIKE "
+        t = t concat if (node.isNot) " NOT LIKE " else " LIKE "
```
alancai98, line 399: (self-review): Lombok's default getter for `boolean` (not the boxed `Boolean`; boxed will still use `get`) will use the `is` prefix rather than `get`. Stylistically, `is` is
- more consistent with what we do in the plan already (e.g. RelUnion's `isAll`)
- default for IntelliJ when creating new getters from private fields
- if we care about conventions, JavaBeans prescribes using `is` over `get`

This change applies to other places we had used a boolean getter (e.g. identifier's `delimited`).

partiql-ast/build.gradle.kts
```diff
@@ -16,6 +16,9 @@
 plugins {
     id(Plugins.conventions)
     id(Plugins.publish)
+    // Need the Kotlin lombok plugin to allow for Kotlin code in partiql-ast to understand Java Lombok annotations.
+    // https://kotlinlang.org/docs/lombok.html
+    id(Plugins.kotlinLombok) version Versions.kotlinLombok
```
alancai98, line 21: (self-review): Kotlin lombok plugin required to allow Kotlin code to understand Java lombok annotations -- https://kotlinlang.org/docs/lombok.html. It is experimental but 
- seems pretty stable so far since its published by the Kotlin team
- we could always use a fixed version or hand-write the getters if we encounter issues without breaking the public API

partiql-ast/src/main/java/org/partiql/ast/DataType.java
```diff
@@ -28,26 +29,30 @@ public class DataType extends AstEnum {
...
-        public final boolean isOptional;
+        @Getter
+        private final boolean optional;
 
         @Nullable
-        public final List<AttributeConstraint> constraints;
+        private final List<AttributeConstraint> constraints;
         @Nullable
-        public final String comment;
+        private final String comment;
```
johnedquinn, line 45: These need `@Getter`

partiql-ast/src/main/java/org/partiql/ast/SelectItem.java
```diff
@@ -71,5 +73,15 @@ public List<AstNode> getChildren() {
...
+
+        @NotNull
+        public org.partiql.ast.expr.Expr getExpr() {
+            return this.expr;
+        }
+
+        @Nullable
+        public Identifier getAsAlias() {
+            return this.asAlias;
+        }
```
johnedquinn, line 85: Is there a reason you didn't use Lombok here?

partiql-ast/src/main/java/org/partiql/ast/ddl/TableConstraint.java
```diff
@@ -13,7 +14,8 @@
 
 public abstract class TableConstraint extends AstNode {
     @Nullable
-    public final IdentifierChain name;
+    @Getter
+    protected final IdentifierChain name;
```
johnedquinn, line 18: If you want to truly hide this field, you could just handwrite these getters. Same with PathStep.

partiql-ast/src/main/java/org/partiql/ast/QueryBody.java
```diff
@@ -69,19 +77,21 @@ public <R, C> R accept(@NotNull AstVisitor<R, C> visitor, C ctx) {
...
 
-        public final boolean isOuter;
+        @Getter
+        private final boolean outer;
 
         @NotNull
         public Expr lhs;
 
         @NotNull
         public Expr rhs;
```
johnedquinn, line 90: These need `@Getter` and private.

johnedquinn (changes requested):

partiql-ast/src/main/java/org/partiql/ast/DataType.java
alancai98: Thanks, forgot those

partiql-ast/src/main/java/org/partiql/ast/SelectItem.java
alancai98: No reason. Think I was just testing out the de-lomboked code here. I'll change to use Lombok

partiql-ast/src/main/java/org/partiql/ast/ddl/TableConstraint.java
alancai98: Made the field private (same w/ `PathStep`). Lombok getter seems to work fine, so I just used that for now.

johnedquinn (approved):

PR #1688, opened 2024-12-19
[v1] Remove UNKNOWN AST enum variant

alancai98:
## Relevant Issues

## Description
- Removes the `UNKNOWN` AST enum variant
- Renumbers enums to start from 0. We can always add back `UNKNOWN` or some "OTHER" variant in the future.
- Renames `TruthValue.UNK` to `TruthValue.UNKOWN`

### DIFF
--- a/partiql-ast/src/main/java/org/partiql/ast/DatetimeField.java
+++ b/partiql-ast/src/main/java/org/partiql/ast/DatetimeField.java
@@ -11,19 +11,14 @@ import java.util.List;
  */
 @EqualsAndHashCode(callSuper = false)
 public class DatetimeField extends AstEnum {
-    public static final int UNKNOWN = 0;
-    public static final int YEAR = 1;
-    public static final int MONTH = 2;
-    public static final int DAY = 3;
-    public static final int HOUR = 4;
-    public static final int MINUTE = 5;
-    public static final int SECOND = 6;
-    public static final int TIMEZONE_HOUR = 7;
-    public static final int TIMEZONE_MINUTE = 8;
-
-    public static DatetimeField UNKNOWN() {
-        return new DatetimeField(UNKNOWN);
-    }
+    public static final int YEAR = 0;
+    public static final int MONTH = 1;
+    public static final int DAY = 2;
+    public static final int HOUR = 3;
+    public static final int MINUTE = 4;
+    public static final int SECOND = 5;
+    public static final int TIMEZONE_HOUR = 6;
+    public static final int TIMEZONE_MINUTE = 7;
 
     public static DatetimeField YEAR() {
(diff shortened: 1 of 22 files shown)


partiql-ast/src/main/java/org/partiql/ast/DatetimeField.java
```diff
@@ -107,7 +102,7 @@ public static DatetimeField parse(@NotNull String value) {
             case "SECOND": return SECOND();
             case "TIMEZONE_HOUR": return TIMEZONE_HOUR();
             case "TIMEZONE_MINUTE": return TIMEZONE_MINUTE();
-            default: return UNKNOWN();
+            default: throw new IllegalArgumentException("No enum constant DatetimeField." + value);
```
alancai98, line 105: (self-review): behavior follows what Kotlin does for its `valueOf()` method, which throws an `IllegalArgumentException` if the string value does not match any of the enum variants.

partiql-ast/src/main/java/org/partiql/ast/expr/TruthValue.java
```diff
@@ -13,21 +13,16 @@
  */
 @EqualsAndHashCode(callSuper = false)
 public class TruthValue extends AstEnum {
-    public static final int UNKNOWN = 0;
-    public static final int TRUE = 1;
-    public static final int FALSE = 2;
-    public static final int UNK = 3;
+    public static final int TRUE = 0;
+    public static final int FALSE = 1;
+    public static final int UNKNOWN = 2;
```
alancai98, line 18: (self-review): change of `UNK` -> `UNKNOWN` is in reference to this recent PR comment -- https://github.com/partiql/partiql-lang-kotlin/pull/1679#discussion_r1884644109.

partiql-ast/src/main/java/org/partiql/ast/expr/Scope.java
```diff
@@ -14,13 +14,8 @@
  */
 @EqualsAndHashCode(callSuper = false)
 public class Scope extends AstEnum {
```
RCHowell, line 16: Where is Scope used in the AST? This might be a holdover from @T vs T which I think shouldn't be "scope" but more like some kind of qualifier boolean on the ExprVar/ExprId or whatever we are calling it.

RCHowell (approved):

partiql-ast/src/main/java/org/partiql/ast/expr/Scope.java
alancai98: Yeah, `ExprVarRef` is the only place it's used. In a future PR (probably when I update the docs), I'll remove the `Scope` enum opt for a qualifier boolean.

Issue #1690, opened 2024-12-24
Redefine non-standard DATE_ADD and DATE_DIFF to interval arithmetic


RCHowell:
## Description

Once https://github.com/partiql/partiql-lang-kotlin/issues/1657 is completed, then we can redefine DATE_ADD and DATE_DIFF to be special syntax for the standard SQL datetime arithmetic operators.

This is outside the scope of updating Datum to use java.time, so existing DATE_ADD and DATE_DIFF only matches 0.14 behavior which is not ideal and wrong in some places.

xd1313113:
datetime arithmetic operators are not fully equivalent to DATE_ADD and DATE_DIFF, close as the decision is to keep those functions.

Gaps.
Date_Add supports 
```
date_add(DAY, long/expr, DATE) // Interval literal does not support expr. Also internal literal expects int while long is used in date_add
date_diff(DAY, date,date) // produces long type while `date - date` produce interval and the function parameter can be expr
```

Issue #1646, opened 2024-11-14
[Tracking] Converting function implementations with provider interfaces


johnedquinn:
## Relevant Issues/PRs
- #1598
- #1603

## Description
As I was going to replace a single function (the plus operator) with the new modeling for functions, I realized that some aspects of the plumbing weren't fully flushed out. This led to all tests using the old plumbing to fail. Therefore, we need to update all functions.

Here are some items to track:
- [ ] Arithmetic operators
  - [ ] PLUS, BITWISE_AND
  - [ ] MINUS, MULT, DIV, etc
- [ ] Predicates
- [ ] Logical Ops
- [ ] Comparison Ops

PR #1695, opened 2025-01-03
Renames all hidden built-in functions using unicode non-character

johnedquinn:
## Description
- Prefixes all hidden built-in functions using unicode non-character
- If there is a customer need for overriding these built-ins, we can standardize/publicize this and/or make it more user-friendly

## License Information

### DIFF
--- a/partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
+++ b/partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
@@ -387,6 +387,16 @@ internal class PartiQLParserDefault : PartiQLParser {
 
         companion object {
 
+            /**
+             * The internal system prefix is '\uFDEF', one of unicode's 'internal-use' non-characters. This allows us to "hide"
+             * certain functions from being directly invocable via PartiQL text.
+             * See:
+             * - http://www.unicode.org/faq/private_use.html#nonchar1
+             * - http://www.unicode.org/versions/Unicode5.2.0/ch16.pdf#G19635
+             * - http://www.unicode.org/versions/corrigendum9.html
+             */
+            private const val SYSTEM_PREFIX_INTERNAL: String = "\uFDEF"
+
             private val rules = GeneratedParser.ruleNames.asList()
 
             /**
@@ -1751,9 +1761,10 @@ internal class PartiQLParserDefault : PartiQLParser {
             val rhs = visitExpr(ctx.expr(1))
             val fieldLit = ctx.dt.text.lowercase()
             // TODO error on invalid datetime fields like TIMEZONE_HOUR and TIMEZONE_MINUTE
+            // TODO: This should (likely) be parsed into its own node. The planner should convert this into a call.
             when {
-                ctx.DATE_ADD() != null -> exprCall(identifierChain(identifier("date_add_$fieldLit", false), null), listOf(lhs, rhs), null)
-                ctx.DATE_DIFF() != null -> exprCall(identifierChain(identifier("date_diff_$fieldLit", false), null), listOf(lhs, rhs), null)
+                ctx.DATE_ADD() != null -> exprCall(identifierChain(identifier("${SYSTEM_PREFIX_INTERNAL}date_add_$fieldLit", false), null), listOf(lhs, rhs), null)
+                ctx.DATE_DIFF() != null -> exprCall(identifierChain(identifier("${SYSTEM_PREFIX_INTERNAL}date_diff_$fieldLit", false), null), listOf(lhs, rhs), null)
                 else -> throw error(ctx, "Expected DATE_ADD or DATE_DIFF")
             }
         }
--- a/partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
+++ b/partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
@@ -1761,7 +1761,7 @@ internal class PartiQLParserDefault : PartiQLParser {
             val rhs = visitExpr(ctx.expr(1))
             val fieldLit = ctx.dt.text.lowercase()
             // TODO error on invalid datetime fields like TIMEZONE_HOUR and TIMEZONE_MINUTE
-            // TODO: This should (likely) be parsed into its own node. The planner should convert this into a call.
+            // TODO: This should (maybe) be parsed into its own node. We could convert this into an operator. See https://github.com/partiql/partiql-lang-kotlin/issues/1690.
             when {
                 ctx.DATE_ADD() != null -> exprCall(identifierChain(identifier("${SYSTEM_PREFIX_INTERNAL}date_add_$fieldLit", false), null), listOf(lhs, rhs), null)
                 ctx.DATE_DIFF() != null -> exprCall(identifierChain(identifier("${SYSTEM_PREFIX_INTERNAL}date_diff_$fieldLit", false), null), listOf(lhs, rhs), null)
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
@@ -90,6 +90,7 @@ import org.partiql.planner.internal.ir.rexOpVarUnresolved
 import org.partiql.planner.internal.typer.CompilerType
 import org.partiql.planner.internal.typer.PlanTyper.Companion.toCType
 import org.partiql.planner.internal.util.DateTimeUtils
+import org.partiql.planner.internal.utils.FunctionUtils
 import org.partiql.spi.catalog.Identifier
 import org.partiql.spi.value.Datum
 import org.partiql.types.PType
@@ -286,8 +287,8 @@ internal object RexConverter {
             // Fn
             val name = when (symbol) {
                 // TODO move hard-coded operator resolution into SPI
-                "+" -> "pos"
-                "-" -> "neg"
+                "+" -> FunctionUtils.OP_POS
+                "-" -> FunctionUtils.OP_NEG
                 else -> error("unsupported unary op $symbol")
             }
             val id = Identifier.delimited(name)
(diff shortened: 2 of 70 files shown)


partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -1752,9 +1762,10 @@ internal class PartiQLParserDefault : PartiQLParser {
             val rhs = visitExpr(ctx.expr(1))
             val fieldLit = ctx.dt.text.lowercase()
             // TODO error on invalid datetime fields like TIMEZONE_HOUR and TIMEZONE_MINUTE
+            // TODO: This should (likely) be parsed into its own node. The planner should convert this into a call.
```
RCHowell, line 1765: Or better yet, rewritten using SQL standard operators. https://github.com/partiql/partiql-lang-kotlin/issues/1690

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnDateAddMonth.kt
```diff
@@ -65,7 +65,7 @@ internal val Fn_DATE_ADD_MONTH__INT_DATE__DATE = Function.static(
     Datum.date(datetimeValue.plusMonths(intervalValue))
 }
 
-internal val Fn_DATE_ADD_MONTH__INT32_TIMESTAMP__TIMESTAMP = Function.static(
+internal val Fn_DATE_ADD_MONTH__INT32_TIMESTAMP__TIMESTAMP = FunctionUtils.hidden(
```
RCHowell, line 68: See these, I recommend just adding ".hidden" or ".system" to Function rather than FunctionUtils, but I also don't know if you had to deal with visibility modifier issues.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnEq.kt
```diff
@@ -25,12 +26,13 @@ import org.partiql.types.PType
 internal object FnEq : Function {
 
     // Memoize shared variables
+    private const val NAME = FunctionUtils.SYSTEM_PREFIX_INTERNAL + "eq"
```
RCHowell, line 29: Rather than "computing" the name in every function definition, make it a helper.

```
val name = FunctionUtils.hide("eq")
```

partiql-spi/src/main/kotlin/org/partiql/spi/function/utils/FunctionUtils.kt
```diff
@@ -0,0 +1,62 @@
...
+     */
+    fun hidden(
+        name: String,
+        parameters: Array<Parameter>,
+        returns: PType,
+        isNullCall: Boolean = true,
+        isMissingCall: Boolean = true,
+        invoke: (Array<Datum>) -> Datum,
+    ): Function {
+        val hiddenName = "$SYSTEM_PREFIX_INTERNAL$name"
```
RCHowell, line 34: Use the same "hide" function for computing name so we don't have the logic in N different places. If you were to change to something like "$_" then you'd have to update every function as well as this. See if you can get everything to be behind one method.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
```diff
@@ -325,24 +326,24 @@ internal object RexConverter {
...
-                        ">=" -> "gte"
-                        "=" -> "eq"
-                        "||" -> "concat"
-                        "+" -> "plus"
-                        "-" -> "minus"
-                        "*" -> "times"
-                        "/" -> "divide"
-                        "%" -> "modulo"
-                        "&" -> "bitwise_and"
+                        "<" -> FunctionUtils.OP_LT
```
RCHowell, line 335: How about a "System.get(..)" pattern here rather than having a big list of hardcoded strings? Maybe it's still hardcoded strings, but at least packaged within something that could later be expanded upon.

RCHowell (changes requested):

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnExists.kt
```diff
@@ -33,6 +33,7 @@ internal val Fn_EXISTS__LIST__BOOL = Function.static(
     Datum.bool(exists)
 }
 
+// TODO: Should this exist?
 internal val Fn_EXISTS__STRUCT__BOOL = Function.static(
```
alancai98, line 37: What's this TODO for?

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrimTrailing.kt
```diff
@@ -3,13 +3,13 @@
...
 package org.partiql.spi.function.builtins
 
-import org.partiql.spi.function.Function
 import org.partiql.spi.function.Parameter
+import org.partiql.spi.function.utils.FunctionUtils
 import org.partiql.spi.function.utils.StringUtils.codepointTrimTrailing
 import org.partiql.spi.value.Datum
 import org.partiql.types.PType
 
-internal val Fn_TRIM_TRAILING__STRING__STRING = Function.static(
```
alancai98, line 12: Seems like some other functions aren't hidden yet. Taking a look at the https://github.com/johnedquinn/partiql-lang-jvm/blob/main-builtins-names/partiql-spi/src/main/kotlin/org/partiql/spi/function/Builtins.kt#L21,
- abs
- bit_length
- cardinality
- char_length
- lower
- (some others)

partiql-spi/src/main/kotlin/org/partiql/spi/function/utils/FunctionUtils.kt
```diff
@@ -0,0 +1,62 @@
...
+     * - http://www.unicode.org/faq/private_use.html#nonchar1
+     * - http://www.unicode.org/versions/Unicode5.2.0/ch16.pdf#G19635
+     * - http://www.unicode.org/versions/corrigendum9.html
+     */
+    const val SYSTEM_PREFIX_INTERNAL: String = "\uFDEF"
+
+    /**
+     * Returns an implementation of a [Function] that has the [name] obfuscated via the [SYSTEM_PREFIX_INTERNAL].
+     */
+    fun hidden(
```
alancai98, line 26: Will aggregations also be hidden w/ the special prefix?

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnNot.kt
```diff
@@ -5,12 +5,16 @@ package org.partiql.spi.function.builtins
...
 import org.partiql.spi.value.Datum
 import org.partiql.types.PType
 
+/**
+ * This is the boolean NOT predicate. Its name is obfuscated via the use of [FunctionUtils.SYSTEM_PREFIX_INTERNAL].
+ */
 internal val Fn_NOT__BOOL__BOOL = object : Function {
 
-    private var name = "not"
+    private val name = FunctionUtils.SYSTEM_PREFIX_INTERNAL + "not"
```
alancai98, line 17: There are some functions (like here and `IS_MISSING`, `IS_NULL`, etc) that do not directly use the `FunctionUtils.hidden` method. Is that because they haven't been migrated to use the new interface (from https://github.com/partiql/partiql-lang-kotlin/issues/1646)?

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnIsMissing.kt
```diff
@@ -5,15 +5,16 @@ package org.partiql.spi.function.builtins
 
 import org.partiql.spi.function.Function
 import org.partiql.spi.function.Parameter
+import org.partiql.spi.function.utils.FunctionUtils
 import org.partiql.spi.value.Datum
 import org.partiql.types.PType
 
 /**
- * Function (operator) for the `IS MISSING` special form.
+ * Function (operator) for the `IS MISSING` special form. Its name is obfuscated via [FunctionUtils.SYSTEM_PREFIX_INTERNAL].
```
alancai98, line 13: Similarly for some other comments.

```suggestion
 * Function (operator) for the `IS MISSING` special form. Its name is hidden via [FunctionUtils.SYSTEM_PREFIX_INTERNAL].
```

alancai98 (changes requested):

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrimTrailing.kt
johnedquinn: These are public in SQL. See Section 20.70 and 6.17.

partiql-spi/src/main/kotlin/org/partiql/spi/function/utils/FunctionUtils.kt
johnedquinn: No. See https://github.com/partiql/partiql-lang-kotlin/pull/1695#discussion_r1905854234. Same reasoning.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnNot.kt
johnedquinn: Not necessarily. They don't have a bunch of overloads, so they don't need to directly implement the new interface.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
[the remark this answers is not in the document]
johnedquinn: True. For now, I'm just adding the prefix and leaving the TODO. Post V1, it might be a good idea to reintroduce those nodes or go with the translation that Robert mentioned. I don't have a preference.

RCHowell:
Overall LGTM and gets that job done, but you may want to consider not having a FunctionUtils and not listing out the names of functions. It would be better to just reference the name on the operator itself. It's not ideal to have the names floating around in multiple places because now the process of adding operators is fragile and requires updating in multiple places which does not scale well. I recommend brainstorming better ways to register and reference builtins that does not require juggling

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrimTrailing.kt
alancai98: I see. Perhaps could be good to add a comment at some place (perhaps in the FunctionUtils.kt) that these SQL99-defined scalar and aggregation function names are purposely not hidden.

RCHowell (approved):

alancai98 (approved):

PR #1700, opened 2025-01-08
[v1] Remodel AST paths exprs and identifiers; rename SPI's Identifier part to simple

alancai98:
## Relevant Issues
- #1610 

## Description
- Remodels AST path expressions to be more consistent with exclude modeling
- Remodels AST identifiers to be more consistent with our parsing rules and SPI modeling

## License Information


partiql-ast/src/main/java/org/partiql/ast/expr/ExprPath.java
```diff
@@ -19,22 +18,20 @@ public final class ExprPath extends Expr {
     @NotNull
     private final Expr root;
 
-    @Nullable
-    private final PathStep next;
+    @NotNull
+    private final List<PathStep> steps;
```
alancai98, line 22: (self-review) `ExprPath` modeling changed to be more alike the EXCLUDE path modeling. The prior linked list approach could be slightly annoying to use compared to just a normal Java list.

partiql-ast/src/main/java/org/partiql/ast/Identifier.java
```diff
@@ -8,25 +8,29 @@
...
 @Builder(builderClassName = "Builder")
 @EqualsAndHashCode(callSuper = false)
 public final class Identifier extends AstNode {
     @NotNull
-    private final String symbol;
+    private final List<Part> qualifier;
 
-    private final boolean delimited;
+    @NotNull
+    private final Part base;
```
alancai98, line 20: (self-review) modeling more closely follows the parser rule (https://github.com/partiql/partiql-lang-kotlin/blob/ast-cleanup/partiql-parser/src/main/antlr/PartiQLParser.g4#L60-L61) and the SPI modeling for Identifiers -- https://github.com/partiql/partiql-lang-kotlin/blob/ast-cleanup/partiql-spi/src/main/kotlin/org/partiql/spi/catalog/Identifier.kt#L12-L15. This modeling ended up simplifying a lot of the conversion code since the prior linked list modeling could be cumbersome to work with.

New modeling follows SQL1999's BNF rule defined in section 5.4:
```
<local or schema qualified name> ::=
[ <local or schema qualifier> <period> ] <qualified identifier>
```

I wasn't too sure on the differences between this `<local or schema qualified name>` and `<identifier chain>` but it seems like `<local or schema qualified name>` is used in more places such as FROM source identifiers and CREATE TABLE table names.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/helpers/ToBinder.kt
```diff
@@ -38,37 +38,22 @@ internal fun Expr.toBinder(index: () -> Int): Identifier = when (this) {
...
-private fun IdentifierChain.toBinder(): Identifier {
-    if (next == null) return root.symbol.toBinder()
-    var cur = next
-    var prev = cur
-    while (cur != null) {
-        prev = cur
-        cur = cur.next
-    }
-    return prev!!.root.symbol.toBinder()
-}
```
alancai98, line 56: (self-review) example of the complicated logic to extract the "base" of a potentially qualified identifier. This is much simpler with the new modeling -- `this.base.toBinder()`

partiql-ast/src/main/java/org/partiql/ast/Identifier.java
RCHowell: Thanks for the recap. Alan and I discussed offline, and we agreed this modeling most closely aligns with how we discuss identifiers and their qualifier. I don't love the "Identifier.Part" but SQL BNF does not give us a good name here ("actual identifier" is meh) and "part" makes sense and is short and sweet.

Happy to hear this reduced conversion code as we've battled that across all different models of identifiers, names, binding paths, path expressions, exclusion paths etc.

partiql-ast/src/main/java/org/partiql/ast/Identifier.java
```diff
@@ -35,11 +39,49 @@ public <R, C> R accept(@NotNull AstVisitor<R, C> visitor, C ctx) {
...
+    public List<Part> getQualifier() {
+        return this.qualifier;
     }
 
-    public boolean isDelimited() {
-        return this.delimited;
+    @NotNull
+    public Part getBase() {
+        return this.base;
+    }
```
RCHowell, line 49: While "getIdentifier()" and "identifier" is not ideal (maybe base is better?) I recommend going with "getIdentifier" only for SPI consistency, but I understand what you are going for

partiql-ast/src/main/java/org/partiql/ast/Identifier.java
```diff
@@ -35,11 +39,49 @@ public <R, C> R accept(@NotNull AstVisitor<R, C> visitor, C ctx) {
...
+
+        @NotNull
+        public String getSymbol() {
+            return this.symbol;
+        }
+
+        public boolean isDelimited() {
+            return this.delimited;
+        }
     }
```
RCHowell, line 86: Did you consider adding the same helper methods that SPI has for creating identifiers? It is cumbersome to create them, but the helper methods are an easy addition with a lot of utility.

partiql-ast/src/main/java/org/partiql/ast/Let.java
```diff
@@ -47,9 +47,9 @@ public static class Binding extends AstNode {
         private final Expr expr;
 
         @NotNull
-        private final Identifier asAlias;
+        private final Identifier.Part asAlias;
```
RCHowell, line 50: Hmm seeing all of this makes me thing "Identifier.Single" or "Identifier.Simple" is better than Part. I leave it entirely up to your opinion and John's – my only requirement is that any change is also made to SPI so that these are the same. As I said earlier, I don't love "part" and seeing the single/simple identifier used in many more places has inspired a better definition

RCHowell (dismissed):

partiql-ast/src/main/java/org/partiql/ast/Identifier.java
alancai98: Yeah, I was planning to add them. I didn't get a chance to add them yesterday

partiql-ast/src/main/java/org/partiql/ast/Let.java
alancai98: Naming is definitely hard 😅. I'll try to get some consensus here before adjusting in the AST + SPI

alancai98: Just my opinion though they are all reasonable options:
- part
  - nothing wrong w/ it,
  - makes it clear that it's just a segment of an Identifier
  - an `Identifier` is built up from a list of parts (a qualifier that's a list of parts and a base part)
- single
  - could be confused for an individual identifier which could be qualified
- simple
  - I think slightly better than `single`
  - better for the base identifier but may be more confusing for the qualifier (i.e. a qualifier is a list of simple identifiers)

johnedquinn: SQL's `<identifier>` is unqualified. And `<table name>`, `<local or schema qualified name>`, and `<identifier chain>` are all (potentially) qualified identifiers.

If we want `Identifier` to be represent the qualified identifier, then I think I like `Identifer.Simple` with some good Javadocs telling the difference between the two. An alternative is two have two classes: `IdentQualified/IdentifierQualified` and `Ident/Identifier`. Though, I'm fine with whatever approach you take @alancai98.

partiql-ast/src/main/java/org/partiql/ast/Identifier.java
alancai98: Yeah, `base` is too generic -- changed to `identifier`.

partiql-ast/src/main/java/org/partiql/ast/Let.java
alancai98: Changed to `Identifier.Simple` in the AST and SPI. Perhaps it's overkill to create two separate classes (like the v0.14.* modeling), so kept it as a nested class within `Identifier`.

partiql-ast/src/main/java/org/partiql/ast/Identifier.java
```diff
@@ -35,11 +41,131 @@ public <R, C> R accept(@NotNull AstVisitor<R, C> visitor, C ctx) {
...
-    public String getSymbol() {
-        return this.symbol;
+    public List<Simple> getQualifier() {
+        return this.qualifier;
+    }
+
+    @NotNull
+    public Simple getIdentifier() {
+        return this.identifier;
+    }
```
johnedquinn, line 51: Can you indicate that this is the right-most identifier in the Javadocs? AKA:
> For `a.b.c`, this method would return `c`.

johnedquinn (approved):

partiql-ast/src/main/java/org/partiql/ast/Identifier.java
alancai98: Yeah, I have another PR coming soon that updates the docs before the final release. I'll add this note.

alancai98: https://github.com/partiql/partiql-lang-kotlin/pull/1704 adds the comment/

PR #1703, opened 2025-01-09
SPI & Types Package Audit

johnedquinn:
## Description
- SPI Package Audit. Moves classes around.

## License Information


gradle.properties
```diff
@@ -1,5 +1,5 @@
 group=org.partiql
-version=1.0.0-rc.4-SNAPSHOT
+version=1.0.0-rc.4
```
alancai98, line 2: Snapshot removed but maybe we should just remove it while we're doing the package audits? We're going to release 1.0.0 soon anyways

partiql-types/README.adoc
```diff
@@ -1,317 +0,0 @@
-= PartiQL Types & Values
```
alancai98, line 1: Should this be copied over to spi as well?

partiql-plan/src/main/java/org/partiql/plan/JoinType.java
```diff
@@ -14,6 +14,22 @@ public class JoinType extends Enum {
...
+            case INNER:
+                return "INNER";
+            case LEFT:
+                return "LEFT";
+            case RIGHT:
+                return "RIGHT";
+            case FULL:
+                return "FULL";
+        }
+        return "UNKNOWN";
```
alancai98, line 30: Similar to https://github.com/partiql/partiql-lang-kotlin/pull/1688, do we see uses for an `UNKNOWN` variant yet? If not, perhaps we can remove it for now to be consistent with the AST enums?

alancai98: Also some of the `name` functions follow a different implementation (i.e. some use `String.valueOf(code);` for the default). We should align them to the same implementation

johnedquinn: Removed them! And, I made the `name()` contract a bit more concrete. See the update.

gradle.properties
johnedquinn: Yeah that's fine by me 👍

partiql-types/README.adoc
johnedquinn: Eh, it's largely out of date with the use of PartiQLValue. Also, documentation for types will be on the website, especially with the most recent contribution.

alancai98 (approved):

PR #1709, opened 2025-01-10
Audits eval package

johnedquinn:
## Description
- Audits eval package
- Removes the public `getValues()` on `Row`.

## License Information


partiql-eval/src/main/java/org/partiql/eval/Row.java
```diff
@@ -10,14 +10,32 @@
  */
 public class Row {
 
+    private final Datum[] values;
```
alancai98, line 13: nit: perhaps could still be helpful to have a method to get all the `values`?

alancai98 (approved):
Just one suggestion but looks good otherwise

Issue #1680, opened 2024-12-13
[v1] Function data type mismatches (known at compile-time) return missing in strict mode


alancai98:
## Description
Currently, `PError.FUNCTION_TYPE_MISMATCH` has a default severity of warning. Because of this, the standard error listener used during planning will not give an error in strict mode when there's a function data type mismatch. The function data type mismatch will create a `RexError` node -- https://github.com/partiql/partiql-lang-kotlin/blob/b65141557d5545e41abae6568e9d82f76e76333c/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt#L757-L759. After planning, that error node will be converted into an `ExprMissing` node.

## To Reproduce
Run the following query in strict mode

```
'foo' || 123
```
^ will output missing

If the type involved in the function/operation is not always a data type mismatch, the query will give the expected error at evaluation time:
```
SELECT v.a + 2 FROM [{'a':1}, {'a':'foo'}] v
```

^ will error:
```
org.partiql.errors.TypeCheckException: Could not find function plus with types: [STRING, INTEGER].
```

## Expected Behavior
- Expect the query `'foo' || 123` to give an error in strict mode
- I think we should also make function data type mismatches (`PError.FUNCTION_TYPE_MISMATCH`) an error by default so the standard error listener will give an error at plan-time

## Additional Context
- Java version: 11
- PartiQL version: v1 rc.3
- Add any other context about the problem here.

RCHowell:
We will need to re-enable the spec tests so this gets caught now that we better understand the desired behavior here.

PR #1711, opened 2025-01-13
Fixes decimal/numeric arithmetic & mistyped function bugs

johnedquinn:
## Description

Fixes 45 conformance tests. There are two major commits in this PR. They should be looked at individually to aid with the review.
1. Fixes arithmetic conformance tests (6 tests).
  a. Did this by flushing out NUMERIC.
  b. Fixed some overflow for the arithmetic operations of divide/multiply/etc.
  c. Made the "MOD" function public. SQL:1999 has it user-accessible.
3. Fixes typing of functions conformance tests (39 tests).
  a. I just introduced RexError to error out in STRICT mode.
  b. I also fixed the typing of operations/functions where one of the operands is the MISSING value.

## License Information

### DIFF
--- a/partiql-eval/src/main/java/org/partiql/eval/Row.java
+++ b/partiql-eval/src/main/java/org/partiql/eval/Row.java
@@ -10,32 +10,14 @@ import java.util.Objects;
  */
 public class Row {
 
-    private final Datum[] values;
-
     /**
-     * @param index the requested index
-     * @return the value at the given index
+     * TODO internalize values.
      */
-    public Datum get(int index) {
-        return values[index];
-    }
+    public final Datum[] values;
 
     /**
-     * @param index the requested index
-     * @param value the value to insert
-     */
-    public void set(int index, Datum value) {
-        values[index] = value;
-    }
-
-    /**
-     * @return the number of values in the record
-     */
-    public int getSize() {
-        return values.length;
-    }
-
-    /**
+     * TODO keep ??
+     *
      * @param values the values
      * @return the record
      */
--- a/partiql-eval/src/main/java/org/partiql/eval/compiler/Match.java
+++ b/partiql-eval/src/main/java/org/partiql/eval/compiler/Match.java
@@ -4,7 +4,7 @@ import org.jetbrains.annotations.NotNull;
 import org.partiql.plan.Operand;
 
 /**
- * Match represents a subtree match to be sent to the {@link Strategy}.
+ * Match represents a subtree match to be sent to the
  */
 public class Match {
 
--- a/partiql-eval/src/main/java/org/partiql/eval/compiler/PartiQLCompiler.java
+++ b/partiql-eval/src/main/java/org/partiql/eval/compiler/PartiQLCompiler.java
@@ -26,14 +26,6 @@ public interface PartiQLCompiler {
         return prepare(plan, mode, Context.standard());
     }
 
-    /**
-     * Prepares the given plan into an executable PartiQL statement.
-     *
-     * @param plan The plan to compile.
-     * @param mode The mode to execute in.
-     * @param ctx The shared context object.
-     * @return The prepared statement.
-     */
     @NotNull
     public Statement prepare(@NotNull Plan plan, @NotNull Mode mode, @NotNull Context ctx);
 
--- a/partiql-eval/src/main/java/org/partiql/eval/compiler/Pattern.java
+++ b/partiql-eval/src/main/java/org/partiql/eval/compiler/Pattern.java
@@ -38,14 +38,11 @@ public class Pattern {
         this.predicate = predicate;
     }
 
-    /**
-     * @param operator the operator to match against.
-     * @return whether the operator matches the pattern.
-     */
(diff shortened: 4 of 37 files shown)


partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprError.kt
```diff
@@ -0,0 +1,12 @@
+package org.partiql.eval.internal.operator.rex
+
+import org.partiql.eval.Environment
+import org.partiql.eval.ExprValue
+import org.partiql.spi.errors.TypeCheckException
+import org.partiql.spi.value.Datum
+
+internal class ExprError : ExprValue {
+    override fun eval(env: Environment): Datum {
+        throw TypeCheckException("A warning has been converted into an error.")
```
alancai98, line 10: I was testing the CLI using the example provided in https://github.com/partiql/partiql-lang-kotlin/issues/1680. Works as expected:

```
-- running in strict mode
partiql ▶ 'foo' || 123
   | 
w: [semantic] Undefined function "﷯concat"(STRING, INTEGER). Did you mean: 
- ﷯concat(lhs: CHAR(256), rhs: CHAR(256))
- ﷯concat(lhs: VARCHAR(256), rhs: VARCHAR(256))
- ﷯concat(lhs: STRING, rhs: STRING)
- ﷯concat(lhs: CLOB(2147483647), rhs: CLOB(2147483647))

org.partiql.spi.errors.TypeCheckException: A warning has been converted into an error.
```

Wonder if we could plumb through the specific warning rather than just saying a warning has been changed to an error.

alancai98 (approved):
Nice conformance bump -- https://github.com/partiql/partiql-lang-kotlin/actions/runs/12778578996/attempts/1#summary-35621802787

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprError.kt
johnedquinn: We totally can. I'll tackle this in a follow-up PR. I've added the task to the requirements pre-V1 launch list.

partiql-eval/src/main/java/org/partiql/eval/Row.java
```diff
@@ -10,32 +10,14 @@
...
-     * @return the value at the given index
-     */
-    public Datum get(int index) {
-        return values[index];
-    }
-
     /**
-     * @param index the requested index
-     * @param value the value to insert
+     * TODO internalize values.
```
alancai98, line 14: mind if we move this to a public getter?

partiql-eval/src/main/java/org/partiql/eval/compiler/Match.java
```diff
@@ -4,7 +4,7 @@
 import org.partiql.plan.Operand;
 
 /**
- * Match represents a subtree match to be sent to the {@link Strategy}.
```
alancai98, line 7: part of the comment got deleted here

partiql-eval/src/main/java/org/partiql/eval/Row.java
```diff
@@ -10,32 +10,14 @@
...
-     * @return the number of values in the record
-     */
-    public int getSize() {
-        return values.length;
-    }
+    public final Datum[] values;
 
     /**
+     * TODO keep ??
+     *
```
alancai98, line 20: probably can keep?

partiql-eval/src/main/java/org/partiql/eval/compiler/PartiQLCompiler.java
```diff
@@ -26,14 +26,6 @@ default Statement prepare(@NotNull Plan plan, @NotNull Mode mode) {
         return prepare(plan, mode, Context.standard());
     }
 
-    /**
```
alancai98, line 29: comment was deleted

partiql-eval/src/main/java/org/partiql/eval/compiler/Pattern.java
```diff
@@ -38,14 +38,11 @@ protected Pattern(@NotNull Class<? extends Operator> clazz, @Nullable Predicate<
         this.predicate = predicate;
     }
 
-    /**
```
alancai98, line 41: similarly here comment was deleted

partiql-eval/src/main/java/org/partiql/eval/Row.java
johnedquinn: See https://github.com/partiql/partiql-lang-kotlin/pull/1711#discussion_r1917191059

partiql-eval/src/main/java/org/partiql/eval/compiler/Pattern.java
johnedquinn: See https://github.com/partiql/partiql-lang-kotlin/pull/1711#discussion_r1917191059

alancai98 (approved):

PR #1717, opened 2025-01-15
Flushes out PartiQL's runtime exceptions

johnedquinn:
## Description
- Flushes out PartiQL's runtime exceptions
- Added several error codes to remove the previous JVM exceptions
- There is one additional conformance test failing, but I think it's likely wrong. See the test in the report.

## License Information


partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/internal/Accumulator.kt
```diff
@@ -49,7 +48,7 @@ internal fun checkIsNumberType(funcName: String, value: Datum) {
         return checkIsNumberType(funcName, value.lower())
     }
     if (!value.type.isNumber()) {
-        throw TypeCheckException("Expected NUMBER but received ${value.type}.")
+        throw PErrors.unexpectedTypeException(value.type, listOf(PType.tinyint(), PType.smallint(), PType.integer(), PType.bigint(), PType.decimal(), PType.numeric(), PType.real(), PType.doublePrecision()))
```
alancai98, line 51: nit: could have some internal val within SPI that lists out the numeric types to make this less hard-coded

partiql-cli/src/main/kotlin/org/partiql/cli/pipeline/ErrorMessageFormatter.kt
```diff
@@ -74,6 +80,69 @@ object ErrorMessageFormatter {
...
+
+    /**
+     * @see PError.NUMERIC_VALUE_OUT_OF_RANGE
+     */
+    private fun numericValueOutOfRange(error: PError): String {
+        val value = error.getOrNull("VALUE", String::class.java)
+        val valueStr = prepare(value.toString(), " (", ")")
+        val type = error.getOrNull("TYPE", PType::class.java)
+        val typeString = prepare(type.toString(), "for type ")
+        return "Numeric value$valueStr is out of range$typeString."
```
alancai98, line 107: spacing is slightly off

```suggestion
        val typeString = prepare(type.toString(), " for type ")
        return "Numeric value$valueStr is out of range$typeString."
```

alancai98: may be better (for readability and consistency) to have the returned string template control the whitespace rather than the inner values:

e.g. 

```suggestion
        val typeString = prepare(type.toString(), "for type ")
        return "Numeric value $valueStr is out of range $typeString."
```
where `$valueStr` and `$typeString` don't have the space prefix.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprError.kt
```diff
@@ -2,11 +2,12 @@ package org.partiql.eval.internal.operator.rex
...
 import org.partiql.eval.ExprValue
-import org.partiql.spi.errors.TypeCheckException
+import org.partiql.eval.internal.helpers.PErrors
 import org.partiql.spi.value.Datum
 
 internal class ExprError : ExprValue {
     override fun eval(env: Environment): Datum {
-        throw TypeCheckException("A warning has been converted into an error.")
+        // TODO: Add error to ExprError
+        throw PErrors.internalErrorException(IllegalStateException("A warning has been converted into an error."))
```
alancai98, line 11: probably should be updated?

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprStructStrict.kt
```diff
@@ -13,7 +12,7 @@ internal class ExprStructStrict(private val fields: List<ExprStructField>) :
         val fields = fields.mapNotNull {
             val key = it.key.eval(env)
             if (key.isNull || key.isMissing) {
-                throw TypeCheckException("Struct key was absent.")
+                throw NullPointerException("Struct key was null or missing.")
```
alancai98, line 15: should this be some other exception rather than Java's `NullPointerException`?

partiql-eval/src/main/kotlin/org/partiql/eval/internal/helpers/PErrors.kt
```diff
@@ -0,0 +1,226 @@
...
+
+internal object PErrors {
+
+    /**
+     * Returns a PErrorException with code: [PError.NUMERIC_VALUE_OUT_OF_RANGE].
+     */
+    fun numericOutOfRangeException(value: String, type: PType): PErrorException {
+        val pError = numericOutOfRange(value, type)
+        return PErrorException(pError)
+    }
```
alancai98, line 18: wonder if we'll ever make these constructor functions public within SPI? there is a lot of overlap between the versions

partiql-cli/src/main/kotlin/org/partiql/cli/pipeline/ErrorMessageFormatter.kt
johnedquinn: Agree with the first comment.

The second comment wouldn't work if the type/value were null (which they can be). The string would look like:
```
Numeric value  is out of range .
```

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnDateAddDay.kt
```diff
@@ -57,11 +55,7 @@ internal val Fn_DATE_ADD_DAY__INT_DATE__DATE = FunctionUtils.hidden(
     val interval = args[0].bigDecimal
     val datetime = args[1].localDate
     val datetimeValue = datetime
-    val intervalValue = try {
-        interval.toLong()
-    } catch (e: DataException) {
-        throw TypeCheckException()
-    }
+    val intervalValue = interval.toLong()
```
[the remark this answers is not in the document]
johnedquinn: Actually, this is a remnant of when NUMERIC was meant to represent a BigInteger. Removing all of these :)

partiql-spi/src/main/java/org/partiql/spi/value/PErrors.java
```diff
@@ -0,0 +1,45 @@
...
+                new HashMap<String, Object>() {{
+                    put("VALUE", value);
+                    put("TYPE", type);
+                }}
+        );
+        return new PErrorException(pError);
+    }
+
+    @NotNull
+    static PErrorException wrappedException(@NotNull Throwable t) {
```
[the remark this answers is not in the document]
johnedquinn: This is just a package-private utility method to return a PError with INTERNAL_ERROR.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprError.kt
johnedquinn: This can be a follow-up, as it would be a backwards-compatible addition to ExprError.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/helpers/PErrors.kt
johnedquinn: The intention is to hide it for now unless a customer explicitly wants it. Trying to keep our public API minimal.

partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
```diff
@@ -433,15 +433,16 @@ static Datum decimal(@NotNull BigDecimal value) {
      * @param precision the precision to coerce the value to
      * @param scale the scale to coerce the value to
      * @return a value of type {@link PType#DECIMAL} with the requested precision/scale
-     * @throws DataException if the value could not fit into the requested precision/scale
+     * @throws PErrorException if the value could not fit into the requested precision/scale
```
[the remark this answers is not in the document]
johnedquinn: Adding. Though, for some (such as the length overflow of character strings), we don't have those codes yet. I'll be leaving those untouched, but for those with codes specified, I'm adding.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprStructStrict.kt
johnedquinn: See my latest commit. LMK if you're aligned.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/helpers/PErrors.kt
alancai98: Agree w/ keeping the public API minimal. We can always add it in the future. I guess we'll see how the usages evolve over time as we add more helpers.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprStructStrict.kt
alancai98: Returning null on null key and missing on missing key seems reasonable. Tbh not sure if the spec defines this behavior

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/internal/Accumulator.kt
alancai98: not sure if this has been addressed but non-blocking

alancai98 (approved):

PR #1716, opened 2025-01-15
[v1] Fix ABS integer overflow; add parsing of TINYINT

alancai98:
## Relevant Issues
- Closes https://github.com/partiql/partiql-lang-kotlin/issues/1714
- Closes https://github.com/partiql/partiql-lang-kotlin/issues/1698

## Description
- Adds parsing of TINYINT
- Fix ABS integer overflow
- Enables some previous TINYINT overflow tests

## License Information

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1715, opened 2025-01-14
[v1] Change integer types to error on overflow; give explicit data exception on div/mod by 0

alancai98:
## Relevant Issues
- Closes https://github.com/partiql/partiql-lang-kotlin/issues/1697

## Description
- Changes integer types to error on overflow
- Changes integer types division and modulo by 0 to give an explicit data exception

## License Information

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1696, opened 2025-01-06
[v1] Re-enable some randomized tests

alancai98:
## Relevant Issues
- Closes https://github.com/partiql/partiql-lang-kotlin/issues/1601
- https://github.com/partiql/partiql-lang-kotlin/issues/1697 (int overflow randomized test should be tracked in this other issue)

## Description
- Re-enables some randomized tests, specifically the datetime syntax and eval tests
- Deletes the `TimestampTemporalAccessor` tests since that class was removed from src files
- Adds todos for the integer overflow randomized tests (still disabled) since the implementation + existing tests were incorrect https://github.com/partiql/partiql-lang-kotlin/issues/1697

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1718, opened 2025-01-16
[v1] Fix unary minus with numeric literal parsing; add data exception for unary neg overflow

alancai98:
## Relevant Issues
- None, but similar to recent PRs
  - https://github.com/partiql/partiql-lang-kotlin/pull/1715
  - https://github.com/partiql/partiql-lang-kotlin/pull/1696
  - https://github.com/partiql/partiql-lang-kotlin/pull/1716
- Similar to this old PR https://github.com/partiql/partiql-lang-kotlin/pull/1484

## Description
- Fixes unary minus w/ numeric literal parsing -- now will parse as a signed int literal rather than unary minus + unsigned int literal
  - Previously, it would be impossible to parse `-2147483648` as an INT4
- Add data exception for unary negation overflow

## License Information

### DIFF
--- a/partiql-eval/src/test/kotlin/org/partiql/eval/internal/DataExceptionTest.kt
+++ b/partiql-eval/src/test/kotlin/org/partiql/eval/internal/DataExceptionTest.kt
@@ -226,6 +226,7 @@ class DataExceptionTest {
                 input = "-CAST(${Long.MIN_VALUE} AS BIGINT)"
             ),
             // No explicit casts
+            // Double `-`
             // INT
             FailureTestCase(
                 input = "- ${Integer.MIN_VALUE}" // space needed since `--` turns into a comment
@@ -233,6 +234,15 @@ class DataExceptionTest {
             // BIGINT
             FailureTestCase(
                 input = "- ${Long.MIN_VALUE}" // space needed since `--` turns into a comment
+            ),
+            // Triple `-`
+            // INT
+            FailureTestCase(
+                input = "- - ${Integer.MIN_VALUE}"
+            ),
+            // BIGINT
+            FailureTestCase(
+                input = "- - ${Long.MIN_VALUE}"
             )
         )
     }
--- a/partiql-eval/src/test/kotlin/org/partiql/eval/internal/DataExceptionTest.kt
+++ b/partiql-eval/src/test/kotlin/org/partiql/eval/internal/DataExceptionTest.kt
@@ -224,6 +224,15 @@ class DataExceptionTest {
             // BIGINT
             FailureTestCase(
                 input = "-CAST(${Long.MIN_VALUE} AS BIGINT)"
+            ),
+            // No explicit casts
+            // INT
+            FailureTestCase(
+                input = "- ${Integer.MIN_VALUE}" // space needed since `--` turns into a comment
+            ),
+            // BIGINT
+            FailureTestCase(
+                input = "- ${Long.MIN_VALUE}" // space needed since `--` turns into a comment
             )
         )
     }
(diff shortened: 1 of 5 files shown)


partiql-eval/src/test/kotlin/org/partiql/eval/internal/DataExceptionTest.kt
```diff
@@ -100,6 +104,14 @@ class DataExceptionTest {
             ),
             FailureTestCase(
                 input = "CAST(${Long.MIN_VALUE} AS BIGINT) - CAST(1 AS BIGINT);"
+            ),
+            // Make sure we parse Integer.MIN_VALUE as an INT rather than BIGINT
```
johnedquinn, line 108: Can this work also have a test for the following: `--${Integer.MIN_VALUE}`. This should fail.

johnedquinn (changes requested):

partiql-eval/src/test/kotlin/org/partiql/eval/internal/DataExceptionTest.kt
alancai98: > Can this work also have a test for the following: `--${Integer.MIN_VALUE}`. This should fail.

I assume you mean something like `- -2147483648`? `--${Integer.MIN_VALUE}` would be `---2147483648` which is parsed as a comment. Added.

alancai98: > From my reading of the PR, this previously would fail and this might make it "pass" (though that would be wrong. _If_ my assumption is correct, a different approach would be to update our ANTLR file instead.

The query `- -2147483648` would actually pass previously and return back `2147483648`. `- -2147483648` would get parsed as a two nested unary minuses with an unsigned INT_NUM literal `2147483648` (i.e. same thing as `-(-(2147483648))`). This would get converted into the plan as a two nested unary neg with a **big int** value of `2147483648`, which evaluates to `2147483648`.

With this PR's change, `- -2147483648` would now give the data exception. `- -2147483648` now gets parsed as a single unary minus with a signed INT_NUM literal `-2147483648` (i.e. same thing as `-(-2147483648)`. Converting that into the plan would give a unary neg with a **int** value of `-2147483648`, which will give an error when evaluated.

---

For the other comment about the ANTLR rule, I played around with this a while back with the literal modeling and couldn't seem get all the cases to parse correctly. For now, modifying it in the ANTLR visitor seems good enough but I added a TODO to look into it more in the future.

johnedquinn: Apologies. `- - - 2147483648`

alancai98: > - - - 2147483648

Error behavior is the same as `- - 2147483648`
- both would fail in this PR
- previously, they would succeed. `- - - 2147483648` => `- 2147483648` and `- - 2147483648` => `2147483648`

I can add another test for regressions though

johnedquinn: `- - - 2147483648` = `-(-(-2147483648))`. Which, should fail. With the original commit, I believe this wouldn't fail. With your latest commit, I think it will fail 👍 . Though, a test would give me confidence.

alancai98: Yeah good catch, the original commit had a bug (should only add the `-` sign to a literal numeric string if and only if there was no prior `-` added).

Added the triple `-` test in [`7df4329` (#1718)](https://github.com/partiql/partiql-lang-kotlin/pull/1718/commits/7df432906424b5bf7189e3df69c693a2e778042a).

johnedquinn (approved):

PR #1719, opened 2025-01-17
Creates PRuntimeException and audits eval package

johnedquinn:
## Description
- Audits eval package. Adds Javadocs and exceptions.
- Creates PRuntimeException in favor of PErrorListenerException and PErrorException. Adds some documentation there.

## License Information


partiql-eval/src/main/kotlin/org/partiql/eval/internal/compiler/StandardCompiler.kt
```diff
@@ -158,7 +159,15 @@ internal class StandardCompiler(strategies: List<Strategy>) : PartiQLCompiler {
...
             // execute with no parameters
-            override fun execute(): Datum = root.eval(Environment())
+            override fun execute(): Datum {
+                return try {
+                    root.eval(Environment())
+                } catch (e: PRuntimeException) {
+                    throw e
+                } catch (t: Throwable) {
+                    throw PErrors.internalErrorException(t)
+                }
```
alancai98, line 169: I haven't checked this locally but shouldn't the non-`PRuntimeException`s be caught? I guess not a big deal to add an additional catch just in case.

partiql-eval/src/main/java/org/partiql/eval/Row.java
```diff
@@ -10,14 +10,18 @@
...
 
+    private final Datum[] values;
+
     /**
-     * TODO internalize values.
+     * Get the values in this record.
+     * @return the values
      */
-    public final Datum[] values;
+    public final Datum[] getValues() {
```
alancai98, line 19: thanks for adding back the stuff reverted from prior package audit. think there's one last minor thing that was reverted for Match.java

[`ff15aa2` (#1711)](https://github.com/partiql/partiql-lang-kotlin/pull/1711/commits/ff15aa2a53c5e40ac4790c0209ddd8093adc8799#r1917175319)

partiql-eval/src/main/kotlin/org/partiql/eval/internal/compiler/StandardCompiler.kt
alancai98: To rephrase, I was initially thinking that all the `Throwable` exceptions were already wrapped in a `PRuntimeException`, but seems like we need this additional catch in case there's some exception that we haven't already wrapped in a `PRuntimeException`. For example (this specific case has been fixed by https://github.com/partiql/partiql-lang-kotlin/pull/1720), `1.0 / 0.0` would give a "java.lang.ArithmeticException: / by zero" which would be wrapped in a `PRuntimeException` with code `PError.INTERNAL_ERROR` rather than a `PRuntimeException` with code `PError.DIVISION_BY_ZERO`. So the following two cases would give different error codes:
```
1 / 0 -- division by zero error
1.0 / 0.0 -- internal error
```

I think this catch(throwable) is fine for now, but we should make sure exceptions we throw try to use the existing `PRuntimeException` codes.

alancai98 (approved):

PR #1722, opened 2025-01-21
Adds Fn, FnProvider, Agg, and AggProvider APIs

johnedquinn:
## Description
- Cleans up the Function and Aggregation APIs by creating distinct APIs for providing functions as well as implementing functions.
- Adds Javadocs to each class/method. Performed the audit.
- There is one more failing conformance test (MYSQL_SELECT_29), however, this doesn't pass in the old evaluator. See the report. It passes locally, but it times out on the remote runner.

## Reviewing
- The functionality is the same. The APIs are now cleaned up to remove unnecessary methods, and the naming is more indicative of the underlying class's definition.

## License Information


partiql-spi/src/main/java/org/partiql/spi/function/AggProvider.java
```diff
@@ -0,0 +1,184 @@
...
+            this.name = name;
+        }
+
+        /**
+         * Adds a {@link Parameter} to the {@link AggProvider} with the given {@link PType}. The name of the {@link
+         * Parameter} is automatically generated.
+         * @param type The type of the parameter.
+         * @return the {@link Builder} instance.
+         */
+        @SuppressWarnings("UnusedReturnValue")
```
alancai98, line 72: unused suppression

partiql-spi/src/main/java/org/partiql/spi/function/FnProvider.java
```diff
@@ -0,0 +1,214 @@
...
+
+/**
+ * <p>
+ * This provides potentially overloaded implementations of a particular scalar function (e.g. {@code ABS(int)},
+ * {@code ABS(decimal(p, s))}, etc.).
+ * </p>
+ * @see Builder
+ * @see Fn
+ */
+public abstract class FnProvider {
```
alancai98, line 22: naming-wise, I think `...Provider` is somewhat non-descriptive. Perhaps a more apt name is `...Overload` since that's how the javadoc describes the `FnProvider` and `AggProvider`?

partiql-spi/src/main/kotlin/org/partiql/spi/function/Aggregation.kt
```diff
@@ -1,79 +1,30 @@
...
-        override fun getParameters(): Array<Parameter> = parameters
-        override fun getReturnType(args: Array<PType>): PType = returns
-        override fun getAccumulator(args: Array<PType>): Accumulator = accumulator()
-        override fun toString(): String {
-            val parameters = parameters.joinToString("__") { it.getType().name() }
-            val returnType = returns.name()
-            return "FN_${name}___${parameters}___$returnType"
-        }
+    @JvmStatic
+    fun static(
```
alancai98, line 18: could rename to `overload`? I don't know what `static` is referring to here

partiql-spi/src/main/kotlin/org/partiql/spi/function/Function.kt
```diff
@@ -4,128 +4,59 @@ import org.partiql.spi.types.PType
...
-        override fun getInstance(args: Array<PType>): Instance = instance
-        override fun toString(): String {
-            // TODO debug strings for SqlTypeFamily
-            // val parameters = parameters.joinToString("__") { it.getType().PType.name }
-            val returnType = returns.name()
-            return "FN_${name}___${parameters}___$returnType"
-        }
-    }
+    @JvmStatic
+    fun static(
```
alancai98, line 48: nit since internal: could rename to `overload`? Not sure what `static` means here

partiql-spi/src/main/kotlin/org/partiql/spi/function/Builtins.kt
```diff
@@ -205,12 +205,10 @@ internal object Builtins {
...
-
-        Fn_SUBSTRING__STRING_INT32__STRING,
-        Fn_SUBSTRING__STRING_INT32_INT32__STRING,
-        Fn_SUBSTRING__CLOB_INT64__CLOB,
-        Fn_SUBSTRING__CLOB_INT64_INT64__CLOB,
-
+        Fn_SUBSTRING_2STR,
+        Fn_SUBSTRING_2CLOB,
+        Fn_SUBSTRING_3STR,
+        Fn_SUBSTRING_3CLOB,
```
alancai98, line 211: any reason for these to be renamed? I think keeping things consistently named would be clearer

partiql-spi/src/main/java/org/partiql/spi/function/RoutineProviderSignature.java
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * <p>
+ * This represents the signature of a routine provider. This is distinct from {@link RoutineSignature}, as it is specific
+ * to the provider, and not the instance. The provider signature is used to determine if a routine provider is applicable
+ * to a given call site, and if so, which routine provider to use.
+ * </p>
+ * <p>
+ * While this is currently the same as a {@link RoutineSignature}, additional methods <i>may</i> be added to this class
+ * for future use, such as for retrieving all representative signatures of an implementation (for debugging purposes
+ * or for error-messaging).
```
alancai98, line 16: from offline, these two classes, `RoutineProviderSignature` and `RoutineSignature`, differ. `RoutineProviderSignature` does not have a property for the return type, isNullCall, and isMissingCall. It only really uses the types of the input parameters.

partiql-spi/src/main/java/org/partiql/spi/function/Fn.java
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        /**
+         * Adds a {@link Parameter} to the {@link Fn} with the given {@link PType}. The name of the {@link Parameter} is
+         * automatically generated.
+         * @param type the type of the parameter.
+         * @return the {@link Builder} instance.
+         */
+        @NotNull
+        @SuppressWarnings("UnusedReturnValue")
```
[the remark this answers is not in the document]
johnedquinn: According to my IDE, it is used.

partiql-spi/src/main/java/org/partiql/spi/function/Parameter.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    /**
+     * Returns the name of the parameter.
+     * @return the name of the parameter.
+     */
+    @NotNull
+    public String getName() {
+        return name;
+    }
```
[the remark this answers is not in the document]
johnedquinn: Discussed offline. The name is part of the signature (see SQL:1999 Section 21.25 "PARAMETERS base table").

```
CREATE FUNCTION FOO(bar INT) RETURNS INT ...;
```

partiql-spi/src/main/kotlin/org/partiql/spi/function/Aggregation.kt
johnedquinn: This is an internal name that was already being used. Didn't change it as it is used many times in our codebase and would create an unnecessary diff for an internal API.

partiql-spi/src/main/kotlin/org/partiql/spi/function/Function.kt
johnedquinn: See https://github.com/partiql/partiql-lang-kotlin/pull/1722#discussion_r1925591763

partiql-spi/src/main/java/org/partiql/spi/function/FnProvider.java
```diff
@@ -0,0 +1,214 @@
...
+    @NotNull
+    public abstract RoutineProviderSignature getSignature();
+
+    /**
+     * Returns an instance of the function for the given argument types.
+     * @param args the argument types.
+     * @return an instance of the function for the given argument types, or {@code null} if no such function exists.
+     */
+    @Nullable
+    public abstract Fn getInstance(PType[] args);
```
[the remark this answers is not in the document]
johnedquinn: For any instances of the internal planner logic using it, it's internal and can be updated. As for execution, I've updated internal logic to use arrays in the latest commit. See https://github.com/partiql/partiql-lang-kotlin/pull/1722/commits/0ab406bbcc049b2c75917e263f6203290468e7ee.

partiql-spi/src/main/kotlin/org/partiql/spi/function/Aggregation.kt
alancai98: Similarly for `Function.kt`'s `static`

partiql-spi/src/main/java/org/partiql/spi/function/Fn.java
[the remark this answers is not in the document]
alancai98: that's strange. my IDE's not giving any warning. not a big deal then

alancai98 (commented):
latest changes lgtm. seems like there's a slight conformance regression with this PR -- https://github.com/partiql/partiql-lang-kotlin/actions/runs/12917164270. could we verify we're not introducing any behavior changes?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/Env.kt
```diff
@@ -312,9 +312,11 @@ internal class Env(private val session: Session) {
...
             val a = args[i]
             val p = parameters[i]
-            when (p.type.code() == a.code()) {
-                false -> mapping[i] = coercion(a, p.type) ?: return null
-                true -> continue
+            when {
+                p.type.code() == a.code() -> continue
+                a.code() == PType.DYNAMIC -> mapping[i] = Ref.Cast(a.toCType(), p.type.toCType(), Ref.Cast.Safety.COERCION, true)
+                p.type.code() == PType.DYNAMIC -> continue
+                else -> mapping[i] = coercion(a, p.type) ?: return null
```
alancai98, line 319: could we add a small comment/description of what these additional branches are for?

alancai98 (approved):

alancai98 (approved):

PR #1729, opened 2025-01-24
Adds `Datum#row()`

johnedquinn:
## Description

Adds the ability to create rows.

Note: we already implicitly did this under the hood, but now we're making it apparent and more user-friendly. When writing the usage guide on how to create implementations of `Table`, I realized we couldn't outright create a ROW, even though we specified the type to be a ROW. So, adding this so that I can call the right API in the usage guide.

## License Information

### DIFF
--- a/partiql-spi/src/main/java/org/partiql/spi/value/DatumRow.java
+++ b/partiql-spi/src/main/java/org/partiql/spi/value/DatumRow.java
@@ -98,7 +98,7 @@ class DatumRow implements Datum {
     @Override
     public String toString() {
         StringBuilder sb = new StringBuilder();
-        sb.append("struct::{ ");
+        sb.append("row::{ ");
         for (Map.Entry<String, List<Datum>> entry : _delegate.entrySet()) {
             sb.append(entry.getKey());
             sb.append(": ");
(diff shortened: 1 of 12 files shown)


partiql-spi/src/main/java/org/partiql/spi/value/DatumRow.java
```diff
@@ -0,0 +1,111 @@
...
+    @NotNull
+    @Override
+    public PType getType() {
+        return _type;
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("struct::{ ");
```
alancai98, line 101: Should be row?

```suggestion
        sb.append("row::{ ");
```

partiql-spi/src/main/java/org/partiql/spi/value/DatumRow.java
```diff
@@ -0,0 +1,111 @@
...
+        List<org.partiql.spi.types.Field> fieldTypes = new ArrayList<>();
+        fields.forEach((f) -> {
+            PType fType = f.getValue().getType();
+            org.partiql.spi.types.Field typeField = org.partiql.spi.types.Field.of(f.getName(), fType);
+            fieldTypes.add(typeField);
+        });
+        return PType.row(fieldTypes);
+    }
+
+    private void addFieldToStruct(Map<String, List<Datum>> struct, String key, Datum value) {
```
alancai98, line 52: nit: functions and vars could be renamed to use row rather than struct

alancai98 (approved):
couple minor comments. otherwise looks good

alancai98 (approved):

PR #1736, opened 2025-02-05
Adds parsing and modeling of CTEs

johnedquinn:
## Description
- Adds parsing and ast modeling of the WITH clause.
  - Updates the visitor and AST rewriter
- This PR also (largely) fixes the grammar related to SELECT as it pertains to the expression tree.
  - Before, `exprBagOp` and `exprSelect` were at the top of the `expr` grammar. This isn't right.
  - Now, `expr` and `selectStatement` are distinct variants of `dql`. 
    - `selectStatement` is now _almost_ identical to the SQL:1999 and SQL:2023 EBNF. I've noted all of the rules and their origin (and where they deviate). Notably, a `simpleTable`, in PartiQL, can be an arbitrary expression -- which is not allowed in SQL. This allows for the UNION (and other bag ops) of arbitrary expressions. Note that the visitor prohibits non-bag-ops and non-select expressions coupled with the with clause, order by clause, limit clause, and offset clause.
    - There needed to be some slight modifications, such as the introduction of `subquery` (which is a parenthesized query expression) and the allowance for the direct use of SELECT in a function call (to handle the specification's `COLL_AGG(SELECT ...)` without the use of additional parentheses).
- This PR sends an error to the PErrorListener if planning a ExprQuerySet containing a with clause. A future PR should remove this error when implementing the planning/evaluation.

## API Changes and Release

There are NO API breaking changes. This feature addition will go in the `v1.1.0` release.

This PR adds `With` and `WithListElement`. It also adds JoinTypes that allow for LEFT CROSS JOIN, RIGHT CROSS JOIN, FULL CROSS JOIN, and INNER CROSS JOIN. The LEFT CROSS JOIN was already present in the previous implementation. There are parsing conformance tests that also use the remaining.

## Upcoming Changes

Before the release of `v1.1.0`, a follow-up PR will add for planning and evaluation of the WITH clause.

## License Information


jpschorr (commented):
Is it worth modeling `RECURSIVE` now? If nothing else, we could emit an error for now, but we wouldn't need to touch the parser/AST again.

partiql-ast/src/main/java/org/partiql/ast/With.java
```diff
@@ -0,0 +1,53 @@
...
+ * Represents a PartiQL WITH clause.
+ * </p>
+ * <p>{@code <with clause>    ::=   WITH [ RECURSIVE ] <with list>}</p>
+ * <p>{@code <with list>    ::=   <with list element> [ { <comma> <with list element> }... ]}</p>
+ * @see WithListElement
+ * @see org.partiql.ast.expr.ExprQuerySet
+ */
+@Builder(builderClassName = "Builder")
+@EqualsAndHashCode(callSuper = false)
+public final class With extends AstNode {
```
jpschorr, line 21: Should there be a helper constructor for `With` in `org.partiql.ast.Ast`?

johnedquinn:
Updated PR to address comments @jpschorr

jpschorr (approved):

PR #1744, opened 2025-02-26
Adds parsing support for window functions

johnedquinn:
## DRAFT
Marking this as a draft while I push forward to implementing planning and evaluation. Once I have greater confidence in the modeling (and I add documentation and the factory methods), then I'll mark for ready for review (along with the planning/evaluation PRs).

## Relevant Issues
- None

## Description
- Adds parsing support for window functions (RANK, DENSE_RANK, PERCENT_RANK, CUME_DIST, ROW_NUMBER, LAG, and LEAD). We are implementing SQL:2023 verbatim.
- Adds AST modeling for window functions. We are implementing SQL:2023 verbatim. Note that this PR intentionally makes some classes extendable to allow for future extension. AKA, `WindowPartitionClause` is its own distinct clause due to potential differences from DDL's `PARTITION BY`. Right now, the elements are abstract to allow for potential partitioning on arbitrary expressions, however, for now, they are literal integers. We can extend to include arbitrary expressions in the future.
- Deprecates old window function AST APIs due to incompatibility. The old representation was wrong. We can just deprecate, no big deal. When planning, we can still handle both.

## License Information


johnedquinn:
Closing in favor of the end-to-end PR: #1746

PR #1746, opened 2025-03-10
[DRAFT] Adds RANK, DENSE_RANK, LAG, LEAD, & ROW_NUMBER Window Functions & Operator

johnedquinn:
## DRAFT
- TODO

## Relevant Issues
- TODO

## Description
- TODO

## Other Information
- Updated Unreleased Section in CHANGELOG: **NO**: Not yet
- Any backward-incompatible changes? **NO**: Not technically.
- Do your changes comply with the [Contributing Guidelines][contributing-guidelines] and [Code Style Guidelines][code-style]? **YES**

## License Information

[contributing-guidelines]: https://github.com/partiql/partiql-lang-kotlin/blob/main/CONTRIBUTING.md
[code-style]: https://github.com/partiql/partiql-lang-kotlin/blob/main/CODE_STYLE.md


partiql-parser/src/main/antlr/PartiQLParser.g4
```diff
@@ -982,18 +986,99 @@ overlay
...
+ * <window specification> ::= <left paren> <window specification details> <right paren>
+ * <window specification details> ::= [ <existing window name> ] [ <window partition clause> ] [ <window order clause> ] [ <window frame clause> ]
+ * <existing window name> ::= <window name>
+ * <window frame clause> ::= ...
+ */
+windowSpecification
+    : PAREN_LEFT
+        existingWindowName=symbolPrimitive?
+        partition=windowPartitionClause?
+        order=orderByClause?
```
yliuuuu, line 1063: This is interesting: you can further partition/order an existing window...

partiql-parser/src/main/antlr/PartiQLParser.g4
```diff
@@ -201,6 +201,26 @@ querySpecification
...
+    ;
+
+/**
+ * EBNF 2023:
+ * <window clause> ::= WINDOW <window definition list>
+ * <window definition list> ::= <window definition> [ { <comma> <window definition> }... ]
+ */
+windowClause
+    : WINDOW windowDefinition (COMMA windowDefinition)*
+    ;
```
yliuuuu, line 214: Just a personal notes to remind myself when review further section: 

It seems like we can parse statement like 
```
SELECT w1 FROM employee AS t WINDOW w1 AS (PARTITION BY t.b ORDER BY t.c);
```

Curious if we support this.

partiql-ast/src/main/java/org/partiql/ast/QueryBody.java
```diff
@@ -49,61 +86,100 @@ public SFW(@NotNull Select select, @Nullable Exclude exclude, @NotNull From from
             this.where = where;
             this.groupBy = groupBy;
             this.having = having;
+            this.windowClause = null;
```
yliuuuu, line 89: Backward compatible. Very nice : )

partiql-ast/src/main/java/org/partiql/ast/WindowFunctionType.java
```diff
@@ -0,0 +1,160 @@
...
+ */
+public abstract class WindowFunctionType extends AstNode {
+    /**
+     * A window function that takes no arguments.
+     * @see ExprWindowFunction#getFunctionType()
+     * @see WindowFunctionSimpleName
+     */
+    @Builder(builderClassName = "Builder")
+    @EqualsAndHashCode(callSuper = false)
+    public static final class NoArg extends WindowFunctionType {
```
yliuuuu, line 25: Hmmm, mind evaluating why the choice of `NoArg` vs say `Ranking` and `RowNumber`. 

So seemingly for one we are modeling using number of args, for the other we are modeling using "function characteristic" for lack of better words. 

I understand that both `Ranking` and `RowNumber` must not contains argument... And no arg is, well, no arg... and you don't have to worry about argument type....

But perhaps using `NoArg` is bit over-smart, in that it might cause difficulty when (especially a user is) constructing the ast.

partiql-ast/src/main/java/org/partiql/ast/WindowReference.java
```diff
@@ -0,0 +1,96 @@
...
+import org.partiql.ast.expr.ExprWindowFunction;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Represents a reference to a window.
+ * @see ExprWindowFunction#getWindowReference()
+ */
+public abstract class WindowReference extends AstNode {
```
yliuuuu, line 15: For the ast node? can we inline the two? 

i.e., just use `WindowSpecification`
```
    @Nullable
    private final Identifier.Simple existingName;

    @Nullable
    private final WindowPartitionClause partitionClause;

    @Nullable
    private final WindowOrderClause orderClause;
```

yliuuuu (commented):
Finished Parser and Ast Package. As always, very delightful code : ). 

One preliminary comment: 

Perhaps it is worth to track the feature gap between components if anything exists. 

i.e., if something is supported in parsing but not supported in planning, etc.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -1962,24 +1987,74 @@ internal class PartiQLParserDefault : PartiQLParser {
...
-            exprWindowOver(partitions, sorts)
+        override fun visitLeadOrLagFunction(ctx: GeneratedParser.LeadOrLagFunctionContext) = translate(ctx) {
+            val isLead = when (ctx.name.type) {
+                GeneratedLexer.LEAD -> true
+                GeneratedLexer.LAG -> false
+                else -> throw error(ctx, "Expected LEAD or LAG")
+            }
+            val extent = visitExpr(ctx.extent)
+            val offset = ctx.offset.text.toLong()
+            val default = ctx.default_.let { visitExpr(it) }
```
yliuuuu, line 2028:
```suggestion
            val offset = ctx.offset?.text?.toLong()
            val default = ctx.default_?.let { visitExpr(it) }
```

partiql-ast/src/main/java/org/partiql/ast/WindowReference.java
johnedquinn: I'm not exactly sure what you mean.

If you're asking why can't we just move all those attributes inside a non-abstract WindowReference, well they are syntactically different. So, we couldn't accurately unparse window references. Though, I do get what you mean, because the existingName, if not null, could just be the reference.

If it will never matter to unparse, then we can collapse. LMK your thoughts.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
johnedquinn: Nice catch 👍

CHANGELOG.md
```diff
@@ -23,6 +23,47 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Contributors
+Thank you to all who have contributed!
+
+## [1.2.0](https://github.com/partiql/partiql-lang-kotlin/releases/tag/v1.2.0) - 2025-03-12
+
+### Added
+- Added parsing, planning, and execution of the interval type and interval values.
+- Added multiple operations involving datetime and interval values.
+- Added planning and evaluation of the LET clause
```
alancai98, line 48: (side-comment) this PR probably needs to be rebased off of `main`. currently, this PR is showing changes that I assume were not yet in `main` (e.g. interval, let clause) at the time the PR was created, leading to a larger than expected diff

partiql-ast/src/main/java/org/partiql/ast/WindowSpecification.java
```diff
@@ -0,0 +1,83 @@
...
+@Builder(builderClassName = "Builder")
+@EqualsAndHashCode(callSuper = false)
+public final class WindowSpecification extends AstNode {
+    // TODO: In the future: @Nullable private final WindowFrameClause frameClause;
+
+    @Nullable
+    private final Identifier.Simple existingName;
+
+    @Nullable
+    private final WindowPartitionClause partitionClause;
```
alancai98, line 25: Perhaps this additional `WindowPartitionClause` is unnecessary and we could just have this be a `NonNull` `List<WindowPartition>`? I get that the SQL2023 spec defines the clause as optional under `<window specification details>`, but if you'll notice the `<window partition column reference list>` will **always** contain at least one `<window partition column reference>`. That may be harder to enforce with this AST modeling since you could define a non-null `partitionClause` that has an empty list of `WindowPartition`s.

Also, it can be confusing on the proper way to define a `WindowSpecification` without any `WindowPartition`s with this modeling. Is the correct way
1. a `WindowSpecification` with a null `partitionClause` or
2. a `WindowSpecification` with a non-null `partitionClause` that has an empty `WindowPartition`s list

partiql-ast/src/main/java/org/partiql/ast/WindowFunctionType.java
```diff
@@ -0,0 +1,160 @@
...
+import org.partiql.ast.expr.ExprWindowFunction;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Represents a window function type.
+ * @see ExprWindowFunction#getFunctionType() 
+ */
+public abstract class WindowFunctionType extends AstNode {
```
alancai98, line 17: Similar to the parser comment, I'm wondering if we should change the window functions to not be as hard-coded within the AST modeling and follow something more like scalar and aggregate function calls. IMO this can make it a lot more straightforward to add future support for user-defined window functions.

Trino for example models window functions (along with scalars and aggregates) using [FunctionsCall](https://github.com/trinodb/trino/blob/579d325cbc005ec0702a2e63de19921775676ec9/core/trino-parser/src/main/java/io/trino/sql/tree/FunctionCall.java). For all of the function names, they just use a `QualifiedName` rather than any sort of hard-coded names within their AST, leaving the validation to a later stage. I'm not sure if we should lump all three of those under `ExprCall` but modeling window functions with just an identifier name rather than the `WindowFunctionType` could be a cleaner modeling as more window functions are added over time.

partiql-ast/src/main/java/org/partiql/ast/WindowReference.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * Represents a reference to an in-line window specification.
+     * @see ExprWindowFunction#getWindowReference()
+     * @see WindowSpecification
+     */
+    @Builder(builderClassName = "Builder")
+    @EqualsAndHashCode(callSuper = false)
+    public static final class InLineSpecification extends WindowReference {
```
alancai98, line 63: It's not clear to me why we need a separate in-line window specification class. Perhaps a simpler modeling could follow what Trino does with their [Window](https://github.com/trinodb/trino/blob/579d325cbc005ec0702a2e63de19921775676ec9/core/trino-parser/src/main/java/io/trino/sql/tree/Window.java) interface.

The `Window` interface is implemented by two separate classes
- [WindowReference](https://github.com/trinodb/trino/blob/579d325cbc005ec0702a2e63de19921775676ec9/core/trino-parser/src/main/java/io/trino/sql/tree/WindowReference.java) and
- [WindowSpecification](https://github.com/trinodb/trino/blob/579d325cbc005ec0702a2e63de19921775676ec9/core/trino-parser/src/main/java/io/trino/sql/tree/WindowSpecification.java)

Following a similar modeling, we could have a `Window` abstract class that is used by the `ExprWindowFunction`.

partiql-ast/src/main/java/org/partiql/ast/WindowFunctionType.java
```diff
@@ -0,0 +1,160 @@
...
+     * Represents either the LAG or LEAD window function.
+     * @see ExprWindowFunction#getFunctionType()
+     */
+    @Builder(builderClassName = "Builder")
+    @EqualsAndHashCode(callSuper = false)
+    public static final class LeadOrLag extends WindowFunctionType {
+        private final boolean isLead;
+        private final Expr extent;
+        private final Long offset;
+        private final Expr defaultValue;
```
alancai98, line 69: If window functions were modeled more closely to `ExprCall`, these could all be function arguments rather than hard-coded properties.

partiql-ast/src/main/java/org/partiql/ast/WindowReference.java
yliuuuu: Let's see: 

Currently: 
```
rowNumber() OVER w1 // AST: window name
rowNumber() OVER (PARTITION BY ... ORDER BY ....) // AST: InLineSpecification
```

If we remove window reference and just use `WindowSpecification`
```
rowNumber() OVER w1 // windowSpec(existing window: w1, partitionBy: null, orderBy: null)
rowNumber() OVER (PARTITION BY ... ORDER BY ....) // AST: // windowSpec(existing window: null, partitionBy: ..., orderBy: ...)
```

Un-parsing: 
```
rowNumber() OVER (w1) // extra parenthesis. 
rowNumber() OVER (PARTITION BY ... ORDER BY ....) // AST: // windowSpec(existing window: null, partitionBy: ..., orderBy: ...)
```

Perhaps an more apparent example would be: 
// Original input
```
rowNumber() OVER w1
rowNumber() OVER (w1)
```

// Un-parsing result: 
```
rowNumber() OVER (w1)
rowNumber() OVER (w1)
```

yliuuuu: I personally lean over towards combining those for simplicity in the AST. 

But I will leave the decision to other members who more actively working on the un-parsing use case...

partiql-eval/src/test/kotlin/org/partiql/eval/internal/WindowTests.kt
```diff
@@ -0,0 +1,368 @@
...
+            Global(
+                name = "employee",
+                value = Datum.bag(employees.map { it.toDatum() })
+            )
+        )
+
+        private const val FALLBACK: String = "UNKNOWN"
+
+        @JvmStatic
+        fun successTestCases() = listOf(
```
yliuuuu, line 80: A couple more test I think would be beneficial here: 

Using window function with group by: [ref](https://github.com/partiql/partiql-lang-kotlin/blob/c26536c149bb58d081b4f41544f1ebfd532f26a0/lang/src/test/kotlin/org/partiql/lang/eval/builtins/windowFunctions/WindowFunctionTests.kt#L111-L127)

Using window function with subquery: [ref](https://github.com/partiql/partiql-lang-kotlin/blob/c26536c149bb58d081b4f41544f1ebfd532f26a0/lang/src/test/kotlin/org/partiql/lang/eval/builtins/windowFunctions/WindowFunctionTests.kt#L149-L157)

Window order does not affect outer order: [ref](https://github.com/partiql/partiql-lang-kotlin/blob/c26536c149bb58d081b4f41544f1ebfd532f26a0/lang/src/test/kotlin/org/partiql/lang/eval/builtins/windowFunctions/WindowFunctionTests.kt#L91-L108)

partiql-eval/src/main/kotlin/org/partiql/eval/internal/window/NavigationFunction.kt
```diff
@@ -0,0 +1,38 @@
...
+     * Evaluate the function for the current row.
+     * @param env the environment to use for evaluation
+     * @return the result of the evaluation
+     */
+    abstract fun eval(env: Environment): Datum
+
+    /**
+     * Reset the function to its initial state.
+     */
+    abstract fun reset()
```
yliuuuu, line 26: Could you please elaborate on why is this necessary?

partiql-parser/src/main/antlr/PartiQLParser.g4
```diff
@@ -982,18 +986,99 @@ overlay
...
+ *   <rank function type> <left paren> <right paren>
+ *   | ROW_NUMBER <left paren> <right paren>
+ *   | <lead or lag function>
+ *   | and more...
+ */
+windowFunctionType
+    : rankFunctionType PAREN_LEFT PAREN_RIGHT # WindowFunctionTypeRank
+    | ROW_NUMBER PAREN_LEFT PAREN_RIGHT       # WindowFunctionTypeRowNumber
+    | leadOrLagFunction                       # WindowFunctionTypeLeadOrLag
+    ;
```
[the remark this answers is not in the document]
johnedquinn: I'd say let's do that when we get there. The EBNF does not contribute to API breaking changes. WindowFunctionType also is not a sealed interface. If we want to make it more explicit, I can make NoArg specific to RowNumber and Rank. Then, if we ever decided to allow user-defined window functions, we can do so by adding a variant specifically for this.

partiql-ast/src/main/java/org/partiql/ast/WindowSpecification.java
```diff
@@ -0,0 +1,83 @@
...
+ * @see WindowClause.WindowDefinition#getSpecification()
+ * @see WindowReference.InLineSpecification
+ */
+@Builder(builderClassName = "Builder")
+@EqualsAndHashCode(callSuper = false)
+public final class WindowSpecification extends AstNode {
+    // TODO: In the future: @Nullable private final WindowFrameClause frameClause;
+
+    @Nullable
+    private final Identifier.Simple existingName;
```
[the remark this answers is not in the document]
johnedquinn: Personally, I like the fact that it matches the EBNF and that it makes it apparent that the WindowSpecification operates against a pre-defined (aka existing) window, effectively creating a new window.

partiql-ast/src/main/java/org/partiql/ast/WindowOrderClause.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Represents a window specification's order by clause.
+ * @see WindowSpecification#getOrderClause() 
+ */
+@Builder(builderClassName = "Builder")
+@EqualsAndHashCode(callSuper = false)
+public final class WindowOrderClause extends AstNode {
+    @NotNull
+    private final List<Sort> sorts;
```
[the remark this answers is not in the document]
johnedquinn: Yes. Can do. Only did this to match SQL's EBNF

partiql-ast/src/main/java/org/partiql/ast/WindowFunctionType.java
johnedquinn: I'd be more keen to add a user-defined variant in the future. See the [2003 EBNF](https://ronsavage.github.io/SQL/sql-2003-2.bnf.html#window%20function%20type). [Trino](https://trino.io/docs/current/functions/window.html#functions-window--page-root) doesn't have support for over half of the specified variants of the window functions.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -501,6 +517,177 @@ internal object RelConverter {
...
+                return Pair(select, input)
+            }
+
+            // Grab window definitions and add to internal data structure
+            val windows = InternalWindowDefinitions(env)
+            select.window?.definitions?.map {
+                it.specification.existingName?.let {
+                    env.listener.report(PErrors.featureNotSupported("Window referencing other window"))
+                }
+                windows.addWindowDef(it.name.text, it.specification)
```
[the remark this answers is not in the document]
johnedquinn: Addressing in next rev.

partiql-ast/src/main/java/org/partiql/ast/WindowReference.java
johnedquinn: We could put WindowSpecification to be a variant of WindowReference (and make it a nested class, or leave it out separately, doesn't matter to me). Alternatively, extract a WindowReference interface. Did this locally, and while it's nice, I'm really despising how AstNode is an abstract class. Should've just created a public AstNodeBase. I'll publish a rev, and it should be apparent why it's a burden.

johnedquinn: I'm actually going to take @yliuuuu's recommendation of removing WindowReference entirely. It removes all the shenanigans, and we get what you're describing also 😄

partiql-ast/src/main/java/org/partiql/ast/WindowReference.java
johnedquinn: After playing around with some modeling, I agree.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/window/NavigationFunction.kt
johnedquinn: It actually isn't necessary. There's another `reset` method being leveraged. This must've been a remnant of the past when I was experimenting with it. 😄

johnedquinn: It actually isn't necessary. There's another `reset` method being leveraged. This must've been a remnant of the past when I was experimenting with it. 😄 Anyways, it's internal. Removing.

partiql-ast/src/main/java/org/partiql/ast/WindowFunctionType.java
johnedquinn: Decided to go ahead with your observation. Trying to thread the needle between your comments and @alancai98's.

partiql-eval/src/test/kotlin/org/partiql/eval/internal/WindowTests.kt
johnedquinn: Added those tests in latest rev. I've cross-verified the test cases with PostgreSQL, and seemingly all 3 original expected results are wrong. Here's a DB fiddle for you to check: https://www.db-fiddle.com/f/xvd8JwWTZwSj5BkeaPCSot/7

partiql-ast/src/main/java/org/partiql/ast/WindowFunctionType.java
```diff
@@ -0,0 +1,221 @@
...
+        }
+    }
+
+    /**
+     * Represents either the LAG or LEAD window function.
+     * @see ExprWindowFunction#getFunctionType()
+     */
+    @Builder(builderClassName = "Builder")
+    @EqualsAndHashCode(callSuper = false)
+    public static final class LeadOrLag extends WindowFunctionType {
```
yliuuuu: No objection here, but might argue that by breaking it down into two classes, the API is more consistent.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelOpWindow.kt
```diff
@@ -0,0 +1,165 @@
...
+        functions.map { it.reset(_partition) }
+    }
+
+    override fun peek(): Row? {
+        // Check if there is an existing partition. If so, evaluate and return.
+        _partitionPeekingNumber++
+        if (_partition.size() > _partitionPeekingNumber) {
+            return produceResult()
+        }
+        _partitionPeekingNumber = 0L
```
yliuuuu, line 59: Notes: When we reached here, it means we either at the very first peek call, or we drained the previous partition.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelOpWindow.kt
```diff
@@ -0,0 +1,165 @@
...
+            }
+        }
+        var previousInfoIndex = newLocalPartition.add(OrderingInfo(partitionCreationIndex))
+        newLocalPartition.add(firstRow, previousInfoIndex)
+        val newEnv = _env.push(firstRow)
+        val firstRowPartitionKeys = Array(partitionBy.size) { partitionBy[it].eval(newEnv) }
+        var previousRowSortKeys = Array(sortBy.size) { sortBy[it].expr.eval(newEnv) }
+
+        // Add partition's remaining rows
+        while (input.hasNext()) {
```
yliuuuu, line 82: Notes: Finish up the entire partition creation and cache it.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelOpWindow.kt
```diff
@@ -0,0 +1,165 @@
...
+            val nextRow = input.next()
+            val nextEnv = _env.push(nextRow)
+
+            // Stop (and save spillover row) if at next partition
+            val nextPartitionKeys = Array(partitionBy.size) { partitionBy[it].eval(nextEnv) }
+            val isNewPartition = comparator.compare(firstRowPartitionKeys, nextPartitionKeys) != 0
+            if (isNewPartition) {
+                leftoverRow = nextRow
+                break
+            }
```
yliuuuu, line 93: Notes: we reached the end of partition. Since we have already evaluated the first row of next partition, cache it.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/window/WindowFunctionSignatureProvider.kt
```diff
@@ -0,0 +1,56 @@
...
+     *  to return a nullable [org.partiql.plan.WindowFunctionNode], not a [WindowFunctionSignature].
+     * @return null if unable to create a signature.
+     */
+    private fun lagOrLead(name: String, expr: Rex, offset: Rex, default: Rex, ignoreNulls: Boolean): WindowFunctionSignature? {
+        if (offset.type.code() != PType.BIGINT) {
+            return null
+        }
+        val returnTyper = DynamicTyper()
+        returnTyper.accumulate(expr)
+        returnTyper.accumulate(default)
```
yliuuuu, line 53: Do we want to accumulate or throw on mismatch?

johnedquinn: I might be missing something. The lag/lead's output type will be the expr or default if the offset row doesn't exist, so they don't need to match. In the case of coercible types, maybe we coerce them to their common super type, hence the TODO. But, I don't see any problem with the output type being ANY if the types aren't coercible.

The only validation I'd say is required is that the offset by coercible to BIGINT (or for now just BIGINT).

partiql-ast/src/main/java/org/partiql/ast/WindowFunctionType.java
johnedquinn: Yeah, I mean I'm fine with that too 😄  Whatever floats your boat

test/partiql-tests
alancai98, line 1: partiql-tests submodule change should be reverted

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ Thank you to all who have contributed!
 
 ### Added
 - partiql-ast: add `With` and `WithListElement` to `AstVisitor` and `SqlDialect`
+- Adds end-to-end support for window functions including RANK, DENSE_RANK, LAG, LEAD, and ROW_NUMBER functions alongside the WINDOW clause.
```
alancai98: We should call out that the new window clause and window function support is experimental.

partiql-ast/src/main/java/org/partiql/ast/WindowClause.java
```diff
@@ -0,0 +1,107 @@
...
+        return visitor.visitWindowClause(this, ctx);
+    }
+    
+    /**
+     * Represents a window definition in a PartiQL WINDOW clause.
+     * @see WindowClause
+     */
+    @lombok.Builder(builderClassName = "Builder")
+    @EqualsAndHashCode(callSuper = false)
+    public static final class WindowDefinition extends AstNode {
```
alancai98: nit: can unnest from the `WindowClause` or just call this `Definition`. Currently the type is `WindowClause.WindowDefinition`, which is pretty lengthy

alancai98: also need a deprecated/experimental notice

partiql-ast/src/main/java/org/partiql/ast/WindowPartition.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+
+/**
+ * Represents a partition specified by a window specification.
+ * @see Name
+ * @see WindowSpecification#getPartitionClause()
+ * @deprecated This feature is experimental and is subject to change.
+ */
+@Deprecated
+public abstract class WindowPartition extends AstNode {
```
alancai98: Why is the `WindowPartition` `abstract`? If I'm matching this up properly to the SQL2023 spec, this lines up with `<window partition column reference>`, which will have a `<column reference>` and an optional `<collate clause>`. Perhaps a simpler modeling is just as a final class with a `columnReference` field.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ Thank you to all who have contributed!
...
 - **EXPERIMENTAL** partiql-plan: add representation of `RelWith` and `WithListElement` to the plan and the 
 `OperatorVisitor`
 - **EXPERIMENTAL** partiql-planner: add planner builder function to control whether `With` table references are
```
alancai98, line 54: Should probably also call out that the prior modeling of window is deprecated and replaced with the newer modeling.

partiql-ast/src/main/java/org/partiql/ast/QueryBody.java
```diff
@@ -40,6 +40,43 @@ public static class SFW extends QueryBody {
         @Nullable
         private final Expr having;
 
+        @Nullable
+        private final WindowClause windowClause;
+
+        /**
```
alancai98, line 46: Constructor should be experimental since the `window` is an experimental feature

partiql-eval/src/main/java/org/partiql/eval/WindowFunction.java
```diff
@@ -0,0 +1,27 @@
+package org.partiql.eval;
+
+import org.jetbrains.annotations.NotNull;
+import org.partiql.spi.value.Datum;
+
+/**
+ * Represents an executable window function.
+ * @see org.partiql.plan.WindowFunctionNode
+ */
+public interface WindowFunction {
```
alancai98, line 12: Needs an experimental notice

partiql-eval/src/main/java/org/partiql/eval/WindowPartition.java
```diff
@@ -0,0 +1,24 @@
+package org.partiql.eval;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Represents a partition of rows within a window.
+ */
+public interface WindowPartition {
```
alancai98, line 10: Needs an experimental notice

partiql-plan/src/main/java/org/partiql/plan/OperatorVisitor.java
```diff
@@ -128,6 +129,10 @@ default R visitSort(@NotNull RelSort rel, C ctx) {
         return defaultVisit(rel, ctx);
     }
 
+    default R visitWindow(@NotNull RelWindow rel, C ctx) {
```
alancai98, line 136: Probably need notice

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -502,6 +518,183 @@ internal object RelConverter {
             return Pair(sel, rel)
         }
 
+        /**
```
alancai98, line 531: nit: since `RelConverter` is already getting somewhat large as a file, could we move the logic specific to window functions and clause to a separate window util file?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -502,6 +518,183 @@ internal object RelConverter {
...
+            }
+
+            fun getAll(): List<InternalWindowDefinition> {
+                return definitions.values + inLineWindowDefs
+            }
+
+            private fun assertKeyDoesNotExist(name: String) {
+                if (definitions.containsKey(name)) {
+                    val id = org.partiql.spi.catalog.Identifier.delimited(name)
+                    // TODO: We may want to introduce a VAR_DECL_AMBIGUOUS or NAME_SHADOW warning.
```
alancai98: Is this TODO still relevant? The following line is reporting an ambiguous var ref error

partiql-planner/src/main/kotlin/org/partiql/planner/internal/window/WindowFunctionSignatureProvider.kt
```diff
@@ -0,0 +1,56 @@
...
+     * Creates the LAG window function signature.
+     * TODO: We may eventually want/need to handle coercion of the expr/default. This would require this class
+     *  to return a nullable [org.partiql.plan.WindowFunctionNode], not a [WindowFunctionSignature].
+     * @return null if unable to create a signature.
+     */
+    private fun lagOrLead(name: String, expr: Rex, offset: Rex, default: Rex, ignoreNulls: Boolean): WindowFunctionSignature? {
+        if (offset.type.code() != PType.BIGINT) {
+            return null
+        }
+        val returnTyper = DynamicTyper()
```
alancai98, line 51: May be helpful to provide some clarification why we are using the dynamic typer for `lag` and `lead`

partiql-spi/src/main/java/org/partiql/spi/errors/PRuntimeException.java
```diff
@@ -55,9 +55,30 @@ public int hashCode() {
...
+     * @return null
+     */
+    @Override
+    public synchronized Throwable getCause() {
+        // DEVELOPERS: You may TEMPORARILY uncomment the following 4 lines during debugging. Do not commit the lines uncommented.
+        // Throwable t = error.getOrNull("CAUSE", Throwable.class);
+        // if (t != null) {
+        //     return t;
+        // }
+        return null;
```
alancai98, line 72: Why are we returning `null` for the cause here?

partiql-spi/src/main/java/org/partiql/spi/value/DatumString.java
```diff
@@ -35,7 +35,7 @@ public PType getType() {
     public String toString() {
         return "DatumString{" +
                 "_type=" + _type +
-                ", _value=" + _value +
+                ", _value='" + _value + '\'' +
```
alancai98, line 38: nit: for consistency and don't need the escape if we use double-quotes

```suggestion
                ", _value='" + _value + "'" +
```

partiql-ast/src/main/java/org/partiql/ast/AstVisitor.java
```diff
@@ -324,6 +325,106 @@ public R visitExprArray(ExprArray node, C ctx) {
         return defaultVisit(node, ctx);
     }
 
+    /**
```
alancai98, line 328: For these functions, probably need an experimental notice.

partiql-ast/src/main/java/org/partiql/ast/WindowSpecification.java
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Represents a window specification or a window reference.
+ * @see WindowClause.WindowDefinition#getSpecification()
+ * @deprecated This feature is experimental and is subject to change.
+ */
+@Builder(builderClassName = "Builder")
+@EqualsAndHashCode(callSuper = false)
+@Deprecated
+public final class WindowSpecification extends AstNode {
+    // TODO: In the future: @Nullable private final WindowFrameClause frameClause;
```
alancai98, line 20: Could we create a tracking issue for the TODOs brought up in this PR? In the code and in PR description there are a lot of TODOs mentioned (e.g. window frame clause, unsupported functions in `WindowBuiltins`) but may be hard to discover unless we do a code search of the repo.

partiql-parser/src/main/antlr/PartiQLParser.g4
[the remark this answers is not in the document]
alancai98: I'm still not fully-convinced that the modeling for the window functions needs to be hard-coded in the parser and AST. In the future if we are to support more window functions, we would need additional rules in the parser and variants within the `WindowFunctionType`, which may be redundant. Trino, for instance, defines all scalar, aggregate, and window functions as a [FunctionCall](https://github.com/trinodb/trino/blob/d1501ee536c0c6883ed960a184ba7b3f9e36a6c3/core/trino-parser/src/main/java/io/trino/sql/tree/FunctionCall.java) (see their [grammar](https://github.com/trinodb/trino/blob/d1501ee536c0c6883ed960a184ba7b3f9e36a6c3/core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4#L594-L597)).

Since we've marked all the public APIs as experimental though, I'm still fine w/ proceeding and experimenting further as we get feedback from customers.

alancai98 (commented):
Thanks for updating the PR and adding the experimental notices! Mostly left some minor comments/feedback that should be quick to address before we merge. A couple other things to highlight:
- I think we should try to use a separate tracking issue for things that are left as TODOs. Otherwise, the TODOs may be hidden in the code and hard to prioritize.
- The current modeling of window functions in the parser/ast may need further experimentation as we integrate w/ customers. Since everything that's public will be marked w/ experimental, I think we can still proceed with merging in its current modeling.

partiql-ast/src/main/java/org/partiql/ast/WindowPartition.java
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Represents a partition specified by a name.
+     * @see WindowSpecification#getPartitionClause()
+     * @deprecated This feature is experimental and is subject to change.
+     */
+    @Builder(builderClassName = "Builder")
+    @EqualsAndHashCode(callSuper = false)
+    @Deprecated
```
yliuuuu: NIT: For this an other place applicable, can we add a message clarifying that the @Deprecate annotation is for experimental feature usage.

yliuuuu:
```suggestion
    @Deprecated("This is an experimental feature and may be changing without major version upgrade")
```

johnedquinn: It's in the Javadoc. Unfortunately, Java's "@Deprecated" annotation doesn't have a description like Kotlin's.

partiql-ast/src/main/java/org/partiql/ast/AstVisitor.java
```diff
@@ -532,6 +641,14 @@ public R visitLet(Let node, C ctx) {
         return defaultVisit(node, ctx);
     }
 
+    public R visitWindowClause(WindowClause node, C ctx) {
+        return defaultVisit(node, ctx);
+    }
+
+    public R visitWindowDefinition(WindowClause.Definition node, C ctx) {
+        return defaultVisit(node, ctx);
+    }
```
alancai98, line 690: Missing some deprecated/warning notices on these functions as well

partiql-ast/src/main/java/org/partiql/ast/WindowFunctionType.java
```diff
@@ -0,0 +1,309 @@
...
+    /**
+     * The RANK window function.
+     * @see ExprWindowFunction#getFunctionType()
+     * @deprecated This feature is experimental and is subject to change.
+     */
+    @Builder(builderClassName = "Builder")
+    @EqualsAndHashCode(callSuper = false)
+    @Deprecated
+    public static final class Rank extends WindowFunctionType {
+        private int _type;
```
alancai98: nit: this kind of feels a lot like `AstEnum`. If we can't make `Rank` extend `AstEnum` as well (I think this is not permitted in java to extend multiple abstract classes), perhaps we can just have these as their own variant of `WindowFunctionType` since there's no shared properties. Like

```
public static final class Rank extends WindowFunctionType { ... }
public static final class DenseRank extends WindowFunctionType { ... }
public static final class PercentRank extends WindowFunctionType { ... }
```

alancai98 (approved):
A couple final minor comments. Otherwise looks good!

partiql-ast/src/main/java/org/partiql/ast/WindowSpecification.java
alancai98: Issue created by John -- https://github.com/partiql/partiql-lang-kotlin/issues/1837

PR #1749, opened 2025-04-18
fix: misc issues when working with PLK-1.x

alancai98:
## Relevant Issues
- None

## Description
- Fixing some miscellaneous issues encountered while working to upgrade Scribe to use PLK-1.x. See the CHANGELOG or the self-review for more details.

## Other Information
- Updated Unreleased Section in CHANGELOG: **[YES/NO]**: Yes
- Any backward-incompatible changes? **[YES/NO]**: No
- Any new external dependencies? **[YES/NO]**: No
- Do your changes comply with the [contributing][cg] and [code style][csg] guidelines? **[YES/NO]** Yes

## License Information

### DIFF
--- a/partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
+++ b/partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
@@ -128,8 +128,15 @@ public abstract class SqlDialect : AstVisitor<SqlBlock, SqlBlock>() {
      * @param node
      * @param tail
      */
-    public open fun visitExprWrapped(node: Expr, tail: SqlBlock): SqlBlock = when (node) {
-        is ExprQuerySet -> {
+    public open fun visitExprWrapped(node: Expr, tail: SqlBlock, operator: Boolean = false): SqlBlock = when {
+        node is ExprQuerySet -> {
+            var t = tail
+            t = t concat "("
+            t = visit(node, t)
+            t = t concat ")"
+            t
+        }
+        node is ExprOperator && operator -> {
             var t = tail
             t = t concat "("
             t = visit(node, t)
@@ -264,15 +271,14 @@ public abstract class SqlDialect : AstVisitor<SqlBlock, SqlBlock>() {
         val lhs = node.lhs
         return if (lhs != null) {
             var t = tail
-            t = visitExprWrapped(lhs, t)
+            t = visitExprWrapped(lhs, t, operator = true)
             t = t concat " ${node.symbol} "
-            t = visitExprWrapped(node.rhs, t)
+            t = visitExprWrapped(node.rhs, t, operator = true)
             t
         } else {
             var t = tail
-            t = t concat node.symbol + "("
-            t = visitExprWrapped(node.rhs, t)
-            t = t concat ")"
+            t = t concat node.symbol
+            t = visitExprWrapped(node.rhs, t, operator = true)
             return t
         }
     }
--- a/partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
+++ b/partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
@@ -535,14 +535,22 @@ public abstract class SqlDialect : AstVisitor<SqlBlock, SqlBlock>() {
         t = t concat "TRIM("
         // [LEADING|TRAILING|BOTH]
         val trimSpec = node.trimSpec
-        if (trimSpec != null) {
-            t = t concat "${trimSpec.name()} "
-        }
-        // [<chars> FROM]
         val chars = node.chars
-        if (chars != null) {
-            t = visitExprWrapped(chars, t)
-            t = t concat " FROM "
+        when {
+            trimSpec != null && chars != null -> {
+                t = t concat trimSpec.name()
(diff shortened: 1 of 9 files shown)


partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -884,8 +884,12 @@ internal class PlanTyper(private val env: Env, config: Context) {
         private fun replaceCaseBranch(originalRex: Rex, outputType: CompilerType, function: DynamicTyper.Mapping): Rex {
             return when (function) {
                 is DynamicTyper.Mapping.Coercion -> {
-                    val cast = env.resolveCast(originalRex, function.target)!!
-                    Rex(outputType, cast)
+                    if (originalRex.type == function.target) {
```
alancai98, line 887: self-review: we were adding extra `CAST`s which weren't necessary (e.g. when the `CASE WHEN` branches had the same type). e.g. 

```
CASE WHEN ... THEN 1 ELSE 2 END
```

would get typed as

```
CASE WHEN ... THEN CAST(1 AS INT) ELSE CAST(1 AS INT) END
```

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
```diff
@@ -851,7 +851,12 @@ internal object RexConverter {
                 DataType.CHARACTER, DataType.CHAR -> call(FunctionUtils.OP_IS_CHAR, targetType.length.toRex(), arg0)
                 DataType.CHARACTER_VARYING, DataType.VARCHAR -> call(FunctionUtils.OP_IS_VARCHAR, targetType.length.toRex(), arg0)
                 DataType.CLOB -> call(FunctionUtils.OP_IS_CLOB, arg0)
-                DataType.STRING -> call(FunctionUtils.OP_IS_STRING, targetType.length.toRex(), arg0)
+                DataType.STRING -> {
```
alancai98, line 854: self-review: the function calls for `IS STRING` and `IS DECIMAL`/`IS NUMERIC` without any type arguments parameters were getting mistranslated to the wrong function call

partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
```diff
@@ -128,8 +128,15 @@ public abstract class SqlDialect : AstVisitor<SqlBlock, SqlBlock>() {
      * @param node
      * @param tail
      */
-    public open fun visitExprWrapped(node: Expr, tail: SqlBlock): SqlBlock = when (node) {
-        is ExprQuerySet -> {
+    public open fun visitExprWrapped(node: Expr, tail: SqlBlock, operator: Boolean = false): SqlBlock = when {
```
alancai98, line 131: self-review: not quite a fix but more of adjustment to follow how the pretty-printer for Scribe wraps expressions in parens. shouldn't be a breaking change since we provide a default value for `operator`

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
alancai98: previously: IS STRING -> IS STRING(<null>)
fixed: IS STIRNG -> IS STRING

similarly for decimal
previously: IS DECIMAL -> IS DECIMAL(<null>, <null>)
fixed: IS DECIMAL -> IS DECIMAL

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
```diff
@@ -860,13 +865,20 @@ internal object RexConverter {
...
-                DataType.NUMERIC -> call(FunctionUtils.OP_IS_NUMERIC, targetType.precision.toRex(), targetType.scale.toRex(), arg0)
-                DataType.DEC, DataType.DECIMAL -> call(
-                    FunctionUtils.OP_IS_DECIMAL,
-                    targetType.precision.toRex(),
-                    targetType.scale.toRex(),
-                    arg0
-                )
+                DataType.NUMERIC, DataType.DEC, DataType.DECIMAL -> {
+                    when {
+                        targetType.precision == null && targetType.scale == null -> call(
```
jpschorr, line 870: if precision is non-null, then scale is guaranteed to be non-null?

jpschorr (dismissed):

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
alancai98: following our conversation, I checked this again and seems like I was wrong. the null scale conversion only applied for `CAST`s and not for `IS TYPE`. in the latest commit, I added the conversion for `null` scale to be a scale of `0` for `IS TYPE` -- [`65fef33` (#1749)](https://github.com/partiql/partiql-lang-kotlin/pull/1749/commits/65fef335b2126f341696eab808e331b5a1c36aa7)

jpschorr (approved):

PR #1755, opened 2025-05-19
fix: BETWEEN operator function overloads

alancai98:
## Relevant Issues
- None

## Description
- `BETWEEN` operator was not mapping to the correct function in the case of mixed numeric types

## Other Information

## License Information


alancai98:
The following were previously failing but now pass
```
between mixed types int int decimal false, compileOption: PERMISSIVE
between mixed types int int decimal false, compileOption: STRICT
between mixed types decimal int int false, compileOption: PERMISSIVE
between mixed types decimal int int false, compileOption: STRICT
```

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/DiadicComparisonOperator.kt
```diff
@@ -39,7 +39,7 @@ internal abstract class DiadicComparisonOperator(name: String) : DiadicOperator(
 
     abstract fun getNumberComparison(lhs: Number, rhs: Number): Boolean
 
-    private fun Datum.getNumber(): Number {
+    internal fun Datum.getNumber(): Number {
```
jpschorr, line 42: Maybe we ought to move this extension function?

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnBetween.kt
```diff
@@ -1,230 +1,152 @@
...
-@file:Suppress("ClassName")
-
 package org.partiql.spi.function.builtins
 
+import org.partiql.spi.function.Fn
+import org.partiql.spi.function.FnOverload
+import org.partiql.spi.function.Function.instance
 import org.partiql.spi.function.Parameter
+import org.partiql.spi.function.RoutineOverloadSignature
+import org.partiql.spi.function.builtins.FnGt.getNumber
```
jpschorr, line 8: I think this is an incorrect import?

jpschorr (commented):
This is changing `BETWEEN` from a statically dispatched op into a dynamically dispatched op like other comparison operators are currently, correct?

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/DiadicComparisonOperator.kt
alancai98: Yea, moved to a different Datum util file as an extension function [`3cfeb1d` (#1755)](https://github.com/partiql/partiql-lang-kotlin/pull/1755/commits/3cfeb1df6b0ec2dde306c075bcf9f12798e5eb0f)

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnBetween.kt
alancai98: Yea, not too sure what happened with this import. Seems like it was importing the `getNumber` through the `FnGt` object which extends `DiadicComparisonOperator`. Fixed in [`3cfeb1d` (#1755)](https://github.com/partiql/partiql-lang-kotlin/pull/1755/commits/3cfeb1df6b0ec2dde306c075bcf9f12798e5eb0f).

jpschorr (approved):

PR #1760, opened 2025-05-20
Enable definition of aggregation functions in the Catalog

jpschorr:
## Relevant Issues
- N/A

## Description
Previously, the list of aggregation functions known to planning was hard-coded. This PR looks up aggregation functions in the catalog, allowing planning to find both built-ins and catalog-defined functions.

## Other Information

## License Information

### DIFF
--- a/partiql-eval/src/test/kotlin/org/partiql/eval/CatalogTest.kt
+++ b/partiql-eval/src/test/kotlin/org/partiql/eval/CatalogTest.kt
@@ -1,0 +1,163 @@
+package org.partiql.eval
+
+import org.junit.jupiter.api.Test
+import org.partiql.eval.compiler.PartiQLCompiler
+import org.partiql.parser.PartiQLParser
+import org.partiql.planner.PartiQLPlanner
+import org.partiql.spi.Context
+import org.partiql.spi.catalog.Catalog
+import org.partiql.spi.catalog.Session
+import org.partiql.spi.errors.PError
+import org.partiql.spi.errors.PErrorListener
+import org.partiql.spi.function.AggOverload
+import org.partiql.spi.function.FnOverload
+import org.partiql.spi.function.Parameter
+import org.partiql.spi.types.PType
+import org.partiql.spi.value.Datum
+import org.partiql.spi.value.Field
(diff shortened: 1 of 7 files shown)


partiql-eval/src/test/kotlin/org/partiql/eval/CatalogTest.kt
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    @Test
+    fun scalar_extension_function() {
+        val query = "SELECT abs(x) as a, foo_bar_baz(x) as b, FOO_BAR_BAZ(x) as C FROM << {'x':1}, {'x':2} >>"
+        val catalog = object : Catalog {
+            override fun getName(): String = "default"
+            override fun getFunctions(session: Session, name: String): Collection<FnOverload> {
+                return listOf(
+                    // Just a `abs(F64, F64) -> F64` with a different name
```
alancai98, line 49: nit: fn signature should be a single argument?

```suggestion
                    // Just a `abs(F64) -> F64` with a different name
```

partiql-eval/src/test/kotlin/org/partiql/eval/CatalogTest.kt
```diff
@@ -0,0 +1,163 @@
...
+        val datum = (result as TestResult.Success).result
+        val expected = Datum.bagVararg(
+            Datum.struct(
+                Field.of("a", Datum.doublePrecision(3.0)),
+                Field.of("b", Datum.doublePrecision(3.0)),
+                Field.of("C", Datum.doublePrecision(3.0))
+            ),
+        )
+        assertEquals(0, Datum.comparator().compare(expected, datum))
+    }
```
alancai98, line 137: wonder if we could also test a few other scenarios:
1. aggregate UDF has same name as a builtin scalar (e.g. UPPER)
2. scalar UDF has same name as a builtin aggregate (e.g. COUNT)
3. scalar UDF and aggregate UDF have same name

alancai98 (commented):
implementation looks good. glad that we could finally get rid of the hard-coded aggregates. left a couple minor comments

PR #1768, opened 2025-07-09
feat: experimental WITH representation in the plan

alancai98:
## Relevant Issues
- None

## Description
- Adds an experimental, limited `With` representation in the partiql-plan
  - This is currently marked as experimental since we may choose a different modeling of CTEs in the future.
- Adds an experimental `With` configuration to the planner that controls whether `With` variable references are rewritten to their query representation.
  - This PR preserves the original behavior to rewrite `With` variable references while also providing the option to not rewrite the var refs.
- Adds some missing implementations of `With` and `WithListElement` in `SqlDialect` and `AstVisitor`

## Other Information

## License Information

### DIFF
--- a/partiql-eval/src/main/java/org/partiql/eval/Environment.java
+++ b/partiql-eval/src/main/java/org/partiql/eval/Environment.java
@@ -63,7 +63,7 @@ public class Environment {
         try {
             return stack[depth].getValues()[offset];
         } catch (IndexOutOfBoundsException ex) {
-            throw new RuntimeException("Invalid variable reference [$depth:$offset]\n$this");
+            throw new RuntimeException(String.format("Invalid variable reference [%d:%d]\n%s", depth, offset, this));
         }
     }
 
--- a/partiql-plan/src/main/java/org/partiql/plan/rel/RelWith.java
+++ b/partiql-plan/src/main/java/org/partiql/plan/rel/RelWith.java
@@ -1,0 +1,105 @@
+package org.partiql.plan.rel;
+
+import org.jetbrains.annotations.NotNull;
+import org.partiql.plan.Operand;
+import org.partiql.plan.OperatorVisitor;
+import org.partiql.plan.WithListElement;
+import org.partiql.plan.rex.Rex;
+
+import java.util.List;
+
+/**
+ * <p>
+ * <b>NOTE:</b> This is experimental and subject to change without prior notice!
+ * </p>
+ * <p>
+ * WITH relation expression. This is currently experimental and is currently missing some
+ * features like recursive CTEs.
+ */
+public abstract class RelWith extends RelBase {
+
+    /**
+     * Creates a new {@link RelWith} instance.
+     *
+     * @param input input rel (operand 0)
+     * @param elements with list elements
+     * @return new {@link RelWith} instance
+     */
+    @NotNull
+    public static RelWith create(@NotNull Rel input, @NotNull List<WithListElement> elements) {
+        return new Impl(input, elements);
+    }
+
+    /**
+     * Gets the input rel (operand 0).
+     * @return input rel (operand 0)
+     */
+    @NotNull
+    public abstract Rel getInput();
+
+    /**
+     * Gets the WITH list elements.
+     * @return WITH list elements.
+     */
+    @NotNull
+    public abstract List<WithListElement> getElements();
+
+    @NotNull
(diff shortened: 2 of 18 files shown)


partiql-eval/src/main/java/org/partiql/eval/Environment.java
```diff
@@ -63,7 +63,7 @@ public Datum get(int depth, int offset) {
         try {
             return stack[depth].getValues()[offset];
         } catch (IndexOutOfBoundsException ex) {
-            throw new RuntimeException("Invalid variable reference [$depth:$offset]\n$this");
+            throw new RuntimeException(String.format("Invalid variable reference [%d:%d]\n%s", depth, offset, this));
```
alancai98, line 66: self-review: this was an improperly formatted java string. it was previously using Kotlin's syntax to construct a formatted string

partiql-planner/src/main/kotlin/org/partiql/planner/builder/PartiQLPlannerBuilder.kt
```diff
@@ -18,7 +18,7 @@ import org.partiql.planner.internal.SqlPlanner
  */
 public class PartiQLPlannerBuilder {
 
-    private val flags: MutableSet<PlannerFlag> = mutableSetOf()
+    private val flags: MutableSet<PlannerFlag> = mutableSetOf(PlannerFlag.REPLACE_WITH_REFS)
```
alancai98: self-review:  by default, preserve prior behavior to rewrite all WITH variable references to the query representation. using the below `replaceWithReferences` with a value of false, will not rewrite the variable references

partiql-plan/src/main/java/org/partiql/plan/rel/RelWith.java
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * <p>
+ * <b>NOTE:</b> This is experimental and subject to change without prior notice!
+ * </p>
+ * <p>
+ * WITH relation expression. This is currently experimental and is currently missing some
+ * features like recursive CTEs.
+ */
+public abstract class RelWith extends RelBase {
```
alancai98, line 19: self-review: `WITH`'s representation in the partiql-plan is marked as experimental for now. We may make public API changes as we add the remaining SQL features related to WITH.

partiql-planner/src/main/kotlin/org/partiql/planner/builder/PartiQLPlannerBuilder.kt
```diff
@@ -61,4 +61,21 @@ public class PartiQLPlannerBuilder {
...
+
+    /**
+     * **NOTE** This is experimental and subject to change without prior notice!
+     *
+     * Experimental planner mode to control whether WITH variable references are replaced with their definitions.
+     *
+     * @param replaceWith denotes whether to replace WITH variable references with their definitions.
+     * @return
+     */
+    public fun replaceWithReferences(replaceWith: Boolean = true): PartiQLPlannerBuilder {
```
jpschorr: Having a clause named `With` is unfortunate, in that it makes this read ambiguously.

Essentially this is something like `forceInlineWIthClause`, right?

jpschorr (commented):
Perhaps we should rename `REPLACE_WITH_REFS`, et. al. to something like `INLINE_FORCE_WITH_CLAUSE`.

Some of the constants and methods read ambiguously. 
e.g., is it:
- replace `WITH` refs, or
- replace with refs

partiql-planner/src/main/kotlin/org/partiql/planner/builder/PartiQLPlannerBuilder.kt
alancai98: Yeah, can be confusing `WITH` can refer to the `WITH` clause and the preposition. Also in Kotlin there's a [`with`](https://kotlinlang.org/docs/scope-functions.html#with) scope function which adds more confusion.

I'll change to your suggestions, which are a lot clearer.

alancai98: Updated the naming in [`e388628` (#1768)](https://github.com/partiql/partiql-lang-kotlin/pull/1768/commits/e388628e9ac2deab514f68274fb5400a03e6bd40)

Issue #1742, opened 2025-02-20
Comma join on values without any aliases results in variable not found warnings


alancai98:
## Description
- Comma joining two values in a from source without any alias results in variable not found warnings from the planner

## To Reproduce
```
PartiQL Shell
Version: 1.1.0-742e1501c
partiql ▶ SELECT * FROM <<{'a': 1}>>, <<{'a': 2}>>;
w: [semantic] Variable reference (_0) could not be found in the database environment or in the set of available locals. Locals: [_0].
w: [semantic] Expression always returns missing
w: [semantic] Variable reference (_0) could not be found in the database environment or in the set of available locals. Locals: [_0].
w: [semantic] Variable reference (_0) could not be found in the database environment or in the set of available locals. Locals: [_0].
w: [semantic] Variable reference (_0) could not be found in the database environment or in the set of available locals. Locals: [_0].
w: [semantic] Expression always returns missing
w: [semantic] Variable reference (_0) could not be found in the database environment or in the set of available locals. Locals: [_0].
w: [semantic] Variable reference (_0) could not be found in the database environment or in the set of available locals. Locals: [_0].

=== RESULT ===
<<
  {}
>>
```

## Expected Behavior
When doing a normal explicit CROSS JOIN or providing an alias to one or more values, it works as expected

Using a CROSS JOIN
```
partiql ▶ SELECT * FROM <<{'a': 1}>> CROSS JOIN <<{'a': 2}>>;

=== RESULT ===
<<
  {
    a: 1,
    a: 2
  }
>>

OK!
```
Providing an alias to the comma-join values:
```
partiql ▶ SELECT * FROM <<{'a': 1}>> AS t1, <<{'a': 2}>> AS t2;

=== RESULT ===
<<
  {
    a: 1,
    a: 2
  }
>>
```

## Additional Context
- Java version: 17
- PartiQL version: 1.1.0-742e1501c
- Add any other context about the problem here.
  - Pre-1.0 CLIs do not exhibit this behavior.

PR #1775, opened 2025-07-16
fix: alias collision for multiple table references

XuechunHHH:
## Relevant Issues
- [Closes/Related To] Issue #1742 

## Description
This PR addresses alias collision problem that occurs when multiple table references need auto-generated aliases.

### Current Behavior 

```
partiql ▶ SELECT * FROM <<{'a': 1}>>, <<{'a': 2}>>;

=== RESULT ===
<<
  {
    'a': 1,
    'a': 2
  }
>>

OK!
partiql ▶ SELECT * FROM <<{'a': 1}>> CROSS JOIN <<{'a': 2}>> CROSS JOIN <<{'a': 3}>>;

=== RESULT ===
<<
  {
    'a': 1,
    'a': 2,
    'a': 3
  }
>>

OK!
partiql ▶ SELECT * FROM <<{'a': 1}>> CROSS JOIN <<{'a': 2}>>, <<{'a': 3}>>;

=== RESULT ===
<<
  {
    'a': 1,
    'a': 2,
    'a': 3
  }
>>

OK!
```

## Other Information

## License Information

### DIFF
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
@@ -1213,7 +1213,6 @@ internal object RexConverter {
                 DataType.USER_DEFINED -> TODO("Custom type not supported ")
                 // <interval type>
                 DataType.INTERVAL -> {
-                    // TODO meta for interval when precision is unspecified
                     when (val q = type.intervalQualifier) {
                         is IntervalQualifier.Single -> {
                             val precision = q.precision ?: INTERVAL_DEFAULT_PRECISON
@@ -1226,6 +1225,13 @@ internal object RexConverter {
                                 DatetimeField.MINUTE -> PType.intervalMinute(precision)
                                 DatetimeField.SECOND -> PType.intervalSecond(precision, fractionalPrecision)
                                 else -> error("Unsupported DatetimeField: ${q.field}")
+                            }.also {
+                                if (q.precision == null) {
+                                    it.setUnspecifiedPrecisionMeta()
+                                }
+                                if (q.fractionalPrecision == null) {
+                                    it.setUnspecifiedFractionalPrecisionMeta()
+                                }
                             }
                         }
                         is IntervalQualifier.Range -> {
@@ -1242,6 +1248,13 @@ internal object RexConverter {
                                 DatetimeField.HOUR to DatetimeField.SECOND -> PType.intervalHourSecond(precision, scale)
                                 DatetimeField.MINUTE to DatetimeField.SECOND -> PType.intervalMinuteSecond(precision, scale)
                                 else -> error("Unsupported DatetimeField: $lhsField to $rhsField")
+                            }.also {
+                                if (q.startFieldPrecision == null) {
+                                    it.setUnspecifiedPrecisionMeta()
+                                }
+                                if (q.endFieldFractionalPrecision == null) {
+                                    it.setUnspecifiedFractionalPrecisionMeta()
+                                }
                             }
                         }
                         else -> error("Unsupported IntervalQualifier: $q")


partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/NormalizeFromSource.kt
```diff
@@ -32,17 +32,35 @@ import org.partiql.planner.internal.util.BinderUtils.toBinder
...
  */
 internal object NormalizeFromSource : AstPass {
-
     override fun apply(statement: Statement): Statement = statement.accept(Visitor, 0) as Statement
 
     private object Visitor : AstRewriter<Int>() {
-
         // Each SFW starts the ctx count again.
-        override fun visitQueryBodySFW(node: QueryBody.SFW, ctx: Int): AstNode = super.visitQueryBodySFW(node, 0)
+        override fun visitQueryBodySFW(
```
alancai98: Taking a look at this AstPass, I'm noticing some existing bugs beyond just the comma-join case mentioned in #1742.

Consider when we do an explicit CROSS JOIN over three `FromExpr`:

```
partiql ▶ SELECT * FROM <<{'a': 1}>> CROSS JOIN <<{'a': 2}>> CROSS JOIN <<{'a': 3}>>;
w: [semantic] Variable reference (_1) could not be found in the database environment or in the set of available locals. Locals: [_0, _1].
w: [semantic] Expression always returns missing
w: [semantic] Variable reference (_1) could not be found in the database environment or in the set of available locals. Locals: [_0, _1].
w: [semantic] Variable reference (_1) could not be found in the database environment or in the set of available locals. Locals: [_0, _1].
w: [semantic] Variable reference (_1) could not be found in the database environment or in the set of available locals. Locals: [_0, _1].
w: [semantic] Expression always returns missing
w: [semantic] Variable reference (_1) could not be found in the database environment or in the set of available locals. Locals: [_0, _1].
w: [semantic] Variable reference (_1) could not be found in the database environment or in the set of available locals. Locals: [_0, _1].

=== RESULT ===
<<
  {
    'a': 1
  }
>>

OK!
```

My expectation is that the above query would have output 
```
<<
  {
    'a': 1,
    'a': 2,
    'a': 3
  }
>>
```

I think the real issue is that when we pass in an updated `ctx` to a visit function for `JOIN`s and table references, we currently assume that the `ctx` has been incremented by a fixed amount. For example, in `visitFromJoin`, we increment the `ctx` by 1 after visiting the `lhs` - https://github.com/XuechunHHH/partiql-lang-kotlin/blob/7e57827baf838d3a94a2f94100269100d837814a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/NormalizeFromSource.kt#L64-L65.

One possible fix is to create a class just storing the current count and use that as the `ctx` rather than just an `Int`. Something like:
```Kotlin
class IntCounter(var counter: Int = 0)
```

Then whenever a `FromExpr` encounters a null `AS` alias or `AT` alias, we increment this counter.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/NormalizeFromSource.kt
alancai98, line 1: Let's add some new end-to-end evaluation tests for this change. There are a couple ways to test evaluation:
1. Add a unit test within a class such as [PartiQLEvaluatorTest](https://github.com/XuechunHHH/partiql-lang-kotlin/blob/7e57827baf838d3a94a2f94100269100d837814a/partiql-eval/src/test/kotlin/org/partiql/eval/internal/PartiQLEvaluatorTest.kt)
2. Add to the conformance tests -- https://github.com/partiql/partiql-tests

It's generally recommended to add any parsing and evaluation tests to the conformance tests since we can ensure the same behavior across PartiQL implementations. For PLK, these test can be run locally and as part of the GitHub Actions. The process to add more conformance tests can be slightly more involved so LMK if you have any questions. It'll normally follow this process:
1. Cut a PR to partiql-tests with your new tests
2. Once approved, update the Git submodule for PLK
3. PR the updated Git submodule in a separate PR or you can also do it in this PR

Regarding some tests we can add, I think we can start with the one mentioned in the GH issue. We can also add
- The cross join case I mentioned in the other comment `SELECT * FROM <<{'a': 1}>> CROSS JOIN <<{'a': 2}>> CROSS JOIN <<{'a': 3}>>`
- A mix of cross joins and comma joins `SELECT * FROM <<{'a': 1}>> CROSS JOIN <<{'a': 2}>>, <<{'a': 3}>>`

alancai98 (commented):
Thanks for taking on this issue! The underlying issue is a bit more involved than the GH issue described. Feel free to lmk if you have some followup questions.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/NormalizeFromSource.kt
XuechunHHH: I see the real issue here. Thanks for bringing it up.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/NormalizeFromSource.kt
XuechunHHH: Thanks for the detailed guidance on adding conformance tests! I understand the process and will add the appropriate tests to `partiql-tests` accordingly.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/NormalizeFromSource.kt
```diff
@@ -73,4 +85,7 @@ internal object NormalizeFromSource : AstPass {
             }
         }
     }
+
+    fun visitStatement(node: Statement): Statement =
+        node.accept(Visitor, IntCounter(0)) as Statement
```
alancai98: This appears unused

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/NormalizeFromSource.kt
```diff
@@ -32,19 +32,32 @@ import org.partiql.planner.internal.util.BinderUtils.toBinder
...
  */
 internal object NormalizeFromSource : AstPass {
+    override fun apply(statement: Statement): Statement = statement.accept(Visitor, IntCounter(0)) as Statement
 
-    override fun apply(statement: Statement): Statement = statement.accept(Visitor, 0) as Statement
-
-    private object Visitor : AstRewriter<Int>() {
+    internal class IntCounter(var counter: Int = 0) {
+        fun next(): Int = counter++
+    }
```
alancai98, line 38: Nice, updated context looks correct.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/NormalizeFromSource.kt
```diff
@@ -32,19 +32,32 @@ import org.partiql.planner.internal.util.BinderUtils.toBinder
...
+            // Handle the case where From contains multiple table references
+            if (node.tableRefs.size > 1) {
+                val newTableRefs = node.tableRefs.map { tableRef ->
+                    tableRef.accept(this, ctx) as FromTableRef // assigning different context values to each table reference
+                }
+                return From(newTableRefs)
+            }
+            // Handle single table
+            return super.visitFrom(node, ctx) as From
+        }
```
alancai98: With the context changes, I believe we can actually get rid of this overridden visit function.

XuechunHHH: Yep, thanks for pointing it out. The `IntCounter` approach makes the override unnecessary. I will remove it.

test/partiql-tests-runner/README.md
```diff
@@ -12,11 +12,11 @@ This package enables:
 
 ```shell
 # default, test data from partiql-tests submodule will be used
-./gradlew :test:partiql-tests-runner:ConformanceTestReport
+./gradlew :test:partiql-tests-runner:generateTestReport
```
alancai98, line 15: Thanks for updating the readme!

alancai98 (approved):
Nice work and thanks for tackling that issue!

Issue #1777, opened 2025-07-18
REPL exits for `INTERNAL` datum in some case


zyfy29:
## Description
REPL exits when `INTERNAL` datum underflows.

## To Reproduce
run such a command of below, cli will throw a `java.lang.StringIndexOutOfBoundsException` and exit

```
interval '2:3.00000005' minute to second(6);
interval '3:4' minute to second(3);
```

## Expected Behavior
Give the expected result. 
If this case is not allowed, print a error and continue processing.

## Additional Context
- Java version: 23
- PartiQL version: 1.2.0-4ac06dcb3

PR #1778, opened 2025-07-18
fix: handle the formatting when the nanoseconds of INTERVAL datum underflow

zyfy29:
## Relevant Issues
- fix Issue #1777 

## Description
For the `INTERVAL` datum, when the `value` of nanos is less than the `10**scale`, formatter function will throw a exception, which causes a unexpected REPL exit.
I changed the formatter function logic to handle this case correctly.

## Other Information

## License Information

### DIFF
--- a/partiql-cli/src/test/kotlin/org/partiql/cli/io/DatumWriterTextPrettyTests.kt
+++ b/partiql-cli/src/test/kotlin/org/partiql/cli/io/DatumWriterTextPrettyTests.kt
@@ -1,6 +1,8 @@
 package org.partiql.cli.io
 
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.MethodSource
 import org.partiql.spi.types.PType
 import org.partiql.spi.value.Datum
 import org.partiql.spi.value.Field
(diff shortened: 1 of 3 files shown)


partiql-cli/src/test/kotlin/org/partiql/cli/io/DatumWriterTextPrettyTests.kt
```diff
@@ -160,10 +164,22 @@ class DatumWriterTextPrettyTests {
...
+        fun fractionalSecondsCases(): List<Any> {
+            return listOf(
+                arrayOf(12345678, 9, ".012345678"),
+                arrayOf(123456789, 9, ".123456789"),
+                arrayOf(300000000, 1, ".3"),
+                arrayOf(500000000, 2, ".50"),
+                arrayOf(0, 2, ".00"),
+                arrayOf(20000, 3, ".000"),
+            )
+        }
```
alancai98: Since these tests use an array of different types, I'm getting this warning:
> Type argument for reified type parameter 'T' was inferred to the intersection of ['Comparable<*>' & 'Serializable']. Reification of an intersection type results in the common supertype being used. This may lead to subtle issues and an explicit type argument is encouraged. This will become an error in a future release.

Rather than fixing this warning, could we actually make these tests follow the format of the other Datum -> expected string output tests in this file? That way we don't need to use introduce custom testing logic and use the `@VisibleForTesting` annotation. Doing so can make these tests more maintainable in the long-term.

zyfy29:
OK, I did forgot to check the `@VisibleForTesting` usage. That is better for the code consistency indeedly.

alancai98:
Thanks again for the issue and PR!

Issue #1779, opened 2025-07-18
`INTERVAL` times and divide with numerics


alancai98:
SQL spec defines a few more operations that are not yet implemented in partiql-lang-kotlin. These are also pointed out in https://github.com/partiql/partiql-lang-kotlin/pull/1741's description.

Operations include
- [ ] INTERVAL * NUMERIC -> INTERVAL
- [ ] INTERVAL / NUMERIC -> INTERVAL
- [ ] NUMERIC * INTERVAL -> INTERVAL

Example expected query behavior:
```
-- note: the following query currently returns `MISSING`
partiql ▶ INTERVAL '20' YEAR * 2;

=== RESULT ===
INTERVAL '40' YEAR (2)

OK!
```

alancai98:
Implemented in #1792.

Issue #1784, opened 2025-07-18
Investigate `INTERVAL` and DateTime arithmetic behavior


alancai98:
Tracking issue for behavior of datetime operations with `INTERVAL` output type and result. There's some ambiguity in the SQL spec and in our current implementation for
1. date +/- day time interval
2. time +/- year month interval

PR #1741, opened 2025-02-18
[DRAFT] Adds support for the interval type

johnedquinn:
## Relevant Issues
- Closes https://github.com/partiql/partiql-lang-kotlin/issues/1058
- Closes https://github.com/partiql/partiql-lang-kotlin/issues/1657

## Description
- Adds support for the interval type

## License Information

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1792, opened 2025-07-25
Add support INTERVAL times or divide with numerics #1779

xd1313113:
## Relevant Issues
- Closes Issue #1779

## Description
- Add support `INTERVAL` times or divide with numerics 

## Other Information
- Any backward-incompatible changes? **NO**:
- Any new external dependencies? **NO**:

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ rewritten to their query representation
 - Fix alias collision problem when multiple table references need auto-generated aliases
 - Add `ABS` function support on `INTERVAL` field
 - Display of `Ion` in PartiQL CLI output
+- Support `INTERVAL` times and divide with numerics
 
 ### Removed
 
--- a/partiql-eval/src/test/kotlin/org/partiql/eval/internal/IntervalDivideTests.kt
+++ b/partiql-eval/src/test/kotlin/org/partiql/eval/internal/IntervalDivideTests.kt
@@ -6,6 +6,15 @@ import org.junit.jupiter.params.ParameterizedTest
 import org.junit.jupiter.params.provider.MethodSource
 import org.partiql.spi.value.Datum
 
+/**
+ * Test cases for interval division operations.
+ * Tests division of intervals by numeric values including integers, decimals, doubles,
+ * negative numbers, null, missing values, and edge cases like division by zero.
+ *
+ * Related Github issues:
+ * https://github.com/partiql/partiql-lang/issues/100
+ * https://github.com/partiql/partiql-lang-kotlin/issues/1784
+ */
 class IntervalDivideTests {
 
     @ParameterizedTest
--- a/partiql-eval/src/test/kotlin/org/partiql/eval/internal/IntervalDivideTests.kt
+++ b/partiql-eval/src/test/kotlin/org/partiql/eval/internal/IntervalDivideTests.kt
@@ -1,0 +1,86 @@
+package org.partiql.eval.internal
+
+import org.junit.jupiter.api.parallel.Execution
+import org.junit.jupiter.api.parallel.ExecutionMode
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.MethodSource
+import org.partiql.spi.value.Datum
+
+class IntervalDivideTests {
+
+    @ParameterizedTest
+    @MethodSource("intervalDivideNumberCases")
+    @Execution(ExecutionMode.CONCURRENT)
+    fun intervalDivideNumber(tc: SuccessTestCase) = tc.run()
+
+    @ParameterizedTest
+    @MethodSource("intervalDivideNegativeCases")
+    @Execution(ExecutionMode.CONCURRENT)
+    fun intervalDivideNegative(tc: SuccessTestCase) = tc.run()
+
+    @ParameterizedTest
+    @MethodSource("intervalDivideZeroCases")
+    @Execution(ExecutionMode.CONCURRENT)
+    fun intervalDivideZero(tc: FailureTestCase) = tc.run()
+
+    companion object {
+
+        private const val INTERVAL_Y = "INTERVAL '3' YEAR"
+        private const val INTERVAL_M = "INTERVAL '7' MONTH"
+        private const val INTERVAL_YM = "INTERVAL '1-5' YEAR TO MONTH"
+
+        private const val INTERVAL_D = "INTERVAL '2' DAY"
+        private const val INTERVAL_H = "INTERVAL '4' HOUR"
+        private const val INTERVAL_MIN = "INTERVAL '5' MINUTE"
+        private const val INTERVAL_S = "INTERVAL '10.5' SECOND"
+        private const val INTERVAL_DTS = "INTERVAL '2 4:5:10.5' DAY TO SECOND"
+
+        private class Input(
+            val arg0: String,
+            val arg1: String,
+            val expected: Datum
+        )
+
+        @JvmStatic
+        fun intervalDivideNumberCases() = listOf(
+            Input(INTERVAL_Y, "3", Datum.intervalYearMonth(1, 0, 2)),
+            Input(INTERVAL_M, "3", Datum.intervalYearMonth(0, 2, 2)),
+            Input(INTERVAL_YM, "3", Datum.intervalYearMonth(0, 5, 2)),
+
+            Input(INTERVAL_D, "3", Datum.intervalDaySecond(0, 16, 0, 0, 0, 2, 6)),
+            Input(INTERVAL_H, "3", Datum.intervalDaySecond(0, 1, 20, 0, 0, 2, 6)),
+            Input(INTERVAL_MIN, "3", Datum.intervalDaySecond(0, 0, 1, 40, 0, 2, 6)),
+            Input(INTERVAL_S, "3", Datum.intervalDaySecond(0, 0, 0, 3, 5 * 1000000000, 2, 6)),
+            Input(INTERVAL_DTS, "3", Datum.intervalDaySecond(0, 17, 21, 43, 5 * 1000000000, 2, 6))
+        ).map { case ->
+            SuccessTestCase("${case.arg0} / ${case.arg1}", case.expected)
+        }
+
+        @JvmStatic
+        fun intervalDivideNegativeCases() = listOf(
+            Input(INTERVAL_Y, "-3", Datum.intervalYearMonth(0, 0, 2)),
+            Input(INTERVAL_M, "-3", Datum.intervalYearMonth(0, 0, 2)),
+            Input(INTERVAL_YM, "-3", Datum.intervalYearMonth(0, 0, 2)),
+
+            Input(INTERVAL_D, "-3", Datum.intervalDaySecond(0, 0, 0, 0, 0, 2, 6)),
+            Input(INTERVAL_H, "-3", Datum.intervalDaySecond(0, 0, 0, 0, 0, 2, 6)),
+            Input(INTERVAL_MIN, "-3", Datum.intervalDaySecond(0, 0, 0, 0, 0, 2, 6)),
+            Input(INTERVAL_S, "-3", Datum.intervalDaySecond(0, 0, 0, 0, 0, 2, 6)),
+            Input(INTERVAL_DTS, "-3", Datum.intervalDaySecond(0, 0, 0, 0, 0, 2, 6))
+        ).map { case ->
+            SuccessTestCase("${case.arg0} / ${case.arg1}", case.expected)
+        }
+
+        @JvmStatic
+        fun intervalDivideZeroCases() = listOf(
+            INTERVAL_Y,
+            INTERVAL_M,
+            INTERVAL_D,
+            INTERVAL_H,
+            INTERVAL_MIN,
+            INTERVAL_S
+        ).map { interval ->
+            FailureTestCase("$interval / 0")
+        }
+    }
+}
--- a/partiql-eval/src/test/kotlin/org/partiql/eval/internal/IntervalDivideTests.kt
+++ b/partiql-eval/src/test/kotlin/org/partiql/eval/internal/IntervalDivideTests.kt
@@ -34,6 +34,7 @@ class IntervalDivideTests {
         private const val INTERVAL_MIN = "INTERVAL '5' MINUTE"
         private const val INTERVAL_S = "INTERVAL '10.5' SECOND"
         private const val INTERVAL_DTS = "INTERVAL '2 4:5:10.5' DAY TO SECOND"
+        private const val NANOS_PER_SECOND = 1000000000
 
         private class Input(
             val arg0: String,
@@ -43,30 +44,30 @@ class IntervalDivideTests {
 
         @JvmStatic
         fun intervalDivideNumberCases() = listOf(
-            Input(INTERVAL_Y, "3", Datum.intervalYearMonth(1, 0, 2)),
-            Input(INTERVAL_M, "3", Datum.intervalYearMonth(0, 2, 2)),
+            // Input(INTERVAL_Y, "3", Datum.intervalYearMonth(1, 0, 2)),
+            // Input(INTERVAL_M, "3", Datum.intervalYearMonth(0, 2, 2)),
             Input(INTERVAL_YM, "3", Datum.intervalYearMonth(0, 5, 2)),
 
             Input(INTERVAL_D, "3", Datum.intervalDaySecond(0, 16, 0, 0, 0, 2, 6)),
             Input(INTERVAL_H, "3", Datum.intervalDaySecond(0, 1, 20, 0, 0, 2, 6)),
             Input(INTERVAL_MIN, "3", Datum.intervalDaySecond(0, 0, 1, 40, 0, 2, 6)),
-            Input(INTERVAL_S, "3", Datum.intervalDaySecond(0, 0, 0, 3, 5 * 1000000000, 2, 6)),
-            Input(INTERVAL_DTS, "3", Datum.intervalDaySecond(0, 17, 21, 43, 5 * 1000000000, 2, 6))
+            Input(INTERVAL_S, "3", Datum.intervalDaySecond(0, 0, 0, 3, (0.5 * NANOS_PER_SECOND).toInt(), 2, 6)),
+            Input(INTERVAL_DTS, "3", Datum.intervalDaySecond(0, 17, 21, 43, (0.5 * NANOS_PER_SECOND).toInt(), 2, 6))
         ).map { case ->
             SuccessTestCase("${case.arg0} / ${case.arg1}", case.expected)
         }
 
         @JvmStatic
         fun intervalDivideNegativeCases() = listOf(
-            Input(INTERVAL_Y, "-3", Datum.intervalYearMonth(0, 0, 2)),
-            Input(INTERVAL_M, "-3", Datum.intervalYearMonth(0, 0, 2)),
-            Input(INTERVAL_YM, "-3", Datum.intervalYearMonth(0, 0, 2)),
+            Input(INTERVAL_Y, "-3", Datum.intervalYearMonth(-1, 0, 2)),
+            Input(INTERVAL_M, "-3", Datum.intervalYearMonth(0, -2, 2)),
+            Input(INTERVAL_YM, "-3", Datum.intervalYearMonth(0, -5, 2)),
 
-            Input(INTERVAL_D, "-3", Datum.intervalDaySecond(0, 0, 0, 0, 0, 2, 6)),
-            Input(INTERVAL_H, "-3", Datum.intervalDaySecond(0, 0, 0, 0, 0, 2, 6)),
-            Input(INTERVAL_MIN, "-3", Datum.intervalDaySecond(0, 0, 0, 0, 0, 2, 6)),
-            Input(INTERVAL_S, "-3", Datum.intervalDaySecond(0, 0, 0, 0, 0, 2, 6)),
-            Input(INTERVAL_DTS, "-3", Datum.intervalDaySecond(0, 0, 0, 0, 0, 2, 6))
+            Input(INTERVAL_D, "-3", Datum.intervalDaySecond(0, -16, 0, 0, 0, 2, 6)),
+            Input(INTERVAL_H, "-3", Datum.intervalDaySecond(0, -1, -20, 0, 0, 2, 6)),
+            Input(INTERVAL_MIN, "-3", Datum.intervalDaySecond(0, 0, -1, -40, 0, 2, 6)),
+            Input(INTERVAL_S, "-3", Datum.intervalDaySecond(0, 0, 0, -3, (-0.5 * NANOS_PER_SECOND).toInt(), 2, 6)),
+            Input(INTERVAL_DTS, "-3", Datum.intervalDaySecond(0, -17, -21, -43, (-0.5 * NANOS_PER_SECOND).toInt(), 2, 6))
         ).map { case ->
             SuccessTestCase("${case.arg0} / ${case.arg1}", case.expected)
         }
@@ -75,10 +76,12 @@ class IntervalDivideTests {
         fun intervalDivideZeroCases() = listOf(
             INTERVAL_Y,
             INTERVAL_M,
+            INTERVAL_YM,
             INTERVAL_D,
             INTERVAL_H,
             INTERVAL_MIN,
-            INTERVAL_S
+            INTERVAL_S,
+            INTERVAL_DTS
         ).map { interval ->
             FailureTestCase("$interval / 0")
         }
--- a/partiql-eval/src/test/kotlin/org/partiql/eval/internal/IntervalDivideTests.kt
+++ b/partiql-eval/src/test/kotlin/org/partiql/eval/internal/IntervalDivideTests.kt
@@ -23,6 +23,25 @@ class IntervalDivideTests {
     @Execution(ExecutionMode.CONCURRENT)
     fun intervalDivideZero(tc: FailureTestCase) = tc.run()
 
+    @ParameterizedTest
+    @MethodSource("intervalDivideNullCases")
+    @Execution(ExecutionMode.CONCURRENT)
+    fun intervalDivideNull(tc: SuccessTestCase) = tc.run()
+
+    @ParameterizedTest
+    @MethodSource("intervalDivideMissingCases")
+    @Execution(ExecutionMode.CONCURRENT)
+    fun intervalDivideMissing(tc: SuccessTestCase) = tc.run()
+
+    @ParameterizedTest
+    @MethodSource("intervalDivideDecimalCases")
+    @Execution(ExecutionMode.CONCURRENT)
+    fun intervalDivideDecimal(tc: SuccessTestCase) = tc.run()
+    @ParameterizedTest
+    @MethodSource("intervalDivideWithPrecisionOverflowCases")
+    @Execution(ExecutionMode.CONCURRENT)
+    fun intervalDivideWithPrecisionOverflowCases(tc: FailureTestCase) = tc.run()
+
     companion object {
 
         private const val INTERVAL_Y = "INTERVAL '3' YEAR"
@@ -44,8 +63,8 @@ class IntervalDivideTests {
 
         @JvmStatic
         fun intervalDivideNumberCases() = listOf(
-            // Input(INTERVAL_Y, "3", Datum.intervalYearMonth(1, 0, 2)),
-            // Input(INTERVAL_M, "3", Datum.intervalYearMonth(0, 2, 2)),
+            Input(INTERVAL_Y, "3", Datum.intervalYearMonth(1, 0, 2)),
+            Input(INTERVAL_M, "3", Datum.intervalYearMonth(0, 2, 2)),
             Input(INTERVAL_YM, "3", Datum.intervalYearMonth(0, 5, 2)),
 
             Input(INTERVAL_D, "3", Datum.intervalDaySecond(0, 16, 0, 0, 0, 2, 6)),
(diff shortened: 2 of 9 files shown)


partiql-spi/src/main/java/org/partiql/spi/types/PType.java
```diff
@@ -326,7 +326,7 @@ public static int[] codes() {
     public static final int BIGINT = 5;
 
     /**
-     * PartiQL's big integer type.
+     * PartiQL's numberic type.
```
alancai98:
```suggestion
     * SQL's numeric type.
```

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/DiadicOperator.kt
```diff
@@ -458,24 +541,7 @@ internal abstract class DiadicOperator(
...
-        fillTable(PType.TIMESTAMPZ, PType.INTERVAL_YM, ::getTimestampIntervalInstance) // TODO: Create a specific timezone impl
-        fillTable(PType.TIMESTAMPZ, PType.INTERVAL_DT, ::getTimestampIntervalInstance) // TODO: Create a specific timezone impl
-        fillTable(PType.INTERVAL_YM, PType.DATE, ::getIntervalDateInstance)
-        fillTable(PType.INTERVAL_DT, PType.DATE, ::getIntervalDateInstance)
-        fillTable(PType.INTERVAL_DT, PType.TIME, ::getIntervalTimeInstance)
-        fillTable(PType.INTERVAL_DT, PType.TIMEZ, ::getIntervalTimeInstance) // TODO: Create a specific timezone impl
-        fillTable(PType.INTERVAL_YM, PType.TIMESTAMP, ::getIntervalTimestampInstance)
-        fillTable(PType.INTERVAL_DT, PType.TIMESTAMP, ::getIntervalTimestampInstance)
-        fillTable(PType.INTERVAL_YM, PType.TIMESTAMPZ, ::getIntervalTimestampInstance) // TODO: Create a specific timezone impl
-        fillTable(PType.INTERVAL_DT, PType.TIMESTAMPZ, ::getIntervalTimestampInstance) // TODO: Create a specific timezone impl
```
alancai98, line 478: We're missing `PType.INTERVAL_DT` with `PType.TIMESTAMPZ` in the `fillIntervalTable` function

```
fillTable(PType.INTERVAL_DT, PType.TIMESTAMPZ, ::getIntervalTimestampInstance) // TODO: Create a specific timezone impl
```

CHANGELOG.md
alancai98, line 1: Taking a look at the test failures, you'll need to update the plan typing tests in https://github.com/partiql/partiql-lang-kotlin/blob/dixiao/intervalarithmatic/partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/operator/OpArithmeticTest.kt. In the original `INTERVAL` PR that added + and - support for `INTERVAL`, the `+` and minus tests were broken into a separate testing function -- https://github.com/partiql/partiql-lang-kotlin/pull/1741/files#r1962254559. You'll need to do a similar thing for `expr-40` (for *) and `expr-41` (for /) since `INTERVAL` now supports * and / 

https://github.com/partiql/partiql-lang-kotlin/blob/51e816b9fa1cb2ae802b7688b0f107e69450df36/partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/operator/OpArithmeticTest.kt#L26-L27.

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ rewritten to their query representation
 - CLI printing of negative interval literals
 - Fix alias collision problem when multiple table references need auto-generated aliases
 - Add `ABS` function support on `INTERVAL` field
+- Support `INTERVAL` times and divide with numerics
```
alancai98, line 51: Side-note: PR will need a rebase off of `main` since there's a conflict in the CHANGELOG.

partiql-spi/src/main/java/org/partiql/spi/types/PType.java
xd1313113: fixed

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/DiadicOperator.kt
xd1313113: fixed

CHANGELOG.md
xd1313113: merged

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    fun intervalMultiply(interval: PType): ((Datum, BigDecimal) -> Datum)? {
+        return when (interval.intervalCode) {
+            IntervalCode.YEAR,
+            IntervalCode.MONTH,
+            IntervalCode.YEAR_MONTH -> { i, number ->
+                val totalMonths = i.years * MONTHS_PER_YEAR + i.months
+                // Double is good to avoid calculation inaccuracy as we truncate fraction part
```
alancai98: Why are we converting to a double?

partiql-eval/src/test/kotlin/org/partiql/eval/internal/IntervalDivideTests.kt
```diff
@@ -0,0 +1,89 @@
...
+        private class Input(
+            val arg0: String,
+            val arg1: String,
+            val expected: Datum
+        )
+
+        @JvmStatic
+        fun intervalDivideNumberCases() = listOf(
+            // Input(INTERVAL_Y, "3", Datum.intervalYearMonth(1, 0, 2)),
+            // Input(INTERVAL_M, "3", Datum.intervalYearMonth(0, 2, 2)),
```
alancai98: Any reason these tests are commented out?

partiql-spi/src/main/java/org/partiql/spi/types/PType.java
```diff
@@ -326,7 +326,7 @@ public static int[] codes() {
     public static final int BIGINT = 5;
 
     /**
-     * PartiQL's big integer type.
+     * SQL's numberic type.
```
alancai98: This still has a typo

```suggestion
     * SQL's numeric type.
```

partiql-eval/src/test/kotlin/org/partiql/eval/internal/IntervalTimesTests.kt
```diff
@@ -0,0 +1,125 @@
...
+        private const val INTERVAL_YM_LARGE = "INTERVAL '111111-1' YEAR(6) TO MONTH"
+        private const val INTERVAL_D = "INTERVAL '2' DAY"
+        private const val INTERVAL_H = "INTERVAL '4' HOUR"
+        private const val INTERVAL_MIN = "INTERVAL '5' MINUTE"
+        private const val INTERVAL_S = "INTERVAL '10.5' SECOND"
+        private const val INTERVAL_DTS = "INTERVAL '2 4:5:10.5' DAY TO SECOND"
+        private const val INTERVAL_DTS_LARGE = "INTERVAL '111111 1:1:1.111111' DAY(6) TO SECOND(6)"
+        private const val NANOS_PER_SECOND = 1_000_000_000
+
+        private class Input(
```
alancai98, line 79: Can we also include tests for `null` and `missing` arguments?
- NULL * INTERVAL -> NULL
- MISSING * INTERVAL -> MISSING
- INTERVAL * NULL -> NULL
- INTERVAL * MISSING -> MISSING
- INTERVAL / NULL -> NULL
- INTERVAL / MISSING -> MISSING

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
```diff
@@ -0,0 +1,119 @@
...
+import java.math.RoundingMode
+
+internal object IntervalUtils {
+    private const val MONTHS_PER_YEAR = 12
+    private const val SECONDS_PER_MINUTE = 60
+    private const val MINUTES_PER_HOUR = 60
+    private const val HOURS_PER_DAY = 24
+    private const val SECONDS_PER_HOUR = SECONDS_PER_MINUTE * MINUTES_PER_HOUR
+    private const val SECONDS_PER_DAY = SECONDS_PER_HOUR * HOURS_PER_DAY
+    private const val NANOS_PER_SECOND = 1_000_000_000
```
alancai98: nit: these constants are converted to longs in the code. rather than needing to do that conversion everywhere these are used, you can define a Kotlin long literal w/ an `L` suffix

https://kotlinlang.org/docs/numbers.html#integer-types

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
```diff
@@ -0,0 +1,119 @@
...
+
+    fun intervalDivide(interval: PType): ((Datum, BigDecimal) -> Datum)? {
```
alancai98: For this function and `intervalMultiply`, it may be helpful to provide more of the thought process on what conversions you're doing and any assumptions you're making (e.g. related to truncation vs rounding) since it's not as simple as `return interval * number`. As it's currently implemented, there are a lot of number conversions that I don't quite understand.

Also what does a null return value for these functions represent?

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
```diff
@@ -0,0 +1,119 @@
...
+        return when (interval.intervalCode) {
+            IntervalCode.YEAR,
+            IntervalCode.MONTH,
+            IntervalCode.YEAR_MONTH -> { i, number ->
+                val totalMonths = i.years * MONTHS_PER_YEAR + i.months
+                // Double is good to avoid calculation inaccuracy as we truncate fraction part
```
alancai98: There's no double in this computation?

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
```diff
@@ -0,0 +1,119 @@
...
+                val totalMonthsInDecimal = BigDecimal(totalMonths).divide(number, RoundingMode.DOWN)
```
alancai98: Any reason `RoundingMode.DOWN` is chosen here? From my read of the SQL spec, it's implementation-defined whether we round or truncate. Regardless of what we choose, I think it should follow the behavior of other parts of the codebase. Let's use this constant: https://github.com/partiql/partiql-lang-kotlin/blob/52000305a625b6e2613728daeb7a40d499aa7e80/partiql-spi/src/main/kotlin/org/partiql/spi/utils/NumberUtils.kt#L28.

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
```diff
@@ -0,0 +1,119 @@
...
+            IntervalCode.DAY_MINUTE,
+            IntervalCode.DAY_SECOND,
+            IntervalCode.HOUR,
+            IntervalCode.HOUR_MINUTE,
+            IntervalCode.HOUR_SECOND,
+            IntervalCode.MINUTE,
+            IntervalCode.MINUTE_SECOND,
+            IntervalCode.SECOND -> { i, number ->
+                val totalSecondsInBigDecimal = toSeconds(i)
+                val resultInBigDecimal = totalSecondsInBigDecimal.divide(number)
```
alancai98: BigDecimal divide without a rounding mode specified will result in a runtime error if the result cannot be represented in a decimal:

```
partiql ▶ INTERVAL '1' SECOND (9) / 3;

e: [runtime] Unexpected failure encountered. Caused by: java.lang.ArithmeticException: Non-terminating decimal expansion; no exact representable decimal result..
java.lang.ArithmeticException: Non-terminating decimal expansion; no exact representable decimal result.
        at java.base/java.math.BigDecimal.divide(BigDecimal.java:1780)
        at org.partiql.spi.utils.IntervalUtils$intervalDivide$2.invoke(IntervalUtils.kt:43)
```

Let's change it to use the constant defined in `NumberUtils` -- https://github.com/partiql/partiql-lang-kotlin/blob/52000305a625b6e2613728daeb7a40d499aa7e80/partiql-spi/src/main/kotlin/org/partiql/spi/utils/NumberUtils.kt#L28.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTimes.kt
```diff
@@ -116,4 +119,22 @@ internal object FnTimes : DiadicArithmeticOperator("times") {
...
+            op(interval, BigDecimal(number.toString()))
+        }
+    }
+
+    override fun getNumberIntervalInstance(lhs: PType, rhs: PType): Fn? {
+        val op = IntervalUtils.intervalMultiply(rhs) ?: return null
+        return basic(rhs, lhs, rhs) { args ->
+            val number = args[0].getNumber()
+            val interval = args[1]
+            op(interval, BigDecimal(number.toString()))
```
alancai98: Let's use our `BigDecimal` helper function to construct rather than converting toString first -- https://github.com/partiql/partiql-lang-kotlin/blob/52000305a625b6e2613728daeb7a40d499aa7e80/partiql-spi/src/main/kotlin/org/partiql/spi/utils/NumberUtils.kt#L34.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTimes.kt
```diff
@@ -116,4 +119,22 @@ internal object FnTimes : DiadicArithmeticOperator("times") {
             Datum.doublePrecision(arg0 * arg1)
         }
     }
+
+    override fun getIntervalNumberInstance(lhs: PType, rhs: PType): Fn? {
+        val op = IntervalUtils.intervalMultiply(lhs) ?: return null
```
alancai98: Why are we returning `null` here?

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnDivide.kt
```diff
@@ -129,4 +132,17 @@ internal object FnDivide : DiadicArithmeticOperator("divide") {
...
+    override fun getIntervalNumberInstance(lhs: PType, rhs: PType): Fn? {
+        val op = IntervalUtils.intervalDivide(lhs) ?: return null
+        return basic(lhs, lhs, rhs) { args ->
+            val interval = args[0]
+            val number = args[1].getNumber()
+            if (number.isZero()) {
+                throw PErrors.divisionByZeroException(interval, lhs)
+            }
+
+            op(interval, BigDecimal(number.toString()))
```
alancai98: Similar comments as the multiple equivalents:
- let's use the `BigDecimal` constructor we've defined in `NumberUtils` -- https://github.com/partiql/partiql-lang-kotlin/blob/52000305a625b6e2613728daeb7a40d499aa7e80/partiql-spi/src/main/kotlin/org/partiql/spi/utils/NumberUtils.kt#L34
- change return type to not be nullable in the override

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
```diff
@@ -0,0 +1,119 @@
...
+            IntervalCode.HOUR_SECOND,
+            IntervalCode.MINUTE,
+            IntervalCode.MINUTE_SECOND,
+            IntervalCode.SECOND -> { i, number ->
+                val totalSecondsInBigDecimal = toSeconds(i)
+                val resultInBigDecimal = totalSecondsInBigDecimal * number
+                fromSecond(resultInBigDecimal, interval.precision, interval.fractionalPrecision)
+            }
+
+            else -> { num, i -> Datum.missing() }
```
alancai98: We should throw an error here rather than silently returning missing. Similarly for divide.

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
```diff
@@ -0,0 +1,119 @@
...
+            val totalSeconds = daysInSeconds + hoursInSeconds + minutesInSeconds + i.seconds
+            return BigDecimal.valueOf(totalSeconds).add(BigDecimal.valueOf(i.nanos.toLong(), 9))
+        } else {
+            throw UnsupportedOperationException("Unable to convert non DayToSeconds type")
+        }
+    }
+
+    private fun fromSecond(totalSeconds: BigDecimal, precision: Int, fractionalPrecision: Int): Datum {
+        val resultTotalNanos: BigInteger = totalSeconds.movePointRight(9).toBigIntegerExact()
+        val divRem = resultTotalNanos.divideAndRemainder(BigInteger.valueOf(NANOS_PER_SECOND.toLong()))
```
alancai98: Perhaps you can directly call `BigDecimal`'s `divideAndRemainder` [function](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#divideAndRemainder-java.math.BigDecimal-java.math.MathContext-) rather than performing these conversions to `BigInteger`?

partiql-eval/src/test/kotlin/org/partiql/eval/internal/IntervalDivideTests.kt
```diff
@@ -0,0 +1,89 @@
...
+            Input(INTERVAL_D, "-3", Datum.intervalDaySecond(0, -16, 0, 0, 0, 2, 6)),
+            Input(INTERVAL_H, "-3", Datum.intervalDaySecond(0, -1, -20, 0, 0, 2, 6)),
+            Input(INTERVAL_MIN, "-3", Datum.intervalDaySecond(0, 0, -1, -40, 0, 2, 6)),
+            Input(INTERVAL_S, "-3", Datum.intervalDaySecond(0, 0, 0, -3, (-0.5 * NANOS_PER_SECOND).toInt(), 2, 6)),
+            Input(INTERVAL_DTS, "-3", Datum.intervalDaySecond(0, -17, -21, -43, (-0.5 * NANOS_PER_SECOND).toInt(), 2, 6))
+        ).map { case ->
+            SuccessTestCase("${case.arg0} / ${case.arg1}", case.expected)
+        }
+
+        @JvmStatic
```
alancai98, line 108: We should also test some other cases:
- Interval multiplied/divided by a floating point value
- Interval where a range is expanded beyond the interval precision argument's precision. E.g. `INTERVAL '10' YEAR * 10`. These cases are currently expected to fail which we should track as the current behavior. Feel free to also link https://github.com/partiql/partiql-lang-kotlin/issues/1784 and https://github.com/partiql/partiql-lang/issues/100.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/operator/OpArithmeticTest.kt
```diff
@@ -142,4 +160,87 @@ class OpArithmeticTest : PartiQLTyperTestBase() {
...
+                val arg1 = args[1]
+                val output = when {
+                    arg0 in allNumberPType && arg1 in allIntervalYMType -> PType.intervalYearMonth(9)
+                    arg0 in allNumberPType && arg1 in allIntervalDTType -> PType.intervalDaySecond(9, 6)
+
+                    arg0 in allIntervalYMType && arg1 in allNumberPType -> PType.intervalYearMonth(9)
+                    arg0 in allIntervalDTType && arg1 in allNumberPType -> PType.intervalDaySecond(9, 6)
+                    arg0 == arg1 -> arg1
+                    // TODO arg0 == StaticType.DECIMAL && arg1 == StaticType.FLOAT -> arg1 // TODO: The cast table is wrong. Honestly, it should be deleted.
+                    // TODO arg1 == StaticType.DECIMAL && arg0 == StaticType.FLOAT -> arg0 // TODO: The cast table is wrong
```
alancai98: This TODO doesn't apply for this test. Can be deleted. Similarly for divide.

alancai98 (commented):
This feature is definitely more complicated than I had initially anticipated. Think we're on the right track in this PR. Let's add some more tests and make it clearer how we're thinking about this computation since the SQL definition is a bit obfuscated.

partiql-eval/src/test/kotlin/org/partiql/eval/internal/IntervalDivideTests.kt
xd1313113: Add test to cover both cases

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
xd1313113: It is not supported

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTimes.kt
xd1313113: Remove the null return

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
xd1313113: I choose truncation to handling the interval calculation, RoundingMode.DOWN behavior same as truncate.

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
xd1313113: I choose truncation to handling the interval calculation, RoundingMode.DOWN behavior same as truncate.

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
xd1313113: I got a different error which does not reach decimal calculaltion. Let us sync tomorrow.

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
xd1313113: removed stale comment

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
xd1313113: added more comments

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTimes.kt
```diff
@@ -116,4 +119,22 @@ internal object FnTimes : DiadicArithmeticOperator("times") {
...
+    override fun getIntervalNumberInstance(lhs: PType, rhs: PType): Fn? {
+        val op = IntervalUtils.intervalMultiply(lhs)
+        return basic(lhs, lhs, rhs) { args ->
+            val interval = args[0]
+            val number = args[1].getNumber()
+            op(interval, NumberUtils.bigDecimalOf(number))
+        }
+    }
+
+    override fun getNumberIntervalInstance(lhs: PType, rhs: PType): Fn? {
```
alancai98: These do not need to be nullable https://kotlinlang.org/docs/null-safety.html

```suggestion
    override fun getIntervalNumberInstance(lhs: PType, rhs: PType): Fn {
        val op = IntervalUtils.intervalMultiply(lhs)
        return basic(lhs, lhs, rhs) { args ->
            val interval = args[0]
            val number = args[1].getNumber()
            op(interval, NumberUtils.bigDecimalOf(number))
        }
    }

    override fun getNumberIntervalInstance(lhs: PType, rhs: PType): Fn {
```

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnDivide.kt
```diff
@@ -129,4 +132,17 @@ internal object FnDivide : DiadicArithmeticOperator("divide") {
             Datum.doublePrecision(arg0 / arg1)
         }
     }
+
+    override fun getIntervalNumberInstance(lhs: PType, rhs: PType): Fn? {
```
alancai98: Similarly here. Does not need to be nullable

```suggestion
    override fun getIntervalNumberInstance(lhs: PType, rhs: PType): Fn {
```

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
alancai98: > I choose truncation to handling the interval calculation, RoundingMode.DOWN behavior same as truncate.

I understand that you chose truncation for the interval calculation. My original question was about what led you to choose truncation compared to rounding? Is there a part of the SQL spec that led you to choose this as the default? Or some other database implementation that we're following here? My read of the SQL spec here is that it's implementation-defined.

If it's implementation-defined, we should call out that we chose a certain behavior (either truncation or rounding). May also be helpful to add a TODO to enable future configuration.

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
```diff
@@ -0,0 +1,140 @@
...
+    fun intervalDivide(interval: PType): ((Datum, BigDecimal) -> Datum) {
+        return when (interval.intervalCode) {
+            IntervalCode.YEAR,
+            IntervalCode.MONTH,
+            IntervalCode.YEAR_MONTH -> { i, number ->
+                // get total months from the interval as long as it can hold the interval with precision = 9.
+                val totalMonths: Long = i.years * MONTHS_PER_YEAR + i.months
+                // use big decimal to do the calculation as we can control the round mode.
+                // RoundingMode.DOWN mode is used to truncate fraction part e.g. 2.9 months will be 2 months.
+                val totalMonthsInDecimal = BigDecimal(totalMonths).divide(number, 0, RoundingMode.DOWN)
```
alancai98: Let's define a reusable constant for the rounding mode so that it can be used for both `divide` calls. Doing so can ensure consistency across our BigDecimal operations and allow us to more easily change the behavior (e.g. if we support future configurability of the rounding behavior).

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
alancai98: Which error were you getting?

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
alancai98: Can you clarify that it's not supported? `BigDecimal`'s `divideAndRemainder` has been in Java since 1.5 -- https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#divideAndRemainder-java.math.BigDecimal-java.math.MathContext-

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
```diff
@@ -0,0 +1,140 @@
...
+    private const val MINUTES_PER_HOUR = 60L
+    private const val HOURS_PER_DAY = 24L
+    private const val SECONDS_PER_HOUR = SECONDS_PER_MINUTE * MINUTES_PER_HOUR
+    private const val SECONDS_PER_DAY = SECONDS_PER_HOUR * HOURS_PER_DAY
+    private const val NANOS_PER_SECOND = 1_000_000_000L
+
+    /**
+     * Divides an interval by a numeric value.
+     * For YEAR_MONTH Interval, We maintain the integer part (precision 0) and truncate the floating part
+     * For DAY_SECOND Interval, We maintain the precision of 9 for nanoseconds and truncate the additional part
```
alancai98: I don't quite understand what these comments are stating (similarly for multiply). I think it would be helpful to state something along the lines of "for year-month intervals, we convert the year and month components into the total number of months and then ... \<state what you do for the division operator\>". Similarly for the day-second interval.

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
```diff
@@ -0,0 +1,140 @@
...
+     *
+     * @param interval The interval type to divide
+     * @return A function that takes an interval Datum and BigDecimal divisor, returns the divided interval
+     */
+    fun intervalDivide(interval: PType): ((Datum, BigDecimal) -> Datum) {
+        return when (interval.intervalCode) {
+            IntervalCode.YEAR,
+            IntervalCode.MONTH,
+            IntervalCode.YEAR_MONTH -> { i, number ->
+                // get total months from the interval as long as it can hold the interval with precision = 9.
```
alancai98: What is this comment saying w/ the precision = 9?

partiql-eval/src/test/kotlin/org/partiql/eval/internal/IntervalDivideTests.kt
alancai98: We're still missing some tests for the multiply/divided by floating point value. We already have tests for integer and decimal but not floating point. The test should be on a DatumFloat/DatumDouble -- https://github.com/partiql/partiql-lang-kotlin/blob/6eadf1301ab0589c041458c6bfca567f88e8c2ae/partiql-spi/src/main/java/org/partiql/spi/value/DatumDouble.java.

Also, can you please also link https://github.com/partiql/partiql-lang-kotlin/issues/1784 and https://github.com/partiql/partiql-lang/issues/100 for the precision overflow tests.

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
xd1313113: Sorry, I misunderstand your question, the primary intent is to calculate integer second and  integer nano second by moving the point in one divide call. FYI, I followed the implementation of java line 1003 
https://github.com/frohoff/jdk8u-jdk/blob/master/src/share/classes/java/time/Duration.java#L991

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTimes.kt
xd1313113: Thanks. fixed. it is interesting that the override can have different function signature in Kotlin.

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
xd1313113: Sorry, I did not make it clear. I tested a few DBs and eventually follow the ORACLE implementation, which looks sticks to SQL specification. Let me know where I should document this. 

Also it is true that we should investigate the current arithmetic behavior of interval  and date calculation and make them consistent.

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
xd1313113: I rephrased and make it clearly that the max precision of year is 9

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
xd1313113: Yeah, it is confusing. Removing those function comment. Let us focus on the comment inside code.

partiql-eval/src/test/kotlin/org/partiql/eval/internal/IntervalDivideTests.kt
xd1313113: I see, You are asking to test Datum.float/double. Not sure how to initialize in test code. Will sync you offline.

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
```diff
@@ -0,0 +1,159 @@
...
+     * @return A function that takes an interval Datum and BigDecimal divisor, returns the divided interval
+     */
+    fun intervalDivide(interval: PType): ((Datum, Number) -> Datum) {
+        return when (interval.intervalCode) {
+            IntervalCode.YEAR,
+            IntervalCode.MONTH,
+            IntervalCode.YEAR_MONTH -> { i, number ->
+                // get total months from the interval as Long Type as it can hold the interval with max precision = 9.
+                val totalMonths: Long = i.years * MONTHS_PER_YEAR + i.months
+                val totalMonthsInDouble = totalMonths / number.toDouble()
```
alancai98: I think we should only keep the `number` argument a double if it's already an approximate number (float/double). We shouldn't convert it to double if it's already an exact number (int, decimal).

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
```diff
@@ -0,0 +1,159 @@
...
+     * @return A function that takes an interval Datum and BigDecimal multiplier, returns the multiplied interval
+     */
+    fun intervalMultiply(interval: PType): ((Datum, Number) -> Datum) {
+        return when (interval.intervalCode) {
+            IntervalCode.YEAR,
+            IntervalCode.MONTH,
+            IntervalCode.YEAR_MONTH -> { i, number ->
+                // get total months from the interval as Long Type as it can hold the interval with max precision = 9.
+                val totalMonths: Long = i.years * MONTHS_PER_YEAR + i.months
+                val totalMonthsInDouble = totalMonths * number.toDouble()
```
alancai98: Similarly here. We should not convert exact number representations to approximate if `number` is already exact (int, decimal, etc).

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
```diff
@@ -0,0 +1,159 @@
...
+            IntervalCode.MINUTE_SECOND,
+            IntervalCode.HOUR_SECOND,
+            IntervalCode.DAY_SECOND -> interval.fractionalPrecision
+            IntervalCode.DAY,
+            IntervalCode.DAY_HOUR,
+            IntervalCode.DAY_MINUTE,
+            IntervalCode.HOUR,
+            IntervalCode.HOUR_MINUTE,
+            IntervalCode.MINUTE -> INTERVAL_DEFAULT_FRACTIONAL_PRECISION
+            else -> throw IllegalArgumentException("Cannot get fraction precision for Non-INTERVAL_DT type")
```
alancai98:
```suggestion
            else -> throw IllegalArgumentException("Cannot get fractional precision for non-INTERVAL_DT type")
```

partiql-spi/src/main/kotlin/org/partiql/spi/utils/IntervalUtils.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    private fun toSeconds(i: Datum): BigDecimal {
+        if (i.type.code() == PType.INTERVAL_DT) {
+            val daysInSeconds = i.days * SECONDS_PER_DAY
+            val hoursInSeconds = i.hours * SECONDS_PER_HOUR
+            val minutesInSeconds = i.minutes * SECONDS_PER_MINUTE
+            val totalSeconds = daysInSeconds + hoursInSeconds + minutesInSeconds + i.seconds
+            return BigDecimal.valueOf(totalSeconds).add(BigDecimal.valueOf(i.nanos.toLong(), NANO_MAX_PRECISION))
+        } else {
+            throw UnsupportedOperationException("Unable to convert non DayToSeconds type to seconds")
```
alancai98:
```suggestion
            throw UnsupportedOperationException("Unable to convert non-INTERVAL_DT type to seconds")
```

partiql-eval/src/test/kotlin/org/partiql/eval/internal/IntervalDivideTests.kt
alancai98: Can you please also link https://github.com/partiql/partiql-lang-kotlin/issues/1784 and https://github.com/partiql/partiql-lang/issues/100 for the precision overflow tests for divide and multiply? Linking the issues can give more clarity on if the behavior is subject to change in the future.

alancai98 (commented):
Some last comments. Please also reference https://github.com/partiql/partiql-lang-kotlin/issues/1784 and https://github.com/partiql/partiql-lang/issues/100 in the divide and multiply tests.

alancai98 (approved):

Issue #183, opened 2020-02-13
Add support for `OVERLAP` Predicate


therapon:
Add support for `OVERLAP` Predicate

PR #1794, opened 2025-07-29
Add support for `OVERLAPS` predicate 

XuechunHHH:
## Relevant Issues
- [Closes/Related To] Issue #183 

## Description
- Add support for `OVERLAPS` predicate
- Add corresponding conformance tests

## Other Information

## License Information

### DIFF
--- a/partiql-ast/api/partiql-ast.api
+++ b/partiql-ast/api/partiql-ast.api
@@ -67,6 +67,7 @@ public final class org/partiql/ast/Ast {
 	public static final fun exprOperator (Ljava/lang/String;Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/expr/ExprOperator;
 	public static synthetic fun exprOperator$default (Ljava/lang/String;Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;ILjava/lang/Object;)Lorg/partiql/ast/expr/ExprOperator;
 	public static final fun exprOr (Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/expr/ExprOr;
+	public static final fun exprOverlaps (Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/expr/ExprOverlaps;
 	public static final fun exprOverlay (Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/expr/ExprOverlay;
 	public static final fun exprOverlay (Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/expr/ExprOverlay;
 	public static synthetic fun exprOverlay$default (Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;ILjava/lang/Object;)Lorg/partiql/ast/expr/ExprOverlay;
@@ -315,6 +316,8 @@ public abstract class org/partiql/ast/AstRewriter : org/partiql/ast/AstVisitor {
 	public fun visitExprOperator (Lorg/partiql/ast/expr/ExprOperator;Ljava/lang/Object;)Lorg/partiql/ast/AstNode;
 	public synthetic fun visitExprOr (Lorg/partiql/ast/expr/ExprOr;Ljava/lang/Object;)Ljava/lang/Object;
 	public fun visitExprOr (Lorg/partiql/ast/expr/ExprOr;Ljava/lang/Object;)Lorg/partiql/ast/AstNode;
+	public synthetic fun visitExprOverlaps (Lorg/partiql/ast/expr/ExprOverlaps;Ljava/lang/Object;)Ljava/lang/Object;
+	public fun visitExprOverlaps (Lorg/partiql/ast/expr/ExprOverlaps;Ljava/lang/Object;)Lorg/partiql/ast/AstNode;
 	public synthetic fun visitExprOverlay (Lorg/partiql/ast/expr/ExprOverlay;Ljava/lang/Object;)Ljava/lang/Object;
 	public fun visitExprOverlay (Lorg/partiql/ast/expr/ExprOverlay;Ljava/lang/Object;)Lorg/partiql/ast/AstNode;
 	public synthetic fun visitExprParameter (Lorg/partiql/ast/expr/ExprParameter;Ljava/lang/Object;)Ljava/lang/Object;
@@ -496,6 +499,7 @@ public abstract class org/partiql/ast/AstVisitor {
 	public fun visitExprNullPredicate (Lorg/partiql/ast/expr/ExprNullPredicate;Ljava/lang/Object;)Ljava/lang/Object;
 	public fun visitExprOperator (Lorg/partiql/ast/expr/ExprOperator;Ljava/lang/Object;)Ljava/lang/Object;
 	public fun visitExprOr (Lorg/partiql/ast/expr/ExprOr;Ljava/lang/Object;)Ljava/lang/Object;
+	public fun visitExprOverlaps (Lorg/partiql/ast/expr/ExprOverlaps;Ljava/lang/Object;)Ljava/lang/Object;
 	public fun visitExprOverlay (Lorg/partiql/ast/expr/ExprOverlay;Ljava/lang/Object;)Ljava/lang/Object;
 	public fun visitExprParameter (Lorg/partiql/ast/expr/ExprParameter;Ljava/lang/Object;)Ljava/lang/Object;
 	public fun visitExprPath (Lorg/partiql/ast/expr/ExprPath;Ljava/lang/Object;)Ljava/lang/Object;
@@ -2450,6 +2454,24 @@ public class org/partiql/ast/expr/ExprOr$Builder {
 	public fun build ()Lorg/partiql/ast/expr/ExprOr;
 	public fun lhs (Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/expr/ExprOr$Builder;
 	public fun rhs (Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/expr/ExprOr$Builder;
+	public fun toString ()Ljava/lang/String;
+}
+
+public final class org/partiql/ast/expr/ExprOverlaps : org/partiql/ast/expr/Expr {
+	public fun <init> (Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;)V
+	public fun accept (Lorg/partiql/ast/AstVisitor;Ljava/lang/Object;)Ljava/lang/Object;
+	public static fun builder ()Lorg/partiql/ast/expr/ExprOverlaps$Builder;
+	public fun equals (Ljava/lang/Object;)Z
+	public fun getChildren ()Ljava/util/List;
+	public fun getLhs ()Lorg/partiql/ast/expr/Expr;
+	public fun getRhs ()Lorg/partiql/ast/expr/Expr;
+	public fun hashCode ()I
+}
+
+public class org/partiql/ast/expr/ExprOverlaps$Builder {
+	public fun build ()Lorg/partiql/ast/expr/ExprOverlaps;
+	public fun lhs (Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/expr/ExprOverlaps$Builder;
+	public fun rhs (Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/expr/ExprOverlaps$Builder;
 	public fun toString ()Ljava/lang/String;
 }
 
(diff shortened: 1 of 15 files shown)


partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
```diff
@@ -0,0 +1,144 @@
...
+        return Datum.comparator().compare(a, b)
+    }
+
+    private fun validateDateTimeComparison(type1: PType, type2: PType) {
+        val code1 = type1.code()
+        val code2 = type2.code()
+        val compatible = when {
+            // DATE types are comparable with DATE types
+            code1 == PType.DATE && code2 == PType.DATE -> true
+            // TODO(): We currently don't support TIMEZ & TIMESTAMPZ
```
XuechunHHH, line 137: There seems to be a mismatch between AST and SPI layer currently. In parser, `DataType.TIME_WITH_TIME_ZONE()` and `DataType.TIMESTAMP_WITH_TIME_ZONE()` are used while in spi we are using `PType.TIMEZ` and `PType.TIMESTAMPZ`. Shall we add the mapping somewhere to actually support time/timestamp with time zone?

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
```diff
@@ -0,0 +1,144 @@
...
+                if (!d1.isNull && !isDateTimeType(d1.type)) {
+                    throw IllegalArgumentException("OVERLAPS predicate requires datetime start values, but got: ${d1.type}")
+                }
+                if (!d2.isNull && !isDateTimeType(d2.type)) {
+                    throw IllegalArgumentException("OVERLAPS predicate requires datetime start values, but got: ${d2.type}")
+                }
+                // If D1 or D2 is null, return null
+                if (d1.isNull || d2.isNull) {
+                    return@instance Datum.nullValue()
+                }
```
alancai98: I think we should put this check after we validate the other argument. The other argument could be invalid and if this argument is `null`, `null` will be returned rather than giving an error:

```
[null, 1] OVERLAPS [DATE '2023-06-01', DATE '2023-08-01'];
```

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
alancai98: Yeah, it's a gap in the existing implementation. Could you cut an issue to track that conversion and link it in the code?

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
```diff
@@ -0,0 +1,144 @@
...
+internal object FnOverlaps : FnOverload() {
+    val name = FunctionUtils.hide("overlaps")
+    override fun getSignature(): RoutineOverloadSignature {
+        return RoutineOverloadSignature(name, listOf(PType.dynamic(), PType.dynamic()))
+    }
+
+    override fun getInstance(args: Array<out PType>): Fn? {
+        val arg0 = args[0]
+        val arg1 = args[1]
+        // Both arguments must be collections (arrays or bags)
```
alancai98: It's an interesting extension to support bags in addition to arrays. My read of the SQL spec is that the ordering of the arguments doesn't matter for OVERLAPS. Though the SQL spec's definition applies to ROW VALUE expressions which are more aligned with arrays/lists.

alancai98 (commented):
Some minor suggestions. Otherwise looks good. Let's also get the conformance test PR and update the git submodule first.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
XuechunHHH: Here is the issue: https://github.com/partiql/partiql-lang-kotlin/issues/1795

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
XuechunHHH: Exactly. That's why I included bag support here. Would this extension be a deviation from the SQL spec? I can restrict it to arrays only if needed for strict compliance.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
XuechunHHH: Aha... I didn't realize the issue here. Thanks for pointing out.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
alancai98: I think it's fine to include and a logical extension to the SQL spec. We already extend some SQL operations like `IN` to support bags.

XuechunHHH: Sounds good. Will keep the extension support.

partiql-ast/api/partiql-ast.api
```diff
@@ -67,6 +67,7 @@ public final class org/partiql/ast/Ast {
 	public static final fun exprOperator (Ljava/lang/String;Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/expr/ExprOperator;
 	public static synthetic fun exprOperator$default (Ljava/lang/String;Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;ILjava/lang/Object;)Lorg/partiql/ast/expr/ExprOperator;
 	public static final fun exprOr (Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/expr/ExprOr;
+	public static final fun exprOverlaps (Lorg/partiql/ast/expr/Expr;Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/expr/ExprOverlaps;
```
xd1313113, line 70: @alancai98 What is this api file used for? How is it generated?

alancai98: These `.api` files are used to track what we expose as part of our public API. This is important to track since we do not want to
- accidentally make something public that should not be public
- modify an existing public API since it could be a backwards-incompatible change.

You can read more about the Kotlin binary compatibility validator here -- https://github.com/Kotlin/binary-compatibility-validator. The gist is that every gradle build will run the `apiCheck` command to check that the `.api` files match what's actually public. To generate a new `.api` file following some public API changes, you run `apiDump`.

alancai98: Another reference for the Kotlin API compatibility validator -- https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#use-the-binary-compatibility-validator.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
```diff
@@ -0,0 +1,224 @@
...
+                val d1 = period1[0]
+                val field1 = period1[1]
+                // D2: value of the first field of <row value expression 2>
+                val d2 = period2[0]
+                val field2 = period2[1]
+                // For the datetime validation:
+                if (!d1.isNull && !d1.isMissing && !isDateTimeType(d1.type)) {
+                    throw PErrors.unexpectedTypeException(d1.type, listOf(PType.date(), PType.time(), PType.timestamp()))
+                }
+                if (!d2.isNull && !d2.isMissing && !isDateTimeType(d2.type)) {
```
alancai98: nit: checking if a datum is null or missing is a common enough check that perhaps we should add a function to `FunctionUtils` called `isUnknown`

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
```diff
@@ -0,0 +1,224 @@
...
+            aIsUnknown && b.getBoolean() -> Datum.bool(true)
+            bIsUnknown && a.getBoolean() -> Datum.bool(true)
+            else -> Datum.nullValue(PType.bool())
+        }
+    }
+
+    private fun logicalNot(a: Datum): Datum {
+        if (a.isNull || a.isMissing) return Datum.nullValue(PType.bool())
+        return Datum.bool(!a.boolean)
+    }
```
alancai98: These are similar enough to `FnAnd`, `FnOr`, and `FnNot` that I think we can move them to `FunctionUtils` and have `FnOverlaps` as well as `FnAnd`, `FnOr`, and `FnNot` use the internal functions.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
```diff
@@ -0,0 +1,224 @@
...
+
+    private fun addInterval(datetime: Datum, interval: Datum): Datum? {
+        return try {
+            val plusFn = FnPlus.getInstance(arrayOf(datetime.type, interval.type)) ?: return null
+            val result = plusFn.invoke(arrayOf(datetime, interval))
+            if (result.isNull) null else result
+        } catch (_: Exception) {
+            null
+        }
+    }
```
alancai98: We seem to be using `null` to represent some invalid computation here. I think we should be throwing an error when we try to get a `Fn` from `FnPlus` to add the interval to the datetime and similarly for the `Exception` that's getting caught. If we can remove the `null` return type for this function, then we can remove the nullable arguments + return types for a lot of other places such as `calculateEndpoint` and `compareValues`.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
```diff
@@ -0,0 +1,224 @@
...
+
+    private fun logicalOr(a: Datum, b: Datum): Datum {
+        val aIsUnknown = a.isNull || a.isMissing
+        val bIsUnknown = b.isNull || b.isMissing
+        return when {
+            aIsUnknown && bIsUnknown -> Datum.nullValue(PType.bool())
+            !aIsUnknown && !bIsUnknown -> Datum.bool(a.getBoolean() || b.getBoolean())
+            aIsUnknown && b.getBoolean() -> Datum.bool(true)
+            bIsUnknown && a.getBoolean() -> Datum.bool(true)
```
alancai98: nit: for these calls, to `getBoolean` can just use the property access syntax `.boolean`. Applies to some other places too

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
```diff
@@ -0,0 +1,224 @@
...
+                val (s2, t2) = if (d2.isNull || d2.isMissing) {
+                    Pair(e2, d2)
+                } else {
+                    val cmp = compareValues(e2, d2)
+                    if (cmp != null && !cmp.isNull && !cmp.isMissing && cmp.int < 0) {
+                        Pair(e2, d2)
+                    } else {
+                        Pair(d2, e2)
+                    }
+                }
```
alancai98: nit: the same logic is repeated here. perhaps can refactor to a separate function.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
```diff
@@ -0,0 +1,224 @@
...
+                // D1: value of the first field of <row value expression 1>
+                val d1 = period1[0]
+                val field1 = period1[1]
+                // D2: value of the first field of <row value expression 2>
+                val d2 = period2[0]
+                val field2 = period2[1]
+                // For the datetime validation:
+                if (!d1.isNull && !d1.isMissing && !isDateTimeType(d1.type)) {
+                    throw PErrors.unexpectedTypeException(d1.type, listOf(PType.date(), PType.time(), PType.timestamp()))
```
alancai98: For the list of possible types, we can also include PType.timez and PType.timestampz. Also for the field1/field2 list.

Let's also define a variable for the d1/d2 possible type list and a variable for field1/field2 such that we don't need as much repetition.

test/partiql-tests
alancai98, line 1: Now that https://github.com/partiql/partiql-tests/pull/149 is merged in, let's update this submodule

alancai98 (commented):
looks good overall. some minor suggestions for factoring that should cleanup the implementation a bit

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
XuechunHHH: Thanks for the heads-up.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
XuechunHHH: That makes sense. Will refactor the code.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnOverlaps.kt
XuechunHHH: Aha, that's a clever move!

XuechunHHH:
> looks good overall. some minor suggestions for factoring that should cleanup the implementation a bit

Thanks for the comprehensive review and detailed comments. Refactor the code accordingly.

alancai98 (approved):
new factoring looks good. thanks for addressing the comments. think we need to do one last rebase/merge from `main` since there's a conflict

Issue #1796, opened 2025-08-01
`<unknown value> AND FALSE` returns the incorrect result in permissive mode and errors in strict mode


alancai98:
## Description
`MISSING AND FALSE` and `NULL AND FALSE` returns the incorrect result in permissive mode and errors in strict mode

## To Reproduce
In permissive mode
```
Version: 1.2.3-SNAPSHOT-52000305a

partiql ▶ FALSE AND MISSING;

=== RESULT ===
false

OK!
partiql ▶ MISSING AND FALSE;

=== RESULT ===
missing

OK!
```

In strict mode
```
Version: 1.2.3-SNAPSHOT-52000305a

partiql ▶ FALSE AND MISSING;

=== RESULT ===
false

OK!
partiql ▶ MISSING AND FALSE;

e: [runtime] Unexpected failure encountered. Caused by: org.partiql.spi.value.InvalidOperationException: Operation "getBoolean" is not valid for type UNKNOWN..
org.partiql.spi.value.InvalidOperationException: Operation "getBoolean" is not valid for type UNKNOWN.
```

## Expected Behavior
- The above cases should all return `false`. Seems like we're missing a branch in the `FnAnd` implementation

## Additional Context
- Java version: 17
- PartiQL version: 1.2.3-SNAPSHOT
- Add any other context about the problem here.

PR #1797, opened 2025-08-01
fix: logical AND with an unknown value and false

alancai98:
## Relevant Issues
- Closes https://github.com/partiql/partiql-lang-kotlin/issues/1796

## Description
- We were missing a branch in the implementation of `AND`. This resulted in the wrong result for a query such as `NULL AND FALSE` and `MISSING AND FALSE`.

## Other Information

## License Information


test/partiql-tests
alancai98, line 1: self-review: this will be updated after https://github.com/partiql/partiql-tests/pull/150 is merged in

xd1313113: Curious why tests for operator AND is organized in partiql-tests instead of kotlin projects?

xd1313113: And curious why do we define
AndNULLTRUE -> NULL 
AndNULLFALSE -> FALSE

xd1313113 (approved):

test/partiql-tests
alancai98: > Curious why tests for operator AND is organized in partiql-tests instead of kotlin projects?

`partiql-tests` are the conformance tests that can be used independently of the implementation programming language. So we can use those tests in both our Kotlin/Java and Rust implementations (and any future implementations). If we had the tests hard-coded in our Kotlin project, it would be much trickier to share tests and ensure the same behavior across both implementations.

It's preferred, where possible, to add parsing and evaluation tests in the conformance tests. We can't directly some tests such as `INTERVAL` since we have not yet defined the `INTERVAL` data type in the test data model yet (see https://github.com/partiql/partiql-tests/issues/145).

alancai98: > And curious why do we define AndNULLTRUE -> NULL AndNULLFALSE -> FALSE

This is defined in [section 6.30](https://web.cecs.pdx.edu/~len/sql1999.pdf#page=244) of the SQL1999 spec.

XuechunHHH: [partiql-tests](https://github.com/partiql/partiql-tests) contains conformance tests for PLK. That way we can measure conformance across both our Kotlin and Rust implementations. 

For `AND` rules, it is defined in SQL spec: https://web.cecs.pdx.edu/~len/sql1999.pdf#page=342

PR #1798, opened 2025-08-01
Add support to read partiQL env file as data source to CLI

xd1313113:
## Relevant Issues

## Description
- < Explain >

## Other Information
- Updated Unreleased Section in CHANGELOG: **[YES/NO]**: <Explain if NO>
- Any backward-incompatible changes? **[YES/NO]**: <Explain if YES>
- Any new external dependencies? **[YES/NO]**: <Explain if YES>
- Do your changes comply with the [contributing][cg] and [code style][csg] guidelines? **[YES/NO]**

## License Information

### DIFF
--- a/partiql-cli/README.adoc
+++ b/partiql-cli/README.adoc
@@ -13,19 +13,20 @@ partiql                     # No arguments
 partiql -d=/path/to/db      # Launch an interactive shell for the database in /path/to/db
 partiql sample.ion    # Load sample.ion into the database environment as the binding `stdin`
 partiql sample.csv    # '' format derived from extension
+partiql -e sample.env    # Load partiQL literal data into partiQL shell
 
 # stdin
 cat sample.ion | partiql -f ion    # Load from stdin with the specified input format (-f) Ion.
 
 # program string
-partiql -q 'SELECT * FROM stdin' sample.ion
-cat sample.ion | partiql -f ion -q 'SELECT * FROM stdin'
+partiql 'SELECT * FROM stdin' sample.ion
+cat sample.ion | partiql -f ion 'SELECT * FROM stdin'
 
 # program script
 #   > cat script.sql
 #   > SELECT * FROM stdin
 partiql -q script.sql sample.ion
-cat sample.ion | partiql -f ion -q 'SELECT * FROM stdin'
+cat sample.ion | partiql -f ion 'SELECT * FROM stdin'
 ----
 
 [source,shell,subs=normal]
--- a/partiql-cli/README.adoc
+++ b/partiql-cli/README.adoc
@@ -18,14 +18,14 @@ partiql sample.csv    # '' format derived from extension
 cat sample.ion | partiql -f ion    # Load from stdin with the specified input format (-f) Ion.
 
 # program string
-partiql 'SELECT * FROM stdin' sample.ion
-cat sample.ion | partiql -f ion 'SELECT * FROM stdin'
+partiql -q 'SELECT * FROM stdin' sample.ion
+cat sample.ion | partiql -f ion -q 'SELECT * FROM stdin'
 
 # program script
 #   > cat script.sql
 #   > SELECT * FROM stdin
 partiql -q script.sql sample.ion
-cat sample.ion | partiql -f ion 'SELECT * FROM stdin'
+cat sample.ion | partiql -f ion -q 'SELECT * FROM stdin'
 ----
 
 [source,shell,subs=normal]
--- a/partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
+++ b/partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
@@ -19,6 +19,10 @@ import org.partiql.cli.io.Format
 import org.partiql.cli.pipeline.ErrorMessageFormatter
 import org.partiql.cli.pipeline.Pipeline
 import org.partiql.cli.shell.Shell
+import org.partiql.eval.Mode
+import org.partiql.eval.compiler.PartiQLCompiler
+import org.partiql.parser.PartiQLParser
+import org.partiql.planner.PartiQLPlanner
 import org.partiql.spi.catalog.Catalog
 import org.partiql.spi.catalog.Name
 import org.partiql.spi.catalog.Session
@@ -101,6 +105,7 @@ internal class MainCommand : Runnable {
         names = ["-f", "--format"],
         description = ["The data format, using the form <input>[:<output>]."],
         paramLabel = "<input[:output]>",
+        defaultValue = "ION:ION",
         converter = [Format.Converter::class],
     )
     lateinit var format: Pair<Format, Format>
@@ -139,17 +144,15 @@ internal class MainCommand : Runnable {
     )
     lateinit var warningsAsErrors: Array<ErrorCodeString>
 
-    @CommandLine.Parameters(
-        index = "0",
-        arity = "0..1",
+    @CommandLine.Option(
+        names = ["-q", "--query"],
         description = ["An optional PartiQL statement string."],
         paramLabel = "'statement'",
-        converter = [PairConverter::class],
     )
-    var program: Pair<String?, File?>? = null
+    var program: String? = null
 
     @CommandLine.Parameters(
-        index = "1..*",
+        index = "0..*",
         arity = "0..*",
         description = ["An optional list of files to execute the statement against."],
     )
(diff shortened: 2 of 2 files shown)


partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
```diff
@@ -237,7 +247,42 @@ internal class MainCommand : Runnable {
...
+
         // Derive a `default catalog from stdin (or file streams)
+        return listOf(parseIon())
+    }
+
+    private fun parsePartiQL(env: File): Catalog {
+        val stream = env.inputStream()
+        val compiler = PartiQLCompiler.standard()
+        val parser = PartiQLParser.standard()
+        val planner = PartiQLPlanner.standard()
```
alancai98: If possible, let's use `Pipeline.execute` rather than instantiating each of the components -- https://github.com/partiql/partiql-lang-kotlin/blob/7dce1375e4e5d281d779eb75b574666320d4e217/partiql-cli/src/main/kotlin/org/partiql/cli/pipeline/Pipeline.kt#L27-L36. This will keep the logic to execute a PartiQL statement the same as other parts in the CLI.

partiql-cli/README.adoc
```diff
@@ -13,6 +13,7 @@ partiql                     # No arguments
 partiql -d=/path/to/db      # Launch an interactive shell for the database in /path/to/db
 partiql sample.ion    # Load sample.ion into the database environment as the binding `stdin`
 partiql sample.csv    # '' format derived from extension
+partiql -e sample.env    # Load partiQL literal data into partiQL shell
```
alancai98, line 16: Let's also update the below `man partiql` section that provides more details on the CLI/shell options. Also let's add an example using the environment loading.

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
```diff
@@ -237,7 +247,42 @@ internal class MainCommand : Runnable {
...
+        if (env != null) {
+            return listOf(parsePartiQL(env!!))
+        }
+
         // Derive a `default catalog from stdin (or file streams)
+        return listOf(parseIon())
+    }
+
+    private fun parsePartiQL(env: File): Catalog {
+        val stream = env.inputStream()
```
alancai98: We should do some validation that the file path is valid and give a more detailed error message. With how it's currently implemented, an invalid file path will print out the entire stack trace.

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
```diff
@@ -237,7 +247,42 @@ internal class MainCommand : Runnable {
...
+        val compiler = PartiQLCompiler.standard()
+        val parser = PartiQLParser.standard()
+        val planner = PartiQLPlanner.standard()
+        val data = stream.bufferedReader(charset("UTF-8")).use { it.readText() }
+        val parseResult = parser.parse(data)
+        val statement = parseResult.statements[0]
+        val plan = planner.plan(statement, Session.empty()).plan
+        val datum = compiler.prepare(plan, Mode.PERMISSIVE()).execute()
```
alancai98: The execution mode we choose (i.e. strict vs permissive) should follow what is defined for the `strict` such that the results between query execution and environment loading is consistent.

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
```diff
@@ -237,7 +247,42 @@ internal class MainCommand : Runnable {
...
+
+        if (env != null) {
+            return listOf(parsePartiQL(env!!))
+        }
+
         // Derive a `default catalog from stdin (or file streams)
+        return listOf(parseIon())
+    }
+
+    private fun parsePartiQL(env: File): Catalog {
```
alancai98: nit (naming): `parsePartiQL` and `parseIon` should be called something like `loadPartiQL` or `loadIon`. Parsing generally has a different meaning than what these functions accomplish

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
alancai98: We should include some proper error messaging if the data loading fails (e.g. if the input PartiQL literal expression is malformed and fails to parse/evaluate).

partiql-cli/README.adoc
alancai98, line 1: General other comments:
- please include a PR description
- please update the CHANGELOG

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
```diff
@@ -237,7 +247,42 @@ internal class MainCommand : Runnable {
...
+        val stream = env.inputStream()
+        val compiler = PartiQLCompiler.standard()
+        val parser = PartiQLParser.standard()
+        val planner = PartiQLPlanner.standard()
+        val data = stream.bufferedReader(charset("UTF-8")).use { it.readText() }
+        val parseResult = parser.parse(data)
+        val statement = parseResult.statements[0]
```
XuechunHHH: Does it mean only the first statement in the env file is processed for `parseResult.statements[0]`? Is this intentional, meaning we only support a single statement in the env file?

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
```diff
@@ -84,6 +88,12 @@ internal class MainCommand : Runnable {
...
     var dir: File? = null
 
+    @CommandLine.Option(
+        names = ["-e", "--env"],
+        description = ["File containing the global environment"],
+    )
+    var env: File? = null
+
     @CommandLine.Option(
         names = ["--strict"],
```
XuechunHHH, line 97: Has `--strict` option already supported?

alancai98: Yep, it's supported

```
❯ ./partiql-cli/partiql.sh --strict
partiql ▶ {'a': 1}.b;
w: [semantic] Variable reference (b) could not be found in the database environment or in the set of available locals.

e: [runtime] Unexpected failure encountered. Caused by: java.lang.IllegalStateException: A warning has been converted into an error..
java.lang.IllegalStateException: A warning has been converted into an error.
```

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
alancai98: Yeah, this is the way the previous CLI versions worked. We should properly document it somewhere that the `.env` files are expected to be a single struct where each key-value pair corresponds to a new table in the default catalog.

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
xd1313113: Yes, will do

partiql-cli/README.adoc
xd1313113: will do

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
xd1313113: sure added a check

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
xd1313113: Done

xd1313113: The pipeline has a error listener and will throw with message when parsing.

e: [syntax] 1:1:1 Unexpected token (UNRECOGNIZED).
PRuntimeException{error=PError{code=INTERNAL_ERROR, severity=ERROR, kind=EXECUTION, location=null, properties={}}}

partiql-cli/README.adoc
```diff
@@ -55,6 +59,46 @@ man partiql
 
     --strict
         Strict mode (type-checking) throws runtime exceptions on mistyped operations.
+
+**Loading PartiQL Environment files**
```
alancai98: thanks for including the example. let's include this under a separate heading/code block. it's currently within the `man partiql` description which can be confusing

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
```diff
@@ -237,7 +243,42 @@ internal class MainCommand : Runnable {
...
+            return listOf(loadPartiQL(env!!))
+        }
+
         // Derive a `default catalog from stdin (or file streams)
+        return listOf(loadIon())
+    }
+
+    private fun loadPartiQL(env: File): Catalog {
+        if (!env.exists()) {
+            error("File not exists! Please check the env file path.")
```
alancai98: nit: grammar and let's also print out the env file

```suggestion
            error("File '$env' does not exist! Please check the env file path.")
```

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
alancai98: I think we should still output the error from running the parser + evaluator but also include info that the error is coming from the environment file loading. Otherwise, it can be confusing where/how to fix the error.

alancai98 (approved):
nice job adding this functionality back

Issue #1553, opened 2024-08-19
Add toString() for debugging of Datum


johnedquinn:
Should we also add `toString()` overrides for the other Datums? (Could track in a separate issue)

_Originally posted by @alancai98 in https://github.com/partiql/partiql-lang-kotlin/pull/1545#discussion_r1718764791_

alancai98:
Closed by https://github.com/partiql/partiql-lang-kotlin/pull/1802

PR #1802, opened 2025-08-06
Add toString() support for all datum types #1553

xd1313113:
## Relevant Issues
- Closes Issue #1553 

## Description
- Add toString() support for all datum types
- Make toString consistent

## Other Information
- Any new external dependencies? **YES**  Add java Snapshot test and its dependency slf4j

## License Information

### DIFF
--- a/partiql-spi/src/test/kotlin/org/partiql/spi/value/DatumToStringTest.kt
+++ b/partiql-spi/src/test/kotlin/org/partiql/spi/value/DatumToStringTest.kt
@@ -15,10 +15,21 @@ import java.time.ZoneOffset
 
 /**
  * Test cases for toString methods of all Datum implementations.
- * Verifies that each Datum class produces expected string representations.
- *
  * Test methods are in alphabetical order by method name
+ *
+ * The test suite is created using snapshot test. A .snap file was created after first run.
+ * On subsequent test runs, the .snap file is compared with the one procured before.
+ * If not match, the test will fail and a .snap.debug with the conflict is created.
+ * You need to decide if you have introduced a regression or intentionally made the change.
+ *
+ * If you introduced a regression, please fix the code.
+ * Otherwise, you can manually modify the .snap file to make it pass or delete it,
+ * and it will be generated again from scratch.
+ * Once you fix the test, the *.snap.debug file will get deleted.
+ *
+ * For more information, please refer to https://github.com/origin-energy/java-snapshot-testing
  */
+
 @ExtendWith(SnapshotExtension::class)
 class DatumToStringTest {
 


partiql-spi/src/main/java/org/partiql/spi/value/DatumRow.java
```diff
@@ -4,11 +4,7 @@
 import org.partiql.spi.types.PType;
 import org.partiql.spi.types.PTypeField;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
XuechunHHH: Shall we avoid wildcard import here according to the code style? Similar in other files.

partiql-spi/src/main/java/org/partiql/spi/value/DatumStruct.java
```diff
@@ -82,15 +78,25 @@ public PType getType() {
...
+            StringJoiner innerJoiner = new StringJoiner(", ", "[", "]");
+            for (Datum d : value) {
+                innerJoiner.add(d.toString());
+            }
+            sb.append(innerJoiner);
+
+            outerJoiner.add(sb.toString());
+        });
+
+        return "DatumStruct{" +
```
XuechunHHH, line 81: Can we create a helper function like:
```
public static String datumToString(String className, Object type, Object value) {
    return "Datum" + className + "{_type=" + type + ", _value=" + value + '}';
}
```
to avoid the code repetition?

partiql-spi/src/main/java/org/partiql/spi/value/DatumBytes.java
```diff
@@ -34,4 +34,17 @@ public byte[] getBytes() {
     public PType getType() {
         return _type;
     }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        for (byte b : _value) {
+            sb.append(String.format("%02X", b & 0xFF));
```
alancai98: mind clarifying what this printing logic is doing? I think it may be better to just leave it as bytes since the primary use case of `toString` is for our own debugging.

partiql-spi/src/main/java/org/partiql/spi/value/DatumCollection.java
```diff
@@ -39,9 +42,13 @@ public PType getType() {
...
+        StringJoiner joiner = new StringJoiner(", ", "[", "]");
+
+        _value.forEach(d -> joiner.add(d.toString()));
+
         return "DatumCollection{" +
                 "_type=" + _type +
-                ", _value=" + _value +
+                ", _value=" + joiner +
                 '}';
     }
```
alancai98: Doing a `+` here will create a new immutable string, which can be a redundant operation if the collection values' string representation is large. For any container datums (e.g. collection, row, struct), we should try reducing the `+` usage in favor of `StringBuilder`s. For scalar values, they tend to be a small, fixed size so it's not too necessary to migrate to use `StringBuilder`. I'll leave it up to you if you want to migrate those other classes.

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalDayTime.java
```diff
@@ -123,6 +123,9 @@ public int getNanos() {
 
     @Override
     public String toString() {
-        return "INTERVAL '" + days + " " + hours + ":" + minutes + ":" + seconds + "." + nanos + "' " + getType();
+        return "DatumIntervalDayTime{" +
```
alancai98, line 126: Missing a few fields like interval code, precision, and fractional precision.

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalYearMonth.java
```diff
@@ -58,8 +58,10 @@ public int getMonths() throws InvalidOperationException, NullPointerException {
         return _months;
     }
 
-    @Override
     public String toString() {
-        return "INTERVAL '" + _years + "-" + _months + "' " + getType();
+        return "DatumIntervalYearMonth{" +
```
alancai98, line 62: Missing a few fields like precision and interval code

partiql-spi/src/main/java/org/partiql/spi/value/DatumStruct.java
```diff
@@ -82,15 +78,25 @@ public PType getType() {
...
-        return sb.toString();
+
+            StringJoiner innerJoiner = new StringJoiner(", ", "[", "]");
+            for (Datum d : value) {
+                innerJoiner.add(d.toString());
+            }
+            sb.append(innerJoiner);
+
+            outerJoiner.add(sb.toString());
+        });
```
alancai98: This block is the same as what's in `DatumRow`. Let's extract it to a separate method

partiql-spi/src/main/java/org/partiql/spi/value/DatumRow.java
alancai98: Yeah let's get rid of the * imports here. Since the files are written in Java, they aren't caught by ktlint

partiql-spi/src/main/java/org/partiql/spi/value/DatumStruct.java
```diff
@@ -82,15 +78,25 @@ public PType getType() {
...
+            StringBuilder sb = new StringBuilder();
+            sb.append(key);
             sb.append(": ");
-            sb.append(entry.getValue().toString());
-            sb.append(", ");
-        }
-        sb.append(" }");
-        return sb.toString();
+
+            StringJoiner innerJoiner = new StringJoiner(", ", "[", "]");
```
alancai98: I don't think we need to wrap these in the square brackets, `[` & `]`.

partiql-spi/src/test/kotlin/org/partiql/spi/value/DatumToStringTest.kt
```diff
@@ -0,0 +1,256 @@
...
+import java.time.ZoneOffset
+import kotlin.test.assertEquals
+
+/**
+ * Test cases for toString methods of all Datum implementations.
+ * Verifies that each Datum class produces expected string representations.
+ *
+ * Test methods are in alphabetical order by method name
+ */
+class DatumToStringTest {
```
alancai98, line 34: While I think unit tests are important to add for most things, I think these hand-written `toString` unit tests may be annoying to maintain over time. For instance, if we added another field to be printed by each `Datum`, we would have to manually update all of the existing tests to match this string representation. IMO this is slight overkill for `toString` since we'll primarily use it for debugging.

This has been my experience especially with our [plan typing tests](https://github.com/partiql/partiql-lang-kotlin/blob/b5be0aa4f6e7c416c8443c78c53c39984ea772aa/partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/PlanTyperTestsPorted.kt), especially when doing any sort of refactor/migration. Our old evaluation tests had this issue until we transitioned to use the conformance tests, `partiql-tests`.

Perhaps this is a good chance for us to start using snapshot testing -- https://github.com/origin-energy/java-snapshot-testing. I've found it to be a lot simpler to maintain these tests in some other codebases. We already do snapshot testing in our Rust implementation. See here for an example setup using the Java/Kotlin version -- https://github.com/partiql/partiql-graphviz-java/pull/11.

partiql-spi/src/main/java/org/partiql/spi/value/DatumStruct.java
alancai98: IMO, it's not too much repetition and there's enough differences between the `Datum`s. We should refactor some repeated logic we use for struct and row though

partiql-spi/src/main/java/org/partiql/spi/value/DatumRow.java
xd1313113: Thanks. Will fix IDE seems to be the evil.

partiql-spi/src/main/java/org/partiql/spi/value/DatumBytes.java
xd1313113: Will add comment, this will convert to HEX value. I think byte array is more readable in HEX. The java default print number in radix 10. Do you have preference?

partiql-spi/src/main/java/org/partiql/spi/value/DatumCollection.java
xd1313113: The collection Datums are using stringjoiner to avoid immutable string concatenation for value.
Also I am thinking to String.format to reduce + currently in return statement

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalDayTime.java
xd1313113: The _type has already those info. So remove it from value.

partiql-spi/src/main/java/org/partiql/spi/value/DatumStruct.java
xd1313113: Why? The struct is key-value pair and value is List<Datum>. I use "[]" for array/list representation

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalDayTime.java
alancai98: The `_type` does not include the `intervalCode` since the PType only tells you if it's a year-month interval or a day-time interval. The interval code (https://github.com/partiql/partiql-lang-kotlin/blob/dixiao/daturmtoStr/partiql-spi/src/main/java/org/partiql/spi/types/IntervalCode.java#L16-L28) will actually tell you which of the original fields was specified in the datum. Like whether the year-month interval was just a `year` or just a `month`.

The `PType` does include the `precision` and `fractionalPrecision` printed back as a type (like `DAY(3) TO SECOND (5)`) but perhaps it would be more readable as separate fields. I'm less opinionated on these specific fields.

partiql-spi/src/main/java/org/partiql/spi/value/DatumBytes.java
alancai98: IMO, since the `toString` is used for debugging, I think the output should just be as similar to what may be passed in as part of the constructor to a `DatumBytes`. For the example you included in the tests, it's hard to easily understand how `Datum.blob(byteArrayOf(10, 50, 100, 127))` corresponds to the value `0A32647F` without diving into the implementation. I think it would be clearer if the output string representation was a list of the input values `[10, 50, 100, 12]`.

partiql-spi/src/main/java/org/partiql/spi/value/DatumStruct.java
alancai98: Ah I see. Forgot that `DatumStruct`s and `DatumRows` permit duplicate key names. Perhaps can add a test using the duplicate key names to check that the `innerJoiner` works for > 1 element.

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalDayTime.java
xd1313113: It actually does 
 assertEquals("DatumIntervalYearMonth{**_type=YEAR (2) TO MONTH**, _value=INTERVAL '2-6'}", datum.toString())

alancai98: Ah ok, it wasn't clear to me that the `PType` `toString` encodes the `intervalCode`. Let's add a couple explicit tests that don't include all of the fields. E.g.
- `Datum.intervalYear(10, 2)`
- `Datum.intervalDay(15, 2)`

partiql-spi/src/main/java/org/partiql/spi/value/DatumCollection.java
xd1313113: I took a further look. "+" is fine here as compiler actually compiles + into StringBuilder. String.format is actually more expensive. Collection object are built by Stringjoiner which internally use stringbuiler

partiql-spi/src/main/java/org/partiql/spi/value/DatumStruct.java
xd1313113: I took a look that for same key, it get chained in the list of value. duplicate key test is not necessary for toStringTest as the tests focuses on structure of the object, which is covered, flat and nested.

partiql-spi/src/test/kotlin/org/partiql/spi/value/DatumToStringTest.kt
xd1313113: Agree that toString wasn't that important for most use scenarios. my thought is it does not hurt to add it for now as toString is **public** and we customize its implementation and implementation is still under test. For major refactor/migration, they may need to make a decision to keep it, disable or fix later

partiql-spi/src/main/java/org/partiql/spi/value/DatumCollection.java
alancai98: > I took a further look. "+" is fine here as compiler actually compiles + into StringBuilder. String.format is actually more expensive.

Not to disagree with your point, but I'm curious how did you verify this? In the docs or some local testing? My read of this discussion https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java is that they `+` and `StringBuilder` `append` are largely equivalent in more recent versions of Java. Seems like the issue w/ combining `String`s is when using `+=` in a loop.

partiql-spi/src/main/java/org/partiql/spi/value/DatumStruct.java
alancai98: IMO, duplicate keys could still be a good test to add.

partiql-spi/src/test/kotlin/org/partiql/spi/value/DatumToStringTest.kt
alancai98: Could we try using snapshot testing following my other PR to setup (https://github.com/partiql/partiql-graphviz-java/pull/11/files)? A small change to the `toString` such as adding a new property to a `Datum` or modifying the printing of `PType`s could require tedious manual effort in updating the unit test.

It should be easy to set up -- just add a line to the spi build files and annotations to all of the existing tests. It can also give you good practice working with our current project builds (like adding additional dependencies). I can help answer any questions w/ the setup.

partiql-spi/src/main/java/org/partiql/spi/value/DatumCollection.java
xd1313113: yeah, forget to mentioned. I actually changed to StringBuilder, however intellij just remind me not needed as complier will compile static "+" to Stringbuilder. I also did some local test. There seems the performance diff are subtle in my toString implementation. And we only need to avoid += in loop where we use stringjoiner for our collection Datum

partiql-spi/src/main/java/org/partiql/spi/value/DatumStruct.java
xd1313113: added

partiql-spi/src/test/kotlin/org/partiql/spi/value/DatumToStringTest.kt
xd1313113: add snapshot test

buildSrc/src/main/kotlin/partiql.versions.kt
```diff
@@ -53,6 +53,8 @@ object Versions {
     const val junit4 = "4.12"
     const val junit4Params = "1.1.1"
     const val mockito = "4.5.0"
+    const val snapshotTest = "4.0.8"
+    const val slf4j = "2.0.17"
```
alancai98, line 57: Thanks for going about adding the snapshot tests. Can we remove this slf4j dependency? I didn't need to use it in the other PR https://github.com/partiql/partiql-graphviz-java/pull/11.

partiql-spi/src/test/kotlin/org/partiql/spi/value/DatumToStringTest.kt
```diff
@@ -0,0 +1,280 @@
...
+import java.time.LocalTime
+import java.time.OffsetDateTime
+import java.time.OffsetTime
+import java.time.ZoneOffset
+
+/**
+ * Test cases for toString methods of all Datum implementations.
+ * Verifies that each Datum class produces expected string representations.
+ *
+ * Test methods are in alphabetical order by method name
```
alancai98, line 18: nit: let's add some brief docs on how to run the snapshot tests and update the results if there's a divergence

alancai98 (commented):
A couple last minor comments. Thanks for migrating the tests to snapshot testing

alancai98:
Also the PR description probably needs updated to reflect the updated CHANGELOG and the new testing dependency.

buildSrc/src/main/kotlin/partiql.versions.kt
xd1313113: It is not optional, per Quick Start section from https://github.com/origin-energy/java-snapshot-testing

If removed, follow error is thrown
org.slf4j.LoggerFactory
java.lang.ClassNotFoundException: org.slf4j.LoggerFactory
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
	at au.com.origin.snapshots.junit5.SnapshotExtension.<clinit>

partiql-spi/src/test/kotlin/org/partiql/spi/value/DatumToStringTest.kt
xd1313113: added

Issue #1782, opened 2025-07-18
Remove `PartiQLValue` usages in testing


alancai98:
`PartiQLValue` was internalized in the `v1` release to just be used in testing -- https://github.com/partiql/partiql-lang-kotlin/blob/71cda7b21cce1bb143d687ba08c3955b32faa39f/partiql-spi/src/testFixtures/kotlin/org/partiql/value/PartiQLValue.kt. We should removing its usage altogether, since it makes adding new datatypes (such as INTERVAL) challenging.

Issue #973, opened 2023-01-25
`NULLIF(MISSING, MISSING)` should return `MISSING` not `NULL`


alancai98:
## Description
`NULLIF(MISSING, MISSING)` should return `MISSING` not `NULL`. If we reduce `NULLIF` to a `CASE WHEN` expression as specified in the [SQL-92 spec section 6.9 page 141](http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt), `NULLIF(MISSING, MISSING)` is equivalent to:

```
CASE WHEN MISSING = MISSING THEN NULL ELSE MISSING END
```

Since `MISSING = MISSING` is not true, the result is the else branch, `MISSING`.

## To Reproduce
1. Run PartiQL in either typing mode
2. Run query

```
PartiQL> NULLIF(MISSING, MISSING);
===' 
NULL
--- 
OK!
```

## Expected Behavior
Expect above query to return `MISSING` rather than `NULL`.

## Additional Context
- Java version: 11
- PartiQL version: 0.9.2-SNAPSHOT

DoD
- Behavior corrected
- [Function.md](https://github.com/partiql/partiql-lang-kotlin/blob/main/docs/documentation/Functions.md) documentation example for `NULLIF(MISSING, MISSING)` should be updated
- Update CHANGELOG

PR #1805, opened 2025-08-11
Remove `PartiQLValue` usage in conformance tests

XuechunHHH:
## Relevant Issues
- [Closes/Related To] Issue #1782 

## Description
- Migrate `PartiQLValue` to `Datum` in conformance tests
- TODO: Remove the `PartiQLValue` usage in unit tests and other potential places to fully deprecate the data structure 

## Other Information

## License Information


test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/util/DatumUtil.kt
```diff
@@ -0,0 +1,63 @@
...
+import com.amazon.ionelement.api.ionInt
+import com.amazon.ionelement.api.ionListOf
+import com.amazon.ionelement.api.ionNull
+import com.amazon.ionelement.api.ionString
+import com.amazon.ionelement.api.ionStructOf
+import org.partiql.spi.types.PType
+import org.partiql.spi.value.Datum
+
+// For error display only
+internal fun Datum.toIonElement(): IonElement {
```
XuechunHHH, line 20: This conversion is used for report display only and not in comparator.

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ rewritten to their query representation
 - Add cli option -e to load PartiQL literal data as default catalog data
 
 ### Changed
+- Migrate the usage of `PartiQLValue` to `Datum` in conformance tests
```
alancai98: Probably can leave out this entry from the changelog since it doesn't affect the public API or any behaviors of the public APIs

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/util/DatumUtil.kt
```diff
@@ -0,0 +1,63 @@
...
+    return when {
+        this.isNull() -> ionNull()
+        this.isMissing() -> ionNull()
```
alancai98: Think this should be the null annotated with the missing annotation `$missing::null`

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/util/DatumUtil.kt
```diff
@@ -0,0 +1,63 @@
...
+                        val field = fieldIterator.next()
+                        fields.add(field(field.name, field.value.toIonElement()))
+                    }
+                    ionStructOf(fields)
+                }
+                PType.VARIANT -> {
+                    val lowered = this.lower()
+                    lowered.toIonElement().withAnnotations("\$ion")
+                }
+                else -> ionString(this.toString())
```
alancai98, line 84: seems like we're also missing some other datetime types. Timestamp can map to Ion's timestamp. Date and time values can be mapped based on the representations we currently use in partiql-tests (https://github.com/partiql/partiql-tests/blob/main/partiql-tests-data/eval/query/order-by.ion#L248)

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/util/DatumUtil.kt
```diff
@@ -0,0 +1,63 @@
...
+                    ionListOf(elements)
+                }
+                PType.BAG -> {
+                    val elements = mutableListOf<IonElement>()
+                    for (element in this) {
+                        elements.add(element.toIonElement())
+                    }
+                    ionListOf(elements).withAnnotations("\$bag")
+                }
+                PType.STRUCT -> {
```
alancai98: Also can include PType.ROW in this branch of the `when`

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/executor/EvalExecutor.kt
```diff
@@ -93,57 +90,21 @@ class EvalExecutor(
...
-            return true
-        }
-        // TODO: this comparator is designed for order by
-        //  One of the false result it might produce is that
-        //  it treats MISSING and NULL equally.
-        //  we should move to hash or equals in value class once
-        //  we finished implementing those.
-        if (comparator.compare(v1, v2) == 0) {
-            return true
-        }
```
alancai98, line 129: Hmm, I think this is still an issue w/ directly using Datum's `comparator` directly since I believe it treats a comparison of `null` and `missing` as equal -- https://github.com/partiql/partiql-lang-kotlin/blob/f09decb0fca0cd5d3b47664dc48cf0aff329adc1/partiql-spi/src/main/java/org/partiql/spi/value/DatumComparator.java#L136-L138. The default `Datum` `comparator` follows what's defined in section 12.2 of the PartiQL spec (https://partiql.org/assets/PartiQL-Specification.pdf#page=44), which treats `null` and `missing` as the same. That particular behavior is useful for order by and group by, but is probably not what we want for test assertions on `Datum`.

We could opt for a couple approaches here:
1. creating a new Datum comparator just for tests or
2. adding some parameter to the existing Datum comparator to treat `nulls` and `missing` values as different

I'm wondering if we can try out option 2 without breaking any existing public APIs.

XuechunHHH: Thanks for the suggestions. Will add a `distinguishNullMissing` parameter to specifically support the behavior in tests.

partiql-spi/src/main/kotlin/org/partiql/spi/utils/FunctionUtils.kt
```diff
@@ -115,10 +115,13 @@ internal object FunctionUtils {
         val aIsUnknown = isUnknown(a)
         val bIsUnknown = isUnknown(b)
         return when {
+            // If either operand is false, result is false
```
XuechunHHH, line 118: Fix the `AndNullUnknown` logic.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/util/DatumUtil.kt
```diff
@@ -0,0 +1,88 @@
...
+        else -> {
+            when (this.type.code()) {
+                PType.BOOL -> ionBool(this.boolean)
+                PType.INTEGER -> ionInt(this.int.toLong())
+                PType.BIGINT -> ionInt(this.long)
+                PType.STRING -> ionString(this.string)
+                PType.DECIMAL -> ionDecimal(Decimal.valueOf(this.bigDecimal))
+                PType.DOUBLE -> ionFloat(this.double)
+                PType.REAL -> ionFloat(this.float.toDouble())
+                PType.TIMESTAMP -> ionTimestamp(this.localDateTime.toString())
```
alancai98: This branch can also include `PType.TIMESTAMPZ`

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelOpUnionDistinct.kt
```diff
@@ -26,10 +26,11 @@ internal class RelOpUnionDistinct(
 
     override fun peek(): Row? {
         for (record in input) {
+            val originalValues = record.values.copyOf()
             record.values.coerceMissing()
             if (!seen.contains(record.values)) {
                 seen.add(record.values)
-                return Row(record.values)
+                return Row(originalValues)
```
alancai98, line 33: Any reason these are getting changed in this PR?

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprNullIf.kt
```diff
@@ -15,7 +15,7 @@ internal class ExprNullIf(
         val value = valueExpr.eval(env)
         val nullifier = nullifierExpr.eval(env)
         return when (comparator.compare(value, nullifier)) {
-            0 -> Datum.nullValue()
+            0 -> if (value.isMissing || nullifier.isMissing) Datum.missing() else Datum.nullValue()
```
alancai98, line 18: Any reason this is getting changed in this PR?

partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
```diff
@@ -1152,17 +1152,39 @@ static Comparator<Datum> comparator() {
      * operations.
      * </p>
      * @param nullsFirst if true, nulls are ordered before non-null values, otherwise after.
-     * @return the default comparator for {@link Datum}.
+     * @return the default comparator for {@link Datum}. By default, distinguishNullMissing is set to false.
```
alancai98: Since `distinguishNullMissing` is an internal implementation detail. Perhaps you can just say null and missing are treated as equivalent.

```suggestion
     * @return the default comparator for {@link Datum}. By default, null and missing are treated as equivalent.
```

partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
```diff
@@ -1152,17 +1152,39 @@ static Comparator<Datum> comparator() {
...
+     * </p>
+     * @param nullsFirst if true, nulls are ordered before non-null values, otherwise after.
+     * @param distinguishNullMissing if true, null and missing values are distinguished (null < missing), otherwise they are treated as equivalent. By default, this is set to false and only set to true when testing for now.
+     * @return the comparator for {@link Datum}.
+     * @see Datum
+     * @see java.util.TreeSet
+     * @see java.util.TreeMap
+     */
+    @NotNull
+    static Comparator<Datum> comparator(boolean nullsFirst, boolean distinguishNullMissing) {
```
alancai98, line 1183: Let's add a new changelog entry for this new comparator since it's part of the public API

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnEq.kt
```diff
@@ -16,8 +16,8 @@ import org.partiql.spi.value.Datum
  * > Equality never fails in the type-checking mode and never returns MISSING in the permissive mode. Instead, it can
  * compare values of any two types, according to the rules of the PartiQL type system. For example, 5 = 'a' is false.
  *
- * For the existing conformance tests, whenever an operand is NULL or MISSING, the output is NULL. This implementation
- * follows this.
+ * For the existing conformance tests, when an operand is NULL, the output is NULL. When an operand is MISSING,
```
alancai98: Hmm I believe the existing behavior is correct. Equality follows a bit different behavior than other functions/operators much like the logical operators. Perhaps there was a bad conformance test? Which test led to this change?

partiql-spi/src/main/kotlin/org/partiql/spi/utils/FunctionUtils.kt
```diff
@@ -136,7 +139,8 @@ internal object FunctionUtils {
     }
 
     internal fun logicalNot(a: Datum): Datum {
-        if (isUnknown(a)) return Datum.nullValue(PType.bool())
+        if (a.isMissing) return Datum.missing(PType.bool())
+        if (a.isNull) return Datum.nullValue(PType.bool())
```
alancai98: Ah sorry for the confusion here. The original conformance test I had added was incorrect. `NOT MISSING` should return `NULL`. I'll fix the test.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnEq.kt
XuechunHHH: So `missing = null;` should having output `null` in both modes, am I understanding it correctly?

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelOpUnionDistinct.kt
XuechunHHH: `record.values.coerceMissing()` converts `missing` to `null` for consistent deduplication while the output should preserve `missing` according to the test at http://github.com/partiql/partiql-tests/blob/b05c0ca923dff3578acf867812bdf693a3d5fc7e/partiql-tests-data/eval/rfc/0007.ion#L230-L244

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprNullIf.kt
XuechunHHH: `nullif(missing, missing);` should have `missing` as output according to the `missing` propagation behavior. `comparator.compare(missing, missing)` will be 0 and returns `null` before this change.

partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
XuechunHHH: Got it.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnEq.kt
alancai98: Yes. Based on my read of spec section 7.1.1 Equality:
> Equality never fails in the type-checking mode and never returns `MISSING` in the permissive mode

`=` should never return `MISSING`.

partiql-spi/src/main/kotlin/org/partiql/spi/utils/FunctionUtils.kt
alancai98: Test fixed in https://github.com/partiql/partiql-tests/pull/153.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnEq.kt
XuechunHHH: I see the issue here. Will cut a PR to fix the corresponding conformance tests.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprNullIf.kt
alancai98: Ok -- the behavior is now correct with this fix but just to clarify it is due to the `NULLIF` rewrite to `CASE WHEN` as described in section 6.21 of SQL1999 (https://web.cecs.pdx.edu/~len/sql1999.pdf#page=205) not `missing` propagation. `NULLIF` is another of the functions/operators that does not propagate `null`/`missing`.

`NULLIF(MISSING, MISSING)` is equivalent to `CASE WHEN MISSING=MISSING THEN NULL ELSE MISSING END`, which should return `MISSING`.

Feel free to call out this fix in the CHANGELOG as well since we had a prior tracking issue https://github.com/partiql/partiql-lang-kotlin/issues/973.

XuechunHHH: Interesting. Still get a lot to learn about the behavior of `missing` and `null`.

alancai98 (approved):
nice work

PR #1811, opened 2025-08-15
Delete `PartiQLValue`

XuechunHHH:
## Relevant Issues

## Description
- Fully deprecate `PartiQLValue` interface and all of its related implementations

## Other Information
- Updated Unreleased Section in CHANGELOG: **[NO]**: No changes for public interfaces

## License Information


partiql-spi/src/test/kotlin/org/partiql/value/TimestampTest.kt
```diff
@@ -1,928 +0,0 @@
-package org.partiql.value
```
alancai98, line 1: This is a pretty comprehensive suite of tests of the prior `PartiQLValue` implementation. I don't believe we have this replicated for `Datum` or in the conformance tests (if we do, feel free to ignore this comment). I think we should:
1. comment out this file for now
2. cut an issue to replicate these tests somewhere (likely in the conformance tests)
3. link the issue in this file

alancai98 (approved):
Nice job and glad we could finally remove this tech debt. Approving for now but let's comment out the `TimestampTest.kt` rather than delete it. IMO it has some pretty valuable tests that we should port over to the conformance tests at some point.

Issue #1791, opened 2025-07-25
Fix datum interval printing in the CLI


alancai98:
Currently a datum interval that has mixed signs between the datetime range fields is not properly printed in the CLI:

```
partiql ▶ INTERVAL '2-2' YEAR TO MONTH + INTERVAL '-1-3' YEAR TO MONTH;

=== RESULT ===
INTERVAL '1-1' YEAR (2) TO MONTH

OK!
partiql ▶ INTERVAL '2-2' YEAR TO MONTH + INTERVAL '-2-3' YEAR TO MONTH;

=== RESULT ===
INTERVAL '0-1' YEAR (2) TO MONTH

OK!
```

The underlying Datum values are correct (though may need normalized https://github.com/partiql/partiql-lang-kotlin/issues/1790).

Issue #1790, opened 2025-07-25
Normalize/validate datum interval values


alancai98:
Followup to this [thread](https://github.com/partiql/partiql-scribe/pull/106#discussion_r2229834924) in Scribe, we should ensure that constructed Datum interval ranges do not have mixed signs.

E.g. a year-month interval range should either
- normalize 1 year and -1 months to 0 year and 11 months OR
- give an error

zyfy29:
https://github.com/partiql/partiql-lang-kotlin/blob/52000305a625b6e2613728daeb7a40d499aa7e80/partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalYearMonth.java#L19

INTERVALs with mixed signs are allowed up to now. All the `DatumIntervalYearMonth` instances comes here, I think we can normalize it constructor. The same is true for `DatumIntervalDayTime`.

alancai98:
> INTERVALs with mixed signs are allowed up to now.

Yeah, it's an issue w/ the initial implementation.

> All the `DatumIntervalYearMonth` instances comes here, I think we can normalize it in the constructor. The same is true for `DatumIntervalDayTime`.

Normalization in the constructor is probably the approach we'll go with or in the static factories -- https://github.com/partiql/partiql-lang-kotlin/blob/52000305a625b6e2613728daeb7a40d499aa7e80/partiql-spi/src/main/java/org/partiql/spi/value/Datum.java#L912. Still need to investigate the options here since downstream parts of the library that use `INTERVAL` datums (e.g. datum cli printer) may need to perform a similar validation.

PR #1815, opened 2025-08-18
feat: refactor internal INTERVAL `Datum` implementation

alancai98:
## Relevant Issues
- Closes https://github.com/partiql/partiql-lang-kotlin/issues/1790
- Closes https://github.com/partiql/partiql-lang-kotlin/issues/1791

## Description
Refactors the internal `INTERVAL` representation. This new representation is less complicated than the prior implementation and ensures that the user does not need to perform normalization before using any Datum interval. See self-review comments for more details.

## Other Information

## License Information

### DIFF
--- a/partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
+++ b/partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
@@ -287,7 +287,7 @@ public interface Datum extends Iterable<Datum> {
      * <b>! ! ! EXPERIMENTAL ! ! !</b> This is an experimental API under development by the PartiQL maintainers.
      * </p>
      * <p>
-     * Returns the total amount of months for this year-month interval.
+     * Returns the total amount of months for this year-month INTERVAL.
      * </p>
      * <p>
      * This only applies to {@link PType#INTERVAL_YM}.
@@ -305,7 +305,8 @@ public interface Datum extends Iterable<Datum> {
      * <b>! ! ! EXPERIMENTAL ! ! !</b> This is an experimental API under development by the PartiQL maintainers.
      * </p>
      * <p>
-     * Returns the total amount of whole seconds for this day-time interval.
+     * Returns the total amount of whole seconds for this day-time INTERVAL. This does not include the nanosecond
+     * component of the day-time INTERVAL. The nanosecond component can be retrieved by calling {@link Datum#getNanos()}.
      * </p>
      * <p>
      * This only applies to {@link PType#INTERVAL_DT}.


partiql-spi/src/test/java/org/partiql/spi/value/__snapshots__/DatumToStringTest.snap
```diff
@@ -129,7 +129,7 @@ DatumIntervalDayTime{_type=SECOND (2, 6), _value=INTERVAL '0 0:0:45.500000000'}
 
 
 org.partiql.spi.value.DatumToStringTest.testDatumIntervalYearMonthToString[datumMonth]=[
-DatumIntervalYearMonth{_type=MONTH (2), _value=INTERVAL '0-15'}
```
alancai98, line 132: self-review: prior test should have normalized the month part of the year-month interval

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalYearMonth.java
```diff
@@ -10,14 +10,18 @@
  */
 class DatumIntervalYearMonth implements Datum {
 
-    private final int _years;
-    private final int _months;
+    private final long _months;
```
alancai98, line 13: self-review: for year-month intervals, only need to store the total number of months when doing computation. this simplifies the operations on intervals and removes the need for normalization.

Also considered using Java's `Period` https://docs.oracle.com/javase/8/docs/api/java/time/Period.html but it was slight overkill for just two fields.

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalDayTime.java
```diff
@@ -4,17 +4,20 @@
...
  * expect that all values have been normalized and checked already. This also lazily creates its type.
  */
 class DatumIntervalDayTime implements Datum {
 
-    private final int days;
-    private final int hours;
-    private final int minutes;
-    private final int seconds;
-    private final int nanos;
+    private final Duration _duration;
```
alancai98, line 20: self-review: using Java's `Duration` here which provides a lot of the datetime arithmetic operations that we need and performs its own normalization. The normalization slightly differs from what we want hence why there's some additional logic related to negative durations. E.g. Java's duration represents -1 nanoseconds as `-1 seconds` and `999999999 nanoseconds` what we really want is for all of the fields to be of the same sign (i.e. all positive or all negative).

partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
```diff
@@ -279,6 +279,42 @@ default int getMonths() throws InvalidOperationException, NullPointerException {
         throw new InvalidOperationException(getType(), "getMonths");
     }
 
+    /**
```
alancai98, line 285: self-review: marked as experimental for now. I'll see how the Scribe integration goes before making these non-experimental

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMinus.kt
```diff
@@ -71,38 +71,31 @@ internal object FnMinus : DiadicArithmeticOperator("minus") {
                 val p: Int = lhs.precision // TODO: Do we need to calculate a new precision?
                 val s: Int = 6 // TODO: Do we need to calculate a new fractional precision?
                 basic(PType.intervalDaySecond(p, s)) { args ->
-                    val interval0 = args[0]
-                    val interval1 = args[1]
-                    subtractIntervalDayTimes(interval0, interval1, p, s)
+                    val i1 = args[0]
```
alancai98, line 75: self-review: see here and in the other operator files for the simplified implementation of datetime arithmetic with the new representation

partiql-cli/src/main/kotlin/org/partiql/cli/io/DatumWriterTextPretty.kt
```diff
@@ -235,22 +238,25 @@ class DatumWriterTextPretty(
...
                 this.out.print(")")
             }
             IntervalCode.SECOND -> {
                 this.out.print("'")
-                this.out.print(datum.seconds)
-                this.out.print(fractionalSeconds(datum.nanos, type.fractionalPrecision))
+                if (datum.totalSeconds < 0) {
+                    this.out.print("-")
+                }
+                this.out.print((datum.days * SECONDS_PER_DAY + datum.hours * SECONDS_PER_HOUR + datum.minutes * SECONDS_PER_MINUTE + datum.seconds).absoluteValue)
```
xd1313113: Why do we need to do calculation before print? why not just `totalSeconds.absoluteValue`

partiql-spi/src/test/java/org/partiql/spi/value/__snapshots__/DatumToStringTest.snap
xd1313113: This is Month only interval. We should not normalize this one. Otherwise, it looks like a Year-Month interval.

partiql-cli/src/main/kotlin/org/partiql/cli/io/DatumWriterTextPretty.kt
```diff
@@ -288,7 +303,10 @@ class DatumWriterTextPretty(
             }
             IntervalCode.DAY_SECOND -> {
                 this.out.print("'")
-                this.out.print(datum.days)
+                if (datum.totalSeconds < 0 || (datum.totalSeconds == 0L && datum.nanos < 0)) {
```
XuechunHHH: nit: checking `datum.totalSeconds == 0L` seems unnecessary since `datum.nanos < 0` only when `datum.totalSeconds == 0L`.

partiql-cli/src/test/kotlin/org/partiql/cli/io/DatumWriterTextPrettyTests.kt
```diff
@@ -146,6 +165,27 @@ class DatumWriterTextPrettyTests {
...
+            // Value greater than field divisor
+            Datum.intervalMonth(99, 2),
+            Datum.intervalHour(99, 2),
+            Datum.intervalMinute(99, 2),
+            Datum.intervalSecond(99, 0, 2, 2),
+            Datum.intervalSecond(-99, 1, 2, 2),
+            Datum.intervalMinuteSecond(99, 1, 0, 2, 2),
+            Datum.intervalHourSecond(99, 1, 2, 0, 2, 2),
+            Datum.intervalHourMinute(99, 1, 2),
+            Datum.intervalHourSecond(99, 1, 2, 3, 2, 2),
```
XuechunHHH, line 188: Probably can add more tests for truncation.

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalDayTime.java
```diff
@@ -98,34 +98,58 @@ private void initType() {
...
     }
 
     @Override
     public int getHours() {
-        return hours;
+        return _duration.toHoursPart();
     }
 
     @Override
     public int getMinutes() throws InvalidOperationException, NullPointerException {
```
XuechunHHH: nit: looks unnecessary to throw error for `getMinutes()` and `getDays()`

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalDayTime.java
```diff
@@ -98,34 +98,58 @@ private void initType() {
...
+        }
+        else {
+            return _duration.getSeconds();
+        }
     }
 
     @Override
     public String toString() {
+        String sign = "";
+        if (getTotalSeconds() == 0 && getNanos() < 0) {
```
XuechunHHH: Looks like should be `getTotalSeconds() < 0 || getNanos() < 0` here.

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalYearMonth.java
```diff
@@ -50,18 +54,27 @@ private void initType() {
...
+        return (int) (_months / 12);
     }
 
     @Override
     public int getMonths() throws InvalidOperationException, NullPointerException {
+        return (int) (_months % 12);
+    }
+
+    @Override
+    public long getTotalMonths() throws InvalidOperationException, NullPointerException {
```
XuechunHHH: Same here. Seems will never throw an error.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMinus.kt
```diff
@@ -71,38 +71,31 @@ internal object FnMinus : DiadicArithmeticOperator("minus") {
                 val p: Int = lhs.precision // TODO: Do we need to calculate a new precision?
                 val s: Int = 6 // TODO: Do we need to calculate a new fractional precision?
```
XuechunHHH: Why we pick 6 here instead of `maxOf(lhs.fractionalPrecision ?: 6, rhs.fractionalPrecision ?: 6)` and `lhs.precision` instead of `INTERVAL_MAX_PRECISION` (inconsistency with YM implementation)?

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMinus.kt
```diff
@@ -71,38 +71,31 @@ internal object FnMinus : DiadicArithmeticOperator("minus") {
...
                 basic(PType.intervalDaySecond(p, s)) { args ->
-                    val interval0 = args[0]
-                    val interval1 = args[1]
-                    subtractIntervalDayTimes(interval0, interval1, p, s)
+                    val i1 = args[0]
+                    val i2 = args[1]
+                    Datum.intervalDaySecond(
```
XuechunHHH, line 77: Shall we handle possible overflow cases here? e.g. ` INTERVAL '2147483647' DAY - INTERVAL '-1' DAY;`

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMinus.kt
```diff
@@ -71,38 +71,31 @@ internal object FnMinus : DiadicArithmeticOperator("minus") {
...
             lhs.code() == PType.INTERVAL_YM && rhs.code() == PType.INTERVAL_YM -> {
-                val p: Int = lhs.precision // TODO: Do we need to calculate a new precision?
+                val p = INTERVAL_MAX_PRECISION // Based on SQL2023 6.44 SR 2)c), precision is implementation-defined. Set to max
                 basic(PType.intervalYearMonth(p)) { args ->
-                    val interval0 = args[0]
-                    val interval1 = args[1]
-                    subtractIntervalYearMonths(interval0, interval1, p)
+                    val i1 = args[0]
+                    val i2 = args[1]
+                    Datum.intervalYearMonth(i1.years - i2.years, i1.months - i2.months, p)
```
XuechunHHH, line 93: Same here. Do we need to add some error handling?

XuechunHHH (commented):
Nice refactor work on `INTERVAL`. With `duration`, the operator implementation looks much more cleaner and less error-prone. The code is significantly more readable and the logic flows better. Left comments at some places where I am a little confused but I may make some mistakes.

Also, are we planning to add some unit tests for these arithmetic or will we add those to conformance tests after finishing support for `INTERVAL?`

partiql-cli/src/main/kotlin/org/partiql/cli/io/DatumWriterTextPretty.kt
alancai98: Both branches are actually necessary. `totalSeconds` denotes the number of whole seconds (excluding nano seconds) while `nano` denotes the number of nanoseconds. We want to add the `-` sign whenever either
- total whole seconds is negative or
- when the total whole seconds is 0 but the number of nanoseconds is negative

For the first case, imagine we had `-10` whole seconds, the output should include the `-` sign (e.g. `INTERVAL '-10' SECOND`). The second case is a bit more subtle. Imagine we had `0` whole seconds and `-90000000` (e.g. `INTERVAL -0.9 SECOND`, ignoring any fractional precision).

Through the current representation of day-time intervals, there should not be a mix of signs between `totalSeconds` and `nano`.

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalDayTime.java
alancai98: Nice catch. ~I think it should replicate what we do in the Datum pretty printer, so `getTotalSeconds() < 0 || (getTotalSeconds() == 0 && getNanos() < 0)`.~

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalDayTime.java
alancai98: Yeah, it's unnecessary. Copied over from the interface definition. I'll remove here and other places.

partiql-cli/src/test/kotlin/org/partiql/cli/io/DatumWriterTextPrettyTests.kt
alancai98: IMO, this isn't the right place to add since tests are for the CLI's datum pretty printer. Instead, I think we should add more truncation tests in the conformance tests (https://github.com/partiql/partiql-tests/issues/144). I'll add a list of more tests we should add in that issue.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMinus.kt
alancai98: Ah, I forgot to update the day-time precision and fractional precision. I'll update.

Based on SQL2023 6.44, `precision` for leading field of year-month and day-time interval is implementation-defined, so I'll set it to `INTERVAL_MAX_PRECISION`. The fractional second precision is max of lhs and rhs fractional precision

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMinus.kt
alancai98: Overflow should not occur since the max precision of an interval field is `9`. The max number representable by an interval of precision `9` is less than `2147483647`, so overflow should not occur. There's already an existing function to check if an interval field exceeds the set precision https://github.com/partiql/partiql-lang-kotlin/blob/fd5493f616677aee9104c76c2c1dccaf331ad859/partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalHelpers.java#L20-L25.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMinus.kt
alancai98: Similar to other comment for day-time interval subtraction. The max values are restricted by the max precision of `9`, which is less than the overflow threshold.

partiql-cli/src/test/kotlin/org/partiql/cli/io/DatumWriterTextPrettyTests.kt
alancai98: Added to the issue (https://github.com/partiql/partiql-tests/issues/144) description.

alancai98:
> Also, are we planning to add some unit tests for these arithmetic or will we add those to conformance tests after finishing support for `INTERVAL?` 

We have existing unit tests in partiql-lang-kotlin for arithmetic +, -, *, / (see https://github.com/partiql/partiql-lang-kotlin/blob/fd5493f616677aee9104c76c2c1dccaf331ad859/partiql-eval/src/test/kotlin/org/partiql/eval/internal/IntervalPlusTests.kt for example). There very well could be some edge cases we aren't covering. Since we're going to migrate the evaluation tests to the conformance tests in the near future, I think we should track any `INTERVAL` testing gaps in https://github.com/partiql/partiql-tests/issues/144.

partiql-cli/src/main/kotlin/org/partiql/cli/io/DatumWriterTextPretty.kt
XuechunHHH: Emm, since there should not be a mix of sings between `totalSeconds` and `nano`, I think checking `datum.totalSeconds < 0 || datum.nanos < 0` is enough here. Both cases you mentioned above and a possible third case `-10.9` (`-10` whole seconds and `-90000000`) could be all included here.

partiql-cli/src/test/kotlin/org/partiql/cli/io/DatumWriterTextPrettyTests.kt
XuechunHHH: Ah yes, it would be more appropriate to be added in conformance tests.

partiql-cli/src/main/kotlin/org/partiql/cli/io/DatumWriterTextPretty.kt
```diff
@@ -235,22 +238,25 @@ class DatumWriterTextPretty(
             }
             IntervalCode.HOUR -> {
                 this.out.print("'")
-                this.out.print(datum.hours)
+                this.out.print(datum.totalSeconds / SECONDS_PER_HOUR)
```
xd1313113, line 240: I think we should keep months, hours, minutes etc as normalized value. Private setter and public getter. 
It is a bit odd here to calculate the value by caller itself.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMinus.kt
XuechunHHH: Nice. That makes sense.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMinus.kt
XuechunHHH: I see. Thanks for clarification.

partiql-cli/src/main/kotlin/org/partiql/cli/io/DatumWriterTextPretty.kt
```diff
@@ -235,22 +238,25 @@ class DatumWriterTextPretty(
...
+                this.out.print(datum.totalSeconds / SECONDS_PER_MINUTE)
                 this.out.print("' MINUTE (")
                 this.out.print(type.precision)
                 this.out.print(")")
             }
             IntervalCode.SECOND -> {
                 this.out.print("'")
-                this.out.print(datum.seconds)
-                this.out.print(fractionalSeconds(datum.nanos, type.fractionalPrecision))
+                if (datum.totalSeconds < 0) {
```
XuechunHHH: Seems to be `datum.totalSeconds < 0 || datum.nanos < 0)` here.

partiql-cli/src/main/kotlin/org/partiql/cli/io/DatumWriterTextPretty.kt
XuechunHHH: nit: use `datum.totalSeconds.absoluteValue` instead

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalYearMonth.java
```diff
@@ -49,19 +53,28 @@ private void initType() {
     }
 
     @Override
-    public int getYears() throws InvalidOperationException, NullPointerException {
-        return _years;
+    public int getYears() {
+        return (int) (_months / 12);
```
xd1313113, line 57: What should we return for getyears if it is Month interval or year-Month interval respectively?
E.g.  For 16 months,  if it is interval Month, the output should be "0-16" and if it is Interval YearMonth, we should output  "1-4", which is 1 year and 4 months.

partiql-cli/src/test/kotlin/org/partiql/cli/io/DatumWriterTextPrettyTests.kt
```diff
@@ -74,6 +74,25 @@ class DatumWriterTextPrettyTests {
           INTERVAL '-1:2.30' MINUTE (2) TO SECOND (2),
           INTERVAL '3:4.00' MINUTE (2) TO SECOND (2),
           INTERVAL '3:4.012345678' MINUTE (2) TO SECOND (9),
```
XuechunHHH, line 76: Can probably add more zero leading tests like: `INTERVAL '-0.10' SECOND (2,2)` accordingly here.

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalDayTime.java
```diff
@@ -97,35 +97,59 @@ private void initType() {
     }
 
     @Override
-    public int getDays() throws InvalidOperationException, NullPointerException {
-        return days;
+    public int getDays() {
```
xd1313113, line 100: Not sure if getDays should return different result based on interval type. And writeDatum should use those properties for print rather than calculating on its owns.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnAbs.kt
```diff
@@ -118,22 +117,11 @@ internal val Fn_ABS__INTERVAL_YM__INTERVAL_YM = Function.overload(
     parameters = arrayOf(Parameter("value", PType.intervalYearMonth(2))),
```
XuechunHHH: Also use `INTERVAL_MAX_PRECISION` instead here

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnAbs.kt
```diff
@@ -142,66 +130,13 @@ internal val Fn_ABS__INTERVAL_DT__INTERVAL_DT = Function.overload(
     parameters = arrayOf(Parameter("value", PType.intervalDaySecond(2, 6))),
```
XuechunHHH: Also use `INTERVAL_MAX_PRECISION` and `NANO_MAX_PRECISION` instead here

partiql-cli/src/main/kotlin/org/partiql/cli/io/DatumWriterTextPretty.kt
XuechunHHH: Agree to use `datum.totalSeconds.absoluteValue` to simplify the code. Still doing calculation at print time tho.

partiql-cli/src/main/kotlin/org/partiql/cli/io/DatumWriterTextPretty.kt
XuechunHHH: > This is Month only interval. We should not normalize this one. Otherwise, it looks like a Year-Month interval.

`INTERVAL '0-15' YEAR TO MONTH` is not allowed now.

XuechunHHH: > I think we should keep months, hours, minutes etc as normalized value. Private setter and public getter.
> It is a bit odd here to calculate the value by caller itself.

We are currently using `Duration` as internal representation for `INTERVAL` and don't maintain months, hours, minutes separately any more.

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalYearMonth.java
XuechunHHH: I think the `getYears()` is designed for getting `years` in `_months`. Can refer to:
> self-review: for year-month intervals, only need to store the total number of months when doing computation. this simplifies the operations on intervals and removes the need for normalization.

In the example you mentioned, it is used to get the `1 year`.

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalDayTime.java
XuechunHHH: Same for `getDays()` here. It is for getting `days` in `_duration`.

partiql-cli/src/main/kotlin/org/partiql/cli/io/DatumWriterTextPretty.kt
alancai98: Ah my bad, you're right. In a previous version of this refactor, I had modeled `totalSeconds` and `nano` more similarly to Java's `Duration`, which _can_ have mixed signs (e.g. a negative `totalSeconds` and a positive `nano` field). I will revise to your suggestion.

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalDayTime.java
alancai98: I'll revise to your suggestion.

partiql-cli/src/main/kotlin/org/partiql/cli/io/DatumWriterTextPretty.kt
alancai98: Fixed.

partiql-cli/src/main/kotlin/org/partiql/cli/io/DatumWriterTextPretty.kt
alancai98: My bad, this was some leftover logic from the prior implementation I was trying out. We can just use `datum.totalSeconds.absoluteValue`.

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalYearMonth.java
alancai98: > What should we return for getyears if it is Month interval or year-Month interval respectively?
E.g. For 16 months, if it is interval Month, the output should be "0-16" and if it is Interval YearMonth, we should output "1-4", which is 1 year and 4 months.

It should be from the normalized value. For both of the above it will output `1` year.

partiql-spi/src/main/java/org/partiql/spi/value/DatumIntervalDayTime.java
alancai98: > Not sure if getDays should return different result based on interval type. And writeDatum should use those properties for print rather than calculating on its owns.

Leaving the values unnormalized may make the printing slightly easier but it will make any other computation more complicated since a user of the API would need to check the interval type. IMO, when it comes to computation, all we really care about is the normalized value. E.g. `INTERVAL '2' YEAR` is effectively equivalent to `INTERVAL '24' MONTH`.

partiql-cli/src/main/kotlin/org/partiql/cli/io/DatumWriterTextPretty.kt
alancai98: > I think we should keep months, hours, minutes etc as normalized value. Private setter and public getter.
It is a bit odd here to calculate the value by caller itself.

What do you mean by normalized value? In the other comments you were suggesting not using the normalized value.

partiql-spi/src/test/java/org/partiql/spi/value/__snapshots__/DatumToStringTest.snap
alancai98: In the `.toString()` implementation, we already print out the interval type (i.e. that it's a month interval).

partiql-cli/src/test/kotlin/org/partiql/cli/io/DatumWriterTextPrettyTests.kt
alancai98: I already added some leading zero tests. See https://github.com/partiql/partiql-lang-kotlin/blob/1068fffb48581f87e0c53d3dc93b74a9da7b9576/partiql-cli/src/test/kotlin/org/partiql/cli/io/DatumWriterTextPrettyTests.kt#L77-L86

partiql-spi/src/test/java/org/partiql/spi/value/__snapshots__/DatumToStringTest.snap
xd1313113: Discussed offline, we will normalize the getter values. And remind users to pull full format to get accurate value.

partiql-cli/src/main/kotlin/org/partiql/cli/io/DatumWriterTextPretty.kt
```diff
@@ -327,7 +350,10 @@ class DatumWriterTextPretty(
             }
             IntervalCode.MINUTE_SECOND -> {
                 this.out.print("'")
-                this.out.print(datum.minutes)
+                if (datum.totalSeconds < 0 || datum.nanos < 0) {
```
xd1313113, line 353: The negative condition seems counter-intuitive due to java implementation, do we need expose an API in datum?

partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
```diff
@@ -279,6 +279,42 @@ default int getMonths() throws InvalidOperationException, NullPointerException {
...
+     * Returns the total amount of whole seconds for this day-time interval.
+     * </p>
+     * <p>
+     * This only applies to {@link PType#INTERVAL_DT}.
+     * </p>
+     * @return the total number of whole seconds.
+     * @throws InvalidOperationException when the type is not {@link PType#INTERVAL_DT}.
+     * @throws NullPointerException when {@link #isNull()} is true.
+     */
+    default long getTotalSeconds() throws InvalidOperationException, NullPointerException {
```
xd1313113, line 318: getTotalSeconds does not return total duration. Do you expect user to call `datum.nano` as well?

https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#:~:text=The%20total%20duration%20is%20defined%20by%20calling%20this%20method%20and%20getNano().

partiql-cli/src/main/kotlin/org/partiql/cli/io/DatumWriterTextPretty.kt
alancai98: IMO it's not counterintuitive. I'd rather not expose more functions in the public Datum API unless we need to.

partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
alancai98: Yes that's the expectation. It's included in the javadoc already that this returns just the whole seconds:

> Returns the total amount of whole seconds for this day-time interval.

Added to the javadoc about `nanos` in [`baec7f4` (#1815)](https://github.com/partiql/partiql-lang-kotlin/pull/1815/commits/baec7f47a19090098baf49eae21977016e55e325)

partiql-spi/src/main/java/org/partiql/spi/value/DatumComparator.java
```diff
@@ -558,24 +558,12 @@ private static DatumComparison[] fillTimeComparator(DatumComparison[] comps) {
...
-            long totalMonths = self.getYears() *  (long) 12 + self.getMonths();
-            long otherTotalMonths = other.getYears() * (long) 12 + other.getMonths();
-            return Long.compare(totalMonths, otherTotalMonths);
-        };
+        comps[INTERVAL_YM] = (self, other, comp) -> Long.compare(self.getTotalMonths(), self.getTotalMonths());
         comps[INTERVAL_DT] = (self, other, comp) -> {
-            long totalSeconds = self.getDays() * SECONDS_PER_DAY + self.getHours() * SECONDS_PER_HOUR + self.getMinutes() * SECONDS_PER_MINUTE + self.getSeconds();
-            long otherTotalSeconds = other.getDays() * SECONDS_PER_DAY + other.getHours() * SECONDS_PER_HOUR + other.getMinutes() * SECONDS_PER_MINUTE + other.getSeconds();
-            int comparison = Long.compare(totalSeconds, otherTotalSeconds);
+            int comparison = Long.compare(self.getTotalSeconds(), other.getTotalSeconds());
```
alancai98, line 566: self-review: this was another place we can simplify the computation between intervals

PR #1817, opened 2025-08-20
Add exception handling migration guide from v0.14.9 to v1

xd1313113:
## Relevant Issues

## Description
- Add exception handling migration guide from v0.14.9 to v1

## Other Information
Updated Unreleased Section in CHANGELOG: No, No feature released/change
Any backward-incompatible changes? NO
Any new external dependencies? NO
Do your changes comply with the [contributing](https://github.com/partiql/partiql-lang-kotlin/blob/main/CONTRIBUTING.md) and [code style](https://github.com/partiql/partiql-lang-kotlin/blob/main/CODE_STYLE.md) guidelines? YES

## License Information

### DIFF
--- a/docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
+++ b/docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
@@ -1,67 +1,66 @@
 package examples
 
 import org.junit.jupiter.api.assertThrows
-import org.partiql.lang.eval.EvaluationSession
 import org.partiql.parser.PartiQLParser
 import org.partiql.parser.PartiQLParserException
 import org.partiql.parser.SourceLocation
 import org.partiql.planner.PartiQLPlanner
-import java.util.*
+import java.util.Random
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertTrue
 
 class ErrorHandlingChanges {
-
+    /**
+     * The example shows when using {@link PartiQLParser} to parse string statement,
+     * the {@link PartiQLParserException} will be thrown. The parser will stop and throw the exception
+     *  when an error occurs.
+     */
     @Test
-    fun `Parser syntax error handling`() {
+    fun `Parser error handling`() {
         val parser = PartiQLParser.default()
 
         val exception = assertThrows<PartiQLParserException> {
-            parser.parse("SELECT * FRM abd")
+            try {
+                val parseResult = parser.parse("Create Table 'mytable")
+            } catch (e: PartiQLParserException) {
+                // rethrow for exception validation in test
+                throw e
+            }
         }
 
-        assertTrue { exception.message.contains("mismatched input 'FRM'") }
-        assertEquals(SourceLocation(1, 10, 52, 3), exception.location)
-    }
-
-    @Test
-    fun `Planner function type mismatch error`() {
-        val parser = PartiQLParser.default()
-        val planner = PartiQLPlanner.default()
-        val sess = PartiQLPlanner.Session(
-            queryId = UUID.randomUUID().toString(),
-            userId = "debug"
-        )
-
-        val parseResult = parser.parse("SELECT UPPER(123)")
-        val exception = assertThrows<PartiQLParserException> {
-            val result = planner.plan(parseResult.root, sess).plan
-        }
-
-        assertTrue { exception.message.contains("extraneous input '''") }
+        assertTrue { exception.message.contains("extraneous input") }
+        assertEquals("UNRECOGNIZED", exception.tokenType)
         assertEquals(SourceLocation(1, 14, 62, 1), exception.location)
     }
 
+    /**
+     * The example shows when using {@link PartiQLPlanner} to make plan from AST.
+     * There is no specific exception defined in the planner, so the planner will throw any exception
+     *  when an error occurs.
+     */
     @Test
-    fun `Expression evaluation error`() {
-
+    fun `PLanner evaluation error`() {
         val parser = PartiQLParser.default()
         val planner = PartiQLPlanner.default()
-        val sess = PartiQLPlanner.Session(
-            queryId = UUID.randomUUID().toString(),
-            userId = "debug"
+        val session = PartiQLPlanner.Session(
+            queryId = Random().nextInt().toString(),
+            userId = "test-user",
+            currentCatalog = "default",
+            catalogs = mapOf()
         )
 
-        val session = EvaluationSession.builder().build()
-
-        val parseResult = parser.parse("SELECT UPPER(123)")
-        val exception = assertThrows<PartiQLParserException> {
-            val result = planner.plan(parseResult.root, sess).plan
-            result
+        val exception = assertThrows<Exception> {
+            try {
+                val parseResult = parser.parse("Create Table mytable")
+                val planResult = planner.plan(parseResult.root, session)
+            } catch (e: Exception) {
+                // PartiQLParserException will be thrown during PartiQL process the query in the major PartiQL components
+                throw e
+            }
         }
 
-        assertTrue { exception.message.contains("extraneous input '''") }
-        assertEquals(SourceLocation(1, 14, 62, 1), exception.location)
+        assertTrue(exception is IllegalArgumentException)
+        assertTrue(exception.message!!.contains("Unsupported statement"))
     }
 }
--- a/docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
+++ b/docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
@@ -17,8 +17,8 @@ import kotlin.test.assertTrue
 
 class ErrorHandlingChanges {
     /**
-     * The example shows the error handling when using [PartiQLParser] to parse a string statement,
-     * When an error occurs, the parser will stop the parsing process by throwing the [PartiQLParserException].
+     * The example shows error handling when using [PartiQLParser] to parse a string statement,
+     * When an error occurs, the parser will stop the parsing process by throwing [PartiQLParserException].
      */
     @Test
     fun `PartiQLParser default error handling`() {
@@ -29,12 +29,12 @@ class ErrorHandlingChanges {
                 // Extra single quote is added
                 val parseResult = parser.parse("SELECT * FROM 'mytable")
             } catch (e: PartiQLParserException) {
-                // Rethrow for validating the exception in test
+                // Rethrow for validating the exception in the test
                 throw e
             }
         }
 
-        // This is a pain point with using the v0.14.9 parser. There's no proper way to extract out error information
+        // This is a pain point when using the v0.14.9 parser. There's no proper way to extract error information
         // other than looking at the error message string.
         assertTrue { exception.message.contains("extraneous input") }
         assertEquals("UNRECOGNIZED", exception.tokenType)
@@ -43,11 +43,11 @@ class ErrorHandlingChanges {
 
     /**
      *  [PartiQLPlanner] transforms an Abstract Syntax Tree (AST) from the parser into a logical query plan.
-     *  The planner does not throw for most errors. If errors occur, planner will collect multiple errors during the
+     *  The planner does not throw for most errors. If errors occur, the planner will collect multiple errors during the
      * planning phase. The planner provides an optional callback for you to handle those errors.
      *
-     * But the planner may still throw exception in certain circumstances.
-     * This example shows the error handling behavior when making a plan from a valid AST using the [PartiQLPlanner]
+     * However, the planner may still throw exception in certain circumstances.
+     * This example shows the error handling behavior when making a plan from a valid AST using [PartiQLPlanner]
      */
     @Test
     fun `PartiQLPlanner default error handling`() {
@@ -60,13 +60,13 @@ class ErrorHandlingChanges {
             catalogs = mapOf()
         )
 
-        // CREATE TABLE is not supported by planner currently
+        // The planner does not support CREATE TABLE currently
         val parseResult = parser.parse("CREATE TABLE mytable")
         val exception = assertThrows<Exception> {
             try {
                 val planResult = planner.plan(parseResult.root, session)
             } catch (e: Exception) {
-                // Rethrow for validating the exception in test
+                // Rethrow for validating the exception in the test
                 throw e
             }
         }
@@ -105,7 +105,7 @@ class ErrorHandlingChanges {
     }
 
     /**
-     * This example shows the error handling behavior of PartiQL evaluation phase.
+     * This example shows the error handling behavior of the PartiQL evaluation phase.
      */
     @Test
     fun `PartiQL evaluation phase default error handling`() {
--- a/docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
+++ b/docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
@@ -24,7 +24,7 @@ class ErrorHandlingChanges {
             try {
                 val parseResult = parser.parse("Create Table 'mytable")
             } catch (e: PartiQLParserException) {
-                // rethrow for exception validation in test
+                // Rethrow for validating the exception in test
                 throw e
             }
         }
@@ -55,7 +55,7 @@ class ErrorHandlingChanges {
                 val parseResult = parser.parse("Create Table mytable")
                 val planResult = planner.plan(parseResult.root, session)
             } catch (e: Exception) {
-                // PartiQLParserException will be thrown during PartiQL process the query in the major PartiQL components
+                // Rethrow for validating the exception in test
                 throw e
             }
         }
--- a/docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
+++ b/docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
@@ -1,0 +1,67 @@
+package examples
+
+import org.junit.jupiter.api.assertThrows
+import org.partiql.lang.eval.EvaluationSession
+import org.partiql.parser.PartiQLParser
+import org.partiql.parser.PartiQLParserException
+import org.partiql.parser.SourceLocation
+import org.partiql.planner.PartiQLPlanner
+import java.util.*
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class ErrorHandlingChanges {
+
+    @Test
+    fun `Parser syntax error handling`() {
+        val parser = PartiQLParser.default()
+
+        val exception = assertThrows<PartiQLParserException> {
+            parser.parse("SELECT * FRM abd")
+        }
+
+        assertTrue { exception.message.contains("mismatched input 'FRM'") }
+        assertEquals(SourceLocation(1, 10, 52, 3), exception.location)
+    }
+
+    @Test
+    fun `Planner function type mismatch error`() {
+        val parser = PartiQLParser.default()
+        val planner = PartiQLPlanner.default()
+        val sess = PartiQLPlanner.Session(
+            queryId = UUID.randomUUID().toString(),
+            userId = "debug"
+        )
+
+        val parseResult = parser.parse("SELECT UPPER(123)")
+        val exception = assertThrows<PartiQLParserException> {
+            val result = planner.plan(parseResult.root, sess).plan
+        }
+
+        assertTrue { exception.message.contains("extraneous input '''") }
+        assertEquals(SourceLocation(1, 14, 62, 1), exception.location)
+    }
+
+    @Test
+    fun `Expression evaluation error`() {
+
+        val parser = PartiQLParser.default()
+        val planner = PartiQLPlanner.default()
+        val sess = PartiQLPlanner.Session(
+            queryId = UUID.randomUUID().toString(),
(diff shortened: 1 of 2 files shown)


docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,66 @@
...
+     * The example shows when using {@link PartiQLParser} to parse string statement,
+     * the {@link PartiQLParserException} will be thrown. The parser will stop and throw the exception
+     *  when an error occurs.
+     */
+    @Test
+    fun `Parser error handling`() {
+        val parser = PartiQLParser.default()
+
+        val exception = assertThrows<PartiQLParserException> {
+            try {
```
XuechunHHH, line 24: nit: using `assertFailsWith` instead of the `try-catch` block

XuechunHHH, line 28: nit: using `assertFailsWith` or `assertThrows` instead of the `try-catch` block

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,164 @@
...
+        val planner = PartiQLPlanner.standard()
+        val session = Session.empty()
+
+        val query = """
+            CREATE TABLE 'mytable;
+        """.trimIndent()
+
+        val exception = assertThrows<PRuntimeException> {
+            try {
+                val parseResult = parser.parse(query)
```
XuechunHHH: v0.14 uses `parseResult.root` while v1 uses `parseResult.statements[0]`. Is this an API change or just different usage patterns for the inconsistency?

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,164 @@
...
+        val parser = PartiQLParser.standard()
+        val planner = PartiQLPlanner.standard()
+        val session = Session.empty()
+
+        val query = """
+            CREATE TABLE 'mytable;
+        """.trimIndent()
+
+        val exception = assertThrows<PRuntimeException> {
+            try {
```
XuechunHHH, line 69: nit: same here for the `try-catch` block

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,66 @@
...
+     * The example shows when using {@link PartiQLPlanner} to make plan from AST.
+     * There is no specific exception defined in the planner, so the planner will throw any exception
+     *  when an error occurs.
+     */
+    @Test
+    fun `PLanner evaluation error`() {
+        val parser = PartiQLParser.default()
+        val planner = PartiQLPlanner.default()
+        val session = PartiQLPlanner.Session(
+            queryId = Random().nextInt().toString(),
```
XuechunHHH: Why we put a `Random()` instead of a fixed id here?

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,66 @@
...
+            userId = "test-user",
+            currentCatalog = "default",
+            catalogs = mapOf()
+        )
+
+        val exception = assertThrows<Exception> {
+            try {
```
XuechunHHH, line 66: nit: same here for the `try-catch` block

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,66 @@
...
+        assertEquals(SourceLocation(1, 14, 62, 1), exception.location)
+    }
+
+    /**
+     * The example shows when using {@link PartiQLPlanner} to make plan from AST.
+     * There is no specific exception defined in the planner, so the planner will throw any exception
+     *  when an error occurs.
+     */
+    @Test
+    fun `PLanner evaluation error`() {
```
XuechunHHH: typo: Planner

XuechunHHH (commented):
Comprehensive coverage and good examples. Left some comments on minor places.

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,66 @@
...
+import org.partiql.planner.PartiQLPlanner
+import java.util.Random
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class ErrorHandlingChanges {
+    /**
+     * The example shows when using {@link PartiQLParser} to parse string statement,
+     * the {@link PartiQLParserException} will be thrown. The parser will stop and throw the exception
```
alancai98: Let's use the kdoc (Kotlin's equivalent for javadoc) here rather than javadoc, so the code references work - https://kotlinlang.org/docs/kotlin-doc.html. Essentially need to wrap the code reference in square brackets e.g. `[PartiQLParser]`.

Similarly for other kdoc in this file and the v1 examples.

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,66 @@
...
+            try {
+                val parseResult = parser.parse("Create Table 'mytable")
+            } catch (e: PartiQLParserException) {
+                // Rethrow for validating the exception in test
+                throw e
+            }
+        }
+
+        assertTrue { exception.message.contains("extraneous input") }
+        assertEquals("UNRECOGNIZED", exception.tokenType)
```
alancai98, line 40: I'd add a comment highlighting that this is a pain point with using the 0.14.9 parser. There's no proper way to extract out error information other than looking at the error message string.

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,66 @@
...
+     * the {@link PartiQLParserException} will be thrown. The parser will stop and throw the exception
+     *  when an error occurs.
+     */
+    @Test
+    fun `Parser error handling`() {
+        val parser = PartiQLParser.default()
+
+        val exception = assertThrows<PartiQLParserException> {
+            try {
+                val parseResult = parser.parse("Create Table 'mytable")
```
alancai98: nit: can add a comment briefly explaining that this query has an error

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,66 @@
...
+        val session = PartiQLPlanner.Session(
+            queryId = Random().nextInt().toString(),
+            userId = "test-user",
+            currentCatalog = "default",
+            catalogs = mapOf()
+        )
+
+        val exception = assertThrows<Exception> {
+            try {
+                val parseResult = parser.parse("Create Table mytable")
```
alancai98: nit: probably a more illustrative example of a planner error would be to use a query rather than a `DDL` statement. something like a call to a function that doesn't exist in the catalog (e.g. `function_does_not_exist()`).

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,66 @@
...
+            }
+        }
+
+        assertTrue { exception.message.contains("extraneous input") }
+        assertEquals("UNRECOGNIZED", exception.tokenType)
+        assertEquals(SourceLocation(1, 14, 62, 1), exception.location)
+    }
+
+    /**
+     * The example shows when using {@link PartiQLPlanner} to make plan from AST.
```
alancai98: Can rephrase this to some like
```
This example shows the error handling behavior when making a plan from a valid AST using the [PartiQLPlanner]
```

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,66 @@
...
+     * There is no specific exception defined in the planner, so the planner will throw any exception
+     *  when an error occurs.
```
alancai98: What do you mean by the first part of this sentence?

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,66 @@
...
+            } catch (e: Exception) {
+                // Rethrow for validating the exception in test
+                throw e
+            }
+        }
+
+        assertTrue(exception is IllegalArgumentException)
+        assertTrue(exception.message!!.contains("Unsupported statement"))
+    }
+}
```
alancai98, line 130: We should also add an error-handling example with the 0.14.9 evaluator.

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,164 @@
...
+
+    /**
+     * The example shows the default behavior of error handling in major PartiQL components {@link PartiQLParser},
+     * {@link PartiQLPlanner} and {@link PartiQLCompiler}. {@link PRuntimeException} will be thrown from the PartiQL system
+     * to trap any parsing errors or warnings. {@link PRuntimeException} wraps a {@link PError} which includes {@link PError#code()},
+     * {@link PError#kind}, {@link PError#severity}, {@link PError#location} and {@link PError#properties}.
+     * {@link PRuntimeException} with some enhanced information allows us for more robust and customized error control.
+     */
+    @Test
+    fun `Default PartiQL error handling`() {
```
alancai98: For these examples, let's try to keep the naming and formatting consistent. In the v0.14.9 examples, you started off with showing a parser example, planner example, (and should also include an evaluation example). It's hard to follow the differences between versions if we change the naming and content of the examples. We can leave the additional example related to custom error handlers as an extra example in v1.

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,164 @@
...
+        val compiler = PartiQLCompiler.standard()
+        val parser = PartiQLParser.standard()
+        val planner = PartiQLPlanner.standard()
+        val session = Session.empty()
```
alancai98: These aren't actually used because the parser will give an error for this example and can be removed. We should keep the content of this example the same as the v0.14.9 `Parser error handling` example.

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
alancai98: This was brought up in the parser migration guide but basically the v1 parser can support parsing multiple statements. Hence why the v1 uses the list syntax to retrieve the first parsed statement.

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,164 @@
...
+        val context = Context.of(listener)
+
+        val compiler = PartiQLCompiler.standard()
+        val parser = PartiQLParser.standard()
+        val planner = PartiQLPlanner.standard()
+        val session = Session.empty()
+
+        val query = """
+            CREATE TABLE mytable;
+            INSERT INTO 'mytable VALUES (2,3);
```
alancai98: Similar to the comment in v0.14.9, let's just use normal queries (e.g. select-from-where, top-level expressions) rather than DDL/DML. PLK's implementation of DDL/DML (especially to everything after parsing) is pretty sparse. The majority of the present customers are just doing queries, so it may be helpful to keep the examples using queries.

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /**
+     * The example shows we registered a custom error handling in the context of major PartiQL components {@link PartiQLParser},
+     * {@link PartiQLPlanner} and {@link PartiQLCompiler} which records the errors or warning instead of throwing for deferred error processing.
+     * You can choose to treat warning as errors, or inhibit certain warnings or even ignore warnings.
+     * It might be a good idea for you to consolidate your logic of converting PErrors to user-presented messages and use that here.
+     */
+    @Test
+    fun `PartiQLParser with custom error handler`() {
```
alancai98: This example is demonstrating too many things IMO. We should break it up to cover a planner error example and have a separate example for a custom error handler.

alancai98 (commented):
We're off to a good start with the included examples. A few main things we should address regarding the content of the examples
- include a parser, planner, and evaluation example for v0.14.9 and v1. currently v0.14.9 does not include any evaluation example
- keep the naming and content of the examples as similar as possible to make it easier to follow. both the v0.14.9 and v1 examples should include the parser, planner, evaluation examples. v1 can also include how to use a custom error listener as another example
- try to use DQL (select-from-where or top-level expressions) rather than DDL/DML in the examples since that's what's better supported in PLK and used by more of our customers

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
xd1313113: This test is more of demonstration. I intentionally include try catch which is similar to user's code to demonstrate it will catch that exception.

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
xd1313113: Same reason for demonstration

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,237 @@
...
+import org.partiql.spi.errors.PErrorListener
+import org.partiql.spi.errors.PRuntimeException
+import org.partiql.spi.errors.Severity
+import org.partiql.spi.types.PType
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+/**
+ * A unified error handling is introduced PartiQL system. [PRuntimeException] will be thrown from all components in PartiQL system
+ * to trap any errors or warnings. [PRuntimeException] warps a [PError] Java class which exposes information for users to write
```
alancai98:
```suggestion
 * to trap any errors or warnings. [PRuntimeException] wraps a [PError] which exposes information for users to write
```

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,237 @@
...
+ * high quality error messages.
+ *
+ * Also [PErrorListener] is introduced to allow customized error handling. You can register a [PErrorListener] in context of
+ * major PartiQL components [PartiQLParser], [PartiQLPlanner] and [PartiQLCompiler]. It allows you to inspect each PError
+ * the component emitted and decide action on error. You may treat all warnings as errors, suppress all warnings, delay the error handling
+ * or throw custom exception with an error message.
+ *
+ * Note for the output when you choose to delay the error handling. The component will not stop and still return an output object
+ * In case the severity is [Severity.ERROR], the output may be incorrect or incomplete. You should discard the result.
```
alancai98: Not always the case you need to discard the result.

```suggestion
 * In case the severity is [Severity.ERROR], the output may be incorrect or incomplete. You may need to discard the result.
```

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,237 @@
...
+class ErrorHandlingChanges {
+    class CustomPErrorListener : PErrorListener {
+        val errorCollection: MutableList<PError> = mutableListOf()
+
+        override fun report(error: PError) {
+            when (error.severity.code()) {
+                // You may choose to customize the error message and then halt the flow by throwing the error, or you can record the error and then throw when it reaches maximum count
+                // In this case, we record only without throwing
+                Severity.ERROR -> errorCollection.add(error)
+                // You may choose to customize the error message, treat warning as an error and then halt the flow by throwing, or you can suppress all warnings.
```
alancai98: nit: let's add some line breaks for these comments. they are currently pretty long

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,237 @@
...
+            } catch (e: PRuntimeException) {
+                // Rethrow for validating the exception in test
+                throw e
+            }
+        }
+
+        assertEquals(exception.error.code(), PError.UNEXPECTED_TOKEN)
+        assertEquals(PErrorKind.SYNTAX, exception.error.kind.code())
+        assertEquals(Severity.ERROR, exception.error.severity.code())
+        assertEquals(SourceLocation(1, 15, 1), exception.error.location)
```
alancai98, line 82: Let's add some comments to highlight that in v1 (as opposed to v0.14.9), we can extract more relevant error information without needing to do string parsing.

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,237 @@
...
+        // Inspect properties from exception
+        assertEquals("UNRECOGNIZED", exception.error.get("TOKEN_NAME", String::class.java))
+    }
+
+    /**
+     * The example shows the error handling with custom [PErrorListener], which delays the error handling
+     * when using [PartiQLParser] to parse multiple string statements.
```
alancai98:
```suggestion
     * The example shows the error handling with the [CustomPErrorListener], which captures multiple
     * errors rather than throwing on the first error.
```

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,237 @@
...
+        val parser = PartiQLParser.standard()
+        val planner = PartiQLPlanner.standard()
+        val session = Session.empty()
+
+        // Invalid expression adding number and string
+        val parseResult = parser.parse("SELECT 1 + 'string' FROM mytable")
+        assertDoesNotThrow {
+            val plan = planner.plan(parseResult.statements[0], session, context).plan
+        }
+        assertEquals(2, listener.errorCollection.size)
```
alancai98, line 200: Should call out somewhere that there are two errors, not just the number and string addition.

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/ErrorHandlingChanges.kt
```diff
@@ -0,0 +1,245 @@
...
+     * This example shows the error handling behavior that the [PartiQLPlanner] can proceed with most errors
+     * or warnings and will not throw.
+     */
+    @Test
+    fun `PartiQLPlanner default error handling which does not throw`() {
+        val parser = PartiQLParser.standard()
+        val planner = PartiQLPlanner.standard()
+        val session = Session.empty()
+
+        // There are two warnings in when .plan is called on planner
```
alancai98:
```suggestion
        // There are two warnings when calling the planner on the following query:
```

PR #1823, opened 2025-08-29
Multiple bug fixes related to timestampz

xd1313113:
## Relevant Issues
- [Related To] partiql/partiql-tests#156 

## Description
- Fix the timestampz print and parsing in conformance tests file in ION

## Other Information
- Any backward-incompatible changes?  NO

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,8 @@ rewritten to their query representation
 - Remove the extra non-character Unicode in the function name when displaying cli errors 
 - **EXPERIMENTAL** Add `Datum.getTotalMonths` to retrieve the total number of months from a year-month `Datum`
 - **EXPERIMENTAL** Add `Datum.getTotalSeconds` to retrieve the total number of seconds from a day-time `Datum`
+- fix the parsing of DATE_ADD/DATE_DIFF with an uncapitalized datetime field argument.
+- fix that wrong offset is set when parsing Ion timestamp with time zone into Datum.
 
 ### Changed
 


partiql-ast/src/main/java/org/partiql/ast/DatetimeField.java
```diff
@@ -119,7 +119,7 @@ public String name() {
 
     @NotNull
     public static DatetimeField parse(@NotNull String value) {
-        switch (value) {
+        switch (value.toUpperCase()) {
```
alancai98: For consistency with the other `parse` calls in the codebase, we should call `.toUpperCase()`/`uppercase` at the call site and not within the `parse` call -- https://github.com/partiql/partiql-lang-kotlin/blob/18541659dbd1129f83ad6e6caa9a3855dea9bb97/partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt#L1874

alancai98: We can probably later decide if we want all of the static `parse` function calls should call `.toUpperCase()` by default so that the call sites don't have to call `.toUpperCase` themselves. That could be a better long-term option.

XuechunHHH (approved):

partiql-ast/src/main/java/org/partiql/ast/DatetimeField.java
xd1313113: agree. we should make it consistent for now and decide the common pattern of capitalization for PartiQL keyword

xd1313113:
> We should also update the CHANGELOG since this is a bug fix to a public API and also fixes the parsing of `DATE_ADD`/`DATE_DIFF` with an uncapitalized datetime field argument.

Added

PR #1827, opened 2025-09-04
add: migration guides for evaluation/Datum

XuechunHHH:
## Relevant Issues

## Description
- Add migration examples for evaluation/Datum changes from v0.14.9 to v1.2.2

## Other Information
- Updated Unreleased Section in CHANGELOG: **[NO]**: No changes in implementation

## License Information

### DIFF
--- a/docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/EvaluationChanges.kt
+++ b/docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/EvaluationChanges.kt
@@ -1,0 +1,225 @@
+package examples
+
+import org.partiql.lang.CompilerPipeline
+import org.partiql.lang.eval.CompileOptions
+import org.partiql.lang.eval.EvaluationSession
+import org.partiql.lang.eval.ExprFunction
+import org.partiql.lang.eval.ExprValue
+import org.partiql.lang.eval.ExprValueType
+import org.partiql.lang.eval.PartiQLResult
+import org.partiql.lang.eval.TypingMode
+import org.partiql.lang.eval.namedValue
+import org.partiql.lang.types.FunctionSignature
+import org.partiql.types.StaticType.Companion.STRING
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+
+class EvaluationChanges {
+    @Test
+    fun `safe exprvalue access pattern`() {
+        // In PLK v0.14.9, values are represented as ExprValue
+        // Safe access requires checking type and using scalar interface
+        fun getDataAsString(value: ExprValue): String? {
+            val type = value.type
+            // Check type and handle null/missing manually
+            return when (type) {
+                ExprValueType.NULL -> "null"
+                ExprValueType.MISSING -> "missing"
+                ExprValueType.INT -> value.scalar.numberValue().toString()
+                ExprValueType.STRING -> value.scalar.stringValue()
+                ExprValueType.BOOL -> value.scalar.booleanValue().toString()
+                ExprValueType.FLOAT -> value.scalar.numberValue().toString()
+                ExprValueType.DECIMAL -> value.scalar.numberValue().toString()
+                else -> throw RuntimeException("Unsupported type: $type")
+            }
+        }
+        assertEquals("42", getDataAsString(ExprValue.newInt(42)))
+        assertEquals("hello", getDataAsString(ExprValue.newString("hello")))
+        assertEquals("null", getDataAsString(ExprValue.nullValue))
+        assertEquals("missing", getDataAsString(ExprValue.missingValue))
+    }
+
+    // Helper functions for orderly materialization test
+    // These helper functions memoize the data after iterating by return an ExprValue,
+    // which ensuring that the data is backed by in-memory data structures instead of PartiQL engine.
+    // Dispatches to appropriate materialization function based on data type
+    fun materializeInOrder(value: ExprValue): ExprValue {
+        val type = value.type
+        return when (type) {
+            ExprValueType.LIST -> materializeList(value)
+            ExprValueType.BAG -> materializeBag(value)
+            ExprValueType.SEXP -> materializeSexp(value)
+            ExprValueType.STRUCT -> materializeStruct(value)
+            else -> materializeScalar(value)
+        }
+    }
+
+    // Materializes list elements in order by iterating through each element
+    // Back the elements with a List
+    fun materializeList(value: ExprValue): ExprValue {
+        val elements = mutableListOf<ExprValue>()
+        value.forEach { element ->
+            elements.add(materializeInOrder(element))
+        }
+        return ExprValue.newList(elements)
+    }
+
+    // Materializes bag elements in order by iterating through each element
+    // Back the elements with a List
+    fun materializeBag(value: ExprValue): ExprValue {
+        val elements = mutableListOf<ExprValue>()
+        value.forEach { element ->
+            elements.add(materializeInOrder(element))
+        }
+        return ExprValue.newBag(elements)
+    }
+
+    // Materializes sexp elements in order by iterating through each element
+    // Back the elements with a List
+    fun materializeSexp(value: ExprValue): ExprValue {
+        val elements = mutableListOf<ExprValue>()
+        value.forEach { element ->
+            elements.add(materializeInOrder(element))
+        }
+        return ExprValue.newSexp(elements)
+    }
+
+    // Materializes struct fields in order by iterating through each field value
+    // Back the fields with a List
+    fun materializeStruct(value: ExprValue): ExprValue {
+        val fields = mutableListOf<ExprValue>()
+        for (field in value) {
+            val materializedField = materializeInOrder(field)
+            fields.add(materializedField)
+        }
+        return ExprValue.newStruct(fields, org.partiql.lang.eval.StructOrdering.UNORDERED)
+    }
+
+    // Materializes scalar values by accessing their primitive data
+    fun materializeScalar(value: ExprValue): ExprValue {
+        if (value.type == ExprValueType.NULL || value.type == ExprValueType.MISSING) {
+            return value
+        }
+        val type = value.type
+        when (type) {
+            ExprValueType.INT -> value.scalar.numberValue()
+            ExprValueType.STRING -> value.scalar.stringValue()
+            ExprValueType.BOOL -> value.scalar.booleanValue()
+            // ... rest of scalar types
+            else -> throw RuntimeException("Unsupported type: $type")
+        }
+        return value
+    }
+
+    @Test
+    fun `orderly materialization of semi-structured data`() {
+        // In PLK 0.14.9, semi-structured data must be materialized in a predictable order
+        // This is critical for stateful processes that depend on consistent traversal patterns
+        // which means for semi-structured type, we need to fully materialize its first element
+        // before the second one (and so on).
+        val bagValue = ExprValue.newBag(
+            listOf(
+                ExprValue.newInt(42),
+                ExprValue.newStruct(
+                    listOf(
+                        ExprValue.newString("value").namedValue(ExprValue.newString("nested"))
+                    ),
+                    org.partiql.lang.eval.StructOrdering.UNORDERED
+                )
+            )
+        )
+        val structValue = ExprValue.newStruct(
+            listOf(
+                ExprValue.newInt(42).namedValue(ExprValue.newString("id")),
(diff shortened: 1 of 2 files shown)


docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/EvaluationChanges.kt
```diff
@@ -0,0 +1,240 @@
...
+                else -> throw RuntimeException("Unsupported type: $type")
+            }
+        }
+        assertEquals("42", getDataAsString(Datum.integer(42)))
+        assertEquals("hello", getDataAsString(Datum.string("hello")))
+        assertEquals("null", getDataAsString(Datum.nullValue()))
+        assertEquals("missing", getDataAsString(Datum.missing()))
+    }
+
+    // Helper functions for orderly materialization test
```
xd1313113, line 68: Is this the same materialization helper with [partiql-eval/src/test/kotlin/org/partiql/eval/internal/DatumMaterialize.kt](https://github.com/partiql/partiql-lang-kotlin/blob/1a0642e172fa96de7dda88f6f10d7085b45174c8/partiql-eval/src/test/kotlin/org/partiql/eval/internal/DatumMaterialize.kt). The current one looks more complete(handling all scalar types) and concise.

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/EvaluationChanges.kt
```diff
@@ -0,0 +1,240 @@
...
+        }
+        return d
+    }
+
+    @Test
+    fun `orderly materialization of semi-structured data`() {
+        // In PLK v1, semi-structured data must be materialized in a predictable order
+        // This is critical for stateful processes that depend on consistent traversal patterns
+        // which means for semi-structured type, we need to fully materialize its first element
+        // before the second one (and so on).
```
xd1313113, line 150: In what use cases, does the user need to materialize the Datum? For comparison?

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/EvaluationChanges.kt
```diff
@@ -0,0 +1,240 @@
...
+        // Access variables by depth and offset
+        val param1 = env2.get(0, 0)
+        val param2 = env2.get(1, 1)
+        assertEquals("b", param1.string)
+        assertEquals(1, param2.int)
+    }
+
+    @Test
+    fun `working with compiler builder and strategies`() {
+        // In PLK v1, PartiQLCompiler can be built with custom strategies
```
xd1313113, line 198: Can you explain more what custom strategies is in complier?

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/EvaluationChanges.kt
XuechunHHH: Yes. Put some of them here intentionally to demonstrate how the materialization actually works.

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/EvaluationChanges.kt
XuechunHHH: Some use cases might make sense:

- When comparing data elements or performing sorting operations
- When processing data backed by stateful operations (like PartiQL queries)
- When transforming data between different formats
- When analyzing data that requires complete element access

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/EvaluationChanges.kt
```diff
@@ -0,0 +1,240 @@
...
+            ) {
+                    override fun apply(
+                        match: Match,
+                        mode: Mode,
+                        callback: Callback
+                    ): Expr {
+                        return ExprValue { Datum.string("custom strategy") }
+                    }
+                }
+            )
```
XuechunHHH, line 196: In this example, creating a custom strategy will give a string result "custom strategy".

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/EvaluationChanges.kt
XuechunHHH: In the following example, creating a custom strategy will give a string result "custom strategy". See:
```
            .addStrategy(object : Strategy(
                Pattern(Operator::class.java)
            ) {
                    override fun apply(
                        match: Match,
                        mode: Mode,
                        callback: Callback
                    ): Expr {
                        return ExprValue { Datum.string("custom strategy") }
                    }
                }
            )

        val result = statement.execute()
        assertEquals("custom strategy", result.string)
```

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/EvaluationChanges.kt
```diff
@@ -0,0 +1,225 @@
...
+            }
+        }
+        assertEquals("42", getDataAsString(ExprValue.newInt(42)))
+        assertEquals("hello", getDataAsString(ExprValue.newString("hello")))
+        assertEquals("null", getDataAsString(ExprValue.nullValue))
+        assertEquals("missing", getDataAsString(ExprValue.missingValue))
+    }
+
+    // Helper functions for orderly materialization test
+    // These helper functions memoize the data after iterating by return an ExprValue,
```
alancai98:
```suggestion
    // These helper functions materialize the data after iterating by return an ExprValue,
```

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/EvaluationChanges.kt
```diff
@@ -0,0 +1,225 @@
...
+        assertEquals("a", param1.scalar.stringValue())
+        assertEquals(2L, param2.scalar.numberValue())
+    }
+
+    @Test
+    fun `working with compiler pipeline builder`() {
+        // In PLK 0.14.9, compiler pipeline customization is done through custom functions and procedures
+        val pipeline = CompilerPipeline.builder()
+            .addFunction(object : ExprFunction {
+                override val signature = FunctionSignature(
```
alancai98, line 188: Would be helpful to add short comment explaining what this function does (i.e. it just returns back the string `"custom strategy"`). Similarly for the `Datum` v1 example

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/EvaluationChanges.kt
```diff
@@ -0,0 +1,225 @@
...
+            val type = value.type
+            // Check type and handle null/missing manually
+            return when (type) {
+                ExprValueType.NULL -> "null"
+                ExprValueType.MISSING -> "missing"
+                ExprValueType.INT -> value.scalar.numberValue().toString()
+                ExprValueType.STRING -> value.scalar.stringValue()
+                ExprValueType.BOOL -> value.scalar.booleanValue().toString()
+                ExprValueType.FLOAT -> value.scalar.numberValue().toString()
+                ExprValueType.DECIMAL -> value.scalar.numberValue().toString()
```
alancai98: May be helpful to show for non-scalar values like arrays/bags/structs

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/EvaluationChanges.kt
```diff
@@ -0,0 +1,240 @@
...
+            }
+        }
+        assertEquals("42", getDataAsString(Datum.integer(42)))
+        assertEquals("hello", getDataAsString(Datum.string("hello")))
+        assertEquals("null", getDataAsString(Datum.nullValue()))
+        assertEquals("missing", getDataAsString(Datum.missing()))
+    }
+
+    // Helper functions for orderly materialization test
+    // These helper functions memoize the data after iterating by return a Datum,
```
alancai98:
```suggestion
    // These helper functions materialize the data after iterating by return a Datum,
```

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/EvaluationChanges.kt
```diff
@@ -0,0 +1,225 @@
...
+        assertEquals("hello", getDataAsString(ExprValue.newString("hello")))
+        assertEquals("null", getDataAsString(ExprValue.nullValue))
+        assertEquals("missing", getDataAsString(ExprValue.missingValue))
+    }
+
+    // Helper functions for orderly materialization test
+    // These helper functions memoize the data after iterating by return an ExprValue,
+    // which ensuring that the data is backed by in-memory data structures instead of PartiQL engine.
+    // Dispatches to appropriate materialization function based on data type
+    fun materializeInOrder(value: ExprValue): ExprValue {
```
alancai98, line 55: I'm not sure I fully understand these `materialize` functions. They seem to be going through all the scalar and nested values but still returning back an `ExprValue`. Perhaps it would be more illustrative to show the materialization of an `ExprValue` or a `Datum` to a string for example like you had done in the earlier example.

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/EvaluationChanges.kt
XuechunHHH: It intentionally to be `memoize` here. The helper functions here not only doing materialization but also using `Datum` (in-memory data structure) to store it. This behavior is recommended to avoid multiple times iterating or data accessing.

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/EvaluationChanges.kt
XuechunHHH: The key idea of this example is that when having a `Datum` that's backed by a PartiQL query (lazy load), we want to fully evaluate and store its contents in memory rather than re-executing the query each time we access it. 

See: https://partiql.org/plk/1.0/datum.html#_orderly_invocation_of_datums_apis

docs/upgrades/v0.14-to-v1-upgrade/upgraded-examples/src/test/kotlin/examples/EvaluationChanges.kt
XuechunHHH: Same here. What I mean is `memoize` here.

docs/upgrades/v0.14-to-v1-upgrade/examples/src/test/kotlin/examples/EvaluationChanges.kt
alancai98: I see. The referenced link provides some more context on where the memoize description comes from. To me, memoize generally means performing some potentially expensive computation (such as evaluating) and then storing the result of the call for future references.

I think it's a fair callout that users should memoize the data after iterating on the `ExprValue`/`Datum`. But the `materializeInOrder` function is not actually storing the result for future reuse. Similarly, the calls to the function are not memoizing the result.

To make the example a bit more clear I'd either
1. Leave out the memoization references in the comments or
2. In the **calls** to this function, actually reuse the memoized result + update the comment to clarify the **calls** to the function are memoizing the result (not the function itself).

XuechunHHH: The calls to this function should actually reuse the 'memoized' result. Seems like I committed the version before adding the 'memoize' behavior. 

And yes, the word 'memoize' is somehow confusing. Do you think using 'cache' or something else will be better here?

alancai98: Ok, adding the reuse of the materialized result makes more sense now.

> And yes, the word 'memoize' is somehow confusing. Do you think using 'cache' or something else will be better here?

Both words are pretty synonymous. I think it's fine to keep it as memoize since the website guide on `Datum` uses memoize as well.

PR #1819, opened 2025-08-25
Redefine non-standard DATE_ADD to interval arithmetic #1690

xd1313113:
## Relevant Issues
- [Closes/Related To] Issue #1690 

## Description
- Reimplemented DATE_ADD with interval arithmetic.

## Other Information
- Updated Unreleased Section in CHANGELOG: YES
- Any backward-incompatible changes? NO
- Any new external dependencies? NO
- Do your changes comply with the [contributing][cg] and [code style][csg] guidelines? YES

## License Information

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

[cg]: https://github.com/partiql/partiql-lang-kotlin/blob/main/CONTRIBUTING.md
[csg]: https://github.com/partiql/partiql-lang-kotlin/blob/main/CODE_STYLE.md

PR #1828, opened 2025-09-05
add extend data into conformance test runner

xd1313113:
## Relevant Issues
- [Related To] Issue #1819 

## Description
- < Explain >

## Other Information
- Updated Unreleased Section in CHANGELOG: No, only tested
- Any backward-incompatible changes? No

## License Information


test/partiql-tests-runner/src/test/resources/config/eval/skip-eval-extended.txt
xd1313113, line 1: the file contains failed tests for now

XuechunHHH:
Have couple of questions on this PR:

- Why do we want to add the extended tests for now? Is it particularly for `DATE_DIFF` and `DATE_ADD`?
- Will it also be triggered for the Github runner?
- Will the generated report include these extended tests?

xd1313113:
> Have couple of questions on this PR:
> 
> * Why do we want to add the extended tests for now? Is it particularly for `DATE_DIFF` and `DATE_ADD`?
> * Will it also be triggered for the Github runner?
> * Will the generated report include these extended tests?

Synced offline. 
1.We want to ensure how non-sql standard are also tracked.
2.YES
3.YES

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/ConformanceTestBase.kt
```diff
@@ -43,6 +44,17 @@ abstract class ConformanceTestBase<T, V> {
         }
     }
 
+    // Tests the eval tests data from extended dataset with the Kotlin implementation
+    @Timeout(value = 500, threadMode = Timeout.ThreadMode.SEPARATE_THREAD)
```
alancai98: Looking at the test runner report (https://github.com/partiql/partiql-lang-kotlin/pull/1828#issuecomment-3259452753), the non-SQL/PartiQL extended tests are now showing as ignored which affects the overall conformance percentage. I think we should not include the extended tests in the existing conformance sections but have a separate section of the conformance report detailing the extended test results.

test/partiql-tests-runner/src/test/resources/config/eval/skip-eval-extended.txt
alancai98: Which extended tests are not included in the skip list? Just `DATE_ADD`?

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/ConformanceTestBase.kt
XuechunHHH: The skip list includes all the failing cases in the extended tests for now. The new added passing tests results in the percentage change. IMO, we can only include the `DATE_DIFF` and `DATE_ADD` to current test runner for a short term usage. In the long run, we can create a separate test runner/report for the extended tests.

xd1313113: all is included now

test/partiql-tests-runner/src/test/resources/config/eval/skip-eval-extended.txt
xd1313113: all extended is included

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ConformanceComparison.kt
```diff
@@ -27,23 +28,43 @@ fun loadReportFile(file: File, engine: String, commitId: String): Report {
...
-    val failingSet = mutableSetOf<String>()
-    val ignoredSet = mutableSetOf<String>()
-    reader.next()
+    val report = Report(engine, commitId, mutableMapOf())
+
+    if (!reportContent.contains("partiql-extended")) {
+        // Since old report is in old format, we need to read it differently.
+        // Read old report to show better view of new report.
+        // TODO, remove this after new report in next PR.
+        reader.next()
```
xd1313113: I changed the testresult.ion format and cause old data is not showing in the PR report. And this to parse old data and should remove in next PR

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ReportAnalyzer.kt
```diff
@@ -101,33 +123,47 @@ class ReportAnalyzer(
...
+
+        out.appendMarkdown("## Now $testStatus Tests $ICON_X")
+        // character count limitation with comments in GitHub
+        // also, not ideal to list out hundreds of test names
+
+        resultList.forEach {
+
+            if (it.value.isEmpty()) return@forEach
+
+            out.appendMarkdown("### ${it.key} data set")
```
xd1313113: Data Set

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ReportAnalyzer.kt
```diff
@@ -139,20 +175,35 @@ class ReportAnalyzer(
...
+        if (resultList.values.all { it.isEmpty() }) return
+
+        out.appendMarkdown("## Now Passing Tests")
+        // character count limitation with comments in GitHub
+        // also, not ideal to list out hundreds of test names
+
+        resultList.forEach {
+            if (it.value.isEmpty()) return@forEach
+
+            out.appendMarkdown("### ${it.key.uppercase()} data set")
```
xd1313113: Date Set

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/Report.kt
```diff
@@ -3,10 +3,13 @@ package org.partiql.runner
...
-    val passingSet: Set<String>,
-    val failingSet: Set<String>,
-    val ignoredSet: Set<String>
+    var testsResults: MutableMap<String, TestResult> = mutableMapOf()
 ) {
+    data class TestResult(
+        var passingSet: MutableSet<String> = mutableSetOf(),
+        var failingSet: MutableSet<String> = mutableSetOf(),
+        var ignoredSet: MutableSet<String> = mutableSetOf()
+    )
```
alancai98: A few questions here
1. Why are we modifying the `Report` here?
2. What is the `testsResults` map mapping between?
3. Why are we making these all mutable and `var`s here?

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ReportAnalyzer.kt
```diff
@@ -5,6 +5,8 @@ class ReportAnalyzer(
     private val first: Report,
     private val second: Report
 ) {
+    private val PARTIQL_TAG = "partiql"
+    private val PARTIQL_EXTENDED_TAG = "partiql-extended"
```
alancai98: nit: rather than using strings to distinguish between the core conformance tests vs extended, can use an enum

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ReportAnalyzer.kt
```diff
@@ -19,54 +21,74 @@ class ReportAnalyzer(
...
+            appendOptionalNowFailureTests(this, resultList.associate { Pair(it.tag, it.passingFirstFailingSecond) }, limit, TestStatus.FAILING)
+            appendOptionalNowFailureTests(this, resultList.associate { Pair(it.tag, it.passingFirstIgnoredSecond) }, limit, TestStatus.IGNORED)
+            appendOptionalNowPassingTests(this, resultList.associate { Pair(it.tag, it.failureFirstPassingSecond) }, limit, TestStatus.FAILING)
+            appendOptionalNowPassingTests(this, resultList.associate { Pair(it.tag, it.ignoredFirstPassingSecond) }, limit, TestStatus.IGNORED)
         }
     }
 
-    private fun appendTitle(out: Appendable) {
-        val icon = if (passingFirstFailingSecond.isEmpty()) ICON_CHECK else ICON_X
+    private fun appendTitle(out: Appendable, resultList: MutableList<ComparisonResult>) {
```
alancai98: `resultList` does not need to be mutable. Also applies to `appendSummary`

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ReportAnalyzer.kt
```diff
@@ -19,54 +21,74 @@ class ReportAnalyzer(
...
+        val icon = if (resultList.all { it.passingFirstFailingSecond.isEmpty() }) ICON_CHECK else ICON_X
         out.appendMarkdown("# $reportTitle $icon")
     }
 
-    private fun appendTable(out: Appendable) {
-        out.appendLine("| | $firstNameShort | $secondNameShort | +/- |")
+    private fun appendTable(out: Appendable, result: ComparisonResult?) {
```
alancai98: In what case is a `result` null?

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ReportAnalyzer.kt
```diff
@@ -101,33 +123,47 @@ class ReportAnalyzer(
         this.appendLine()
     }
 
-    private fun appendSummary(out: Appendable) {
+    private fun appendSummary(out: Appendable, resultList: MutableList<ComparisonResult>) {
         out.appendMarkdown("## Testing Details")
         out.appendLine("- **Base Commit**: ${first.commitId}")
         out.appendLine("- **Base Engine**: ${first.engine.uppercase()}")
```
alancai98: We can also get rid of the mentions of engine in the report and test classes. This was only relevant for the v0.14 to v1 comparisons.

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/Report.kt
xd1313113: 1 and 2. Report is holding all raw test data generated by unit test. Since we have two dataset now. I remodeled report class to support dataset label to dataset result mapping. Another way might be having report class representing one dataset result. But we need a way to find report based on dataset label when we render the github report.
3. Will fix. Should not mutable here.

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ReportAnalyzer.kt
xd1313113: sure

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ReportAnalyzer.kt
xd1313113: will fix

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ReportAnalyzer.kt
xd1313113: will fix. It was my old implementaton which can be null.

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ReportAnalyzer.kt
xd1313113: sure

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/Report.kt
```diff
@@ -3,10 +3,19 @@ package org.partiql.runner
 data class Report(
     val engine: String,
```
alancai98: We don't need this `engine` field anymore

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/Report.kt
alancai98: I feel like the remodeled approach is a lot harder to maintain than the prior modeling of `Report`. We should have separate reports for
- core PartiQL conformance tests and
- extended PartiQL conformance tests

Rather than use a map of all of the tests results, we should have another field in `Report` for the `DataSet` type. I believe this should allow us to keep the `ReportAnalyzer` mostly the same as before. IMO it's cleaner for the reports to be separated like we had for the prior conformance reporting. So

```
<report 1 comparing core conformance tests>
<table of comparison 1>
<list of conformance regressions>
---
<report 2 comparing extended conformance tests>
<table of comparison 2>
<list of conformance regressions>
```

In the PR, we interweave all of the above which is a bit more confusing to read.

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ReportAnalyzer.kt
alancai98: This hasn't been addressed yet. `engine` is still being used in the `ReportGenerator` and `ConformanceComparison`.

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/Report.kt
xd1313113: sure. refactored

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ReportAnalyzer.kt
xd1313113: removed now

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/Report.kt
xd1313113: I repurposed the field as ion report file name. as each test suite class need having a different output file.

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ConformanceComparison.kt
```diff
@@ -20,17 +20,50 @@ fun analyze(file: File, reports: List<Report>, limit: Int, title: String) {
...
+        reader.stepIn()
+
+        while (reader.next() != null) {
+            val tag = reader.fieldName
+            if (tag.equals(dataset.dataSetName, ignoreCase = true)) {
+                val result = readTestResult(reader)
+                reader.close()
+                return Report(dataset, commitId, result)
+            }
+        }
```
alancai98: What is this loop and the tag checking for? Based on how we call `loadReportFile` in `PartiQLTestsRunner.java` the conditional seems unnecessary.

test/partiql-tests-runner/src/main/java/org/partiql/runner/PartiQLTestsRunner.java
```diff
@@ -13,15 +13,6 @@
...
 public class PartiQLTestsRunner implements Runnable {
 
-    @CommandLine.Option(
-            required = false,
-            description = {"Title of comparison report. Example: 'CROSS-COMMIT', 'CROSS-ENGINE', etc"},
-            paramLabel = "<title>",
-            defaultValue = "CONFORMANCE REPORT",
-            names = {"-t", "--title"}
-    )
-    private String title;
```
alancai98: This still seems like a valuable option for the test runner. Are we able to keep it? The other options don't seem as relevant.

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ReportAnalyzer.kt
```diff
@@ -102,12 +105,6 @@ class ReportAnalyzer(
     }
 
     private fun appendSummary(out: Appendable) {
-        out.appendMarkdown("## Testing Details")
-        out.appendLine("- **Base Commit**: ${first.commitId}")
-        out.appendLine("- **Base Engine**: ${first.engine.uppercase()}")
-        out.appendLine("- **Target Commit**: ${second.commitId}")
-        out.appendLine("- **Target Engine**: ${second.engine.uppercase()}")
```
alancai98: We should still keep the base and target commit details.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/report/ReportGenerator.kt
```diff
@@ -10,59 +10,84 @@ import java.nio.file.Files
 import kotlin.io.path.Path
 
 class ReportGenerator(
```
alancai98, line 13: Think we should stick w/ the prior factoring of `ReportGenerator` similar to what we did for `Report` in the prior review. We should have separate files for the test results of the core and extended tests like what we had talked about.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/ConformanceTestBase.kt
```diff
@@ -11,20 +12,26 @@ abstract class ConformanceTestBase<T, V> {
...
+     * This is not a performance test. This is for stop long-running tests during development process in eval engine.
+     * This number can be smaller, but to account for the cold start time and fluctuation of GitHub runner,
+     * I decided to make this number a bit larger than needed.
+     *
+     * The report name needs matching the enum string value from [DataSet.dataSetName] in order to be shown on Github
+     */
     @Timeout(value = 5, threadMode = Timeout.ThreadMode.SEPARATE_THREAD)
     @ParameterizedTest(name = "{arguments}")
     @ArgumentsSource(TestProvider.Eval::class)
+    @Tag("report:partiql-core")
```
alancai98: Do we really need to use `Tag` here? When using `Tag`, you can only provide a string known at compile time, so these have to be hard-coded rather than reference the `DataSet.dataSetName`. So if we change a value in `DataSet.dataSetName`, this would break very easily.

 I feel like we can just pass the test type as another argument to `ReportGenerator`.

test/partiql-tests-runner/src/main/resources/config/legacy/conformance_test_results.ion
alancai98, line 1: These results should be split between two files IMO. One for the core and one for the extended.

xd1313113: done

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/ConformanceTestBase.kt
xd1313113: we removed tag use as part spilt into two test files

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/report/ReportGenerator.kt
xd1313113: done

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ReportAnalyzer.kt
xd1313113: sure

test/partiql-tests-runner/src/main/java/org/partiql/runner/PartiQLTestsRunner.java
xd1313113: reverted

test/partiql-tests-runner/src/main/kotlin/org/partiql/runner/ConformanceComparison.kt
xd1313113: not needed any more

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/ConformanceTestEval.kt
alancai98, line 1: Let's rename this file to `ConformanceTestEvalCore` and add a comment about what's tested by this conformance test runner.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/ConformanceTestEval.kt
```diff
@@ -1,23 +1,55 @@
...
-        val reporter = ReportGenerator("eval")
+        val reporter = ReportGenerator(DataSet.PartiQLCore)
     }
 
     private val factory = EvalExecutor.Factory
     override val runner = TestRunner(factory)
 
-    override val skipListForEvaluation: Set<Pair<String, CompileType>> = getSkipList("/config/eval/skip-eval.txt")
+    private val skipListForEvaluation: Set<Pair<String, CompileType>> = getSkipList("/config/eval/skip-eval.txt")
+    private val skipListForEquivalence: Set<Pair<String, CompileType>> = getSkipList("/config/eval/skip-eval-equiv.txt")
```
alancai98: Let's rename these files to mention core

```suggestion
    private val skipListForEvaluation: Set<Pair<String, CompileType>> = getSkipList("/config/eval/skip-eval-core.txt")
    private val skipListForEquivalence: Set<Pair<String, CompileType>> = getSkipList("/config/eval/skip-eval-equiv-core.txt")
```

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/ConformanceTestEvalExtended.kt
```diff
@@ -0,0 +1,37 @@
...
+import org.junit.jupiter.params.provider.ArgumentsSource
+import org.partiql.eval.Statement
+import org.partiql.runner.executor.EvalExecutor
+import org.partiql.runner.report.ReportGenerator
+import org.partiql.runner.schema.TestCase
+import org.partiql.runner.test.TestProvider
+import org.partiql.runner.test.TestRunner
+import org.partiql.spi.value.Datum
+
+class ConformanceTestEvalExtended : ConformanceTestBase<Statement, Datum>() {
```
alancai98, line 22: Let's add a short comment about what's tested by this class.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/ConformanceTestEval.kt
xd1313113: Done

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/ConformanceTestEval.kt
xd1313113: Done

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/ConformanceTestEvalExtended.kt
xd1313113: Done

alancai98 (approved):

PR #1829, opened 2025-09-09
fix: variable resolution for qualified references with `GROUP BY`

XuechunHHH:
## Description
- FIx the bug for using `GROUP BY` with table aliases, qualified references in the `SELECT` clause fail to resolve to the corresponding GROUP BY bindings. 

### Current Behavior
- Pass the tests in https://github.com/partiql/partiql-tests/pull/158

## Other Information

## License Information


partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/Scope.kt
```diff
@@ -37,9 +37,16 @@ internal data class Scope(
...
-        // Convert any remaining binding names (tail) to an untyped path expression.
-        return if (tail.isEmpty()) r else r.toPath(tail)
+        // Try exact match
+        val r = matchRoot(head)
+        if (r != null) {
+            return if (tail.isEmpty()) r else r.toPath(tail)
+        }
+        // Fallback: for qualified identifiers, try unqualified part for GROUP BY bindings
+        if (path.size == 2) {
+            return matchRoot(path[1])
```
alancai98: Hmm, this does work for the test cases provided in the description but it does not consider the qualifier to the identifier could be different. For example, consider if we had done an `ORDER BY` over a different binding name, `S`:

```
SELECT T.a, MAX(b) FROM <<{'a': 1, 'b': 2}, {'a': 3, 'b': 5}>> T GROUP BY T.a ORDER BY S.a
```

Since we ignore the qualifier, the above would still associate `S.a` with `T.a`, rather than a variable not found warning. With the current implementation, the above query would output

```
[
  {
    'a': 1,
    '_1': 2
  },
  {
    'a': 3,
    '_1': 5
  }
]
```

---

Another issue is if the path size is greater than `2`. Suppose we were doing a group by, select, and order by over a path of length 3:

```
 SELECT T.c.foo, MAX(b) FROM <<{'a': 1, 'b': 2, 'c': {'foo': 3}}, {'a': 3, 'b': 5, 'c': {'foo': 5}}>> T GROUP BY T.c.foo ORDER BY T.c.foo
```

Running the above query will give the variable not found error we were running into earlier:
```
w: [semantic] Variable reference (T.c.foo) could not be found in the database environment or in the set of available locals. Locals: [$agg_0, foo].

=== RESULT ===
[
  {
    '_1': 2,
    'foo': 3
  },
  {
    '_1': 5,
    'foo': 5
  }
]

OK!
```

alancai98: In the `Scope`'s `schema`, we probably need to include additional information about the path expression used to create the binding.

CHANGELOG.md
```diff
@@ -62,6 +62,7 @@ rewritten to their query representation
 - Return `missing` instead of `null` for `NULLIF(MISSING, MISSING)`
 - CLI printing of mixed sign literals
 - Parsing of `TIMESTAMP` and `TIME` literals with timezone offsets
+- Variable resolution for qualified references with `GROUP BY`
```
alancai98, line 65: Would be good to include the additional tests from the PR description as well as some more tests (such as the test I included in the other comment) in the conformance tests.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/Scope.kt
XuechunHHH: Hmm...I was understanding this issue too simplistically. Thanks for catching these edge cases. I will take another look more deeply.

test/partiql-tests
XuechunHHH, line 1: Updated the conformance test submodule.

XuechunHHH (commented):
Failed in the conformance test 
https://github.com/partiql/partiql-tests/blob/b4673562ec17246f1fa3f584e367120850bac3c2/partiql-tests-data/eval/query/group-by/group-by.ion#L8431-L8447
because of the issue existing in missing normalization for `ORDER BY`. Will be fixed in another PR.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -787,6 +798,46 @@ internal object RelConverter {
...
 
+    /**
+     * Infer table qualifier (alias from FROM clause) from input schema
+     */
+    private fun inferTableQualifierFromInput(input: Rel): String? {
+        return input.type.schema.firstOrNull()?.name
+    }
+
+    /**
+     * Extract the qualifier (except for the final column) from a qualified expression
```
alancai98, line 810: May be helpful to explain this function's logic with some examples in the kdoc.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -787,6 +798,46 @@ internal object RelConverter {
...
+                    when (step) {
+                        is PathStep.Field -> step.field.text
+                        else -> null
+                    }
+                }
+                when {
+                    rootName != null && pathSteps.isNotEmpty() -> "$rootName.${pathSteps.joinToString(".")}"
+                    rootName != null -> rootName
+                    else -> null
+                }
```
alancai98: nit: there's probably some way to clean up this logic, especially the last `when` branch. Maybe have some mutable list with all the parts (root and path steps) that you `joinToString` if it's non-empty?

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/ScopeTest.kt
```diff
@@ -45,12 +45,12 @@ internal class ScopeTest {
             ),
             Scope(
                 listOf(
-                    relBinding("A", struct("B" to PType.bool().toCType())),
-                    relBinding("a", struct("b" to PType.bool().toCType())),
-                    relBinding("X", struct(open = true)),
-                    relBinding("x", struct("Y" to PType.bool().toCType(), open = false)), // We currently don't allow for partial schema structs
-                    relBinding("y", struct(open = true)),
-                    relBinding("T", struct("x" to PType.bool().toCType(), "x" to PType.bool().toCType())),
+                    relBinding("A", struct("B" to PType.bool().toCType()), null),
```
alancai98, line 48: Would be helpful to have at least one `Scope` unit test that has a non-null qualifier. I know the conformance tests already test parts of the logic added in this PR but still may be helpful for demonstrating correctness.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/Scope.kt
```diff
@@ -37,9 +37,19 @@ internal data class Scope(
...
+            if (qualifiedMatch != null) {
+                return qualifiedMatch
+            }
+        }
+        // Try exact match
+        val r = matchRoot(head)
+        if (r != null) {
+            return if (tail.isEmpty()) r else r.toPath(tail)
+        }
+        return null
```
alancai98: Wonder if we should perform the exact match first and then proceed with the multi-part match afterwards?

alancai98: also nit: can use the elvis operator (https://kotlinlang.org/docs/null-safety.html#elvis-operator) to simplify some of these conditional checks

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/Scope.kt
```diff
@@ -62,6 +72,29 @@ internal data class Scope(
...
+    private fun matchQualified(identifier: Identifier, depth: Int = 0): Rex? {
+        val path = identifier.toList()
+        val qualifier = path.dropLast(1).joinToString(".") { it.getText() }
+        val bindingName = path.last().getText()
+        for (i in schema.indices) {
+            val local = schema[i]
+            val type = local.type
+            if (local.qualifier != null &&
+                local.qualifier.equals(qualifier, ignoreCase = true) &&
+                local.name.equals(bindingName, ignoreCase = true)
```
alancai98: Hmm, so here we're ignoring the case-sensitivity of the identifier. A query such as `SELECT "t"."A", MAX(T.b) FROM <<{'a': 1, 'b': 2}, {'a': 3, 'b': 5}>> T GROUP BY T.a ORDER BY T.a` where the projection path is case-sensitive and does not match the exact binding + qualifier case, would be expected to fail/return variable not found.

But ignoring the case-sensitivity will incorrectly match the query to the `T.a` variable:

```
partiql ▶ SELECT "t"."A", MAX(T.b) FROM <<{'a': 1, 'b': 2}, {'a': 3, 'b': 5}>> T GROUP BY T.a ORDER BY T.a;

=== RESULT ===
[
  {
    'A': 1,
    '_1': 2
  },
  {
    'A': 3,
    '_1': 5
  }
]
```

I think we need to take the case-sensitivity of the input `identifier` into account.

alancai98 (commented):
Nice work so far. Think we're on the right track. A few things we need to address/fix.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
XuechunHHH: Good call. Will use a mutable list to collect all parts and simplify the logic.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/Scope.kt
XuechunHHH: Thanks for pointing out. Didn't realize this issue.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/Scope.kt
```diff
@@ -62,6 +65,30 @@ internal data class Scope(
...
+        val qualifierParts = path.dropLast(1)
+        val bindingName = path.last()
+        for (i in schema.indices) {
+            val local = schema[i]
+            val type = local.type
+            if (local.qualifier != null && bindingName.matches(local.name)) {
+                val localQualifierParts = local.qualifier.split(".")
+                if (qualifierParts.size == localQualifierParts.size &&
+                    qualifierParts.zip(localQualifierParts).all { (identPart, localPart) ->
+                        identPart.getText().equals(localPart, ignoreCase = identPart.isRegular())
```
XuechunHHH: Use `.isRegular()` to check the case sensitive setting. Current behavior is like:
```
partiql ▶ SELECT "t"."A", MAX(T.b) FROM <<{'a': 1, 'b': 2}, {'a': 3, 'b': 5}>> T GROUP BY T.a ORDER BY T.a;
w: [semantic] Variable reference ("t"."A") could not be found in the database environment or in the set of available locals. Locals: [T.$agg_0, T.a].

=== RESULT ===
[
  {
    '_1': 2
  },
  {
    '_1': 5
  }
]

OK!
partiql ▶ SELECT t.A, MAX(T.b) FROM <<{'a': 1, 'b': 2}, {'a': 3, 'b': 5}>> T GROUP BY T.a ORDER BY T.a;

=== RESULT ===
[
  {
    'A': 1,
    '_1': 2
  },
  {
    'A': 3,
    '_1': 5
  }
]

OK!
```

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/Scope.kt
```diff
@@ -62,6 +65,30 @@ internal data class Scope(
...
+     */
+    private fun matchQualified(identifier: Identifier, depth: Int = 0): Rex? {
+        val path = identifier.toList()
+        val qualifierParts = path.dropLast(1)
+        val bindingName = path.last()
+        for (i in schema.indices) {
+            val local = schema[i]
+            val type = local.type
+            if (local.qualifier != null && bindingName.matches(local.name)) {
+                val localQualifierParts = local.qualifier.split(".")
```
alancai98: Wonder if it would actually make sense to represent the `qualifier` in the internal plan as a list of `Identifier.Simple`? That way we do not need to do the re-parsing (i.e. the `split` call here). I don't believe we use the string representation of `qualifier` other than for printing.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/Scope.kt
```diff
@@ -62,6 +65,30 @@ internal data class Scope(
...
+                if (qualifierParts.size == localQualifierParts.size &&
+                    qualifierParts.zip(localQualifierParts).all { (identPart, localPart) ->
```
alancai98: This is a pretty complicated nesting of `if`s and conditionals. Consider inverting the `if`s to make it more readable -- https://stackoverflow.com/questions/268132/why-should-i-invert-if-statement-to-reduce-nesting.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/Scope.kt
alancai98: Nice. Mind if we add these tests to the conformance tests to ensure they are continually tested?

alancai98: Also nit: can't we just use the `matches` function? https://github.com/partiql/partiql-lang-kotlin/blob/fix-table-ref/partiql-spi/src/main/kotlin/org/partiql/spi/catalog/Identifier.kt#L145-L150

XuechunHHH: > Nice. Mind if we add these tests to the conformance tests to ensure they are continually tested?

NP.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/Scope.kt
XuechunHHH: Yes, I think that makes more sense, will use a list to store `qualifier` instead.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/Scope.kt
XuechunHHH: Tests added in https://github.com/partiql/partiql-tests/pull/160

PR #1831, opened 2025-09-18
Fix the ORDER BY does not resolve alias from SELECT statement.

xd1313113:
## Relevant Issues
Fix the ORDER BY does not resolve alias from SELECT statement.

### Before
```
SELECT b AS agg_alias FROM [{'b':2},{'b':3}] T GROUP BY b ORDER BY agg_alias DESC;
w: [semantic] Variable reference (agg_alias) could not be found in the database environment or in the set of available locals. Locals: [b].

=== RESULT ===
[
  {
    'agg_alias': 2
  },
  {
    'agg_alias': 3
  }
]

OK!
```

### Now
```
partiql ▶ SELECT b AS agg_alias FROM [{'b':2},{'b':3}] T GROUP BY b ORDER BY agg_alias DESC;

=== RESULT ===
[
  {
    'agg_alias': 3
  },
  {
    'agg_alias': 2
  }
]

OK!
```

It also fixed nested query
```
Select agg_alias as c from (SELECT b AS agg_alias FROM [{'b':2},{'b':3}] T GROUP BY b ORDER BY agg_alias) as T group by c order by c;
```

**New issue** Order BY does not work with aggregation function as alias.
```
SELECT a, MAX(b) as agg_alias FROM [{'a':1, 'b':2},{'a':2, 'b':3}] T GROUP BY a ORDER BY a, agg_alias;
```

## Description
- < Explain >

## Other Information

## License Information


CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ rewritten to their query representation
 - fix the parsing of DATE_ADD/DATE_DIFF with an uncapitalized datetime field argument.
 - fix that wrong offset is set when parsing Ion timestamp with time zone into Datum.
 - Reimplemented DATE_ADD with interval plus arithmetic.
+- Fix ORDER BY statement does not recognize alias from SELECT statement
```
alancai98, line 68: Can we add some conformance tests? Hard for me to verify the changes otherwise.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Normalizes ORDER BY expressions by replacing SELECT aliases with their original expressions.
+ * Maintains separate alias maps for each SFW clause layer to handle nested queries correctly.
+ */
+internal object OrderByAliasSupport : AstPass {
+    override fun apply(statement: Statement): Statement {
+        return Visitor.visitStatement(statement, ArrayDeque()) as Statement
+    }
+
+    private object Visitor : AstRewriter<ArrayDeque<MutableMap<String, Expr>>>() {
```
alancai98: Could you explain a bit more why we're using a queue here? An alternative approach could be to keep a counter global to this visitor/rewriter that you can update.

xd1313113: ArrayDeque<MutableMap<String, Expr>> is used to handle nested SFW statements where:
     - Each query level has its own alias scope
     - Inner queries can't reference outer query aliases in ORDER BY
     - The stack ensures proper scoping isolation

CHANGELOG.md
xd1313113: add unit test to cover transformation and conformance test for evaluation

partiql-planner/src/main/kotlin/org/partiql/planner/internal/SqlPlanner.kt
```diff
@@ -69,6 +70,7 @@ internal class SqlPlanner(
         var ast = this
         ast = NormalizeFromSource.apply(ast)
         ast = NormalizeGroupBy.apply(ast)
+        ast = OrderByAliasSupport.apply(ast)
```
XuechunHHH: Can we change the name to `NormalizeOrderBy` and may include logic other than alias in the same file later.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
```diff
@@ -0,0 +1,125 @@
...
+            val transformed = super.visitExprQuerySet(node, ctx)
+            ctx.removeLast()
+            return transformed
+        }
+
+        /**
+         * Collects aliases from SELECT items into the current scope's map.
+         * Only SELECT items with AS aliases are stored.
+         */
+        override fun visitSelectItem(node: SelectItem, ctx: ArrayDeque<MutableMap<String, Expr>>): AstNode {
```
XuechunHHH: As I can recall, in our implementation, all items have alias. e.g. `SELECT a FROM...` will be `SELECT a AS a FROM...`. Pls feel free to correct me if my memory is wrong.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupportTest.kt
```diff
@@ -0,0 +1,133 @@
...
+ */
+
+package org.partiql.planner.internal.transforms
+
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.MethodSource
+import org.partiql.parser.PartiQLParser
+import kotlin.test.assertEquals
+
+class OrderByAliasSupportTest {
```
alancai98, line 29: Can we also port the v0.14 order by transform tests here -- https://github.com/partiql/partiql-lang-kotlin/blob/v0.14.9/partiql-lang/src/test/kotlin/org/partiql/lang/eval/visitors/OrderBySortSpecVisitorTransformTests.kt

And let's include some tests using set operators.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
```diff
@@ -0,0 +1,125 @@
...
+            if (node is SelectItem.Expr && node.asAlias != null) {
+                ctx.last().put(node.asAlias!!.text, node.expr)
+            }
```
alancai98: nit: can define a variable set to `node.asAlias` out of the conditional so that we don't need the `null` assertion `!!` on line 83

CHANGELOG.md
alancai98: There are some additional tests failing in the conformance test runner -- https://github.com/partiql/partiql-lang-kotlin/pull/1831#issuecomment-3309879253. Can you take a look at those added tests?

xd1313113: Yes, some newly added tests(Aggregation in `order by`) are failing, I moved them to skip list for now.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/SqlPlanner.kt
xd1313113: OrderByAliasSupport is better for now. I plan to use NormalizeOrderBy for aggregation normalization.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupportTest.kt
xd1313113: Added tests from 0.14 and set operators. I reimplements the code so that it works with set operators better.

xd1313113: Yes, added tests from 0.14 and set operators.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
xd1313113: I did not notice them when collecting alias map

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
xd1313113: asAlias is not accessible before node is SelectItem.Expr. Not very sure what you mean

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /**
+     * AST visitor that uses a stack-based approach to track parent query scopes.
+     *
+     * Key behaviors:
+     * - Each ExprQuerySet creates its own alias scope on the stack
+     * - SELECT aliases are collected into the current scope's map
+     * - ORDER BY expressions resolve aliases from the appropriate scope
+     * - Set operations (UNION, INTERSECT, EXCEPT) are skipped Order-By alias replacement
```
alancai98: Just for my understanding, why do we skip over the set operators?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * Context for tracking parent query scopes and their alias mappings.
+     *
+     * @property parentStack Stack of ExprQuerySet nodes representing nested query scopes
+     * @property aliasMap Maps each query scope to its SELECT alias definitions
+     */
+    data class Context(
+        val parentStack: ArrayDeque<ExprQuerySet> = ArrayDeque(),
+        val aliasMap: MutableMap<ExprQuerySet, MutableMap<String, Expr>> = mutableMapOf()
```
alancai98: In the prior review, you were using an only `ArrayDeque()` as the context. Why are we switching to use a `parentStack` and an `aliasMap`. From my understanding, it seems like after we exit from some nested context and perform the rewrite over the `ORDER BY` sets we don't need to keep that alias mapping anymore.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupportTest.kt
```diff
@@ -0,0 +1,419 @@
...
+                    SELECT a + b AS b
+                    FROM (
+                      SELECT b + a AS b
+                      FROM e
+                      ORDER BY b + a
+                    )
+                    ORDER BY a + b
+                """,
+                description = "Case insensitive aliases with different cases in ORDER BY"
+            ),
```
alancai98: This test can be removed. It's the same as `complex_projection_same_alias`.

test/partiql-tests-runner/src/test/resources/config/eval/skip-eval-core.txt
```diff
@@ -1,10 +1,14 @@
...
 PERMISSIVE:::group and order by count
 STRICT:::group and order by count
+PERMISSIVE:::order by aggregation function directly
+STRICT:::order by aggregation function directly
+PERMISSIVE:::order by different aggregation function from select
+STRICT:::order by different aggregation function from select
+PERMISSIVE:::order by aggregation alias
+STRICT:::order by aggregation alias
+PERMISSIVE:::order by aggregation alias with group by alias
+STRICT:::order by aggregation alias with group by alias
```
alancai98: We shouldn't be ignoring these tests even though they are currently failing.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
```diff
@@ -0,0 +1,173 @@
...
+                is ExprVarRef -> {
+                    val identifier = expr.identifier.identifier
+                    val orderByName = identifier.text
+                    val isOrderByRegular = identifier.isRegular
+
+                    // Find matching alias considering case sensitivity
+                    val matchingAlias = if (isOrderByRegular) {
+                        // Regular (unquoted) identifier: case-insensitive lookup
+                        aliasMap.entries.find { (aliasName, _) ->
+                            orderByName.equals(aliasName, ignoreCase = true)
```
alancai98: What happens if there are multiple matches from the `SELECT` list? I think we should be consistent with existing SQL/PartiQL behavior and give an error for an ambiguous variable reference.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
XuechunHHH: My understanding is in the following example with `UNION`:

```
SELECT col AS alias FROM t1
UNION
SELECT col AS alias FROM t2
ORDER BY alias
```

the alias in `ORDER BY` refers to the `alias` in `UNION` operation result, not the one in the original `SELECT` statement. So...  make sense to skip the alias in `Set` operations to me.

See the test -- https://github.com/partiql/partiql-lang-kotlin/pull/1831/files#diff-52180aa61a0068ba4bde09c7ba5c2cac170c016af83db5997a17788165117f8bR321-R336

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
```diff
@@ -0,0 +1,173 @@
...
+            // Pop scope when exiting query
+            ctx.parentStack.removeLast()
+            return transformed
+        }
+
+        /**
+         * Collects SELECT aliases into the current query scope's alias map.
+         * Only processes SelectItem.Expr nodes that have AS aliases defined.
+         */
+        override fun visitSelectItem(node: SelectItem, ctx: Context): AstNode {
```
XuechunHHH: What happens if there are multiple `SELECT` items with the same alias? Do we already have uniqueness validation for within same scope?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
```diff
@@ -0,0 +1,173 @@
...
+     * 6. Pop outer scope
+     */
+    private object Visitor : AstRewriter<Context>() {
+        /**
+         * Manages query scope stack for each ExprQuerySet.
+         * Pushes current query to stack on entry, pops on exit to maintain proper nesting.
+         */
+        override fun visitExprQuerySet(node: ExprQuerySet, ctx: Context): AstNode {
+            // Push current query scope onto stack
+            ctx.parentStack.addLast(node)
```
XuechunHHH: Maybe reductant, for these stack operations, do we need error handling?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/SqlPlanner.kt
XuechunHHH: A separate question, should this `OrderByAliasSupport` be applied before or after the aggregation normalization?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
alancai98: I think it's ok to project out multiple items with the same alias. For example in Postgresql:

```sql
SELECT 1 AS a, 2 AS a;
```

Returns back
```
a | a
-- | --
1 | 2
```

The issue though is if you define multiple things with the same alias and then refer to it later on. That's what leads to the ambiguous variable reference.

XuechunHHH: Aha, I see. So the real issue is the one you raised around line 160. We should probably modify the code to detect when there are multiple matches for the same alias and throw an error (not sure if it is the behavior how sql spec define).

alancai98: Yeah, atm I think there are a couple cases we need to consider:
1. `SELECT ... AS fOo, ... AS FOO ... ORDER BY foo` -> two case-insensitive matches to `foo`
2. `SELECT ... AS foo, ... AS foo ... ORDER BY "foo"` -> two case-sensitive matches to `foo`

Both of the above should return an ambiguous variable error.

XuechunHHH: Also a more complex case with subqueries I can come up with:
```
SELECT x AS foo FROM
  (SELECT y AS foo FROM t1
   UNION
   SELECT z AS foo FROM t2)
ORDER BY foo
```
Ambiguous variable references can exist more than different cols in `SELECT`, need to carefully deal with...

partiql-planner/src/main/kotlin/org/partiql/planner/internal/SqlPlanner.kt
xd1313113: OrderByAliasSupport needs to be before aggregation normalization. As after alias replacement, new aggregation show up in order by

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
xd1313113: Thanks Xuechun. The alias used in order by must show in both subquery when set op is present.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
xd1313113: The old approach works fine for nested queries but works 'blindly' for set operators as orderby belongs to a querysetExpr without SFW. The new approach provide a better structure awareness context for orderby and allow it to explicitly skip for set op.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupportTest.kt
xd1313113: They are actually different by one letter `B` while `b` in complex_projection_same_alias

test/partiql-tests-runner/src/test/resources/config/eval/skip-eval-core.txt
xd1313113: sure

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
xd1313113: good point, select allow same projects and order by will error if there are ambiguous reference.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
xd1313113: yeah, Now I emit an error in this case.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
xd1313113: Not sure what you mean. It is guaranteed to be valid.

XuechunHHH: If it is guaranteed, that will be fine. Just wanted to double confirm if checking is necessary here.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
alancai98: I'm still a bit confused on the changes to the rewriter's context (specifically in this commit https://github.com/partiql/partiql-lang-kotlin/pull/1831/commits/6412350bb209e4ab3c91dec62dc9f12fca732be5). Previously was an `ArrayDeque<MutableMap<String, Expr>>` but now you've changed it to a stack and an alias map/list. It seems redundant to store the alias map after visiting an inner SFW where it might not be used anymore in the outer SFW.

> The old approach works fine for nested queries but works 'blindly' for set operators as orderby belongs to a querysetExpr without SFW.

What do you mean by "blindly"? There are ways in the visitor pattern to skip over set ops.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
```diff
@@ -0,0 +1,179 @@
...
+         */
+        private fun resolveExpr(expr: Expr, aliasMap: List<Pair<String, Expr>>): Expr {
+            return when (expr) {
+                is ExprVarRef -> {
+                    val identifier = expr.identifier.identifier
+                    val orderByName = identifier.text
+                    val isOrderByRegular = identifier.isRegular
+
+                    // Find matching alias considering case sensitivity
+                    val candidates = aliasMap.filter { orderByName.equals(it.first, ignoreCase = isOrderByRegular) }
```
alancai98: Doing a `filter` over the list is an O(n) operation. When it's a case-insensitive lookup, there's no easy way we can avoid scanning through all the aliases unless we store another case-normalized map (e.g. all keys are lower-cased). However, when it's a case-sensitive lookup, lookup can be O(1) using something like a multimap (https://en.wikipedia.org/wiki/Multimap).

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
```diff
@@ -0,0 +1,179 @@
...
+                    if (candidates.size == 1) {
+                        candidates[0].second
+                    } else {
+                        if (candidates.size > 1) {
+                            val candidateNames = candidates.map {
+                                val ref = it.second
+                                if (ref is ExprVarRef) {
+                                    ref.identifier.identifier.text
+                                } else {
+                                    "Not a column name or alias"
```
alancai98: These should not be included in the candidates list. Can do something like a `.mapNotNull` https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/map-not-null.html if we want to ignore these cases.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
```diff
@@ -0,0 +1,179 @@
...
+                                }
+                            }.toList()
```
alancai98: nit: `.toList()` is redundant since `.map` should already return back a list here

partiql-planner/src/test/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupportTest.kt
```diff
@@ -0,0 +1,465 @@
...
+                    SELECT x + y AS total FROM t2
+                    ORDER BY total DESC
+                """,
+                description = "UNION ALL with complex expression alias resolution"
+            )
+        )
+
+        @JvmStatic
+        fun failedTestCases() = listOf(
+            FailedTestCase(
```
alancai98: Let's add both cases that were mentioned in the other comment https://github.com/partiql/partiql-lang-kotlin/pull/1831#discussion_r2402971346 -- case-insensitive ambiguity and case-sensitive ambiguity.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
xd1313113: Let me explain a bit in detail.
In `ArrayDeque<MutableMap<String, Expr>> ` old approach,

**Nested query**
```
stack(empty)
ExprQuerySet:(stack(map1))
  body:SFW
    select:
      project 1(add to map1))
    from:
      ExprQuerySet: stack(map1, map2)
        body:SFW
          select:
            project 2 (add to map2))
          from: T
        orderby: Expr 2 (replace from map2)
        stack.pop() // stack(map1)
  orderby:Expr 1 (replace from map1)
stack.pop() // stack(empty)
```

**Setop Query**
```
stack(empty)
ExprQuerySet:(stack(map1))
  body:op
    lhs:ExprQuerySet // stack(map1,map2)
      select:
        project 1(add to map2))
      from: T
    stack.pop()
    rhs:ExprQuerySet stack(map1,map3)
      select:
        project 1(add to map3))
      from: T
    stack.pop() // (stack(map1))
  orderby:Expr 1 (replace from map1) // map1 is empty as there is always no `SELECT`.
stack.pop() // stack(empty)
```

The reason I swithed to have a parent stack is 
- The old approach skipped orderby "blindly" as map is empty and without knowing if it belongs to ExprQuerySet which has setOp body. 

> There are ways in the visitor pattern to skip over set ops.

- We should not skip setop (inner query main contains order by) but skip order by when its parent has a setop body. The new `parentStack` approach give some additional information.
- It is true that for now outer SFW does not need alias context of inner SFW. It looks different sql system has different behavior of the final table when set is present. Some uses the result of first SFW as final table schema which means outer SFW can have reference from first SFW. It is quite ambiguous since we do not have partiql specification for set op. I can potentially can use a stack with parent and alias map instead.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
xd1313113: Thanks. Implements a map based instead

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
xd1313113: sure

partiql-planner/src/test/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupportTest.kt
xd1313113: sure

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
```diff
@@ -0,0 +1,158 @@
...
+
+            // Visit all statements that may have SELECT or ORDER BY
+            val body = node.body.let { visitQueryBody(it, ctx) as QueryBody }
+            val orderBy = node.orderBy?.let {
+                if (body !is QueryBody.SetOp) {
+                    // Skip alias replacement if the query body is set operations
+                    visitOrderBy(it, ctx) as OrderBy?
+                } else {
+                    node.orderBy
+                }
```
alancai98: nit: slightly cleaner factoring + comment placement

```suggestion
                if (body is QueryBody.SetOp) {
                    // Skip alias replacement if the query body is a set op
                    node.orderBy
                } else {
                    visitOrderBy(it, ctx) as OrderBy?
                }
```

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
```diff
@@ -0,0 +1,158 @@
...
+ * Replaces ORDER BY aliases with their corresponding SELECT expressions using stack-based scope tracking.
+ */
+internal class OrderByAliasSupport(val listener: PErrorListener) : AstPass {
+    override fun apply(statement: Statement): Statement {
+        return Visitor(listener).visitStatement(statement, ArrayDeque()) as Statement
+    }
+
+    /**
+     * Maintains alias scope stack for nested queries and resolves ORDER BY aliases.
+     */
```
alancai98, line 70: Let's add a brief description of the context you're using here since it's somewhat complicated. Could also consider defining this context as its own internal class/data class.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
```diff
@@ -0,0 +1,158 @@
...
+         * Regular identifiers use case-insensitive matching, delimited use case-sensitive.
+         */
+        private fun resolveExpr(expr: Expr, aliasMap: Map<String, List<Expr>>): Expr {
+            return when (expr) {
+                is ExprVarRef -> {
+                    val identifier = expr.identifier.identifier
+                    val orderByName = identifier.text
+
+                    val candidates = if (identifier.isRegular) {
+                        aliasMap.filterKeys { it.equals(orderByName, ignoreCase = true) }.values.flatten()
```
alancai98, line 150: May be worth calling out or implementing an alternative approach that is O(1) lookup if we were to have an additional case-insensitive map where all the keys are lower-cased. That approach has the drawback of needing to store an additional map.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
xd1313113: add a new internal data class

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/OrderByAliasSupport.kt
xd1313113: Called out in comments. Agree we may need to consider additional case-insensitive map in case of performance concern. For sql statements, this might be fine for now.

Issue #1130, opened 2023-06-16
Support TIME WITHOUT TIME ZONE syntax


RCHowell:
## Relevant Issue/Bug

We currently do not support the explicit `WITHOUT TIME ZONE` syntax defined by SQL-99

## Requested Solution/Feature

Support SQL-99 grammar for timezone

```bnf
<datetime type> ::= DATE
                   | TIME [ <left paren> <time precision> <right paren> ] [ <with or without time zone> ]
                   | TIMESTAMP [ <left paren> <timestamp precision> <right paren> ] [ <with or without time zone> ]

<with or without time zone> ::= WITH TIME ZONE
                              | WITHOUT TIME ZONE
```

## Describe Alternatives
..

## Additional Context

Today, WITHOUT is implicit. SQL-99 p.29 describes the alternatives

- Implicit WITH `TIME 00:00:00+ZZ:ZZ`
- Implicit WITHOUT `TIME 00:00:00`
- Explicit WITH `TIME WITH TIME ZONE 00:00:00+ZZ:ZZ`
- Explicit WITHOUT `TIME WITHOUT TIME ZONE 00:00:00` !! MISSING !!

## DoD (Definition of Done)

Support syntax for explicit WITHOUT

xd1313113:
Supported now

PR #1832, opened 2025-09-20
add: support `TIME WITHOUT TIME ZONE` syntax

XuechunHHH:
## Relevant Issues
- Closes Issue #1130 

## Description
- Pass tests in https://github.com/partiql/partiql-tests/pull/161

## Other Information

## License Information


test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/util/DatumUtil.kt
```diff
@@ -31,31 +33,55 @@ internal fun Datum.toIonElement(): IonElement {
...
-                    val timezoneMinutes = (totalSeconds % 3600) / 60
-                    fields.add(field("timezone_hour", ionInt(timezoneHours.toLong())))
-                    fields.add(field("timezone_minute", ionInt(timezoneMinutes.toLong())))
+                    val totalSeconds = offsetTime.second.toBigDecimal() + offsetTime.nano.toBigDecimal().divide(nanoPerSecond.toBigDecimal())
+                    fields.add(field("second", ionDecimal(Decimal.valueOf(totalSeconds))))
+                    val offset = offsetTime.offset.totalSeconds / SecondPerMinute
+                    fields.add(field("offset", ionInt(offset.toLong())))
                     ionStructOf(fields).withAnnotations("\$time")
                 }
+                PType.TIMESTAMP -> {
```
XuechunHHH, line 56: self-review: Change from ion timestamp because PartiQL `timestamp` has broader support than ion. This change makes the 6 previously passed conformance tests fail now. We can discuss offline to double check the expected behavior.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/util/DatumUtil.kt
```diff
@@ -31,31 +33,55 @@ internal fun Datum.toIonElement(): IonElement {
...
                 PType.DATE -> ionString(this.localDate.toString()).withAnnotations("\$date")
                 PType.TIME -> {
                     val time = this.localTime
                     val fields = mutableListOf<StructField>()
                     fields.add(field("hour", ionInt(time.hour.toLong())))
                     fields.add(field("minute", ionInt(time.minute.toLong())))
-                    fields.add(field("second", ionDecimal(Decimal.valueOf(time.second.toBigDecimal()))))
+                    val totalSeconds = time.second.toBigDecimal() + time.nano.toBigDecimal().divide(nanoPerSecond.toBigDecimal())
+                    fields.add(field("second", ionDecimal(Decimal.valueOf(totalSeconds))))
+                    fields.add(field("offset", ionInt(0L)))
```
xd1313113: Is offset  required? I think leaving it as default should be better.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/util/DatumUtil.kt
```diff
@@ -31,31 +33,55 @@ internal fun Datum.toIonElement(): IonElement {
...
-                PType.TIMESTAMP -> ionTimestamp(this.localDateTime.toString() + "Z")
-                PType.TIMESTAMPZ -> ionTimestamp(this.offsetDateTime.toString())
                 PType.DATE -> ionString(this.localDate.toString()).withAnnotations("\$date")
                 PType.TIME -> {
                     val time = this.localTime
                     val fields = mutableListOf<StructField>()
                     fields.add(field("hour", ionInt(time.hour.toLong())))
                     fields.add(field("minute", ionInt(time.minute.toLong())))
-                    fields.add(field("second", ionDecimal(Decimal.valueOf(time.second.toBigDecimal()))))
+                    val totalSeconds = time.second.toBigDecimal() + time.nano.toBigDecimal().divide(nanoPerSecond.toBigDecimal())
```
xd1313113: Can Ion handle max Nano precision from partiql?

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/util/DatumUtil.kt
XuechunHHH: In current design, offset is required because the `$time::` schema is shared by time and timez. IMO, it makes more sense to be a required field especially for timez.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/util/DatumUtil.kt
XuechunHHH: Yes, in PLK implementation, nano precision will be truncated to 9 and is supported in Ion.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/util/DatumUtil.kt
xd1313113: If it is shared, how do we distinguish the time with 0 offset and the time with no offset?

XuechunHHH: Aha.. good call. In this case, probably it will be better to set to `null` for no offset. Thanks for pointing out. I will make the change.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -2116,19 +2135,38 @@ internal class PartiQLParserDefault : PartiQLParser {
                 p
             }
             val hasTimezoneLiteral = timestampString.contains(Regex("[+|-]\\d\\d:\\d\\d$"))
-            if (ctx.ZONE() != null && !hasTimezoneLiteral) {
+            val hasWithTimeZone = ctx.WITH() != null && ctx.ZONE() != null
+            val hasWithoutTimeZone = ctx.WITHOUT() != null && ctx.ZONE() != null
```
lucaxchen: question: why do we need to know if the `ctx.zone()` must not be null here if we already know that `ctx.without()` is not null? My intuition tells me that, we might not need this predicate, but I might be wrong, please correct me if so.

XuechunHHH: Nice catch, in this context, `WITH/WITHOUT` can only appear together with `ZONE`. The checking is reductant here. I will remove it.

partiql-ast/src/main/java/org/partiql/ast/DataType.java
```diff
@@ -142,18 +142,20 @@ public String getComment() {
...
-    public static final int ARRAY = 42;
-    public static final int BAG = 43;
-    public static final int SEXP = 44;
+    public static final int LIST = 43;
+    public static final int ARRAY = 44;
+    public static final int BAG = 45;
+    public static final int SEXP = 46;
     // <user defined type>
-    public static final int USER_DEFINED = 45;
+    public static final int USER_DEFINED = 47;
```
alancai98: Hmm I don't think we should re-number these. If a customer is relying on a specific int corresponding to a certain `DataType`, then renumbering will break their code (e.g. if they depend on `INTERVAL` mapping to 38).

We should instead probably number `TIME_WITHOUT_TIME_ZONE` and `TIMESTAMP_WITHOUT_TIME_ZONE` after `USER_DEFINED`.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -2081,19 +2081,38 @@ internal class PartiQLParserDefault : PartiQLParser {
                 p
             }
             val hasTimezoneLiteral = timeString.contains(Regex("[+|-]\\d\\d:\\d\\d$"))
```
alancai98: nit: this isn't part of your PR but the naming of this old variable is slightly confusing. Should probably be named `hasTimezoneOffset`. Similarly for below with timestamp.

`hasTimezoneLiteral` should be `hasTimezoneOffset`

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/io/DatumIonReader.kt
```diff
@@ -34,6 +34,7 @@ class DatumIonReader(
     private val MINUTES_PER_HOUR = 60
     private val INTERVAL_MAX_PRECISION = 9
     private val INTERVAL_MAX_FRACTIONAL_PRECISION = 9
+    private val TIME_PRECISION = 9
```
alancai98: nit: for consistency


```suggestion
    private val TIME_MAX_PRECISION = 9
```

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/util/DatumUtil.kt
alancai98: nit: this computation is common enough that perhaps we should split it to its own private helper function

partiql-ast/src/main/java/org/partiql/ast/DataType.java
XuechunHHH: Makes sense. Will revert the change.

PR #1836, opened 2025-09-30
fix: LOWER(clob) datum return type

alancai98:
## Relevant Issues
- None

## Description
- Fix returned datum for calling `LOWER` on a clob. Previously was returning a string datum when it should have been a clob.
- Behavior now follows what we do for `UPPER`.

## Other Information

## License Information

### DIFF
--- a/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
+++ b/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
@@ -27,7 +27,7 @@ internal val Fn_LOWER__CLOB__CLOB = Function.overload(
     parameters = arrayOf(Parameter("value", PType.clob(Int.MAX_VALUE))),
 
 ) { args ->
-    val string = args[0].string
+    val string = args[0].bytes.toString(Charsets.UTF_8)
     val result = string.lowercase()
     Datum.clob(result.toByteArray())
 }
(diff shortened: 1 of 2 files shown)


partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
```diff
@@ -29,5 +29,5 @@ internal val Fn_LOWER__CLOB__CLOB = Function.overload(
 ) { args ->
     val string = args[0].string
```
mrugeshgharat: I believe we getting input as string or clob. So need to convert to string.

`args[0].bytes.toString(Charsets.UTF_8)`

XuechunHHH (approved):

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
```diff
@@ -27,7 +27,7 @@ internal val Fn_LOWER__CLOB__CLOB = Function.overload(
     parameters = arrayOf(Parameter("value", PType.clob(Int.MAX_VALUE))),
 
 ) { args ->
-    val string = args[0].string
+    val string = args[0].bytes.toString(Charsets.UTF_8)
     val result = string.lowercase()
-    Datum.string(result)
+    Datum.clob(result.toByteArray())
```
xd1313113, line 32: Nit: can use toByteArray(Charsets.UTF_8) for output. It is more understandable. 

Also curious. Is encoding configurable in Partiql?

alancai98: > Nit: can use toByteArray(Charsets.UTF_8) for output. It is more understandable.

For consistency with the `FnUpper` implementation, I'll leave it as it is -- https://github.com/partiql/partiql-lang-kotlin/blob/d290bc7a39eb753eefff2a039837e7bfd5546f35/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnUpper.kt#L30.

> Also curious. Is encoding configurable in Partiql?

Not currently. We haven't had a use case to make it configurable yet.

Issue #1838, opened 2025-10-01
Add string functions support for `CHAR/VARCHAR`


XuechunHHH:
## Description
Currently in PLK-1.+, we are missing the string functions support for `CHAR/VARCHAR` including:

- CHAR_LENGTH
- SUBSTRING
- POSITION
- BIT_LENGTH
- OCTET_LENGTH
- TRIM_CHARS
- TRIM_LEADING
- TRIM_LEADING_CHARS
- TRIM_TRAILING
- TRIM_TRAILING_CHARS
- LIKE
- LIKE_ESCAPE

We also have a temporary walk around function signatures added without actual implementations for string functions including:

- TRIM
- UPPER
- LOWER

And a temporary implementation for string functions including:

- CONCAT

## DoD (Definition of Done)

For `CHAR/VARCHAR` types;

- Handle length argument for CONCAT
- Implement TRIM/UPPER/LOWER
- Register function signatures and implement the rest of string functions

PR #1839, opened 2025-10-01
add: temporary work around string functions support for `CHAR/VARCHAR`

XuechunHHH:
## Relevant Issues
- Related to Issue #1838 

## Description
- Add function signatures for `UPPER/LOWER/TRIM` on `CHAR/VARCHAR` types as a work around.
- Update the conformance test submodule.

## Other Information

## License Information

### DIFF
--- a/partiql-spi/src/main/kotlin/org/partiql/spi/function/Builtins.kt
+++ b/partiql-spi/src/main/kotlin/org/partiql/spi/function/Builtins.kt
@@ -147,6 +147,8 @@ internal object Builtins {
         Fn_LIKE_ESCAPE__STRING_STRING_STRING__BOOL,
         Fn_LIKE_ESCAPE__CLOB_CLOB_CLOB__BOOL,
 
+        Fn_LOWER__CHAR__CHAR,
+        Fn_LOWER__VARCHAR__VARCHAR,
         Fn_LOWER__STRING__STRING,
         Fn_LOWER__CLOB__CLOB,
 
@@ -184,6 +186,8 @@ internal object Builtins {
         Fn_SUBSTRING__CLOB_INT64_INT64__CLOB,
 
         FnTimes,
+        Fn_TRIM__CHAR__CHAR,
+        Fn_TRIM__VARCHAR__VARCHAR,
         Fn_TRIM__STRING__STRING,
         Fn_TRIM__CLOB__CLOB,
 
@@ -202,6 +206,8 @@ internal object Builtins {
         Fn_TRIM_TRAILING_CHARS__STRING_STRING__STRING,
         Fn_TRIM_TRAILING_CHARS__CLOB_CLOB__CLOB,
 
+        Fn_UPPER__CHAR__CHAR,
+        Fn_UPPER__VARCHAR__VARCHAR,
         Fn_UPPER__STRING__STRING,
         Fn_UPPER__CLOB__CLOB,
 
--- a/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
+++ b/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
@@ -8,6 +8,26 @@ import org.partiql.spi.function.Parameter
 import org.partiql.spi.types.PType
 import org.partiql.spi.value.Datum
 
+internal val Fn_LOWER__CHAR__CHAR = Function.overload(
+
+    name = "lower",
+    returns = PType.character(),
+    parameters = arrayOf(Parameter("value", PType.character())),
+
+) { args ->
+    TODO("Not yet implemented")
+}
+
+internal val Fn_LOWER__VARCHAR__VARCHAR = Function.overload(
+
+    name = "lower",
+    returns = PType.varchar(),
+    parameters = arrayOf(Parameter("value", PType.varchar())),
+
+) { args ->
+    TODO("Not yet implemented")
+}
+
 internal val Fn_LOWER__STRING__STRING = Function.overload(
 
     name = "lower",
--- a/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrim.kt
+++ b/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrim.kt
@@ -9,6 +9,26 @@ import org.partiql.spi.utils.FunctionUtils
 import org.partiql.spi.utils.StringUtils.codepointTrim
 import org.partiql.spi.value.Datum
 
+internal val Fn_TRIM__CHAR__CHAR = FunctionUtils.hidden(
+
+    name = "trim",
+    returns = PType.character(),
+    parameters = arrayOf(Parameter("value", PType.character())),
+
+) { args ->
+    TODO("Not yet implemented")
+}
+
+internal val Fn_TRIM__VARCHAR__VARCHAR = FunctionUtils.hidden(
+
+    name = "trim",
+    returns = PType.varchar(),
+    parameters = arrayOf(Parameter("value", PType.varchar())),
+
+) { args ->
+    TODO("Not yet implemented")
+}
+
 /**
  * From section 6.7 of SQL 92 spec:
  * ```
--- a/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnUpper.kt
+++ b/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnUpper.kt
@@ -8,6 +8,26 @@ import org.partiql.spi.function.Parameter
 import org.partiql.spi.types.PType
 import org.partiql.spi.value.Datum
 
+internal val Fn_UPPER__CHAR__CHAR = Function.overload(
+
+    name = "upper",
+    returns = PType.character(),
+    parameters = arrayOf(Parameter("value", PType.character())),
+
+) { args ->
+    TODO("Not yet implemented")
+}
+
+internal val Fn_UPPER__VARCHAR__VARCHAR = Function.overload(
+
+    name = "upper",
+    returns = PType.varchar(),
+    parameters = arrayOf(Parameter("value", PType.varchar())),
+
+) { args ->
+    TODO("Not yet implemented")
+}
+
 internal val Fn_UPPER__STRING__STRING = Function.overload(
 
     name = "upper",


partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
```diff
@@ -8,6 +8,26 @@ import org.partiql.spi.function.Parameter
...
+
+) { args ->
+    TODO("Not yet implemented")
+}
+
+internal val Fn_LOWER__VARCHAR__VARCHAR = Function.overload(
+
+    name = "lower",
+    returns = PType.varchar(),
+    parameters = arrayOf(Parameter("value", PType.varchar())),
```
alancai98: Wonder if we should provide some other default. Currently the default for `PType.character` and `PType.varchar` are `1`. We can change this default in `PType.java`.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
```diff
@@ -8,6 +8,26 @@ import org.partiql.spi.function.Parameter
 import org.partiql.spi.types.PType
 import org.partiql.spi.value.Datum
 
+internal val Fn_LOWER__CHAR__CHAR = Function.overload(
```
alancai98: Also, need a CHANGELOG entry for easier tracking of changes between releases.

CHANGELOG.md
```diff
@@ -26,9 +26,11 @@ Thank you to all who have contributed!
 ## [Unreleased](https://TODO.com) - YYYY-MM-DD
 
 ### Added
+- `UPPER`, `LOWER`, `TRIM` function support for `CHAR/VARCHAR` types
```
XuechunHHH: self-review: Since the length handling should also be done at planning time, add the real function support instead of only siganatures.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnConcat.kt
```diff
@@ -3,63 +3,137 @@
...
-    Datum.string(arg0 + arg1)
+    override fun getInstance(args: Array<PType>): Fn? {
+        val lhsType = args[0]
+        val rhsType = args[1]
+        // Check if both are string types
+        val stringTypes = setOf(PType.CHAR, PType.VARCHAR, PType.STRING, PType.CLOB)
+        if (lhsType.code() !in stringTypes || rhsType.code() !in stringTypes) {
+            return null
+        }
+        // If string types are different, use precedence: CHAR > VARCHAR > STRING > CLOB
```
XuechunHHH: self-review: This follows previous behavior. e.g. `CHAR & VARCHAR ` -> `VARCHAR`, `VARCHAR & STRING` -> `STRING`, `VARCHAR & BLOB` -> `BLOB`...

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnConcat.kt
```diff
@@ -3,63 +3,137 @@
...
+        if (lhsType.code() != rhsType.code()) {
+            val resultType = maxOf(lhsType.code(), rhsType.code())
+            return when (resultType) {
+                PType.VARCHAR -> {
+                    val totalLength = (lhsType.length.toLong() + rhsType.length.toLong()).coerceAtMost(MAXLENGTH.toLong()).toInt()
```
XuechunHHH: self-review: When totalLength exceeds `Int.MAX_VALUE`:

- planning time: return type length with `Int.MAX_VALUE` for type system compatibility
- runtime: throw an error

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnConcat.kt
XuechunHHH, line 1: self-review: Concat is an operator `||` and thus already got tests in `OpConcatTest.kt`, see: https://github.com/partiql/partiql-lang-kotlin/blob/154b9b4a43675caec00d0e9af25292e923e4f5eb/partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/operator/OpConcatTest.kt

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/functions/UpperTest.kt
```diff
@@ -0,0 +1,72 @@
...
+        val planner = PartiQLPlanner.builder().build()
+        val result = planner.plan(ast, session)
+        val query = result.plan.action as Action.Query
+        val actualType = query.rex.type.pType
+        assertEquals(PType.CHAR, actualType.code())
+        assertEquals(5, actualType.length)
+    }
+
+    @Test
+    fun `upper preserves VARCHAR length`() {
```
lucaxchen: nit: maybe `upper preserves VARCHAR length and type`? Since previously we have PTyp.CLOB returned as default. Same applies to others.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
```diff
@@ -3,51 +3,66 @@
...
+                Datum.string(result)
+            }
+            PType.CLOB -> Function.instance(
+                name = "lower",
+                returns = PType.clob(inputType.length),
+                parameters = arrayOf(Parameter("value", inputType)),
+            ) { args ->
+                val string = args[0].bytes.toString(Charsets.UTF_8)
+                val result = string.lowercase()
+                Datum.clob(result.toByteArray())
```
lucaxchen: out of curious, in `Datum`, we treat `PType.CLOB` as a byte array. I might be wrong here, but I think it should be just text by the SQL spec.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/functions/TrimTest.kt
```diff
@@ -0,0 +1,91 @@
...
+            )
+            .build()
+        val parseResult = PartiQLParser.standard().parse("TRIM(CAST('  HELLO  ' AS CHAR(9)))")
+        val ast = parseResult.statements[0]
+        val planner = PartiQLPlanner.builder().build()
+        val result = planner.plan(ast, session)
+        val query = result.plan.action as Action.Query
+        val actualType = query.rex.type.pType
+        assertEquals(PType.CHAR, actualType.code())
+        assertEquals(9, actualType.length)
```
lucaxchen, line 15: question: what will happen to Datum? The type definition will stay the same. but SQL spec 1999 says: 
```
The declared type of the <trim function> is variable-length character string with maximum
length equal to the fixed length or maximum variable length of the <trim source>.
```

lucaxchen: cc @alancai98 for more info

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrim.kt
```diff
@@ -38,42 +42,57 @@ import org.partiql.spi.value.Datum
...
+        val inputType = args[0]
+        return when (inputType.code()) {
+            PType.CHAR -> Function.instance(
+                name = FunctionUtils.hide("trim"),
+                returns = PType.character(inputType.length),
+                parameters = arrayOf(Parameter("value", inputType)),
+            ) { args ->
+                val string = args[0].bytes.toString(Charsets.UTF_8)
+                val result = string.codepointTrim()
+                Datum.character(result, inputType.length)
```
lucaxchen: it returns Datum.char.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrim.kt
```diff
@@ -38,42 +42,57 @@ import org.partiql.spi.value.Datum
...
+                Datum.string(result)
+            }
+            PType.CLOB -> Function.instance(
+                name = FunctionUtils.hide("trim"),
+                returns = PType.clob(inputType.length),
+                parameters = arrayOf(Parameter("value", inputType)),
+            ) { args ->
+                val string = args[0].bytes.toString(Charsets.UTF_8)
+                val result = string.codepointTrim()
+                Datum.clob(result.toByteArray())
```
lucaxchen: same question as upper/lower.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
XuechunHHH: We currently treat both `CLOB`and `BLOB` as `byte[]` in Datum. See https://github.com/partiql/partiql-lang-kotlin/blob/154b9b4a43675caec00d0e9af25292e923e4f5eb/partiql-spi/src/main/java/org/partiql/spi/value/Datum.java#L723-L763

Thanks for the call out. Agree that `CLOB` should be test-based instead. Both `PType` and `Datum` need to be formalized further.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/functions/TrimTest.kt
XuechunHHH: I didn't get the question on `Datum`, do you mean by character string instead (`Char/Varchar`)?

> character string with maximum length equal to the fixed length or maximum variable length of the <trim source>

From my read, the result type of `trim(CHAR(len))` will be:
- len > Int.MAX_VALUE: CHAR(Int.MAX_VALUE) 
- len <= Int.MAX_VALUE: CHAR(len)

And the first case will never happen since the passed in `CHAR(len)` will be checked.

This is its current behavior.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrim.kt
XuechunHHH: Yes. It will preserve the type passed in.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
```diff
@@ -3,31 +3,66 @@
...
-) { args ->
-    val string = args[0].bytes.toString(Charsets.UTF_8)
-    val result = string.lowercase()
-    Datum.clob(result.toByteArray())
-}
+// Keep the old function names for backward compatibility in Builtins.kt
+internal val Fn_LOWER__CHAR__CHAR = FnLower
+internal val Fn_LOWER__VARCHAR__VARCHAR = FnLower
+internal val Fn_LOWER__STRING__STRING = FnLower
+internal val Fn_LOWER__CLOB__CLOB = FnLower
```
alancai98: Since these are all internal, we don't need to preserve the prior variable names. So in `Builtins.kt`, we only need to reference one `FnLower`. Similarly for the other functions.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
```diff
@@ -3,31 +3,66 @@
...
+                val result = string.lowercase()
+                Datum.character(result, inputType.length)
+            }
+            PType.VARCHAR -> Function.instance(
+                name = "lower",
+                returns = PType.varchar(inputType.length),
+                parameters = arrayOf(Parameter("value", inputType)),
+            ) { args ->
+                val string = args[0].bytes.toString(Charsets.UTF_8)
+                val result = string.lowercase()
```
alancai98: The evaluation implementations varchar and char should follow what we do for `PType.STRING` (i.e. no call to `.bytes` and directly just call `.string`).

Btw just noticed for Datum's `.string`, it should also include `VARCHAR`. Could you also update the javadoc to mention `VARCHAR`?

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
```diff
@@ -3,31 +3,66 @@
...
+            PType.CHAR -> Function.instance(
+                name = "lower",
+                returns = PType.character(inputType.length),
+                parameters = arrayOf(Parameter("value", inputType)),
+            ) { args ->
+                val string = args[0].bytes.toString(Charsets.UTF_8)
+                val result = string.lowercase()
+                Datum.character(result, inputType.length)
+            }
+            PType.VARCHAR -> Function.instance(
```
alancai98: May be worth documenting (probably javadoc on `FnLower`) the behavior for type parameters. Specifically for varchar, char, and clob, the `length` parameter is preserved from the input.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrim.kt
```diff
@@ -38,26 +42,57 @@ import org.partiql.spi.value.Datum
...
+                val string = args[0].bytes.toString(Charsets.UTF_8)
+                val result = string.codepointTrim()
+                Datum.character(result, inputType.length)
+            }
+            PType.VARCHAR -> Function.instance(
+                name = FunctionUtils.hide("trim"),
+                returns = PType.varchar(inputType.length),
+                parameters = arrayOf(Parameter("value", inputType)),
+            ) { args ->
+                val string = args[0].bytes.toString(Charsets.UTF_8)
```
alancai98: Similarly for Trim, can call `.string` for varchar and char fn impls

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrim.kt
```diff
@@ -38,26 +42,57 @@ import org.partiql.spi.value.Datum
  *  * `<trim character> ::= <character value expression>`
  *  * `<trim source> ::= <character value expression>`
  */
-internal val Fn_TRIM__STRING__STRING = FunctionUtils.hidden(
+internal object FnTrim : FnOverload() {
```
alancai98, line 34: Can add comment about for trim, we preserve the varchar, char, and clob length from the input text to the return type.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnUpper.kt
```diff
@@ -3,31 +3,66 @@
...
+                val string = args[0].bytes.toString(Charsets.UTF_8)
+                val result = string.uppercase()
+                Datum.character(result, inputType.length)
+            }
+            PType.VARCHAR -> Function.instance(
+                name = "upper",
+                returns = PType.varchar(inputType.length),
+                parameters = arrayOf(Parameter("value", inputType)),
+            ) { args ->
+                val string = args[0].bytes.toString(Charsets.UTF_8)
```
alancai98: And for this case, we don't need the `.bytes` call. Can just call `.string`.

And let's add the similar docs for the length parameter that we added for `lower`.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnConcat.kt
```diff
@@ -3,63 +3,137 @@
...
 
-) { args ->
-    val arg0 = args[0].string
-    val arg1 = args[1].string
-    Datum.string(arg0 + arg1)
+    override fun getInstance(args: Array<PType>): Fn? {
+        val lhsType = args[0]
+        val rhsType = args[1]
+        // Check if both are string types
+        val stringTypes = setOf(PType.CHAR, PType.VARCHAR, PType.STRING, PType.CLOB)
```
alancai98: Can reference `SqlTypeFamily.TEXT` here rather than creating this inline set https://github.com/partiql/partiql-lang-kotlin/blob/add--string-fn-support-on-char/partiql-spi/src/main/kotlin/org/partiql/spi/internal/SqlTypeFamily.kt#L51.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnConcat.kt
alancai98: Where was the previous type precedence behavior defined?

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnConcat.kt
alancai98: I think if `totalLength` exceeds at compile time, we should give an error and not coerce to `Int.MAX_VALUE`.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnConcat.kt
```diff
@@ -3,63 +3,137 @@
...
+                    parameters = arrayOf(Parameter("lhs", lhsType), Parameter("rhs", rhsType)),
+                ) { args ->
+                    val arg0 = args[0].string
+                    val arg1 = args[1].string
+                    Datum.string(arg0 + arg1)
+                }
+                PType.CLOB -> {
+                    val totalLength = when {
+                        lhsType.code() == PType.STRING || rhsType.code() == PType.STRING -> MAXLENGTH
+                        else -> (lhsType.length.toLong() + rhsType.length.toLong()).coerceAtMost(MAXLENGTH.toLong()).toInt()
```
alancai98: nit: this computation to check for overflow is repeated in a lot of places and is prone to error. May be helpful to break it out into its own helper function. Also can probably simplify the coercion without upcasting to long by simply doing

`arg1 >= 0 && arg2 >= 0 && arg1 + arg2 < 0` => overflow occurred

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/functions/LowerTest.kt
```diff
@@ -0,0 +1,91 @@
...
+            )
+            .build()
+        val parseResult = PartiQLParser.standard().parse("LOWER(CAST('hello' AS CHAR(5)))")
+        val ast = parseResult.statements[0]
+        val planner = PartiQLPlanner.builder().build()
+        val result = planner.plan(ast, session)
+        val query = result.plan.action as Action.Query
+        val actualType = query.rex.type.pType
+        assertEquals(PType.CHAR, actualType.code())
+        assertEquals(5, actualType.length)
```
alancai98, line 15: nit: can move the session, parsing, planning into it's own helper function for reuse. The same logic is repeated for a lot of the tests.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnConcat.kt
alancai98: For full test coverage, we can add a simple test for VARCHAR/CHAR/CLOB with two arguments that when concatenated, exceeds the `Int.MAX_VALUE`. E.g. `CAST('some string' AS VARCHAR(<Int.MAX_VALUE>)) || CAST('a' AS VARCHAR(1))`

partiql-spi/src/main/kotlin/org/partiql/spi/function/Builtins.kt
```diff
@@ -184,6 +186,8 @@ internal object Builtins {
         Fn_SUBSTRING__CLOB_INT64_INT64__CLOB,
 
         FnTimes,
+        Fn_TRIM__CHAR__CHAR,
+        Fn_TRIM__VARCHAR__VARCHAR,
         Fn_TRIM__STRING__STRING,
         Fn_TRIM__CLOB__CLOB,
```
alancai98, line 182: We are also missing implementations for the other `TRIM` functions below this line. The `TRIM` function in text maps to other trim implementation internally.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
XuechunHHH: No problem.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrim.kt
XuechunHHH: We actually don't do that for `clob`. I didn't change any code related to `clob` in this PR. So the string functions behavior for `blob` is using `max_length = Int.MAX_VALUE` for both input and output. Do you think we should also change `clob` behavior to preserve length in this PR?

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnConcat.kt
XuechunHHH: Clear move!

partiql-spi/src/main/kotlin/org/partiql/spi/function/Builtins.kt
XuechunHHH: Yep. Currently, our customer team only need support for this simple `TRIM` function. So only cover this in the PR. The rest of string functions support are tracked by https://github.com/partiql/partiql-lang-kotlin/issues/1838

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrim.kt
alancai98: Yeah, let's do that in this PR for clob since we're already working with these files.

partiql-spi/src/main/kotlin/org/partiql/spi/function/Builtins.kt
alancai98: Ok cool. Maybe let's add a TODO and link the issue in those files for better tracking?

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrim.kt
XuechunHHH: Sounds good.

partiql-spi/src/main/kotlin/org/partiql/spi/function/Builtins.kt
XuechunHHH: no problem

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
XuechunHHH: Yes... I thought I have changed to `.string()` for all `CHAR/VARCHAR`, but it turns out I only did it in `CONCAT`. And I will update the javadoc to include `VARCHAR` in next commit.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnConcat.kt
XuechunHHH: The string types precedence was defined at https://github.com/partiql/partiql-lang-kotlin/blob/154b9b4a43675caec00d0e9af25292e923e4f5eb/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/TypePrecedence.kt#L22-L25

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnConcat.kt
XuechunHHH: Nice call. This will be more efficient.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnConcat.kt
```diff
@@ -3,63 +3,137 @@
...
+        val lhsType = args[0]
+        val rhsType = args[1]
+        // Check if both are string types
+        val stringTypes = setOf(PType.CHAR, PType.VARCHAR, PType.STRING, PType.CLOB)
+        if (lhsType.code() !in stringTypes || rhsType.code() !in stringTypes) {
+            return null
+        }
+        // If string types are different, use precedence: CHAR > VARCHAR > STRING > CLOB
+        if (lhsType.code() != rhsType.code()) {
+            val resultType = maxOf(lhsType.code(), rhsType.code())
```
xd1313113: does maxof(code) determine the precedence?

XuechunHHH: The string types precedence is defined at https://github.com/partiql/partiql-lang-kotlin/blob/154b9b4a43675caec00d0e9af25292e923e4f5eb/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/TypePrecedence.kt#L22-L25

Lower code means higher precedence. So `maxOf()` picks the one with lower precedence.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnUpper.kt
```diff
@@ -3,31 +3,66 @@
...
-    name = "upper",
-    returns = PType.clob(Int.MAX_VALUE),
-    parameters = arrayOf(Parameter("value", PType.clob(Int.MAX_VALUE))),
-
-) { args ->
-    val string = args[0].bytes.toString(Charsets.UTF_8)
-    val result = string.uppercase()
-    Datum.clob(result.toByteArray())
-}
+// Keep the old function names for backward compatibility in Builtins.kt
```
xd1313113: old function only has Fn_UPPER__STRING__STRING and Fn_UPPER__CLOB__CLOB.

do you need add Fn_UPPER__CHAR__CHAR and Fn_UPPER__VARCHAR__VARCHAR ?

Same applies to other functions

CHANGELOG.md
```diff
@@ -27,8 +27,11 @@ Thank you to all who have contributed!
 
 ### Added
 - **EXPERIMENTAL**: Adds end-to-end support for window functions including RANK, DENSE_RANK, LAG, LEAD, and ROW_NUMBER functions alongside the WINDOW clause.
+- `UPPER`, `LOWER`, `TRIM` functions support for `CHAR/VARCHAR`
 
 ### Changed
+- Formalize `UPPER`, `LOWER`, `TRIM` and `CONCAT` length and type handling on string types
+- `CLOB` now supports length parameters with `CLOB(n)` syntax
```
XuechunHHH: Self-review: Added support for max_length parameters in the `CLOB` type, which is `CLOB(n)` instead of `CLOB` now. This change aligns our syntax with the SQL-99 specification, which defines CLOB as optionally accepting a length parameter:

```
 <character string type> ::= 
      CHARACTER [ <left paren> <length> <right paren> ] 
      ...
      | CLOB [ <left paren> <large object length> <right paren> ]
```

partiql-planner/src/test/kotlin/org/partiql/planner/util/Utils.kt
```diff
@@ -72,7 +72,7 @@ val allCharStringPType = setOf(
     PType.character(256), // TODO: Length
     PType.varchar(256), // TODO: Length
     PType.string(),
-    PType.clob(Int.MAX_VALUE), // TODO: Length
+    PType.clob(256), // TODO: Length
```
XuechunHHH, line 75: self-review: This change avoid corner test (overflow) in planner. Previously CLOB actually had fixed param max_length = Int.MAX_VALUE in the implementation.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnUpper.kt
XuechunHHH: > Since these are all internal, we don't need to preserve the prior variable names. So in Builtins.kt, we only need to reference one FnLower. Similarly for the other functions.

Use `FnLower` instead.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
XuechunHHH: I think it's a bug in our code.
cc: @alancai98 for confirmation

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/functions/TrimTest.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.partiql.spi.errors.PRuntimeException
+import org.partiql.spi.types.PType
+import kotlin.test.assertEquals
+
+class TrimTest {
+
+    @Test
+    fun `trim preserves CHAR length and returns VARCHAR type`() {
+        val actualType = FnTestUtils.getQueryResultType("TRIM(CAST('  HELLO  ' AS CHAR(9)))")
+        assertEquals(PType.VARCHAR, actualType.code())
```
lucaxchen, line 14: This looks great!

partiql-planner/src/test/kotlin/org/partiql/planner/util/Utils.kt
lucaxchen: Just wanna know more about the context here, could you elaborate the situation? Got a little confused here, if previously, we set the length to `int.MAX_VALUE`, and have the corner test in the planner, and it got pass, why would it affect now?

partiql-parser/src/main/antlr/PartiQLParser.g4
```diff
@@ -1262,11 +1262,11 @@ type
     : datatype=(
         BOOL | BOOLEAN | TINYINT | SMALLINT | INTEGER2 | INT2 | INTEGER | INT | INTEGER4 | INT4
         | INTEGER8 | INT8 | BIGINT | REAL | CHAR | CHARACTER
-        | STRING | SYMBOL | BLOB | CLOB | DATE | ANY
+        | STRING | SYMBOL | BLOB | DATE | ANY
```
alancai98: Since we're already modifying these files, let's also  move `BLOB` to `TypeArgSingle` + add CHANGELOG callout like we do for CLOB.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnUtils.kt
```diff
@@ -0,0 +1,35 @@
...
+        checkLengthOverflow(length1, length2)
+        return length1 + length2
+    }
+
+    /**
+     * Gets the length of a type, handling STRING types that don't have length constraints.
+     */
+    fun getTypeLength(type: PType): Int {
+        return when (type.code()) {
+            PType.STRING -> 0
```
lucaxchen: for `PType.STRING`, if we call getTypeLength, it returns `0`. It seems odd to me, maybe we should throw an error?

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLower.kt
```diff
@@ -3,31 +3,82 @@
...
+ * According to SQL specification:
+ * - The declared type of the result is the declared type of the <character value expression>
+ * - For CHAR, VARCHAR, and CLOB types, the length parameter is preserved from the input type
+ * - For STRING type, no length parameter is applicable
+ *
+ * Type preservation behavior:
+ * - CHAR(n) → CHAR(n)
+ * - VARCHAR(n) → VARCHAR(n)
+ * - CLOB(n) → CLOB(n)
+ * - STRING → STRING
```
alancai98: nit: SQL doesn't have a STRING type. So let's separate the discussion out from the comment about SQL spec.

Similarly for the other doc comments.

CHANGELOG.md
alancai98: nit: this (along with the BLOB change) are more of a feature addition rather than a changed behavior.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnConcat.kt
```diff
@@ -3,63 +3,104 @@
...
+ * - If either argument is STRING: result is STRING (no length parameter)
+ * - If either argument is CLOB: result is CLOB with length = min(L1 + L2, max_clob_length)
+ * - If either argument is VARCHAR: result is VARCHAR with length = min(L1 + L2, max_varchar_length)
+ * - If both arguments are CHAR: result is CHAR with length = min(L1 + L2, max_char_length)
+ *
+ * Type precedence behavior:
+ * - STRING || any → STRING
+ * - CLOB(n) || CHAR(m)/VARCHAR(m) → CLOB(n + m)
+ * - VARCHAR(n) || CHAR(m) → VARCHAR(n + m)
+ * - CHAR(n) || CHAR(m) → CHAR(n + m)
```
alancai98: nit: for consistency here, let's use `L1` and `L2` rather than `n` and `m`.

partiql-planner/src/test/kotlin/org/partiql/planner/util/Utils.kt
XuechunHHH: Cuz previously, we didn't do any length handling on `PType.CLOB`. In the actual implementation, only `CLOB(Int.MAX_VALUE)` is allowed. So when doing the tests in concat, previously it should be: `CLOB(Int.MAX_VALUE) || CLOB(Int.MAX_VALUE) → CLOB(Int.MAX_VALUE)`, no error will be raised. These tests are only for returning type, not for length handling intention. So just change it to a smaller integer.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnUtils.kt
XuechunHHH: This helper function only used for concat length handling. Intentionally let it be `0` for computation.

alancai98: Hmm probably something is off with the type precedence? I think if we have clob and string, it should return string for concat.

XuechunHHH: I see the issue here. Since SQL spec doesn't have `STRING()` with unlimited length support and `CLOB` should be larger than `STRING`, in our previous PType code [PType.java](https://github.com/partiql/partiql-lang-kotlin/blob/ac49c001cec2e04da56630708277d75e715a8ad6/partiql-spi/src/main/java/org/partiql/spi/types/PType.java#L376-L421), which means`STRING` has higher precedence than `CLOB`. I will update the cast table to make it consistent.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnLike.kt
```diff
@@ -3,54 +3,98 @@
```
XuechunHHH, line 3: Since `like` is predicate, already got planner tests at https://github.com/partiql/partiql-lang-kotlin/blob/ac49c001cec2e04da56630708277d75e715a8ad6/partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/predicate/OpLikeTest.kt

PR #1840, opened 2025-10-06
update: change max/default length support for `PType.CLOB`

XuechunHHH:
## Description
- Extend max/default length support for `PType.CLOB` from `Integer.MAX_VALUE` to `Long.MAX_VALUE`
- Update api `PType.getLength()` return type

## Other Information

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,8 @@ Thank you to all who have contributed!
 ### Added
 
 ### Changed
-- Change max and default length for `PType.CLOB` from `Integer.MAX_VALUE` to `Long.MAX_VALUE` 
+- Change max/default length for `PType.CLOB` from `Integer.MAX_VALUE` to `Long.MAX_VALUE` 
+- Update `PType.getLength` API return value from `int` to `long`
 
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


partiql-spi/src/main/java/org/partiql/spi/types/PType.java
```diff
@@ -963,16 +963,16 @@ public static PType string() {
      */
     @NotNull
     @SuppressWarnings("SameParameterValue")
-    public static PType clob(int length) {
-        return new PTypeWithMaxLength(CLOB, length);
+    public static PType clob(long length) {
```
xd1313113, line 966: Shall we keep PType clob(int length) and cast to long internally?

partiql-spi/src/main/java/org/partiql/spi/types/PTypeWithLongMaxLength.java
```diff
@@ -0,0 +1,35 @@
+package org.partiql.spi.types;
+
+import java.util.Objects;
+
+class PTypeWithLongMaxLength extends PType {
```
xd1313113, line 5: why not retrofit class `PTypeWithMaxLength.java‎` to long directly. Do we want to support both Int and long as length type?

partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
```diff
@@ -732,7 +732,7 @@ static Datum clob(@NotNull byte[] value) throws PRuntimeException {
      * @throws PRuntimeException if the value could not fit into the requested length, or if the requested length is not allowed ({@link org.partiql.spi.errors.PError#INTERNAL_ERROR})
      */
     @NotNull
-    static Datum clob(@NotNull byte[] value, int length) throws PRuntimeException {
+    static Datum clob(@NotNull byte[] value, long length) throws PRuntimeException {
```
xd1313113, line 735: same do we need to keep int function?

PR #1844, opened 2025-10-13
Support aggregation function calls in the `ORDER BY` and `HAVING` clause

xd1313113:
## Relevant Issues

## Description
Support aggregation function calls in the `ORDER BY` and `HAVING` clause, which is supported described in partiql sepc section 12.5

### Before
partiql ▶ SELECT max(b) FROM [{'a':1, 'b':2},{'a':2, 'b':3}] group by a ORDER BY min(b);
w: [semantic] Variable reference (b) could not be found in the database environment or in the set of available locals. Locals: [_0.$agg_0, _0.a].
e: [semantic] Undefined function min(UNKNOWN).

Failed with given input. Please see the above errors.

### After
partiql ▶ SELECT max(b) FROM [{'a':1, 'b':2},{'a':2, 'b':3}] group by a ORDER BY min(b);

=== RESULT ===
[
  {
    '_1': 2
  },
  {
    '_1': 3
  }
]

OK!

## Other Information

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ Thank you to all who have contributed!
 ## [Unreleased](https://TODO.com) - YYYY-MM-DD
 
 ### Added
-- Support aggregation function calls in the `ORDER BY` clause
+- Support aggregation function calls in the `ORDER BY` and `HAVING` clause
 
 ### Changed
 


partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -437,13 +437,17 @@ internal object RelConverter {
          *         1. Ast.Expr.SFW has every Ast.Expr.CallAgg replaced by a synthetic Ast.Expr.Var
          *         2. Rel which has the appropriate Rex.Agg calls and groups
          */
-        private fun convertAgg(input: Rel, select: QueryBody.SFW, groupBy: GroupBy?, env: Env): Pair<QueryBody.SFW, Rel> {
-            // Rewrite and extract all aggregations in the SELECT clause
-            val (sel, aggregations) = AggregationTransform.apply(select, env)
+        private fun convertAgg(input: Rel, body: QueryBody.SFW, orderBy: OrderBy?, env: Env): Triple<QueryBody.SFW, OrderBy?, Rel> {
```
alancai98: The kdoc needs updated for this function following this PR's changes. Also the return is complicated enough that perhaps we should make a dedicated class rather than use `Triple`.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -437,13 +437,17 @@ internal object RelConverter {
          *         1. Ast.Expr.SFW has every Ast.Expr.CallAgg replaced by a synthetic Ast.Expr.Var
          *         2. Rel which has the appropriate Rex.Agg calls and groups
          */
-        private fun convertAgg(input: Rel, select: QueryBody.SFW, groupBy: GroupBy?, env: Env): Pair<QueryBody.SFW, Rel> {
```
alancai98, line 440: nit: for all of the aggregation rewriting, I think we should move `convertAgg` and stuff related to `AggregationTransform` into another file since this file's getting pretty large.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -923,36 +927,17 @@ internal object RelConverter {
     }
 
     /**
-     * Rewrites a SELECT node replacing (and extracting) each aggregation `i` with a synthetic field name `$agg_i`.
```
alancai98, line 926: We shouldn't be deleting the content of the prior comment related to the synthetic field names.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -923,36 +927,17 @@ internal object RelConverter {
...
+     * Base class for rewriting AST nodes and extracting aggregation calls.
      */
-    private object AggregationTransform : AstRewriter<AggregationTransform.Context>() {
-        private data class Context(
+    private abstract class AggregationTransform : AstRewriter<AggregationTransform.Context>() {
+        protected data class Context(
```
alancai98: Could you provide more context for why we need to break the aggregation rewriting for the SFW and ORDER BY into separate rewriters (`SFWAggregationTransform` and `OrderByAggregationTransform`) rather than one rewriter? Just trying to better understand the refactoring of `AggregationTransform` here.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -923,36 +927,17 @@ internal object RelConverter {
     }
 
     /**
-     * Rewrites a SELECT node replacing (and extracting) each aggregation `i` with a synthetic field name `$agg_i`.
+     * Base class for rewriting AST nodes and extracting aggregation calls.
      */
-    private object AggregationTransform : AstRewriter<AggregationTransform.Context>() {
-        private data class Context(
+    private abstract class AggregationTransform : AstRewriter<AggregationTransform.Context>() {
```
alancai98: If we need to keep two separate rewriters extending this `abstract class`, I think this factoring using inheritance is overcomplicated. For instance, the shared `Context` specifies a `keys` field that is only used by the `SFWAggregationTransform`.

There are simpler ways to achieve code reuse of the `ExprCall` visiting and `isAggregateCall` functions that we can talk about if we actually need the two separate rewriters.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Thank you to all who have contributed!
 ## [Unreleased](https://TODO.com) - YYYY-MM-DD
 
 ### Added
+- Fix that the ORDER BY does not work with aggregation function
```
alancai98:
```suggestion
- Support aggregation function calls in the `ORDER BY` clause
```

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -437,13 +437,17 @@ internal object RelConverter {
...
          */
-        private fun convertAgg(input: Rel, select: QueryBody.SFW, groupBy: GroupBy?, env: Env): Pair<QueryBody.SFW, Rel> {
-            // Rewrite and extract all aggregations in the SELECT clause
-            val (sel, aggregations) = AggregationTransform.apply(select, env)
+        private fun convertAgg(input: Rel, body: QueryBody.SFW, orderBy: OrderBy?, env: Env): Triple<QueryBody.SFW, OrderBy?, Rel> {
+            val groupBy = body.groupBy
+            // Rewrite and extract all aggregations in the SFW clause
+            val aggregations = mutableListOf<ExprCall>()
+            val sel = SFWAggregationTransform.apply(body, aggregations, env)
+            // Rewrite and extract all aggregations in the Order By clause
```
XuechunHHH: So, `SELECT COUNT(id) FROM table ORDER BY COUNT(id)` will create two separate calls `[COUNT(id), COUNT(id)]` right? Is it possible to add deduplicate here?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
xd1313113: refactored to avoid using triple

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
xd1313113: Moved AggregationTransform to a different file but kept convertAgg for consistent for now.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
xd1313113: reverted

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
xd1313113: One rewrite works. SFW and ORDER BY are separate node. I was thinking to provide rewrite entry point for each of them.  I reimplemented as one AggregationTransform.

CHANGELOG.md
xd1313113: thanks

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
xd1313113: You are right. inheritance is overcomplicated. Reimplemented AggregationTransform class

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
xd1313113: We do no have a easy to detect duplicate aggregation function. It might be a good optimization.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/AggregationTransform.kt
```diff
@@ -0,0 +1,113 @@
...
+import org.partiql.ast.OrderBy
+import org.partiql.ast.QueryBody
+import org.partiql.ast.Select
+import org.partiql.ast.SelectValue
+import org.partiql.ast.expr.Expr
+import org.partiql.ast.expr.ExprCall
+import org.partiql.planner.internal.Env
+
+/**
+ * Handles aggregation transformation for both SELECT, HAVING and ORDER BY clauses.
```
alancai98:
```suggestion
 * Handles aggregation transformation for both SELECT, HAVING, and ORDER BY clauses.
```

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Thank you to all who have contributed!
 ## [Unreleased](https://TODO.com) - YYYY-MM-DD
 
 ### Added
+- Support aggregation function calls in the `ORDER BY` clause
```
alancai98: Let's also call out in the CHANGELOG + PR description + PR title that aggregations in the `HAVING` clause is also supported w/ this PR's change.

xd1313113: done

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/AggregationTransform.kt
xd1313113: done

CHANGELOG.md
alancai98: Don't think this was addressed yet in the CHANGELOG.

xd1313113: Sorry, miss this file in commit. Just pushed

PR #1849, opened 2025-10-22
add support to transcribe `Overlaps` function

xd1313113:
## Relevant Issues
- [Related To] Issue partiql/partiql-scribe#111

## Description
- add support to transcribe `Overlaps` function

The PR change the AST to SQL for overlaps function 
from
`OVERLAPS([DATE '2023-01-01', DATE '2023-01-05'], [DATE '2023-01-05', DATE '2023-01-10'])`
to 
`[DATE '2023-01-01', DATE '2023-01-05'] OVERLAPS [DATE '2023-01-05', DATE '2023-01-10'];`

## Other Information

## License Information

### DIFF
--- a/partiql-ast/src/test/kotlin/org/partiql/ast/sql/SqlDialectTest.kt
+++ b/partiql-ast/src/test/kotlin/org/partiql/ast/sql/SqlDialectTest.kt
@@ -1198,6 +1198,13 @@ class SqlDialectTest {
                 )
             ),
             expect(
+                "(x, y) OVERLAPS (a, b)",
+                exprOverlaps(
+                    lhs = exprRowValue(listOf(v("x"), v("y"))),
+                    rhs = exprRowValue(listOf(v("a"), v("b")))
+                )
+            ),
+            expect(
                 "[x, y] OVERLAPS [a, b]",
                 exprOverlaps(
                     lhs = exprArray(listOf(v("x"), v("y"))),


partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
```diff
@@ -970,4 +979,17 @@ public abstract class SqlDialect : AstVisitor<SqlBlock, SqlBlock>() {
         true -> text // verbatim ..
         false -> "\"$text\""
     }
+
+    private fun visitExprOverlapsArgs(expr: Expr, tail: SqlBlock): SqlBlock {
```
XuechunHHH: Current implementation can handle simple cases like: `array1 OVERLAPS array2`. But when we have more complex expr comes with operator or parentheses, potential issues might raise. 

So, should there be any parentheses handling around the arguments? And are there any precedence considerations for the `OVERLAPS` operator?

Example can be like: 
```
SELECT * FROM table WHERE col1 + col2 OVERLAPS col3 * col4 AND other_condition
```
Should it be interpreted as something like:
```
SELECT * FROM table WHERE (col1 + col2) OVERLAPS (col3 * col4) AND other_condition
```
or: 
```
SELECT * FROM table WHERE col1 + col2 OVERLAPS col3 * (col4 AND other_condition)
```
Another expr might have issue is:
```
condition1 AND array1 OVERLAPS array2 OR condition2
```
Could be interpreted as:
```
(condition1 AND array1) OVERLAPS (array2 OR condition2) 
```
Instead of:
```
condition1 AND (array1 OVERLAPS array2) OR condition2 
```
Another example with camparison operator:
```
date1 > start_date OVERLAPS date2 < end_date
```
Should be:
```
(date1 > start_date) OVERLAPS (date2 < end_date)
```
instead of :
```
date1 > (start_date OVERLAPS date2) < end_date
```

xd1313113: Synced offline, precedence and parsing has been taken care of parser and planner. We only handle legal cases for exprOverlaps node, which has two arrays/bag as parameters.

am357 (dismissed):

partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
```diff
@@ -638,6 +639,14 @@ public abstract class SqlDialect : AstVisitor<SqlBlock, SqlBlock>() {
         return t
     }
 
+    override fun visitExprOverlaps(node: ExprOverlaps, tail: SqlBlock): SqlBlock {
+        var t = tail
+        t = visitExprOverlapsArg(node.lhs, t)
```
yliuuuu: Could you please help me understand the need of `visitExprOverlapsArg`?

Why would we need to branch out on the type of expr here? 

Or how is `Overlaps` different than `in collection functions` so that `Overlaps`  would requires branching out on the expression type while `InCollection` seemingly does not? 

```
    override fun visitExprInCollection(node: ExprInCollection, tail: SqlBlock): SqlBlock {
        var t = tail
        t = visitExprWrapped(node.lhs, t)
        t = t concat if (node.isNot) " NOT IN " else " IN "
        t = visitExprWrapped(node.rhs, t)
        return t
    }
```

xd1313113: The `overlaps` takes two arguments left and right like below.  We modeled both argument as either an array or bag. 
By default when we transcribe array back to sql, we use bracket parenthesis`[]`. However overlaps requires around parenthesis `()`, Thus we need special handling to override.

`(DATE '2023-01-01', DATE '2023-01-05') OVERLAPS (DATE '2023-01-05', DATE '2023-01-10');`

yliuuuu: I see. In that case, it is a bit weird to see the two parameters get modeled as array....

Like for array: `(DATE '2023-01-01', DATE '2023-01-05')` and `[DATE '2023-01-01', DATE '2023-01-05']` should both work, and therefore transitively `[DATE '2023-01-01', DATE '2023-01-05'] OVERLAPS [DATE '2023-01-05', DATE '2023-01-10'];` must work. 

So interesting to understand is this support is unintentional.

partiql-ast/src/test/kotlin/org/partiql/ast/sql/SqlDialectTest.kt
```diff
@@ -1198,14 +1198,14 @@ class SqlDialectTest {
...
-                "(x, y) OVERLAPS (a, b)",
+                "[x, y] OVERLAPS [a, b]",
                 exprOverlaps(
                     lhs = exprArray(listOf(v("x"), v("y"))),
                     rhs = exprArray(listOf(v("a"), v("b")))
                 )
             ),
             expect(
-                "(x, y) OVERLAPS (a, b)",
+                "<<x, y>> OVERLAPS <<a, b>>",
```
am357, line 1215: By the most recent change, do we mean that we want to support the following:

- `[x, y] OVERLAPS [a, b]`
- `<<x, y>> OVERLAPS <<a, b>>`

xd1313113: yes, get some clarification from Xuechun and Yingtao, array `[]` and bag `<<>>` are supported and modeled internally.  RowValue expr `()` from SQL spec are also supported in parsing but coerced to array internally during planning. So, we use default expr visitor to print for PartiQL. But will override to `()` for other SQL dialects

partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
XuechunHHH: > (DATE '2023-01-01', DATE '2023-01-05') and [DATE '2023-01-01', DATE '2023-01-05']

This is supported intentionally. See - https://github.com/partiql/partiql-tests/blob/67d22dcd5a7eeaaa2624be9d1bb3da6a8b9faf0a/partiql-tests-data/eval/primitives/operators/overlaps.ion#L409

partiql-ast/src/test/kotlin/org/partiql/ast/sql/SqlDialectTest.kt
XuechunHHH: So, we actually support `<row values expression> OVERLAPS <row values expression>` as in SQL-spec:

<row values expression> in PartiQL:
- `[x, y]` list
-  `(x, y)` row
-  `ROW(x, y)` row

We also extendedly add support to bag `<<>>`, which doesn't have indexing due to the order for OVERLAPS doesn't matter.

Which means, any combinations of these four syntax are all supported. See more tests at - https://github.com/partiql/partiql-tests/blob/67d22dcd5a7eeaaa2624be9d1bb3da6a8b9faf0a/partiql-tests-data/eval/primitives/operators/overlaps.ion

PR #1851, opened 2025-10-31
Add support to transcribe AST to SQL for window function and window clause

xd1313113:
## Relevant Issues
- [Related To] Issue partiql/partiql-scribe#81

## Description
- Add support to transcribe AST to SQL for window function and window clause

## Other Information

## License Information

### DIFF
--- a/partiql-ast/api/partiql-ast.api
+++ b/partiql-ast/api/partiql-ast.api
@@ -219,7 +219,7 @@ public final class org/partiql/ast/Ast {
 	public static synthetic fun upsert$default (Lorg/partiql/ast/Identifier;Lorg/partiql/ast/Identifier$Simple;Lorg/partiql/ast/dml/InsertSource;ILjava/lang/Object;)Lorg/partiql/ast/dml/Upsert;
 	public static final fun windowClause (Ljava/util/List;)Lorg/partiql/ast/WindowClause;
 	public static final fun windowClauseDefinition (Lorg/partiql/ast/Identifier$Simple;Lorg/partiql/ast/WindowSpecification;)Lorg/partiql/ast/WindowClause$Definition;
-	public static final fun windowPartition (Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/WindowPartition;
+	public static final fun windowPartition (Lorg/partiql/ast/Identifier;)Lorg/partiql/ast/WindowPartition;
 	public static final fun windowSpecification (Lorg/partiql/ast/Identifier$Simple;Ljava/util/List;Lorg/partiql/ast/OrderBy;)Lorg/partiql/ast/WindowSpecification;
 	public static final fun with (Ljava/util/List;Z)Lorg/partiql/ast/With;
 	public static final fun withListElement (Lorg/partiql/ast/Identifier$Simple;Lorg/partiql/ast/expr/ExprQuerySet;Ljava/util/List;)Lorg/partiql/ast/WithListElement;
@@ -1753,18 +1753,18 @@ public class org/partiql/ast/WindowFunctionType$RowNumber$Builder {
 }
 
 public final class org/partiql/ast/WindowPartition : org/partiql/ast/AstNode {
-	public fun <init> (Lorg/partiql/ast/expr/Expr;)V
+	public fun <init> (Lorg/partiql/ast/Identifier;)V
 	public fun accept (Lorg/partiql/ast/AstVisitor;Ljava/lang/Object;)Ljava/lang/Object;
 	public static fun builder ()Lorg/partiql/ast/WindowPartition$Builder;
 	public fun equals (Ljava/lang/Object;)Z
 	public fun getChildren ()Ljava/util/List;
-	public fun getColumnReference ()Lorg/partiql/ast/expr/Expr;
+	public fun getColumnReference ()Lorg/partiql/ast/Identifier;
 	public fun hashCode ()I
 }
 
 public class org/partiql/ast/WindowPartition$Builder {
 	public fun build ()Lorg/partiql/ast/WindowPartition;
-	public fun columnReference (Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/WindowPartition$Builder;
+	public fun columnReference (Lorg/partiql/ast/Identifier;)Lorg/partiql/ast/WindowPartition$Builder;
 	public fun toString ()Ljava/lang/String;
 }
 
--- a/partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
+++ b/partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
@@ -253,7 +253,7 @@ public abstract class SqlDialect : AstVisitor<SqlBlock, SqlBlock>() {
 
     @Deprecated("This feature is experimental and is subject to change.")
     override fun visitWindowPartition(node: WindowPartition, tail: SqlBlock): SqlBlock {
-        return visitExprWrapped(node.columnReference, tail)
+        return visitIdentifier(node.columnReference, tail)
     }
 
     @Deprecated("This feature is experimental and is subject to change.")
--- a/partiql-parser/src/main/antlr/PartiQLParser.g4
+++ b/partiql-parser/src/main/antlr/PartiQLParser.g4
@@ -1077,7 +1077,7 @@ windowPartitionClause: PARTITION BY col+=windowPartitionColumnReference (COMMA c
  *   <basic identifier chain>
  *   | MODULE <period> <qualified identifier> <period> <column name>
  */
-windowPartitionColumnReference: expr;
+windowPartitionColumnReference: qualifiedName;
 
 cast
     : CAST PAREN_LEFT expr AS type PAREN_RIGHT;
--- a/partiql-ast/src/main/java/org/partiql/ast/WindowPartition.java
+++ b/partiql-ast/src/main/java/org/partiql/ast/WindowPartition.java
@@ -3,7 +3,6 @@ package org.partiql.ast;
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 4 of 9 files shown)


partiql-planner/src/main/resources/partiql_plan_internal.ion
```diff
@@ -245,6 +245,7 @@ rel::{
     },
 
     window::{
+      name: optional::string,
```
xd1313113, line 248: Self-review: window name need to be preserved for transcribe purpose.

partiql-parser/src/main/antlr/PartiQLParser.g4
```diff
@@ -1077,7 +1077,7 @@ windowPartitionClause: PARTITION BY col+=windowPartitionColumnReference (COMMA c
  *   <basic identifier chain>
  *   | MODULE <period> <qualified identifier> <period> <column name>
  */
-windowPartitionColumnReference: qualifiedName;
+windowPartitionColumnReference: expr;
```
xd1313113: The reason windowPartitionColumnReference was remodel from qualifiedName to expr is 

1. Sql Spec defines `<window partition column reference> ::= <column reference> [ <collate clause> ] `which has two types. qualifiedName works for partiql for now as collate is not supported. Expr is more extensible and requires no interface change if we adopt collate in the future.
2. Scribe show qualifiedName as `T.b` instead of `T[b]` for partiql dialect, which is very inconsistent with other column based expression like order by, group by

XuechunHHH: Curious how Scribe model order by / group by expression with nested columns?

partiql-planner/src/main/resources/partiql_plan_internal.ion
XuechunHHH: How will scribe represent it when there is no name? A string like `window_function_${index}`?

partiql-parser/src/main/antlr/PartiQLParser.g4
xd1313113: PLK determine the model. 
Order by model it as OrderBy->Sort->Expr, Order by support column and ordinal index
https://github.com/partiql/partiql-lang-kotlin/blob/4709ec7c4e48bdc528d7755e341a127ff43aab18/partiql-ast/src/main/java/org/partiql/ast/Sort.java#L19

GroupBy -> Key->Expr, group by support column and functions
https://github.com/partiql/partiql-lang-kotlin/blob/4709ec7c4e48bdc528d7755e341a127ff43aab18/partiql-ast/src/main/java/org/partiql/ast/GroupBy.java#L70C1-L71C1

partiql-planner/src/main/resources/partiql_plan_internal.ion
xd1313113: The name is for window clause, not window function, I have to generate window name like window_{index}.

the name is for SELECT a, RANK() OVER `w1`, DENSE_RANK() OVER `w2` FROM T WINDOW `w1` AS (ORDER BY a), `w2` AS (ORDER BY b DESC);

partiql-ast/api/partiql-ast.api
```diff
@@ -1748,18 +1753,18 @@ public class org/partiql/ast/WindowFunctionType$RowNumber$Builder {
...
 	public fun getChildren ()Ljava/util/List;
-	public fun getColumnReference ()Lorg/partiql/ast/Identifier;
+	public fun getColumnReference ()Lorg/partiql/ast/expr/Expr;
 	public fun hashCode ()I
 }
 
 public class org/partiql/ast/WindowPartition$Builder {
 	public fun build ()Lorg/partiql/ast/WindowPartition;
-	public fun columnReference (Lorg/partiql/ast/Identifier;)Lorg/partiql/ast/WindowPartition$Builder;
+	public fun columnReference (Lorg/partiql/ast/expr/Expr;)Lorg/partiql/ast/WindowPartition$Builder;
```
XuechunHHH: Change from `Identifier` to `Expr` is significant. Understood the extensibility concern. But not sure if you have considered the performance implications? Since `Expr` is much more general than Identifier. Will this impact parsing performance for the common case of simple column references?

partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
```diff
@@ -195,6 +200,122 @@ public abstract class SqlDialect : AstVisitor<SqlBlock, SqlBlock>() {
...
+    }
+
+    override fun visitWindowFunctionTypeLead(node: WindowFunctionType.Lead, tail: SqlBlock): SqlBlock {
+        var t = tail concat "LEAD("
+        t = visitExpr(node.extent, t)
+        node.offset?.let {
+            t = t concat ", $it"
+        }
+
+        node.defaultValue.let { defaultValue ->
```
XuechunHHH: Is `defaultValue` always present? Shall we use `node.defaultValue?.let` instead?

partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
```diff
@@ -195,6 +200,122 @@ public abstract class SqlDialect : AstVisitor<SqlBlock, SqlBlock>() {
...
+        t = t concat ")"
+        node.nullTreatment?.let { nullTreatment ->
+            t = t concat " ${nullTreatment.name()}"
+        }
+
+        return t
+    }
+
+    override fun visitWindowFunctionTypeLag(node: WindowFunctionType.Lag, tail: SqlBlock): SqlBlock {
+        var t = tail concat "LAG("
```
XuechunHHH: Looks nearly identical to `Lead`. Can we extract the common code?

partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
```diff
@@ -195,6 +200,122 @@ public abstract class SqlDialect : AstVisitor<SqlBlock, SqlBlock>() {
...
+    }
+
+    override fun visitWindowPartition(node: WindowPartition, tail: SqlBlock): SqlBlock {
+        return visitExprWrapped(node.columnReference, tail)
+    }
+
+    override fun visitWindowSpecification(node: WindowSpecification, tail: SqlBlock): SqlBlock {
+        var t = tail
+        if (node.existingName != null) {
+            t = visitIdentifierSimple(node.existingName!!, t)
```
XuechunHHH, line 263: nit: The null check `if (node.existingName != null)` followed by `node.existingName!!` is redundant to me. Consider also using `node.existingName?.let` here.

partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
```diff
@@ -195,6 +200,122 @@ public abstract class SqlDialect : AstVisitor<SqlBlock, SqlBlock>() {
...
+            if (!node.partitionClause.isNullOrEmpty() || node.orderClause != null) {
+                t = t concat " "
+            }
+        } else {
+            t = t concat "("
+            // PARTITION BY clause
+            if (!node.partitionClause.isNullOrEmpty()) {
```
XuechunHHH, line 270: nit: same here

partiql-parser/src/main/antlr/PartiQLParser.g4
XuechunHHH: Understand the reasoning from your comment. But this change also significantly broadens what's acceptable in a partition clause. 

Have you added validation to ensure only appropriate expressions are used? For example, what happens if a complex expression that doesn't make sense for partitioning?

partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
xd1313113: Not null assertion is required by compiler and let does not apply as it has else condition.

partiql-parser/src/main/antlr/PartiQLParser.g4
xd1313113: I give it a few tests. Surprisingly, there is no validation around `ORDER BY`, `GROUP BY`.  The sql does not fail either and We probably should create an issue to investigating the behavior for `PARTITION BY`, `ORDER BY`, `GROUP BY`.

```
partiql ▶ SELECT max(b) FROM [{'a':1, 'b':2},{'a':2, 'b':3}] group by a ORDER BY 1 > 2;

=== RESULT ===
[
  {
    '_1': 2
  },
  {
    '_1': 3
  }
]

OK!
partiql ▶ SELECT max(b) FROM [{'a':1, 'b':2},{'a':2, 'b':3}] group by a ORDER BY 1 < 2;

=== RESULT ===
[
  {
    '_1': 2
  },
  {
    '_1': 3
  }
]

OK!
```

partiql-ast/api/partiql-ast.api
xd1313113: Regarding performance,`qualifiedName` is more faster and `expr` matching is O(n) matching.  However Path expr has relative high precedence in matching per table here and less alternative. Performance should be identical to `order by` and `group by`.  https://github.com/partiql/partiql-lang-kotlin/blob/52716dfd5f74c09de18c2504ca933b5009eb6e4d/partiql-parser/src/main/antlr/PartiQLParser.g4#L786C38-L786C54

partiql-ast/src/main/java/org/partiql/ast/sql/SqlDialect.kt
xd1313113: it has additional empty check. does not apply

partiql-parser/src/main/antlr/PartiQLParser.g4
XuechunHHH: Yeah, thanks for the investigation. And we definitely need to cut an issue to track this.

XuechunHHH (approved):
LGTM

Issue #1852, opened 2025-11-05
Consider support date and timestamp comparison and minus operator


xd1313113:
## Description
- consider supporting date and timestamp comparison and minus operator. 
- We need to cast internally from date to timestamp or timestamp to date to do comparison. However, different cast may lead different result. We need to have a discussion how to implement it.

```
partiql ▶ cast (date '2000-01-01' as timestamp) >= timestamp '2000-01-01 10:00:00.000'
   | 

=== RESULT ===
false

OK!
partiql ▶ date '2000-01-01' >= cast (timestamp '2000-01-01 10:00:00.000' as date)
   | 

=== RESULT ===
true

OK!
```

## To Reproduce
```
partiql ▶ date '2000-01-02' > timestamp '2000-01-01 01:00:00.000'
   | 
w: [semantic] Undefined function "gt"(DATE, TIMESTAMP(6)).

=== RESULT ===
missing

OK!
partiql ▶ date '2000-01-02' - timestamp '2000-01-01 01:00:00.000'
   | 
w: [semantic] Undefined function "minus"(DATE, TIMESTAMP(6)).

=== RESULT ===
missing
```
## Expected Behavior
- < A clear and concise description of what you expected to happen. >

## Additional Context
- Java version: XXX
- PartiQL version: XXX
- Add any other context about the problem here.

PR #1853, opened 2025-11-06
Add planner support to compare DATE and TIMESTAMP(Z)

xd1313113:
## Relevant Issues
- [Related To] Issue #1852 

## Description
- Add planner support to compare DATE and TIMESTAMP(Z)

## Other Information
- Updated Unreleased Section in CHANGELOG: **Yes**
- Any backward-incompatible changes? **No**
- Any new external dependencies? **No**
- Do your changes comply with the [contributing][cg] and [code style][csg] guidelines? **Yes**

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Thank you to all who have contributed!
 ### Added
 - Support aggregation function calls in the `ORDER BY` and `HAVING` clause
 - Add Support to transcribe `Overlaps` function
+- Add planner support to compare DATE and TIMESTAMP(Z)
 
 ### Changed
 


partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/operator/OpArithmeticTest.kt
```diff
@@ -151,6 +153,8 @@ class OpArithmeticTest : PartiQLTyperTestBase() {
                     // TODO arg1 == StaticType.DECIMAL && arg0 == StaticType.FLOAT -> arg0 // TODO: The cast table is wrong
                     castTablePType(arg1, arg0) == CastType.COERCION -> arg0
                     castTablePType(arg0, arg1) == CastType.COERCION -> arg1
+                    arg0 in allDatePType && arg1 in allTimeStampPType -> PType.intervalDaySecond(9, 0)
```
XuechunHHH, line 156: Can we add this part to `castTablePType` itself instead since we want it to be defined as universal implicit cast rules right?

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMinus.kt
```diff
@@ -296,4 +296,16 @@ internal object FnMinus : DiadicArithmeticOperator("minus") {
...
+        return basic(PType.intervalDaySecond(INTERVAL_MAX_PRECISION, NANO_MAX_PRECISION), dateLhs, timestampRhs) { args ->
+            TODO("https://github.com/partiql/partiql-lang-kotlin/issues/1852")
+        }
+    }
+
+    override fun getTimestampDateInstance(timestampLhs: PType, dateRhs: PType): Fn {
+        return basic(PType.intervalDaySecond(INTERVAL_MAX_PRECISION, NANO_MAX_PRECISION), timestampLhs, dateRhs) { args ->
+            TODO("https://github.com/partiql/partiql-lang-kotlin/issues/1852")
+        }
+    }
```
yliuuuu, line 310: Is it intentional to support this?

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/DiadicOperator.kt
```diff
@@ -152,6 +152,24 @@ internal abstract class DiadicOperator(
...
     }
 
+    /**
+     * @param dateLhs a type of date
+     * @param timestampRhs a type of timestamp
+     * @return an instance of a function
+     */
+    open fun getDateTimestampInstance(dateLhs: PType, timestampRhs: PType): Fn? {
+        return null
+    }
```
yliuuuu, line 162: Why not rely on SQLTypeFamily? https://github.com/partiql/partiql-lang-kotlin/blob/main/partiql-spi/src/main/kotlin/org/partiql/spi/internal/SqlTypeFamily.kt#L104-L113. 

I am not sure if the preferred on should be date, but at the moment I think it would be sufficient for the purpose of allowing comparison between datetime type.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMinus.kt
XuechunHHH: Yes. According to SQL-spec 99:
> The comparison of two datetimes is determined according to the interval resulting from their subtraction.

See - chrome-extension://efaidnbmnnnibpcajpcglclefindmkaj/https://web.cecs.pdx.edu/~len/sql1999.pdf#page=292

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/DiadicOperator.kt
xd1313113: We do not want to support all type in the family. E.g. We should not support date compare to time as they are not comparable due to lack of same datetime fields per SQL spec.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMinus.kt
xd1313113: yes, comparison are based on minus operator

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/DiadicOperator.kt
yliuuuu: I see. Double checked that SQL Spec does not allow coercion from Data to time and vice versa. So make sense to me.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMinus.kt
yliuuuu: I am fine with putting the placeholder now. But I don't think the planner at the moment routes to FnMinus when doing comparison right?

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/operator/OpArithmeticTest.kt
xd1313113: We do not do cast in planning for now. We match directly to getDateTimestampInstance. We may consider adding cast(date to timestamp) to planner to match getTimestampInstance.

XuechunHHH: I see. Fair enough.

PR #1864, opened 2026-01-13
add: Datetime Field Keywords Error Handling

XuechunHHH:
## Problem Statement
Our customer uses datetime field keywords as function arguments:
```
SELECT DATE_ADD(YEAR, 5, timestamp_col) FROM sensor_data
```
These keywords became reserved tokens in PLK's ANTLR grammar, causing parse failures. The solution needed to:

- Allow parsing to succeed with structured error information
- Enable target-specific recovery in downstream

## Action Item(s)
- Grammar Layer (ANTLR) 
   - Added grammar rule to `exprTerm` for datetime field keywords
- AST Layer
   - Created `ExprError` AST node class
   - Added `DATETIME_FIELD_KEYWORD` constant
   - Implemented visitor pattern support
   - Added `visitExprError()` method to visitor interface
- Parser Layer
   - Added `datetimeFieldAsExpression()` error factory method
   - Configured to use `UNEXPECTED_TOKEN` with `ERROR` severity  
   - Added `visitExprTermDatetimeField()` method
   - Converted grammar matches to `ExprError` AST nodes
   - Reported errors to listener
- Planner Layer
   - Convert AST error nodes to plan IR

## Other Information

## License Information

### DIFF
--- a/partiql-ast/src/main/java/org/partiql/ast/expr/ExprError.java
+++ b/partiql-ast/src/main/java/org/partiql/ast/expr/ExprError.java
@@ -2,7 +2,6 @@ package org.partiql.ast.expr;
 
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
-import lombok.Getter;
 import org.jetbrains.annotations.NotNull;
 import org.partiql.ast.AstNode;
 import org.partiql.ast.AstVisitor;
@@ -17,40 +16,25 @@ import java.util.List;
  * such as datetime field keywords (YEAR, MONTH, DAY, HOUR, MINUTE, SECOND) used
  * in expression contexts where they are reserved tokens.
  * <p>
- * The ExprError node preserves the original text for target-specific recovery
- * and includes an error code for categorization.
+ * The ExprError node preserves the original text for target-specific recovery.
  */
 @Builder(builderClassName = "Builder")
 @EqualsAndHashCode(callSuper = false)
 public final class ExprError extends Expr {
 
     /**
-     * Error code for datetime field keywords used in expression contexts.
-     */
-    public static final int DATETIME_FIELD_KEYWORD = 1;
-
-    /**
      * The original text that caused the error.
      */
     @NotNull
-    private final String message;
-
-    /**
-     * The error code categorizing the type of error.
-     * Returns the error code categorizing the type of error.
-     */
-    @Getter
-    private final int code;
+    private final String text;
 
     /**
      * Creates a new ExprError node.
      *
-     * @param message the original text that caused the error
-     * @param code the error code categorizing the type of error
+     * @param text the original text that caused the error
      */
-    public ExprError(@NotNull String message, int code) {
-        this.message = message;
-        this.code = code;
+    public ExprError(@NotNull String text) {
+        this.text = text;
     }
 
     @Override
@@ -67,10 +51,10 @@ public final class ExprError extends Expr {
     /**
      * Returns the original text that caused the error.
      *
-     * @return the error message/original text
+     * @return the original text
      */
     @NotNull
-    public String getMessage() {
-        return this.message;
+    public String getText() {
+        return this.text;
     }
 }
--- a/partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
+++ b/partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
@@ -192,7 +192,6 @@ import org.partiql.ast.dml.UpdateTargetStep
 import org.partiql.ast.expr.Expr
 import org.partiql.ast.expr.ExprArray
 import org.partiql.ast.expr.ExprBag
-import org.partiql.ast.expr.ExprError
 import org.partiql.ast.expr.ExprLit
 import org.partiql.ast.expr.ExprPath
 import org.partiql.ast.expr.ExprQuerySet
@@ -1810,7 +1809,7 @@ internal class PartiQLParserDefault : PartiQLParser {
 
         override fun visitExprTermDatetimeField(ctx: GeneratedParser.ExprTermDatetimeFieldContext) = translate(ctx) {
             val datetimeFieldCtx = ctx.primaryDatetimeField()
-            val fieldText = datetimeFieldCtx.text.uppercase()
+            val fieldText = datetimeFieldCtx.text
             val location = SourceLocation(
                 ctx.start.line.toLong(),
                 ctx.start.charPositionInLine + 1L,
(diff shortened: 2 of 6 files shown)


partiql-ast/src/main/java/org/partiql/ast/expr/ExprError.java
```diff
@@ -0,0 +1,76 @@
...
+        return visitor.visitExprError(this, ctx);
+    }
+
+    /**
+     * Returns the original text that caused the error.
+     *
+     * @return the error message/original text
+     */
+    @NotNull
+    public String getMessage() {
```
johnedquinn: Nit - `getText()` would be more apt. Message denotes that we've curated a message for the consumer.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PErrors.kt
```diff
@@ -72,6 +72,33 @@ internal object PErrors {
...
+        location: SourceLocation?,
+        fieldName: String
+    ): PError {
+        return PError(
+            PError.UNEXPECTED_TOKEN,
+            Severity.ERROR(),
+            PErrorKind.SYNTAX(),
+            location,
+            mapOf(
+                "TOKEN_NAME" to fieldName,
```
johnedquinn, line 96: We normally programmatically provide the token name here: https://github.com/partiql/partiql-lang-kotlin/blob/93cce132c4bb9aec7113df4e6c627ccbe0052374/partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt#L348C30-L348C93

Though, maybe it doesn't fit the use-case. I'm not sure.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
```diff
@@ -153,6 +155,11 @@ internal object RexConverter {
         override fun defaultReturn(node: AstNode, context: Env): Rex =
             throw IllegalArgumentException("unsupported rex $node")
 
+        override fun visitExprError(node: ExprError, ctx: Env): Rex {
+            val type = CompilerType(PType.dynamic())
+            return rex(type, rexOpErr())
```
johnedquinn, line 164: You should emit an error to the listener with an INTERNAL_ERROR code for now.

partiql-ast/src/main/java/org/partiql/ast/expr/ExprError.java
```diff
@@ -0,0 +1,76 @@
...
+ * and includes an error code for categorization.
+ */
+@Builder(builderClassName = "Builder")
+@EqualsAndHashCode(callSuper = false)
+public final class ExprError extends Expr {
+
+    /**
+     * Error code for datetime field keywords used in expression contexts.
+     */
+    public static final int DATETIME_FIELD_KEYWORD = 1;
```
johnedquinn: I don't know if we need this just yet. Just text might be sufficient. Especially since ExprError currently as it stands doesn't support providing the code.

partiql-parser/src/test/kotlin/org/partiql/parser/internal/PartiQLParserDatetimeFieldTests.kt
```diff
@@ -0,0 +1,132 @@
...
+ * when used as expressions, with appropriate error reporting to the listener.
+ */
+class PartiQLParserDatetimeFieldTests {
+
+    private val parser = PartiQLParserDefault()
+
+    /**
+     * Test that YEAR keyword in expression context produces ExprError node.
+     */
+    @Test
```
xd1313113, line 24: Do you need test cases with built in date_add/date_diff function and custom overload function as well?

partiql-ast/src/main/java/org/partiql/ast/expr/ExprError.java
xd1313113: Code looks good to me. text comparison might be buggy e.g. empty, to determine error types.

partiql-parser/src/test/kotlin/org/partiql/parser/internal/PartiQLParserDatetimeFieldTests.kt
XuechunHHH: Those have been covered in the conformance tests already?

xd1313113: date_add/date_diff does. custom overload function may not be covered.

partiql-ast/src/main/java/org/partiql/ast/expr/ExprError.java
XuechunHHH: Good catch. I'll rename it.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PErrors.kt
XuechunHHH: Emmm... I see what you mean here. In this case, I'm providing the uppercase token text `YEAR`, `MONTH`... serving the same purpose. Since these are reserved keywords, I believe the text and symbolic name are effectively the same. 

But I'm open to using the vocabulary lookup if you think it's more consistent with the codebase patterns.

partiql-parser/src/test/kotlin/org/partiql/parser/internal/PartiQLParserDatetimeFieldTests.kt
XuechunHHH: Thanks for the context. Will do.

partiql-ast/src/main/java/org/partiql/ast/expr/ExprError.java
XuechunHHH: Thanks for the call out. Not sure if using code here will be over-engineered? If it is fine, I prefer to keep it this way for better extensibility and easy catch for downstream. 

Let me know if you have strong opinion on this.

XuechunHHH: Emm.. rethink on this and revert it to just text. If we want to extend this later, let's do then.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -1806,6 +1807,19 @@ internal class PartiQLParserDefault : PartiQLParser {
                 exprSessionAttribute(SessionAttribute.CURRENT_DATE())
             }
 
+        override fun visitExprTermDatetimeField(ctx: GeneratedParser.ExprTermDatetimeFieldContext) = translate(ctx) {
+            val datetimeFieldCtx = ctx.primaryDatetimeField()
+            val fieldText = datetimeFieldCtx.text.uppercase()
```
johnedquinn: We probably don't need to uppercase -- as we are returning the offending text that couldn't be parsed.

johnedquinn (dismissed):

PR #1899, opened 2026-04-06
Fix the issue that when `With` statement is present, the relproject is omitted in OperatorRewriter

xd1313113:
## Relevant Issues
- None

## Description
When With statement is present, the relproject is omitted in OperatorRewriter
RelWith
->RelProject

- Add visitWith implementation

## Other Information

## License Information

### DIFF
--- a/partiql-planner/src/test/kotlin/org/partiql/planner/OperatorRewriterTest.kt
+++ b/partiql-planner/src/test/kotlin/org/partiql/planner/OperatorRewriterTest.kt
@@ -1,0 +1,192 @@
+package org.partiql.planner
+
+import org.junit.jupiter.api.parallel.Execution
+import org.junit.jupiter.api.parallel.ExecutionMode
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.MethodSource
+import org.partiql.parser.PartiQLParser
+import org.partiql.plan.Action
+import org.partiql.plan.Operator
+import org.partiql.plan.OperatorRewriter
+import org.partiql.plan.Plan
+import org.partiql.plan.rel.RelProject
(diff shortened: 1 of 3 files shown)


partiql-planner/src/test/kotlin/org/partiql/planner/OperatorRewriterTest.kt
```diff
@@ -0,0 +1,194 @@
...
+ * 3. Plan the [SuccessTestCase.targetQuery]
+ * 4. Assert equivalence via [PlanEquivalenceOperatorVisitor]
+ */
+internal class OperatorRewriterTest {
+
+    data class SuccessTestCase(
+        val name: String,
+        val originalQuery: String,
+        val targetQuery: String,
+        val rewriter: OperatorRewriter<Void?>,
```
johnedquinn: This has a backwards wiring.

SuccessTestCase takes in any impl of an OperatorRewriter, however it uses RunTest (which assumes that ColumnRenamingRewriter is passed here) and validates using that logic. But, really, the logic of the rewriter/validation should be tied together.

Something like:
```
interface OperatorRewriterSuccessCase {
    fun OperatorRewriter getRewriter()

    fun validate(input: String, expected: String)
}
```

Above is just an example, but it's representative of how to tie the rewriter impl to the validation logic.

johnedquinn (changes requested):

partiql-planner/src/test/kotlin/org/partiql/planner/OperatorRewriterTest.kt
xd1313113: Yes, need tie rewriter to validation in test

johnedquinn (approved):

PR #1903, opened 2026-04-13
Feature/cli dir multi format reader

AugustineFu:
## Description
- Add multi-format file (Parquet/CSV/TSV/Ion/JSON) support for the --dir CLI option using lazy iterators instead of loading entire files into memory

## Other Information
- Updated Unreleased Section in CHANGELOG: **[YES/NO]**: NO
- Any new external dependencies? **[YES/NO]**: YES
   -  parquet-hadoop 1.14.1, hadoop-common 3.3.6 (with exclusions), hadoop-mapreduce-client-core 3.3.6 — required by Parquet Java for reading .parquet files.
- Do your changes comply with the [contributing][cg] and [code style][csg] guidelines? **[YES/NO]**: YES

## License Information

### DIFF
--- a/buildSrc/src/main/kotlin/partiql.versions.kt
+++ b/buildSrc/src/main/kotlin/partiql.versions.kt
@@ -44,6 +44,8 @@ object Versions {
     const val kotlinxCoroutinesJdk8 = "1.8.1"
     const val ktlint = "0.42.1" // we're on an old version of ktlint. TODO upgrade https://github.com/partiql/partiql-lang-kotlin/issues/1418
     const val lombok = "1.18.34"
+    const val parquet = "1.14.1"
+    const val hadoop = "3.3.6"
 
     // Testing
     const val assertj = "3.11.0"
@@ -85,6 +87,9 @@ object Deps {
     const val kotlinxCoroutinesJdk8 = "org.jetbrains.kotlinx:kotlinx-coroutines-jdk8:${Versions.kotlinxCoroutinesJdk8}"
     const val ktlint = "com.pinterest.ktlint:ktlint-core:${Versions.ktlint}"
     const val lombok = "org.projectlombok:lombok:${Versions.lombok}"
+    const val parquetHadoop = "org.apache.parquet:parquet-hadoop:${Versions.parquet}"
+    const val hadoopCommon = "org.apache.hadoop:hadoop-common:${Versions.hadoop}"
+    const val hadoopMapreduceClient = "org.apache.hadoop:hadoop-mapreduce-client-core:${Versions.hadoop}"
 
     // Testing
     const val assertj = "org.assertj:assertj-core:${Versions.assertj}"
--- a/partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
+++ b/partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
@@ -477,9 +477,6 @@ internal class MainCommand : Runnable {
             error("Not a directory: ${directory.path}")
         }
         val supportedExtensions = setOf("ion", "pql", "csv", "tsv", "json", "parquet")
-        directory.listFiles()
-            ?.filter { it.isFile && it.extension !in supportedExtensions }
-            ?.forEach { System.err.println("Warning: skipping '${it.name}' (unsupported extension '.${it.extension}'). Supported: $supportedExtensions") }
         return LazyCatalog("default", directory, supportedExtensions) { file ->
             debug("Loading table '${file.nameWithoutExtension}' from ${file.name}")
             when (file.extension) {
--- a/partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
+++ b/partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
@@ -16,6 +16,7 @@
 package org.partiql.cli
 
 import org.partiql.cli.io.DatumCsvReader
+import org.partiql.cli.io.DatumParquetReader
 import org.partiql.cli.io.DatumIonReaderBuilder
 import org.partiql.cli.io.DatumWriterTextPretty
 import org.partiql.cli.io.Format
@@ -479,7 +480,7 @@ internal class MainCommand : Runnable {
         val supportedExtensions = setOf("ion", "pql", "csv", "tsv", "json", "parquet")
         return LazyCatalog("default", directory, supportedExtensions) { file ->
             debug("Loading table '${file.nameWithoutExtension}' from ${file.name}")
-            when (file.extension) {
+            when (file.extension.lowercase()) {
                 "ion", "json" -> {
                     val reader = DatumIonReaderBuilder.standard().build(file.inputStream())
                     val first = try { reader.read() } catch (_: IOException) { return@LazyCatalog Datum.nullValue() }
@@ -508,6 +509,7 @@ internal class MainCommand : Runnable {
                 }
                 "csv" -> DatumCsvReader.read(file.inputStream(), ',')
                 "tsv" -> DatumCsvReader.read(file.inputStream(), '\t')
+                "parquet" -> DatumParquetReader.read(file)
                 "pql" -> {
                     val text = file.readText(Charsets.UTF_8)
                     pipeline().execute(text, Session.empty())
--- a/partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
+++ b/partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
@@ -15,9 +15,11 @@
 
 package org.partiql.cli
 
+import org.partiql.cli.io.DatumCsvReader
 import org.partiql.cli.io.DatumIonReaderBuilder
 import org.partiql.cli.io.DatumWriterTextPretty
 import org.partiql.cli.io.Format
+import org.partiql.cli.io.LazyCatalog
 import org.partiql.cli.pipeline.ErrorMessageFormatter
 import org.partiql.cli.pipeline.Pipeline
 import org.partiql.cli.shell.Shell
@@ -83,8 +85,7 @@ internal class MainCommand : Runnable {
 
     @CommandLine.Option(
         names = ["-d", "--dir"],
-        description = ["Path to the database directory"],
-        hidden = true
+        description = ["Path to the database directory. Each file becomes a table (e.g. users.ion -> table 'users')."],
     )
     var dir: File? = null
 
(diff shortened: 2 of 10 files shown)


partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
```diff
@@ -474,6 +472,54 @@ internal class MainCommand : Runnable {
         }
     }
 
+    private fun loadDir(directory: File): Catalog {
+        if (!directory.isDirectory) {
+            error("Not a directory: ${directory.path}")
+        }
+        val supportedExtensions = setOf("ion", "pql", "csv", "tsv", "json", "parquet")
+        directory.listFiles()
```
johnedquinn: Nice.

partiql-cli/src/main/kotlin/org/partiql/cli/io/LazyCatalog.kt
```diff
@@ -0,0 +1,53 @@
...
+    directory: File,
+    supportedExtensions: Set<String>,
+    loader: (File) -> Datum,
+) : Catalog {
+
+    private val tables: Map<Name, Table>
+
+    init {
+        val files = directory.listFiles()?.filter { it.isFile && it.extension in supportedExtensions } ?: emptyList()
+        tables = files.associate { file ->
```
johnedquinn: This isn't totally lazy. Let's say I start up the PartiQL CLI, run a query against `customers.ion` -- and I realize I should've called it `users.ion`. If in another shell I rename it, this wouldn't pick it up.

Can we make it truly lazy? AKA, `getTable/resolveTable` truly are lazy?

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
```diff
@@ -474,6 +472,54 @@ internal class MainCommand : Runnable {
...
+            ?.filter { it.isFile && it.extension !in supportedExtensions }
+            ?.forEach { System.err.println("Warning: skipping '${it.name}' (unsupported extension '.${it.extension}'). Supported: $supportedExtensions") }
+        return LazyCatalog("default", directory, supportedExtensions) { file ->
+            debug("Loading table '${file.nameWithoutExtension}' from ${file.name}")
+            when (file.extension) {
```
johnedquinn: Can we move this logic to a different location, outside of `Main.kt`? Either directly within `LazyCatalog` or honestly `LazyTable`. I'd like to keep `Main.kt` mostly about parsing and passing onto the handlers.

johnedquinn (changes requested):
Haven't gone through the Parquet reader, but left some initial comments. Overall really glad this work is getting merged in.

buildSrc/src/main/kotlin/partiql.versions.kt
```diff
@@ -85,6 +87,9 @@ object Deps {
     const val kotlinxCoroutinesJdk8 = "org.jetbrains.kotlinx:kotlinx-coroutines-jdk8:${Versions.kotlinxCoroutinesJdk8}"
     const val ktlint = "com.pinterest.ktlint:ktlint-core:${Versions.ktlint}"
     const val lombok = "org.projectlombok:lombok:${Versions.lombok}"
+    const val parquetHadoop = "org.apache.parquet:parquet-hadoop:${Versions.parquet}"
+    const val hadoopCommon = "org.apache.hadoop:hadoop-common:${Versions.hadoop}"
+    const val hadoopMapreduceClient = "org.apache.hadoop:hadoop-mapreduce-client-core:${Versions.hadoop}"
```
xd1313113, line 92: Why does Mapreduce client is needed?

partiql-cli/src/main/kotlin/org/partiql/cli/io/LazyCatalog.kt
AugustineFu: I've modified it, but have a quick question for this: it seems this approach makes getTable run in O(n) time, is the tradeoff here that in normal cli use cases there usually aren't too many queries, so real-time responsiveness matters more?

partiql-cli/src/main/kotlin/org/partiql/cli/io/LazyCatalog.kt
```diff
@@ -0,0 +1,60 @@
...
+        return null
+    }
+
+    private fun findFile(tableName: String): File? {
+        return directory.listFiles()
+            ?.filter { it.isFile && it.extension in supportedExtensions }
+            ?.firstOrNull { it.nameWithoutExtension.equals(tableName, ignoreCase = true) }
+    }
+
+    private fun createTable(name: Name, file: File): Table = when (file.extension) {
```
xd1313113: file.extension to lowercase for comparison

partiql-cli/src/main/kotlin/org/partiql/cli/io/ParquetTable.kt
```diff
@@ -0,0 +1,107 @@
...
+import java.math.BigInteger
+
+/**
+ * A [Table] backed by a Parquet file. Reads only the footer at construction time
+ * to extract schema. Data is streamed Row Group by Row Group on iteration.
+ */
+internal class ParquetTable(
+    private val name: Name,
+    private val file: File,
+) : Table {
```
xd1313113: Should ParquetTable also be a LazyTable instead of Table

buildSrc/src/main/kotlin/partiql.versions.kt
AugustineFu: hadoop-mapreduce-client-core is a transitive dependency of parquet-hadoop 1.14.1, ParquetFileReader.open() internally loads FileInputFormat at class-load time. Removing it will cause NoClassDefFoundError at runtime.

partiql-cli/src/main/kotlin/org/partiql/cli/io/DatumParquetReader.kt
```diff
@@ -0,0 +1,101 @@
+package org.partiql.cli.io
+
+import org.apache.parquet.example.data.simple.SimpleGroup
+import org.apache.parquet.example.data.simple.convert.GroupRecordConverter
```
yliuuuu: I have not yet look at the Parquet code, but are we ok with taking dependency from example package?

partiql-cli/src/main/kotlin/org/partiql/cli/io/LazyCatalog.kt
johnedquinn: Yeah, I'd say real-time responsiveness matters more. I'd rather my tool be reliable than optimized to the microsecond. For a directory with a thousand files, the search will be milliseconds.

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
```diff
@@ -474,6 +473,52 @@ internal class MainCommand : Runnable {
...
 
+    private fun loadDir(directory: File): Catalog {
+        if (!directory.isDirectory) {
+            error("Not a directory: ${directory.path}")
+        }
+        val supportedExtensions = setOf("ion", "pql", "csv", "tsv", "json", "parquet")
+        return LazyCatalog("default", directory, supportedExtensions) { file ->
+            debug("Loading table '${file.nameWithoutExtension}' from ${file.name}")
+            when (file.extension.lowercase()) {
+                "ion", "json" -> {
```
XuechunHHH, line 484: nit: for ion/csv/tsv branches, the InputStream is never closed. May hit trouble if someone ran many queries against many files in an interactive shell session without restarting.

partiql-cli/src/main/kotlin/org/partiql/cli/io/ParquetTable.kt
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * A [LazyTable] backed by a Parquet file. Reads only the footer at construction time
+ * to extract schema. Data reading is delegated to [DatumParquetReader].
+ */
+internal class ParquetTable(
+    name: Name,
+    file: File,
+) : LazyTable(name, file, { DatumParquetReader.read(it) }) {
+
+    private val parquetSchema: MessageType =
```
XuechunHHH: Can we also make it `by lazy` for reading file?

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
```diff
@@ -474,6 +473,52 @@ internal class MainCommand : Runnable {
...
+                    val reader = DatumIonReaderBuilder.standard().build(file.inputStream())
```
XuechunHHH, line 485: For my own curiosity, why using a state machine to iterate instead of just wrapping in a bag?

buildSrc/src/main/kotlin/partiql.versions.kt
XuechunHHH: Hadoop is a heavy dependency for a CLI tool. But guess we don't have workaround :melting_face:

XuechunHHH (commented):
Nice work! Drop some questions for my own understanding.

partiql-cli/src/main/kotlin/org/partiql/cli/io/LazyCatalog.kt
```diff
@@ -0,0 +1,63 @@
...
+    override fun getName(): String = name
+
+    override fun getTable(session: Session, name: Name): Table? {
+        val file = findFile(name.getName()) ?: return null
+        return createTable(name, file)
+    }
+
+    override fun resolveTable(session: Session, identifier: Identifier): Name? {
+        val target = identifier.first()
+        val file = directory.listFiles()
```
johnedquinn: nit -- I believe listFiles accepts a FileFilter. No need to create two lists with listFiles().filter().

johnedquinn (commented):
Overall I like this addition a lot. Few comments remain from others, but I'm good to ship after.

partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
AugustineFu: The original method used readAll() which loads everything into a List eagerly, for a JSON array file, readAll() returns a single Datum.array, the engine can iterate it directly, no wrapping needed. Now I stream lazily (read on demand) here and use a Datum.bag(Iterable) to give the engine something to iterate. But that wrapping causes a JSON array to become one element inside the bag. The first/second pre-read detects single-value files and returns them unwrapped, so I can preserve the original behavior and keep the rest lazy.

johnedquinn (dismissed):
Thanks for all this work! Real excited to get this added functionality out.

xd1313113:
Minor, can you update changelog.md with your cli changes and your github account name

johnedquinn (approved):

PR #1907, opened 2026-04-27
route COALESCE and NULLIF to function call

yliuuuu:
## Relevant Issues
- [Closes/Related To] N/A

## Description
- Route COALESCE and NULLIF to function call. 
- Added a PTestDef to assert on whether a given query can be parsed under SLL mode. 

### Some superficial notes on SLL and LL. 

SLL:  the fast path. The parser first tries SLL prediction, which makes decisions based only on the lookahead tokens without considering the full parser 
call stack (the parsing context). This is fast because it doesn't track which rule invoked the current rule. For unambiguous grammars, SLL always gets the right 
answer.

LL: the fallback. When SLL encounters an ambiguity — two or more grammar alternatives that could match the same input — it can't decide which path to take. At that 
point, the parser falls back to full LL prediction, which considers the complete call stack context. LL can resolve ambiguities that SLL cannot, but it's significantly
more expensive.

The exponential problem with LL:  In LL mode, when the parser hits an ambiguous construct, it must walk through all candidate grammar alternatives to determine if the 
ambiguity is real. For each ambiguous point, it re-examines the sub-expressions beneath it. If those sub-expressions themselves contain ambiguous constructs (e.g., 
nested function calls that trigger the same ambiguity), the parser walks them again for each outer alternative.  

### Literature reference
Post the [LL algorithm  paper](https://www.antlr.org/papers/allstar-techreport.pdf) by ANTLR, so that I, (and others) can go back to deep dive into this when available. 

## Other Information
- Updated Unreleased Section in CHANGELOG: **[YES/NO]**:  Yes. 
- Any backward-incompatible changes? **[YES/NO]**: No. 
- Any new external dependencies? **[YES/NO]**: No.
- Do your changes comply with the [contributing][cg] and [code style][csg] guidelines? **[YES/NO]** Yes. 

## License Information

### DIFF
--- a/partiql-parser/src/main/antlr/PartiQLParser.g4
+++ b/partiql-parser/src/main/antlr/PartiQLParser.g4
@@ -951,12 +951,6 @@ exprTerm
     | tuple                          # ExprTermBase
     ;
 
-nullIf
-    : NULLIF PAREN_LEFT expr COMMA expr PAREN_RIGHT;
-
-coalesce
-    : COALESCE PAREN_LEFT expr ( COMMA expr )* PAREN_RIGHT;
-
 caseExpr
     : CASE case=expr? (WHEN whens+=expr THEN thens+=expr)+ (ELSE else=expr)? END;
 
--- a/partiql-parser/src/main/antlr/PartiQLParser.g4
+++ b/partiql-parser/src/main/antlr/PartiQLParser.g4
@@ -880,24 +880,22 @@ valueExpr
 exprPrimary
     : exprTerm                   # ExprPrimaryBase
     | cast                       # ExprPrimaryBase
-    | sequenceConstructor        # ExprPrimaryBase
-    | substring                  # ExprPrimaryBase
-    | position                   # ExprPrimaryBase
-    | overlay                    # ExprPrimaryBase
+    | sequenceConstructor        # ExprPrimaryBase // TODO: Move to function call
+    | substring                  # ExprPrimaryBase // TODO: Move to function call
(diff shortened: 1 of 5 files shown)


partiql-parser/src/main/antlr/PartiQLParser.g4
```diff
@@ -880,24 +880,22 @@ valueExpr
...
-    | position                   # ExprPrimaryBase
-    | overlay                    # ExprPrimaryBase
+    | sequenceConstructor        # ExprPrimaryBase // TODO: Move to function call
+    | substring                  # ExprPrimaryBase // TODO: Move to function call
+    | position                   # ExprPrimaryBase // TODO: Move the standard form to function call
+    | overlay                    # ExprPrimaryBase // TODO: Move the standard form to function call
     | canCast                    # ExprPrimaryBase  // TODO remove ahead of `v1` release
     | canLosslessCast            # ExprPrimaryBase  // TODO remove ahead of `v1` release
     | extract                    # ExprPrimaryBase
-    | coalesce                   # ExprPrimaryBase
```
johnedquinn, line 890: Can you remove the coalesce and nullif rules if they're not used?

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
johnedquinn: Needs an updated doc

johnedquinn (changes requested):

xd1313113 (approved):

johnedquinn (approved):

PR #1908, opened 2026-04-30
Add interruptible support

AugustineFu:
## Relevant Issues
- [Related To] Issue [[PartiQL Upgrade] Implement scenario tests for dual-engine
](https://issues.amazon.com/issues/WASP-1664?selectedConversation=111a30b8-2ce8-46f7-8b4d-178349da0180)
## Description
   - Adds thread interruption support to the V1 engine. V0 had this via CompileOptions.isInterruptible(true), but it was not carried over to the V1 rewrite. Without it, long-running V1 queries ignore Thread.interrupt() and run until OOM.
   - Adds checkInterrupted() calls across the planner, compiler, all relational operators, and key expression operators. The parser already had support via InterruptibleParser. Always on, no opt-in flag needed.
   - Includes 11 integration tests.

## Other Information
- Updated Unreleased Section in CHANGELOG: **[YES/NO]**: YES
- Any backward-incompatible changes? **[YES/NO]**: NO
- Any new external dependencies? **[YES/NO]**: NO
- Do your changes comply with the [contributing][cg] and [code style][csg] guidelines? **[YES/NO]**: YES

## License Information

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

[cg]: https://github.com/partiql/partiql-lang-kotlin/blob/main/CONTRIBUTING.md
[csg]: https://github.com/partiql/partiql-lang-kotlin/blob/main/CODE_STYLE.md

PR #1913, opened 2026-05-14
Reapply "Add interruption check for V1 (#1908)" (#1911)

AugustineFu:
This reverts commit d501eeaf1f2efbc3b2a5a596d2633920f0494b73.

## Relevant Issues
- [Closes/Related To] [Task](https://taskei.amazon.dev/sprints/509699b0-5769-4c01-9852-458f9149e237?g=assignee&task=541ec58f-8dd2-45ab-ae92-72b3319584e3)

## Description
- Re-applies the thread interruption support for V1 that was reverted in #1911. No changes to the interrupt implementation itself, this is a clean revert of the revert.

## Other Information
- Do your changes comply with the [contributing][cg] and [code style][csg] guidelines? **[YES/NO]**: YES

## License Information

### DIFF
--- a/partiql-eval/src/main/kotlin/org/partiql/eval/internal/compiler/StandardCompiler.kt
+++ b/partiql-eval/src/main/kotlin/org/partiql/eval/internal/compiler/StandardCompiler.kt
@@ -10,6 +10,7 @@ import org.partiql.eval.compiler.Match
 import org.partiql.eval.compiler.PartiQLCompiler
 import org.partiql.eval.compiler.Strategy
 import org.partiql.eval.internal.helpers.PErrors
+import org.partiql.eval.internal.helpers.checkInterrupted
 import org.partiql.eval.internal.operator.Aggregate
 import org.partiql.eval.internal.operator.rel.RelOpAggregate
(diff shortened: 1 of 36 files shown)


johnedquinn (changes requested):
I believe the interruption checks are overkill. Interruption checks are typically small when it comes to performance impacts, but, for retrieving a single result row, if we need to check the interruption flag more than a few times then we'll start to see some impact. Also, the relational operators can really be broken down into two categories: streaming and blocking. Currently, all relational trees end with either a RelOpProject or RelOpPivot (streaming), so one interruption check there should be sufficient. Then, all blocking operators (joins, aggregations, etc) need an interruption check while they accumulate state.

I mentioned this offline, but ideally we'd want a PError type code specifically for this, so that the end-user isn't presented with an internal service exception. Also the error messages for checkInterrupted are not really necessary. The stack trace is maintained and we return PErrors throughout the compilation pipeline, so it's trivial for consumers to know which stage they're receiving the error from. The addition of a PError type code is not strictly necessary, but it's a nice to have.

johnedquinn (approved):
Nice. Thanks for these changes!

PR #1916, opened 2026-05-18
Add `MAP` type api in PartiQL

xd1313113:
## Relevant Issues

## Description
- Add `MAP` type api in PartiQL

## Other Information
- Do your changes comply with the [contributing][cg] and [code style][csg] guidelines? **[YES/NO]**:YES

## License Information

### DIFF
--- a/partiql-spi/src/main/java/org/partiql/spi/types/PType.java
+++ b/partiql-spi/src/main/java/org/partiql/spi/types/PType.java
@@ -117,6 +117,30 @@ public abstract class PType extends org.partiql.spi.Enum {
     }
 
     /**
+     * The key type of the MAP type.
+     *
+     * @return key type of the MAP
+     * @throws UnsupportedOperationException if this is called on a type whose {@link PType#code()} is not:
+     * {@link PType#MAP}
+     */
+    @NotNull
+    public PType getKeyType() throws UnsupportedOperationException {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * The value type of the MAP type.
+     *
+     * @return value type of the MAP
+     * @throws UnsupportedOperationException if this is called on a type whose {@link PType#code()} is not:
+     * {@link PType#MAP}
+     */
+    @NotNull
+    public PType getValueType() throws UnsupportedOperationException {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
      * <p>
      * Gets the interval's code. Each code corresponds with the public static final integers within
      * {@link IntervalCode} (e.g. {@link IntervalCode#YEAR}).
@@ -196,7 +220,8 @@ public abstract class PType extends org.partiql.spi.Enum {
                 PType.UNKNOWN,
                 PType.VARIANT,
                 PType.INTERVAL_YM,
-                PType.INTERVAL_DT
+                PType.INTERVAL_DT,
+                PType.MAP
         };
     }
 
@@ -260,6 +285,8 @@ public abstract class PType extends org.partiql.spi.Enum {
                 return "INTERVAL_YM";
             case PType.INTERVAL_DT:
                 return "INTERVAL_DT";
+            case PType.MAP:
+                return "MAP";
             default:
                 throw new UnsupportedCodeException(code);
         }
@@ -635,6 +662,17 @@ public abstract class PType extends org.partiql.spi.Enum {
      * </p>
      */
     public static final int INTERVAL_DT = 27;
+
+    /**
+     * MAP represents an unordered collection of key-value pairs where keys are unique.
+     * <br>
+     * <br>
+     * <b>Type Syntax</b>: <code>MAP&lt;K, V&gt;</code>
+     * <br>
+     * <b>Applicable methods</b>:
+     * {@link PType#getKeyType()}, {@link PType#getValueType()}
+     */
+    public static final int MAP = 28;
 
     private static final String UNSPECIFIED_LENGTH = "UNSPECIFIED_LENGTH";
     private static final String UNSPECIFIED_PRECISION = "UNSPECIFIED_PRECISION";
@@ -1140,6 +1178,22 @@ public abstract class PType extends org.partiql.spi.Enum {
     }
 
     /**
+     * @return a PartiQL map type with key type DYNAMIC and value type DYNAMIC
+     */
+    @NotNull
+    public static PType map() {
+        return new PTypeMap(PType.dynamic(), PType.dynamic());
+    }
+
+    /**
+     * @return a PartiQL map type with the specified key and value types
+     */
+    @NotNull
+    public static PType map(@NotNull PType keyType, @NotNull PType valueType) {
+        return new PTypeMap(keyType, valueType);
+    }
+
+    /**
      * @return a PartiQL unknown type
      */
     @NotNull
@@ -1219,6 +1273,8 @@ public abstract class PType extends org.partiql.spi.Enum {
                 return unknown();
             case VARIANT:
                 return variant("ion");
+            case MAP:
+                return map();
             default:
                 throw new IllegalArgumentException("Unknown type code: " + code);
         }
--- a/partiql-spi/src/main/java/org/partiql/spi/types/PType.java
+++ b/partiql-spi/src/main/java/org/partiql/spi/types/PType.java
@@ -1184,16 +1184,6 @@ public abstract class PType extends org.partiql.spi.Enum {
     }
 
     /**
-     * @return a PartiQL map type with key type DYNAMIC and value type DYNAMIC
-     * @deprecated This feature is experimental and is subject to change.
-     */
-    @Deprecated
-    @NotNull
-    public static PType map() {
-        return new PTypeMap(PType.dynamic(), PType.dynamic());
-    }
-
-    /**
      * @return a PartiQL map type with the specified key and value types
      * @deprecated This feature is experimental and is subject to change.
      */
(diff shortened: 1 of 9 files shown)


partiql-spi/src/main/java/org/partiql/spi/types/PType.java
```diff
@@ -1139,6 +1183,26 @@ public static PType struct() {
...
     }
 
+    /**
+     * @return a PartiQL map type with key type DYNAMIC and value type DYNAMIC
+     * @deprecated This feature is experimental and is subject to change.
+     */
+    @Deprecated
+    @NotNull
+    public static PType map() {
+        return new PTypeMap(PType.dynamic(), PType.dynamic());
```
johnedquinn: From our discussions, keys of type dynamic should be disallowed.

partiql-spi/src/main/java/org/partiql/spi/types/PTypeMap.java
```diff
@@ -0,0 +1,55 @@
...
+    private final PType _keyType;
+
+    @NotNull
+    private final PType _valueType;
+
+    PTypeMap(@NotNull PType keyType, @NotNull PType valueType) {
+        super(MAP);
+        _keyType = keyType;
+        _valueType = valueType;
+    }
```
johnedquinn, line 26: We should assert that the key is not dynamic.

partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
```diff
@@ -882,6 +919,37 @@ static Datum struct(@NotNull Iterable<Field> values) {
...
+     * Creates a MAP with the given type and entries.
+     * Duplicate keys follow last-write-wins semantics.
+     * @param type the MAP type (must have code MAP)
+     * @param entries the key-value entries
+     * @return a value of type {@link PType#MAP}
+     * @deprecated This feature is experimental and is subject to change.
+     */
+    @Deprecated
+    @NotNull
+    static Datum map(@NotNull PType type, @NotNull Iterable<Entry> entries) {
```
johnedquinn: To avoid people putting in bad types, can we just add a keyType and valueType params? Or add an assertion somewhere? And, check for dynamic key types (not allowed)

partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
```diff
@@ -882,6 +919,37 @@ static Datum struct(@NotNull Iterable<Field> values) {
...
+    // MAP
+
+    /**
+     * Creates an empty MAP with default type MAP&lt;DYNAMIC, DYNAMIC&gt;.
+     * @return a value of type {@link PType#MAP}
+     * @deprecated This feature is experimental and is subject to change.
+     */
+    @Deprecated
+    @NotNull
+    static Datum map() {
```
johnedquinn: To remove, correct?

partiql-spi/src/main/java/org/partiql/spi/value/DatumKey.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @NotNull
+    public Datum getDatum() {
+        return _datum;
+    }
+
+    @Override
+    public int hashCode() {
+        int code = _datum.getType().code();
```
johnedquinn, line 38: Need to check for isNull/isMissing first.

partiql-spi/src/main/java/org/partiql/spi/value/DatumKey.java
```diff
@@ -0,0 +1,89 @@
...
+                return _datum.getLocalDate().hashCode();
+            case PType.TIME:
+                return _datum.getLocalTime().hashCode();
+            case PType.TIMEZ:
+                return _datum.getOffsetTime().hashCode();
+            case PType.TIMESTAMP:
+                return _datum.getLocalDateTime().hashCode();
+            case PType.TIMESTAMPZ:
+                return _datum.getOffsetDateTime().hashCode();
+            // TODO: Support collection types (ARRAY, BAG, STRUCT, ROW, MAP) as MAP keys
```
XuechunHHH, line 72: Shall we throw a clear error for now rather than silently producing wrong behavior?

partiql-spi/src/main/java/org/partiql/spi/value/DatumComparator.java
```diff
@@ -742,6 +748,41 @@ private static DatumComparison[] fillStructComparator(DatumComparison[] comps) {
         return comps;
     }
 
+    @SuppressWarnings({"UnusedReturnValue"})
+    private static DatumComparison[] fillMapComparator(DatumComparison[] comps) {
```
XuechunHHH, line 752: This is sooo expensive for a map... :melting_face:

xd1313113: agree. Possibly optimization can be done if we implement as parallel arrays.

partiql-spi/src/main/java/org/partiql/spi/value/DatumKey.java
xd1313113: It throws `Type not supported` for collection types. Why do you mean by " silently producing wrong behavior?". I do not think user needs to know if it is TODO and Just not supported

partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
xd1313113: I renamed type to keyType and add checks. Is it good? Seems the rest api just use default Dynamic as type.

partiql-spi/src/main/java/org/partiql/spi/types/PType.java
```diff
@@ -116,6 +116,34 @@ public PType getTypeParameter() throws UnsupportedOperationException {
...
+     * @return value type of the MAP
+     * @throws UnsupportedOperationException if this is called on a type whose {@link PType#code()} is not:
+     * {@link PType#MAP}
+     * @deprecated This feature is experimental and is subject to change.
+     */
+    @Deprecated
+    @NotNull
+    public PType getValueType() throws UnsupportedOperationException {
+        throw new UnsupportedOperationException();
+    }
```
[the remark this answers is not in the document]
xd1313113: Planner change not included in this PR

partiql-spi/src/main/java/org/partiql/spi/types/PType.java
```diff
@@ -196,7 +224,8 @@ public static int[] codes() {
                 PType.UNKNOWN,
                 PType.VARIANT,
                 PType.INTERVAL_YM,
-                PType.INTERVAL_DT
+                PType.INTERVAL_DT,
+                PType.MAP
```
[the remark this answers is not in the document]
xd1313113: No Planner change right now.

partiql-spi/src/main/java/org/partiql/spi/types/PType.java
```diff
@@ -1139,6 +1183,16 @@ public static PType struct() {
...
     }
 
+    /**
+     * @return a PartiQL map type with the specified key and value types
+     * @deprecated This feature is experimental and is subject to change.
+     */
+    @Deprecated
+    @NotNull
+    public static PType map(@NotNull PType keyType, @NotNull PType valueType) {
+        return new PTypeMap(keyType, valueType);
```
johnedquinn, line 1194: You'll want to add an `@throws` here, since users won't know to not pass dynamic as the keyType

partiql-spi/src/main/java/org/partiql/spi/value/DatumKey.java
```diff
@@ -0,0 +1,105 @@
...
+                }
+                return BigDecimal.valueOf(f).stripTrailingZeros().hashCode();
+            }
+            case PType.DOUBLE: {
+                double d = _datum.getDouble();
+                if (Double.isNaN(d) || Double.isInfinite(d)) {
+                    throw new IllegalArgumentException("NaN and Infinity are not allowed as MAP keys");
+                }
+                return BigDecimal.valueOf(d).stripTrailingZeros().hashCode();
+            }
```
johnedquinn: Since we don't allow heterogeneous keys, there is no need to normalize to a BigDecimal.

partiql-spi/src/main/java/org/partiql/spi/value/DatumMap.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Override
+    public Datum get(@NotNull Datum key) {
+        DatumKey wrappedKey = new DatumKey(key);
+        Datum value = _entries.get(wrappedKey);
+        // null from HashMap means key not found; a NULL value would be stored as Datum.nullValue()
+        if (value == null) {
+            return Datum.missing();
+        }
```
johnedquinn, line 47: If I recall from the RFC, `my_map["nonexistent_key"]` should error in strict mode. Therefore, this API should likely return `Optional<Datum>`. This way, the engine can programmatically throw/return, rather than the Datum itself.

partiql-spi/src/main/java/org/partiql/spi/value/DatumKey.java
johnedquinn: Now that I'm reading this again, where would you prefer the isNull/isMissing check to be? Before calling hashcode, or in the impl? Sorry for moving back on this, I'm just trying to determine where these checks should go. I'll leave it up to your discretion.

johnedquinn (changes requested):

partiql-spi/src/main/java/org/partiql/spi/value/DatumMap.java
xd1313113: Thanks

johnedquinn (approved):
Thanks for this effort !

PR #1917, opened 2026-05-20
feat: add partiql fmt subcommand

XuechunHHH:
## Description

  Adds a partiql fmt subcommand for pretty-printing PartiQL statements based on [Wadler's "A Prettier Printer" (1997)](https://homepages.inf.ed.ac.uk/wadler/papers/prettier/prettier.pdf).

## Usage
```
  partiql fmt 'SELECT a, b FROM t WHERE x > 1'
  partiql fmt -i query.sql
  partiql fmt -w 120 'SELECT ...'
  cat query.sql | partiql fmt
```

## Architecture

  Input SQL
    → PartiQLParser.parse()
    → DocVisitor.visit(ast)    -- AST-to-Doc visitor (walks AST directly)
    → Doc tree                 -- group/nest/line at every breakpoint
    → render(width=80)         -- greedy best-fit picks flat vs broken
    → formatted SQL string

## Implementation

  - Doc.kt: Wadler document algebra (Nil, Text, Line, SoftBreak, HardLine, Concat, Nest, Union) with greedy best-fit renderer
  - PrettyPrinter.kt: AST-to-Doc visitor that produces Doc trees with breakpoints at:
    - Clause boundaries (SELECT, FROM, WHERE, GROUP BY, HAVING, ORDER BY, LIMIT, OFFSET, LET, EXCLUDE, WINDOW)
    - JOIN boundaries (binary tree flattened to one join per line)
    - Boolean operators (AND/OR chains flattened)
    - Comma-separated lists (SELECT items, function args, GROUP BY keys)
    - Subqueries and collection constructors in brackets
  - Operator flattening: same-precedence chains (a + b + c) stay flat; only inserts parens when precedence differs
  - FmtCommand.kt: picocli subcommand with -i (file input) and -w (width) options

 ### References:
  1. Philip Wadler, "A Prettier Printer" (1997) - http://homepages.inf.ed.ac.uk/wadler/papers/prettier/prettier.pdf
  2. CockroachDB sqlfmt - https://github.com/cockroachdb/cockroach/tree/master/pkg/util/pretty

## Other Information

## License Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ Thank you to all who have contributed!
 ## [Unreleased](https://TODO.com) - YYYY-MM-DD
 
 ### Added
-- Added `--fmt` flag to the CLI for pretty-printing PartiQL statements with width-aware formatting.
+- Added `fmt` subcommand to the CLI (`partiql fmt`) for pretty-printing PartiQL statements with width-aware formatting.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Thank you to all who have contributed!
 ## [Unreleased](https://TODO.com) - YYYY-MM-DD
 
 ### Added
+- Added `--fmt` flag to the CLI for pretty-printing PartiQL statements with width-aware formatting.
 
 ### Changed
 
@@ -39,6 +40,7 @@ Thank you to all who have contributed!
 
 ### Contributors
 Thank you to all who have contributed!
+- @XuechunHHH
 
 ## [1.3.13](https://github.com/partiql/partiql-lang-kotlin/releases/tag/v1.3.13) - 2026-05-07
 
--- a/partiql-cli/partiql.sh
+++ b/partiql-cli/partiql.sh
@@ -16,5 +16,5 @@
 cli_path=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 
 cd "$cli_path"
-../gradlew :partiql-cli:install
+../gradlew :partiql-cli:install < /dev/null
 ../partiql-cli/build/install/partiql-cli/bin/partiql "$@"
--- a/partiql-cli/src/main/kotlin/org/partiql/cli/FmtCommand.kt
+++ b/partiql-cli/src/main/kotlin/org/partiql/cli/FmtCommand.kt
@@ -1,0 +1,90 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ *  You may not use this file except in compliance with the License.
+ * A copy of the License is located at:
+ *
+ *      http://aws.amazon.com/apache2.0/
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+ *  language governing permissions and limitations under the License.
+ */
+
+package org.partiql.cli
+
+import org.partiql.cli.format.pretty
+import org.partiql.parser.PartiQLParser
+import org.partiql.spi.errors.PRuntimeException
+import picocli.CommandLine
+import java.io.File
+import java.io.FileNotFoundException
+import java.io.IOException
+
+/**
+ * The `partiql fmt` subcommand: pretty-prints PartiQL statements.
+ */
+@CommandLine.Command(
+    name = "fmt",
+    mixinStandardHelpOptions = true,
+    description = ["Format (pretty-print) PartiQL statements and print to stdout."]
+)
+internal class FmtCommand : Runnable {
+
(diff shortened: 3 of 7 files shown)


partiql-cli/src/main/kotlin/org/partiql/cli/Main.kt
```diff
@@ -96,6 +98,12 @@ internal class MainCommand : Runnable {
     )
     var env: File? = null
 
+    @CommandLine.Option(
+        names = ["--fmt"],
+        description = ["Format (pretty-print) the input PartiQL statement and print to stdout."],
+    )
+    var fmt: Boolean = false
```
johnedquinn: IMO, this should be a subcommand, since it's a directive to do something outside of querying.

XuechunHHH: Makes sense.

partiql-cli/src/main/kotlin/org/partiql/cli/format/PrettyPrinter.kt
```diff
@@ -0,0 +1,146 @@
...
+ * "SELECT a, b FROM t WHERE x > 1"
+ *
+ * // Long query → clause per line:
+ * "SELECT a, b, c, d, e\nFROM some_long_table_name\nWHERE x > 1 AND y < 100"
+ *
+ * // Very long clause body → keyword + indented body:
+ * "SELECT\n    a, b, c, d, e, f, g, h\nFROM t"
+ * ```
+ */
+internal fun AstNode.pretty(width: Int = 80): String {
```
johnedquinn, line 72: This doesn't require many tests, but maybe just one complex one to make sure we don't break it in the future.

johnedquinn (dismissed):

partiql-cli/src/main/kotlin/org/partiql/cli/FmtCommand.kt
```diff
@@ -0,0 +1,87 @@
...
+import picocli.CommandLine
+import java.io.File
+import java.io.FileNotFoundException
+import java.io.IOException
+
+/**
+ * The `partiql fmt` subcommand: pretty-prints PartiQL statements.
+ */
+@CommandLine.Command(
+    name = "fmt",
```
xd1313113, line 29: I feel full `format` is more clear as subcommand.

partiql-cli/src/test/kotlin/org/partiql/cli/format/PrettyPrinterTest.kt
```diff
@@ -0,0 +1,42 @@
...
+        """.trimIndent()
+        val first = fmt(input)
+        val second = fmt(first)
+        assertEquals(first, second, "Formatting the monster query should be idempotent")
+        assert(first.lines().size > 10) { "Monster query should break across many lines" }
```
xd1313113: Do we need detailed output string for testing to catch possible inner line changes like spaces?

partiql-cli/partiql.sh
```diff
@@ -16,5 +16,5 @@
 cli_path=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 
 cd "$cli_path"
-../gradlew :partiql-cli:install
+../gradlew :partiql-cli:install < /dev/null
```
xd1313113, line 19: Why we need this?

CHANGELOG.md
```diff
@@ -26,7 +26,7 @@ Thank you to all who have contributed!
 ## [Unreleased](https://TODO.com) - YYYY-MM-DD
 
 ### Added
-- Re-applied thread interruption support for the V1 evaluation engine, added `PError.INTERRUPTED` error code for identifying query cancellation via `Thread.interrupt()`
+- Added `--fmt` flag to the CLI for pretty-printing PartiQL statements with width-aware formatting.
```
xd1313113: Do we need update as subcommand

partiql-cli/src/main/kotlin/org/partiql/cli/FmtCommand.kt
XuechunHHH: I personally want to go with `fmt` to follow the convention from `go fmt`, `terraform fmt` and rustfmt. But happy to change to `format` if the team prefers the full word. Let me know if you have a strong preference.

partiql-cli/src/test/kotlin/org/partiql/cli/format/PrettyPrinterTest.kt
XuechunHHH: Emmm, we can have a golden-file test that pins the exact output if you think it worth. Since it is only initial commit for internal usage, I think it is okay to keep as-is for later `fmt` update.

partiql-cli/partiql.sh
XuechunHHH, line 19: When piping stdin to `partiql fmt`, Gradle was consuming stdin before the actual CLI binary could read it. Redirecting Gradle's stdin from `/dev/null` prevents that. so the piped input is preserved for the next line.

johnedquinn (approved):

xd1313113 (approved):

PR #1918, opened 2026-05-20
Add minimal implementation of `MAP`

xd1313113:
## Relevant Issues

## Description
- The minimal implementation of `MAP` type

## Other Information
- Do your changes comply with the [contributing][cg] and [code style][csg] guidelines? **[YES/NO]**:YES

## License Information


lib/sprout/src/main/kotlin/org/partiql/sprout/generator/target/kotlin/KotlinSymbols.kt
```diff
@@ -76,14 +76,29 @@ class KotlinSymbols private constructor(
          * Named constructor somewhat hints at the initial emptiness (when memoized) of the symbol table
          */
         fun init(universe: Universe, options: KotlinOptions): KotlinSymbols = KotlinSymbols(universe, options)
+
+        /**
+         * Kotlin stdlib type names that conflict when used as nested class names.
+         */
+        private val KOTLIN_CONFLICTING_NAMES = setOf(
```
xd1313113, line 83: `map` from [partiql_plan_internal.ion‎](https://github.com/partiql/partiql-lang-kotlin/pull/1918/changes#diff-9528ffa80596b1ac019d7674040f290258e005d9557e8ca0f72492f5659b8158) has conflict with native kotlin lib. escape when conflict

partiql-ast/src/main/java/org/partiql/ast/AstVisitor.java
johnedquinn, line 1: I believe you also need to update the AstRewriter

partiql-ast/src/main/java/org/partiql/ast/DataType.java
```diff
@@ -398,6 +400,14 @@ public static DataType BAG() {
         return new DataType(BAG);
     }
 
+    public static DataType MAP() {
```
johnedquinn: Should be removed. And the keyType should disallow dynamic/any

partiql-ast/src/main/java/org/partiql/ast/DataType.java
```diff
@@ -157,6 +157,8 @@ public String getComment() {
     // <datetime type> continue
     public static final int TIME_WITHOUT_TIME_ZONE = 46;
     public static final int TIMESTAMP_WITHOUT_TIME_ZONE = 47;
+    // <map type>
+    public static final int MAP = 48;
```
johnedquinn, line 164: Nice. For documentation purposes, you may want to note that keyType and elementType methods are used for this code. And link the static constructors to reference this documentation.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprMapConstruct.kt
```diff
@@ -0,0 +1,25 @@
...
+    override fun eval(env: Environment): Datum {
+        val entries = mutableListOf<Entry>()
+        for (field in fields) {
+            val key = field.key.eval(env)
+            if (key.isNull) {
+                throw IllegalArgumentException("NULL is not allowed as a MAP key")
+            }
+            if (key.isMissing) {
+                throw IllegalArgumentException("MISSING is not allowed as a MAP key")
+            }
```
johnedquinn, line 22: Is this expected to happen for both permissive and strict modes?

partiql-eval/src/main/kotlin/org/partiql/eval/internal/compiler/StandardCompiler.kt
```diff
@@ -502,6 +504,15 @@ internal class StandardCompiler(strategies: List<Strategy>) : PartiQLCompiler {
             }
         }
 
+        override fun visitMap(rex: RexMap, ctx: Unit): ExprValue {
+            val entries = rex.getEntries().map {
+                val k = compile(it.key, ctx)
+                val v = compile(it.value, ctx).catch()
```
johnedquinn: See https://github.com/partiql/partiql-lang-kotlin/pull/1918/changes#r3306809957.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprPathIndex.kt
```diff
@@ -14,6 +14,13 @@ internal class ExprPathIndex(
 
     override fun eval(env: Environment): Datum {
         val input = root.eval(env)
+
+        // MAP key lookup
+        if (input.type.code() == PType.MAP) {
+            val k = key.eval(env)
+            return input.get(k)
```
johnedquinn: You should have isNull/isMissing checks

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprPathKey.kt
```diff
@@ -14,12 +14,17 @@ internal class ExprPathKey(
 ) : ExprValue {
 
     override fun eval(env: Environment): Datum {
-        val rootEvaluated = root.eval(env).checkStruct()
+        val rootEvaluated = root.eval(env)
```
johnedquinn: We're still going to need to make sure it's either STRUCT/MAP/ROW.

Also, now that this is essentially acting like hard-coded dynamic dispatch, we _could_ add a new set of ops: `ExprPathKeyStruct`, `ExprPathKeyMap`, `ExprPathKeyDynamic` (which dispatches to the first two).

Also, should there be a root/key isMissing check here? I'm trying to think of whether it should exist here.

partiql-parser/src/main/antlr/PartiQLParser.g4
```diff
@@ -1257,14 +1264,15 @@ type
         | INTEGER8 | INT8 | BIGINT | REAL | CHAR | CHARACTER
         | STRING | SYMBOL | DATE | ANY
       )                                                                                                                # TypeAtomic
-    | datatype=( STRUCT | TUPLE | LIST | ARRAY | SEXP | BAG )                                                          # TypeComplexAtomic
+    | datatype=( STRUCT | TUPLE | LIST | ARRAY | SEXP | BAG | MAP )                                                     # TypeComplexAtomic
```
johnedquinn: Should be disallowed.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -2478,6 +2489,14 @@ internal class PartiQLParserDefault : PartiQLParser {
             DataType.ARRAY(type)
         }
 
+        override fun visitTypeMap(ctx: GeneratedParser.TypeMapContext): DataType = translate(ctx) {
+            val keyType = visitAs<DataType>(ctx.key)
+                .also { isValidTypeDeclarationOrThrow(it, ctx.key) }
```
johnedquinn, line 2495: Should also check for dynamic here, as it's disallowed.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -1272,6 +1286,17 @@ internal class PlanTyper(private val env: Env, config: Context, private val flag
             return rex(type, rexOpStruct(fields))
         }
 
+        override fun visitRexOpMap(node: Rex.Op.`Map`, ctx: CompilerType?): Rex {
+            _listener.report(PErrors.experimental("MAP Constructor"))
```
johnedquinn: We should _probably_ consolidate to one feature name to be used across parsing, planning, compiling. That way, application owners can more easily allow/disallow the feature.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -1272,6 +1286,17 @@ internal class PlanTyper(private val env: Env, config: Context, private val flag
...
+            val entries = node.entries.map {
+                val k = visitRex(it.k, it.k.type)
+                val v = visitRex(it.v, it.v.type)
+                rexOpMapEntry(k, v)
+            }
+            val type = CompilerType(PType.map())
```
johnedquinn: Parameters should be resolved.

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
```diff
@@ -2304,6 +2305,15 @@ internal class PartiQLParserDefault : PartiQLParser {
             exprStruct(fields)
         }
 
+        override fun visitMapConstructor(ctx: GeneratedParser.MapConstructorContext) = translate(ctx) {
```
johnedquinn, line 2308: listener experimental usage perror reporting here ?

johnedquinn: Also, it'd be interesting to see how the CLI can expose the new experimental feature. Could be in a follow-up.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprMapConstruct.kt
```diff
@@ -0,0 +1,25 @@
+package org.partiql.eval.internal.operator.rex
+
+import org.partiql.eval.Environment
+import org.partiql.eval.ExprValue
+import org.partiql.spi.types.PType
+import org.partiql.spi.value.Datum
+import org.partiql.spi.value.Entry
+
+internal class ExprMapConstruct(private val fields: List<ExprStructField>) : ExprValue {
```
johnedquinn: This should accept an optional keyType/valueType here. If none provided, then we'll need to adjust the keyType/valueType returned on line 23.

johnedquinn: * and assert that keyType is not dynamic.

Also, at runtime, it'll be interesting, because we'll need to figure out the common supertype of the key types (excluding dynamic) and value types. Then, cast all non-conforming keys/values accordingly.

johnedquinn: We haven't really discussed the common supertype for the values of entries, though we have for keys. We can discuss this offline.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/compiler/StandardCompiler.kt
xd1313113: It is wrong link? I did not get it.

partiql-ast/src/main/java/org/partiql/ast/AstVisitor.java
xd1313113: thanks

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprMapConstruct.kt
xd1313113: No, throws Runtime exception to allow different behavior in two modes

partiql-parser/src/main/kotlin/org/partiql/parser/internal/PartiQLParserDefault.kt
xd1313113: DataType does not have Dynamic/Any to check

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
xd1313113: use `MAP Type` for now

partiql-ast/src/main/java/org/partiql/ast/DataType.java
johnedquinn: Nit: `/**` for javadocs. Inline comments aren't rendered for consumers.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprMapConstruct.kt
```diff
@@ -0,0 +1,45 @@
...
+        for (field in fields) {
+            val key = field.key.eval(env)
+            if (key.isNull || key.isMissing) {
+                throw PErrors.pathKeyFailureException()
+            }
+            keys.add(key)
+            values.add(field.value.eval(env))
+        }
+        val resolvedKeyType = keyType ?: DynamicTyper.commonSuperType(keys.map { it.type })
+        assert(resolvedKeyType.code() != PType.DYNAMIC) { "MAP key type must not be DYNAMIC" }
```
johnedquinn: nit: eventually this will need to be its own PError type code.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprMapConstruct.kt
```diff
@@ -0,0 +1,45 @@
...
+        val values = mutableListOf<Datum>()
+        for (field in fields) {
+            val key = field.key.eval(env)
+            if (key.isNull || key.isMissing) {
+                throw PErrors.pathKeyFailureException()
+            }
+            keys.add(key)
+            values.add(field.value.eval(env))
+        }
+        val resolvedKeyType = keyType ?: DynamicTyper.commonSuperType(keys.map { it.type })
```
johnedquinn: I'm kind of thinking that this should only occur if we don't have compile-time information. I need to double-check what the result of the map constructor is in the planner.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -1279,6 +1293,40 @@ internal class PlanTyper(private val env: Env, config: Context, private val flag
...
+        override fun visitRexOpMap(node: Rex.Op.Map, ctx: CompilerType?): Rex {
+            _listener.report(PErrors.experimental("MAP type"))
+            val typedEntries = node.entries.map {
+                val k = visitRex(it.k, it.k.type)
+                val v = visitRex(it.v, it.v.type)
+                rexOpMapEntry(k, v)
+            }
+            val keyType = if (typedEntries.isNotEmpty()) {
+                typedEntries.map { it.k.type }.reduce { acc, t ->
+                    (getCommonSuperType(acc, t) ?: error("Incompatible MAP key types: $acc and $t")).toCType()
```
johnedquinn, line 1305: This should check if the common supertype is dynamic. If so, error.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -1279,6 +1293,40 @@ internal class PlanTyper(private val env: Env, config: Context, private val flag
...
+            val valueType = if (typedEntries.isNotEmpty()) {
+                typedEntries.map { it.v.type }.reduce { acc, t ->
+                    (getCommonSuperType(acc, t) ?: acc).toCType()
+                }
+            } else {
+                PType.dynamic()
+            }
+            val entries = typedEntries.map { entry ->
+                val k = if (entry.k.type != keyType) {
+                    val cast = env.resolveCast(entry.k, keyType)
```
johnedquinn, line 1327: You should cast both keys and values to conform with the expected output key/value types. Also, since you have all of this logic, it really would make sense to have both ExprMapConstructor and ExprMapConstructorDynamic, since what you're implementing here is basically dynamic dispatch vs inline compile-time coercions.

Then, your ExprMapConstructor can just assume that the inputs to the constructor are conforming with the input type definitions -- since you're applying the coercions ahead of time.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprPathKeyStruct.kt
```diff
@@ -0,0 +1,28 @@
...
+    @JvmField val key: ExprValue
+) : ExprValue {
+
+    override fun eval(env: Environment): Datum {
+        return evalWithInput(root.eval(env), env)
+    }
+
+    fun evalWithInput(input: Datum, env: Environment): Datum {
+        val k = key.eval(env).check(PType.string())
+        if (input.isNull || k.isNull || k.isMissing) {
```
johnedquinn: or if `input.isMissing`. I would imagine that we return missing if the input/key was missing though.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprPathKeyMap.kt
```diff
@@ -0,0 +1,29 @@
...
+    override fun eval(env: Environment): Datum {
+        return evalWithInput(root.eval(env), env)
+    }
+
+    fun evalWithInput(input: Datum, env: Environment): Datum {
+        if (input.isNull || input.isMissing) {
+            return Datum.nullValue()
+        }
+        val k = key.eval(env).check(PType.string())
+        if (k.isNull || k.isMissing) {
```
johnedquinn: This is different than ExprPathKeyStruct. Is that intentional? I would've imagined:
- input or key are missing --> return missing
- input or key are null --> return null
- entry doesn't exist, throw

Need to double check the spec though.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/compiler/StandardCompiler.kt
johnedquinn: Just has to do with the root/key being missing. What's expected to happen? I have another similar comment.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprPathKeyMap.kt
xd1313113: It should be consistent with struct now.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprPathKeyStruct.kt
xd1313113: fixed

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
xd1313113: fixed

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
xd1313113: we cannot error here, as key expr can be Dynamic type and typing is deferred to evaluation. I will error in when evaluate dynamic map in this case

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprMapConstruct.kt
xd1313113: Now if we don't have compile-time information, ExprMapConstructDynamic will be dispatched. other ExprMapConstruct

partiql-eval/src/main/kotlin/org/partiql/eval/internal/compiler/StandardCompiler.kt
xd1313113: root/key  missing is consistent with struct now.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprMapConstruct.kt
xd1313113: Will check required error code for map in later PRs

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprMapConstruct.kt
```diff
@@ -0,0 +1,31 @@
...
+) : ExprValue {
+    override fun eval(env: Environment): Datum {
+        val entries = mutableListOf<Entry>()
+        for (field in fields) {
+            val key = field.key.eval(env)
+            if (key.isNull) {
+                return Datum.nullValue()
+            }
+
+            if (key.isMissing) {
```
johnedquinn, line 20: Just for context:

https://github.com/partiql/partiql-lang-kotlin/blob/405d2027a514fb868cd170d31d6f2613c157af12/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprStructStrict.kt#L18-L20

Typically, when inputs to an operation/function are unknown, the result is unknown.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprMapConstructDynamic.kt
```diff
@@ -0,0 +1,49 @@
...
+        val values = mutableListOf<Datum>()
+        for (field in fields) {
+            val key = field.key.eval(env)
+            if (key.isNull) {
+                return Datum.nullValue()
+            }
+
+            if (key.isMissing) {
+                throw PErrors.pathKeyFailureException()
+            }
```
johnedquinn, line 25: Same as my other comment in ExprMapConstruct.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprPathIndexCollection.kt
```diff
@@ -0,0 +1,41 @@
...
+    @JvmField val key: ExprValue,
+) : ExprValue {
+
+    override fun eval(env: Environment): Datum {
+        return evalWithInput(root.eval(env), env)
+    }
+
+    fun evalWithInput(input: Datum, env: Environment): Datum {
+        val iterator = when (input.type.code()) {
+            PType.BAG,
```
johnedquinn, line 21: I'm surprised that this was in the original code. The bag indexing.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprPathIndexMap.kt
```diff
@@ -0,0 +1,24 @@
...
+) : ExprValue {
+
+    override fun eval(env: Environment): Datum {
+        return evalWithInput(root.eval(env), env)
+    }
+
+    fun evalWithInput(input: Datum, env: Environment): Datum {
+        val k = key.eval(env)
+        if (input.isNull || k.isNull) {
+            return Datum.nullValue()
```
johnedquinn: This should probably have the type of the map's valueType.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprPathIndexMap.kt
```diff
@@ -0,0 +1,24 @@
...
+        }
+        return input.get(k).orElseThrow { PErrors.pathIndexFailureException() }
```
johnedquinn, line 25: In a follow-up, we probably want a new error code, since indexing refers to grabbing something at a particular index. In this case, it's syntactic sugar for the get operation.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprPathIndexMap.kt
johnedquinn: Also, we should handle missing for both the input and key.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprPathKeyMap.kt
```diff
@@ -0,0 +1,25 @@
...
+    @JvmField val key: ExprValue
+) : ExprValue {
+
+    override fun eval(env: Environment): Datum {
+        return evalWithInput(root.eval(env), env)
+    }
+
+    fun evalWithInput(input: Datum, env: Environment): Datum {
+        val k = key.eval(env)
+        if (input.isNull || k.isNull) {
```
johnedquinn, line 19: Missing value, and type info.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprPathKeyStruct.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    override fun eval(env: Environment): Datum {
+        return evalWithInput(root.eval(env), env)
+    }
+
+    fun evalWithInput(input: Datum, env: Environment): Datum {
+        val checkedRoot = input.checkStruct()
+        val k = key.eval(env).check(PType.string())
+        if (checkedRoot.isNull || k.isNull) {
+            return Datum.nullValue()
```
johnedquinn, line 24: Same here.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt
```diff
@@ -135,6 +136,13 @@ internal class PlanTransform(private val flags: Set<PlannerFlag>) {
             return operators.struct(fields)
         }
 
+        override fun visitRexOpMap(node: IRex.Op.Map, ctx: PType): Any {
+            val entries = node.entries.map { mapEntry(it) }
+            val keyType = if (ctx.code() == PType.MAP) ctx.keyType else PType.dynamic()
```
johnedquinn: This is kind of misleading, since dynamic keyTypes shouldn't be allowed, especially in the public API. This kind of relates to our offline discussion when I was talking about how map constructors can sort-of be represented as functions, since we have dynamic dispatch plumbing already. We don't support varargs, so it would require a work-around, but this also relates to my point about how resolved functions and dynamic dispatch are different nodes. If you wanted it to be clear-cut, you could have a RexOpMapDynamic.

johnedquinn (changes requested):
Nice work thus far. This PR is really shaping up. Some comments that need to be addressed. Also, with all of these new operations, we really need tests.

Where types are known at compile-time, when they're not, etc; Permissive and strict mode; heterogeneous values.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprPathIndexMap.kt
xd1313113: yes. Will think about error code for map

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt
xd1313113: create RexOpMapDynamic, it looks much better dispatching now

partiql-plan/src/main/java/org/partiql/plan/OperatorVisitor.java
```diff
@@ -223,6 +225,18 @@ default R visitStruct(@NotNull RexStruct rex, C ctx) {
         return defaultVisit(rex, ctx);
     }
 
+    default R visitMap(@NotNull RexMap rex, C ctx) {
```
johnedquinn, line 231: This should be marked experimental.

partiql-plan/src/main/kotlin/org/partiql/plan/Operators.kt
```diff
@@ -397,6 +399,16 @@ public interface Operators {
...
+    /**
+     * Create a [RexMap] instance.
+     *
+     * @param fields map entries (key-value pairs)
+     * @return
+     */
+    public fun map(keyType: PType, valueType: PType, entries: List<RexMap.Entry>): RexMap = RexMap.create(keyType, valueType, entries)
+
+    public fun mapDynamic(entries: List<RexMap.Entry>): RexMapDynamic = RexMapDynamic.create(entries)
+
```
johnedquinn, line 419: Should be marked experimental.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprMapConstructDynamic.kt
```diff
@@ -0,0 +1,48 @@
...
+        val keys = mutableListOf<Datum>()
+        val values = mutableListOf<Datum>()
+        for (field in fields) {
+            val key = field.key.eval(env)
+            if (key.isNull) {
+                return Datum.nullValue()
+            }
+
+            if (key.isMissing) {
+                return Datum.missing()
```
johnedquinn, line 24: Technically we still need to compute the type information here for keyType and valueType. This is because if we end up passing the missing/null value to functions, we might resolve to the wrong function. It's a small bug though.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprMapConstruct.kt
```diff
@@ -0,0 +1,29 @@
...
+        for (field in fields) {
+            val key = field.key.eval(env)
+            if (key.isNull) {
+                return Datum.nullValue(PType.map(keyType, valueType))
+            }
+            if (key.isMissing) {
+                return Datum.missing(PType.map(keyType, valueType))
+            }
+            val value = field.value.eval(env)
+            entries.add(Entry.of(key, value))
```
johnedquinn: Oh, one question. I know that MISSING values are stripped from struct construction. I know that in your RFC, missing values are allowed -- https://github.com/partiql/partiql-lang/pull/104/changes . I'm just curious whether we actually want that.

xd1313113: what about map{'a': T.b} and T.b is evaluated as missing? Do we allow this?   Or the whole map will be propagated as Missing

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprMapConstructDynamic.kt
xd1313113: I see. Let me add a TODO to fix it. Will address in later PR.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprMapConstruct.kt
xd1313113: It seems Spec does not allow Missing as Value for a tuple.
I was thinking to distinguish the case key exists, value is missing vs key does not exits at all. However per spec, those two cases should be treated as same and get omitted, I updated my implement to align with spec to strip out missing values in map. I will update MAP spec as well

johnedquinn (approved):
Nice work thus far. Thanks for piecing this together !

PR #1924, opened 2026-06-01
feat: thread-safe cacheable execution plans

johnedquinn:
## Summary

Introduces thread-safe, cacheable execution plans and an opaque `PartiQLVM`.

Plans use integer-referenced table nodes (`RexTableRef`) resolved lazily via user-provided `ExecutionCatalog` at execution time. Functions and aggregates are embedded directly (they're stateless/thread-safe).

## Low-Level Summary

We're introducing `ExecutionPlan` (thread-safe, backed by internal data classes that are immutable) and `PartiQLVM` (internally creates the operator tree given the execution plan). Both of these APIs are **intentionally** opaque. In the future, we may want to leverage `partiql-jni` bytecode as the execution plan -- therefore, `ExecutionPlan` has **zero** public methods. `PartiQLVM` only has one.

To do this, we need a way to reference tables without keeping pointers directly in the executable. This PR introduces a `SymbolTable` that gives consumers a mapping between table qualified names and integers. _Note: Functions and aggregations are still embedded directly, as we assume them to be stateless. If there is a customer-driven need to use unique integer references, then we'll cross that barrier later._

When it comes to compiling, we go from: `Plan (pojo) → PartiQLCompiler -> PlanToExecTransform → ExecutionPlanImpl (pojo)`.

When it comes to evaluation, `ExecutionPlan (pojo) → OperatorCompiler → Evaluable Tree`. Consumers pass in an `ExecutionCatalog`.

Other notable changes: we now use mode (PERMISSIVE/STRICT) at compile time, to avoid the switching overhead at runtime. Also, since we have `Strategy` as a public API for supplying custom `Expr` implementations, I've now added a new API to instead produce `Supplier<Expr>` factories (via `applyFactory()`) for thread-safe custom operators. The other API wasn't thread-safe.

## FAQs

1. **Q**: Why not make the plan itself thread-safe? **A**: A few reasons: our plan as it stands is not even close to being thread-safe. We reference table impls directly, our collections are mutable, node identifiers are mutable, etc. Also, our customers want to cache the final executable to avoid any compiling/planning. This enables that.

### Previous User Experience

```kotlin
val planner = PartiQLPlanner.standard()
val plan = planner.plan(ast, session).plan

val compiler = PartiQLCompiler.standard()
val statement = compiler.prepare(plan, Mode.PERMISSIVE())

// NOT thread-safe — operator tree holds mutable state
val result = statement.execute()
```

### New User Experience

```kotlin
// 1. Plan (with useRefs for thread-safe table references)
val planner = PartiQLPlanner.builder().useRefs().build()
val result = planner.plan(ast, session)
val plan = result.plan         // introspectable, has RexTableRef nodes
val symbols = result.symbols   // maps integer IDs → catalog/table names

// 2. Compile (validates plan is ref-based, bakes mode in)
val compiler = PartiQLCompiler.standard()
val execPlan = compiler.compile(plan, Mode.PERMISSIVE())

// WE CAN NOW CACHE `execPlan`. ALL CODE BELOW CAN BE FROM OTHER THREADS.

// 3. Build ExecutionCatalog from SymbolTable
//    The SymbolTable tells you what each ID means:
//      symbols.getCatalogName(0)    → "my_catalog"
//      symbols.getTables(0)[0]      → TableEntry(id=0, name="users", schema=...)
//      symbols.getTables(0)[1]      → TableEntry(id=1, name="orders", schema=...)
val catalogs = Array(symbols.catalogCount()) { catalogId ->
    ExecutionCatalog { tableId ->
        val entry = symbols.getTables(catalogId)[tableId]
        myDatabase.getTable(entry.name)  // your storage layer
    }
}

// 4. Execute (thread-safe — fresh operator tree per call)
//    Safe to call concurrently from multiple threads with different catalogs
val vm = PartiQLVM.standard()
val datum = vm.execute(execPlan, catalogs)
```

## Backward Compatibility
- Legacy `prepare(plan, mode)` still works with embedded-object plans. It now just invokes the new PartiQLVM under-the-hood.
- New `prepare(plan, mode, catalogs)` for thread-safe Statement convenience
- `compile(plan, mode)` + `PartiQLVM.execute(plan, catalogs)` for explicit VM control
- `RexTable` (embedded) deprecated in favor of `RexTableRef`. You have to opt-in to `useRefs()` so that we don't break current users' code.
- `Statement` deprecated in favor of `PartiQLVM`

## Testing

- [x] Dual-path execution in `SuccessTestCase` / `FailureTestCase` / `TypingTestCase` — all existing eval tests exercise both old and new paths with hard assertions
- [x] `ThreadSafetyTest` — concurrent execution with shared plan, different data per thread
- [x] `ExecutionPlanValidationTest` — compile rejects non-ref plans, accepts ref plans, legacy prepare works both ways, new prepare rejects non-ref
- [x] Conformance test runner updated to use VM path (excluded from default build, 281 gaps from GROUP AS / ordered set ops)
- [x] `StrategyTest` — backward-compat via deprecated `apply()` wrapping into `applyFactory()`
- [x] Full `./gradlew build` passes (including all eval, planner, and plan module tests)


partiql-eval/src/main/java/org/partiql/eval/PartiQLVM.java
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Execute the plan against the given execution catalogs.
+     *
+     * @param plan     the compiled plan (immutable, cacheable, mode baked in)
+     * @param catalogs the execution catalogs indexed by catalog ID from the plan's symbol table
+     * @return the query result
+     * @throws PRuntimeException if an error is encountered during execution
+     */
+    @NotNull
+    Datum execute(@NotNull ExecutionPlan plan, @NotNull ExecutionCatalog[] catalogs) throws PRuntimeException;
```
xd1313113, line 32: for catalogs array, if the plan has a tableid, does it need loop through all catalogs to get the table?

partiql-eval/src/test/kotlin/org/partiql/eval/internal/VMTestUtils.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.partiql.plan.SymbolTable
+import org.partiql.spi.catalog.Catalog
+import org.partiql.spi.catalog.ExecutionCatalog
+import org.partiql.spi.catalog.Session
+import org.partiql.spi.catalog.Table
+
+/**
+ * Builds an array of [ExecutionCatalog] from a [Session] and [SymbolTable] for testing.
+ */
+internal fun buildExecutionCatalogs(symbols: SymbolTable, session: Session): Array<ExecutionCatalog> {
```
XuechunHHH, line 12: Consider promoting it to a public factory, or switch the public API from `ExecutionCatalog[]` array to `Map<String, ExecutionCatalog>`. I feel the public API with an indexed input is fragile.

partiql-eval/src/main/kotlin/org/partiql/eval/ExecutionPlan.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import org.partiql.eval.internal.plan.ExecutionPlanImpl
+
+/**
+ * An opaque, immutable, thread-safe execution plan.
+ *
+ * Produced by [org.partiql.eval.compiler.PartiQLCompiler.compile].
+ * Executed by [PartiQLVM.execute].
+ *
+ * This class has no public methods — internal structure may change without notice.
```
XuechunHHH, line 11: Curious why we make the decision. When something wrong at execution time, there's no way to inspect or what plan was cached without carrying external metadata alongside it in this way.

partiql-eval/src/main/java/org/partiql/eval/compiler/PartiQLCompiler.java
```diff
@@ -18,29 +20,89 @@ public interface PartiQLCompiler {
...
+     * Use this when you want the convenience of {@link Statement} with thread-safe ref-based plans.
+     *
+     * @param plan     The ref-based plan (produced by a planner with {@code useRefs()} enabled).
+     * @param mode     The execution mode (permissive or strict).
+     * @param catalogs The execution catalogs indexed by catalog ID from the plan's symbol table.
+     * @return A thread-safe statement.
+     * @throws PRuntimeException If the plan contains embedded objects or compilation fails.
+     */
+    @NotNull
+    default Statement prepare(@NotNull Plan plan, @NotNull Mode mode, @NotNull ExecutionCatalog[] catalogs) throws PRuntimeException {
```
xd1313113, line 88: Since Statement is deprecated, is it for back-compatible only?

partiql-eval/src/main/kotlin/org/partiql/eval/internal/plan/PExpr.kt
```diff
@@ -0,0 +1,35 @@
...
+import org.partiql.spi.function.FnOverload
+import org.partiql.spi.types.PType
+import org.partiql.spi.value.Datum
+
+internal sealed class PExpr {
+    data class Lit(val value: Datum) : PExpr()
+    data class Var(val depth: Int, val offset: Int) : PExpr()
+    data class TableRef(val catalogId: Int, val tableId: Int) : PExpr()
+    data class TableDirect(val table: Table) : PExpr()
+    data class Call(val fn: Fn, val args: List<PExpr>) : PExpr()
```
XuechunHHH, line 15: Double-check: Thread-safety relies on `Fn` and `Agg` also immutable

johnedquinn: Yeah, that's documented in Javadocs

partiql-eval/src/main/kotlin/org/partiql/eval/ExecutionPlan.kt
johnedquinn: We have an error reporting mechanism, and if it needs to be flushed out to allow even finer-grained information, then that's something we can do. But, this intentionally doesn't expose APIs, following the modeling of PartiQL Rust.

As an example, the JVM itself doesn't give users the "why is this the wrong bytecode". It uses its own error reporting mechanism to give user-actionable information.

partiql-eval/src/test/kotlin/org/partiql/eval/internal/VMTestUtils.kt
johnedquinn: I'm not sure about this. It's not fragile if you have clear docs. It's kind of similar to why we have integer-indexed table references -- for speed. String hashcodes/comparisons introduce a non-zero amount of overhead.

partiql-spi/src/main/java/org/partiql/spi/catalog/ExecutionCatalog.java
```diff
@@ -0,0 +1,27 @@
...
+public interface ExecutionCatalog {
+
+    /**
+     * Returns the table for the given plan-assigned ID.
+     *
+     * @param id the table identifier assigned during planning
+     * @return the table instance
+     */
+    @NotNull
+    Table getTable(int id);
```
xd1313113, line 26: what is the scope of id, is it session unique or catalog unique?

xd1313113 (dismissed):
Looks good to me.

XuechunHHH (dismissed):

partiql-eval/src/main/java/org/partiql/eval/compiler/PartiQLCompiler.java
johnedquinn: Yes, just a means for users to be able to incrementally migrate towards ExecutionCatalogs and the PartiQLVM.

partiql-eval/src/main/java/org/partiql/eval/PartiQLVM.java
johnedquinn: Nope, the compiler sees a RexTableRef and goes: `catalogs[ref.catalogId].getTable(ref.tableId)`. Each of these calls allow for O(1) lookup.

partiql-spi/src/main/java/org/partiql/spi/catalog/ExecutionCatalog.java
johnedquinn: It is tied to the plan itself, so neither per-session nor per-catalog.

xd1313113 (approved):

PR #1934, opened 2026-06-22
Support Correlated Join

yliuuuu:
## Relevant Issues

## Description
Separate correlated joins from non-correlated joins in planner and eval. 
  
The planner now classifies joins into two categories at plan time:
  
  - Correlated (lateral): RHS references LHS bindings → Rel.Op.Correlate
  - Non-correlated: RHS is independent of LHS → Rel.Op.Join
  
The eval layer dispatches them to separate physical operators:
  
  - Correlate → lateral nested-loop (push LHS row into env, open RHS per row)
  - Join → non-lateral nested-loop (materialize RHS once, rescan per LHS row)

## Other Information
- Updated Unreleased Section in CHANGELOG: **[YES/NO]**: Yes.
- Any backward-incompatible changes? **[YES/NO]**: No. 
- Any new external dependencies? **[YES/NO]**: No. 
- Do your changes comply with the [contributing][cg] and [code style][csg] guidelines? **[YES/NO]**: Yes. 

## License Information

### DIFF
--- a/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelOpCorrelateLeft.kt
+++ b/partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelOpCorrelateLeft.kt
@@ -1,0 +1,84 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file.
+ */
+
+package org.partiql.eval.internal.operator.rel
+
+import org.partiql.eval.Environment
+import org.partiql.eval.ExprRelation
+import org.partiql.eval.Row
+import org.partiql.eval.internal.helpers.checkInterrupted
+import org.partiql.plan.rel.RelType
+import org.partiql.spi.value.Datum
+
+/**
+ * Correlated (lateral) left outer join. The RHS is opened per LHS row with the LHS row pushed into the environment.
+ * LHS rows with no RHS match are preserved with NULL-padded RHS.
+ *
+ * Algorithm:
+ * ```
+ * for lhsRecord in lhs:
+ *   matched = false
+ *   for rhsRecord in rhs(lhsRecord):
+ *     matched = true
+ *     yield(lhsRecord + rhsRecord)
+ *   if (!matched):
+ *     yield(lhsRecord + NULL_RECORD)
+ * ```
+ */
+internal class RelOpCorrelateLeft(
+    private val lhs: ExprRelation,
+    private val rhs: ExprRelation,
+    rhsType: RelType,
+) : RelOpPeeking() {
+
+    private val rhsPadded = Row(
+        rhsType.getFields().map { Datum.nullValue(it.type) }.toTypedArray()
+    )
+
+    private lateinit var env: Environment
+    private lateinit var iterator: Iterator<Row>
+
+    override fun openPeeking(env: Environment) {
+        this.env = env
+        lhs.open(env)
+        iterator = implementation()
+    }
+
+    override fun peek(): Row? {
+        return when (iterator.hasNext()) {
+            true -> iterator.next()
+            false -> null
+        }
+    }
+
+    override fun closePeeking() {
+        lhs.close()
+        rhs.close()
+        iterator = emptyList<Row>().iterator()
+    }
+
+    private fun implementation() = iterator {
+        for (lhsRecord in lhs) {
+            var lhsMatched = false
+            rhs.open(env.push(lhsRecord))
+            for (rhsRecord in rhs) {
+                checkInterrupted()
+                lhsMatched = true
+                yield(lhsRecord.concat(rhsRecord))
+            }
+            rhs.close()
+            if (!lhsMatched) {
+                yield(lhsRecord.concat(rhsPadded))
+            }
+        }
+    }
+}
--- a/partiql-eval/src/main/kotlin/org/partiql/eval/internal/plan/PRel.kt
+++ b/partiql-eval/src/main/kotlin/org/partiql/eval/internal/plan/PRel.kt
@@ -15,6 +15,7 @@ internal sealed class PRel {
     data class Filter(val input: PRel, val predicate: PExpr, override val type: RelType? = null) : PRel()
     data class Project(val input: PRel, val projections: List<PExpr>, override val type: RelType? = null) : PRel()
     data class Join(val lhs: PRel, val rhs: PRel, val condition: PExpr, val joinType: PJoinType, override val type: RelType? = null) : PRel()
+    data class Correlate(val lhs: PRel, val rhs: PRel, val joinType: PJoinType, override val type: RelType? = null) : PRel()
     data class Sort(val input: PRel, val collations: List<PCollation>, override val type: RelType? = null) : PRel()
     data class Distinct(val input: PRel, override val type: RelType? = null) : PRel()
     data class Limit(val input: PRel, val limit: PExpr, override val type: RelType? = null) : PRel()
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt
@@ -329,6 +329,16 @@ internal class PlanTransform(private val flags: Set<PlannerFlag>, private val us
             return operators.join(lhs, rhs, condition, joinType)
         }
 
+        override fun visitRelOpCorrelate(node: IRel.Op.Correlate, ctx: PType): Any {
+            val lhs = visitRel(node.lhs, ctx)
+            val rhs = visitRel(node.rhs, ctx)
+            val joinType = when (node.type) {
+                IRel.Op.Correlate.Type.INNER -> JoinType.INNER()
+                IRel.Op.Correlate.Type.LEFT -> JoinType.LEFT()
+            }
+            return operators.correlate(lhs, rhs, joinType)
+        }
+
         override fun visitRelOpExclude(node: IRel.Op.Exclude, ctx: PType): Any {
             val input = visitRel(node.input, ctx)
             val paths = node.paths.mapNotNull { visitRelOpExcludePath(it, ctx) }
--- a/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
+++ b/partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
@@ -192,6 +192,70 @@ internal class PlanTyper(private val env: Env, config: Context, private val flag
     }
 
     /**
+     * Returns true if [node] contains any Rex.Op.Var.Local that references the join's LHS scope.
+     *
+     * We want to be able to assert on whether a variable introduced in LHS is referenced in RHS.
+     * For example:
+     *  1. Direct path reference: `FROM t, t.items`.
+     *  2. Subquery `FROM t INNER JOIN (SELECT ... FROM ... WHERE t.n > ...)
+     *
+     * Recall that RelTyper has a `List<Scope>` field, representing all scopes visible from enclosing contexts.
+     * We call the RelType's `List<Scope>` field `outer`.
+     *
+     * Now consider `FROM LHS, RHS`, LHS introduce a new Scope, `Scope(lhs.type.schema, outer)`,
+     * and when resolving the RHS, the environment we used are `outer + lhsScope`.
+     *
+     * It is entirely possible that RHS itself introduces additional scope (i.e, Subquery).
+     * To measure the scope introduced between the RHS root and node v, we defined a function `nesting`,
+     * nesting(v): For a node v in the typed RHS tree, defined recursively:
+     *  - nesting(RHS root) = 0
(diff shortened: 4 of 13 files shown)


partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/CorrelatedJoinClassificationTest.kt
```diff
@@ -0,0 +1,269 @@
...
+        for (child in node.children) {
+            val found = findFirstJoinOrCorrelate(child)
+            if (found != null) return found
+        }
+        return null
+    }
+
+    private fun assertCorrelate(stmt: Statement, msg: String) {
+        val op = findFirstJoinOrCorrelate(stmt)
+        assert(op is Rel.Op.Correlate) { "$msg — expected Correlate but got ${op?.javaClass?.simpleName}" }
```
xd1313113, line 68: Are we able to assert the operand with nested level

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/CorrelatedJoinClassificationTest.kt
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    private fun assertJoin(stmt: Statement, msg: String) {
+        val op = findFirstJoinOrCorrelate(stmt)
+        assert(op is Rel.Op.Join) { "$msg — expected Join but got ${op?.javaClass?.simpleName}" }
```
xd1313113, line 73: Same

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt
```diff
@@ -329,6 +329,16 @@ internal class PlanTransform(private val flags: Set<PlannerFlag>, private val us
             return operators.join(lhs, rhs, condition, joinType)
         }
 
+        override fun visitRelOpCorrelate(node: IRel.Op.Correlate, ctx: PType): Any {
```
xd1313113, line 332: Can you create a test in sqldialecttest.kt to see you if you need any change there to support Scribe

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -191,6 +191,70 @@ internal class PlanTyper(private val env: Env, config: Context, private val flag
...
+     *
+     * It is entirely possible that RHS itself introduces additional scope (i.e, Subquery).
+     * To measure the scope introduced between the RHS root and node v, we defined a function `nesting`,
+     * nesting(v): For a node v in the typed RHS tree, defined recursively:
+     *  - nesting(RHS root) = 0
+     *  - nesting(v) = nesting(parent) + 1, if parent introduced an additional scope.
+     *  - nesting(v) = nesting(parent) otherwise.
+     *
+     * We claim: A Rex.Op.Var.Local(depth, ref) at node v in the RHS references the LHS scope
+     * iff depth == nesting(v) + 1.
```
xd1313113, line 216: nit: type `iff`

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rel/RelOpCorrelateLeft.kt
```diff
@@ -0,0 +1,84 @@
...
+
+import org.partiql.eval.Environment
+import org.partiql.eval.ExprRelation
+import org.partiql.eval.Row
+import org.partiql.eval.internal.helpers.checkInterrupted
+import org.partiql.plan.rel.RelType
+import org.partiql.spi.value.Datum
+
+/**
+ * Correlated (lateral) left outer join. The RHS is opened per LHS row with the LHS row pushed into the environment.
```
xd1313113, line 23: Do you consider use RelOpLateral rather RelOpCorrelate? Correlate join may be confused with Correlate subquery?

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/CorrelatedJoinClassificationTest.kt
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    @Test
+    fun `non-correlated - subquery references outer query not LHS`() {
+        // The subquery inside the join references t1 from the enclosing SELECT, NOT the join's LHS 'a'.
+        // This exercises nesting > 0: inside the subquery, t1 is at depth=3 and nesting=1,
+        // so depth != nesting + 1 (3 != 2), correctly classified as non-correlated.
+        val stmt = plan(
+            "SELECT VALUE (SELECT VALUE x FROM << 1, 2 >> AS a INNER JOIN (SELECT VALUE i FROM << 1, 2, 3 >> AS i WHERE i <= t1) AS x ON true) FROM << 2 >> AS t1"
+        )
```
xd1313113, line 107: A good one, this one is correlated subquery. We may should create plan for CorrelatedSubquery in the future as well.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/CorrelatedJoinClassificationTest.kt
```diff
@@ -0,0 +1,269 @@
...
+
+    @Test
+    fun `correlated - nested subquery both referencing LHS`() {
+        // Two levels of subqueries, both referencing t.n from the join's LHS.
+        // Outer subquery: nesting=1, depth=2 → 2 == 1+1 ✓
+        // Inner subquery: nesting=2, depth=3 → 3 == 2+1 ✓
+        val stmt = plan(
+            "SELECT VALUE x FROM << {'n': 3} >> AS t INNER JOIN (SELECT VALUE (SELECT VALUE j FROM << 1 >> AS j WHERE j <= t.n) FROM << 1, 2, 3 >> AS i WHERE i <= t.n) AS x ON true"
+        )
+        assertCorrelate(stmt, "Nested subqueries both referencing LHS at different nesting levels")
```
xd1313113, line 166: what about skip level reference in nested subquery?

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/CorrelatedJoinClassificationTest.kt
```diff
@@ -0,0 +1,269 @@
...
+        assert(op.rhs.op is Rel.Op.Filter) {
+            "Expected RHS to be Filter (condition always pushed) but got ${op.rhs.op.javaClass.simpleName}"
+        }
+    }
+
+    @Test
+    fun `correlated - comma join no ON condition`() {
+        // FROM t, t.items AS x — no explicit ON clause. Parser generates ON TRUE implicitly.
+        val stmt = plan("SELECT VALUE x FROM << {'items': [1, 2]} >> AS t, t.items AS x")
+        assertCorrelate(stmt, "Comma lateral join with no ON condition")
```
xd1313113, line 209: Does comma join plan as INNER join on true? Should it still include filter

yliuuuu: Whether a join a correlated depends on if the RHS reference variable from LHS. 

`t1, t2` -> Non Correlated Join , Inner
`t1, t1.p` -> Correlated, Inner. 

For Correlated join, a filter is moved on top of the RHS.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/CorrelatedJoinClassificationTest.kt
yliuuuu: If we define a correlatedSubquery as a query that refers columns from the outer query, not necessary in the FROM clause, i.e., 

```
SELECT c.customer_name,
       (SELECT o.order_date 
        FROM orders o 
        WHERE o.customer_id = c.customer_id 
        ORDER BY o.order_date DESC LIMIT 1) as latest_order
FROM customers c;
```

Then at planning time it is always possible to rewrite the query expression to lateral join. 

```
SELECT c.customer_name, o.order_date
FROM customers c
LEFT JOIN LATERAL (
    SELECT order_date 
    FROM orders 
    WHERE customer_id = c.customer_id  -- Explicit correlation allowed here
    ORDER BY order_date DESC 
    LIMIT 1
) o ON TRUE;
```

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
yliuuuu: If and only if.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/CorrelatedJoinClassificationTest.kt
yliuuuu: We might able to do something like this: 

```
val op = findFirstJoinOrCorrelate(stmt)
        assert(op is Rel.Op.Correlate) { "Outermost join should be Correlate (x refs t)" }
        // The inner join should be Join (a is independent)
        val innerOp = (op as Rel.Op.Correlate).lhs.op
        assert(innerOp is Rel.Op.Join) { "Inner join should be Join (a is independent) but got ${innerOp.javaClass.simpleName}" }
```

Let me know if there is a better way to assert on the plan in the code base.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/PlanTransform.kt
yliuuuu: We most certainly would need changes to support Scribe, but as we now actively uses RelOpCorrelated which is not in use before. 

To the printing functionality, i.e., 

```
public fun AstNode.sql(
    layout: SqlLayout = SqlLayout.STANDARD,
    dialect: SqlDialect = SqlDialect.STANDARD,
): String = dialect.transform(this).sql(layout)
```

This takes in an `AstNode`, which is not touched in this PR.

xd1313113 (dismissed):

partiql-eval/src/main/kotlin/org/partiql/eval/internal/plan/PRel.kt
```diff
@@ -15,6 +15,7 @@ internal sealed class PRel {
     data class Filter(val input: PRel, val predicate: PExpr, override val type: RelType? = null) : PRel()
     data class Project(val input: PRel, val projections: List<PExpr>, override val type: RelType? = null) : PRel()
     data class Join(val lhs: PRel, val rhs: PRel, val condition: PExpr, val joinType: PJoinType, override val type: RelType? = null) : PRel()
+    data class Correlate(val lhs: PRel, val rhs: PRel, val joinType: PJoinType, override val type: RelType? = null) : PRel()
```
johnedquinn, line 18: I believe a better name would be "Lateral". Also, condition should be supported. Though I do see that RelCorrelate already exists as a public node -- so can disregard.

partiql-planner/src/main/resources/partiql_plan_internal.ion
```diff
@@ -359,6 +359,16 @@ rel::{
       ],
     },
 
+    // Correlated (lateral) join. The RHS references bindings from the LHS.
```
johnedquinn, line 362: While this is an internal API, why not just an is_correlated boolean on the join? The diff on this PR would've been quite small.

johnedquinn (commented):
Is there a reason driving these changes? I am in full support for separating out lateral joins vs not due to the performance impact, but didn't see the context.

Also, since we will now be producing RelCorrelate nodes, will this have an effect on our existing customer's code when we go to release?

partiql-eval/src/main/kotlin/org/partiql/eval/internal/plan/PRel.kt
yliuuuu: Condition is achieved by moving filter on top of the RHS. 

```
   Correlate (INNER)
   ├── LHS: Scan(t)
   └── RHS: Filter(x.id = t.id)        ← ON condition pushed here
            └── Scan(t.items AS x)      ← references LHS binding `
```

```
   Join (INNER, condition: x.val = y.val)   ← ON condition lives here
   ├── LHS: Scan(t AS x)
   └── RHS: Scan(s AS y)                    ← independent of LHS
```

partiql-planner/src/main/resources/partiql_plan_internal.ion
yliuuuu: I was think about this.

1. The code gen does not produce a default . i.e., isLateral = false. Which leads to slightly more versbosed changed. 

2. It is intentionally that we don't have a condition on correlated node. So having a boolean flag requires slightly more complicated rewrite later.

partiql-planner/src/test/kotlin/org/partiql/planner/internal/typer/CorrelatedJoinClassificationTest.kt
johnedquinn, line 1: I believe all of these use explicit references to the LHS binding. Can you add a test that does the following:
```
SELECT a, b FROM << { 'a': 1 } >> t, a AS b
```

Notably, `a` on the RHS. PLK today outputs:
```
<<
  {
    'a': 1,
    'b': 1
  }
>>
```

johnedquinn: From your implementation in PlanTyper, it looks fine. But just want to make sure here.

johnedquinn (commented):
LGTM! Can approve after last comment.

johnedquinn (approved):

PR #1935, opened 2026-06-24
Multiple Map Bug fixes

xd1313113:
## Relevant Issues

### Summary

  Bug fixes and enhancements for MAP type operations — key type safety, null/missing semantics, planner typing, and conformance tests.

- Related conformance tests PR partiql/partiql-tests#169

### Changes
  
  - Fix MAP null key propagation: bracket access with NULL key now returns NULL (three-value logic), consistent with map_get and map_contains_key
  - Add implicit key type coercion for MAP lookup operations (map_get, map_contains_key, bracket access) — compatible numeric types are implicitly cast, incompatible types
  return MISSING (permissive) or error (strict)
  - Fix IS MAP type check to use default isNullCall/isMissingCall behavior
  - Separate MAP constructor into permissive/strict mode classes:
    - Permissive: drops NULL/MISSING keys, last-write-wins for duplicates
    - Strict: returns NULL/MISSING for NULL/MISSING keys, throws MAP_DUPLICATE_KEY error for duplicates
  - Add MAP_DUPLICATE_KEY and MAP_KEY_TYPE_MISMATCH error codes
  - Add CoercionFamily helper for determining type compatibility in MAP key lookups
  - Add conformance tests for MAP equality, ORDER BY, GROUP BY, and MISSING IS MAP

## Other Information
- Updated Unreleased Section in CHANGELOG: **[YES/NO]**: NO
- Do your changes comply with the [contributing][cg] and [code
style][csg] guidelines? **[YES/NO]**:YES

## License Information


partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
```diff
@@ -204,7 +132,9 @@ internal class ExprCallDynamic(
                 }
                 val argType = arg.type
                 val paramType = function.signature.parameters[i]
-                when (paramType.type == argType) {
+                // Skip cast for MAP params — function signatures default to MAP<STRING, DYNAMIC>,
+                // but the actual map may have non-string keys. The function body handles key casting.
+                when (paramType.type.code() == PType.MAP || paramType.type == argType) {
```
johnedquinn: This might be the wrong place for this escape hatch. From what I remember, `FnOverload` gives you the ability to provide a `getInstance`, which can be used by function providers to give an exact function implementation for a requested set of arguments. As for the `RoutineOverloadSignature`, I believe it's only used for sorting.

So, the special handling of MAP doesn't need to be distributed across the codebase. It can be specially handled by the few functions that use it.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
```diff
@@ -275,23 +276,27 @@ internal class PlanTyper(private val env: Env, config: Context, private val flag
...
             // Check Root (Dynamic)
             if (rex.type.code() == PType.DYNAMIC || rex.type.code() == PType.VARIANT) {
                 val type = ctx!!.copyWithSchema(listOf(kType, PType.dynamic()).toCType())
                 return rel(type, op)
             }
 
-            // Check Root
+            val resolvedKeyType = when (rex.type.code()) {
+                PType.MAP -> rex.type.keyType
+                else -> kType
```
johnedquinn: Wouldn't this be a string?

partiql-spi/src/main/kotlin/org/partiql/spi/internal/SqlTypeFamily.kt
```diff
@@ -41,6 +41,11 @@ internal class SqlTypeFamily private constructor(
             PType.INTEGER, PType.TINYINT, PType.SMALLINT, PType.BIGINT, PType.REAL, PType.DOUBLE, PType.DECIMAL, PType.NUMERIC -> NUMBER
             PType.CHAR, PType.VARCHAR, PType.STRING, PType.CLOB -> TEXT
             PType.BAG, PType.ARRAY -> COLLECTION
+            PType.BOOL -> BOOLEAN
+            PType.DATE, PType.TIMESTAMP, PType.TIMESTAMPZ -> DATE_TIMESTAMP
```
johnedquinn: This is used for implicit coercions right?

"A datetime is assignable to a site only if the source and target of the assignment are both of type DATE,
or both of type TIME (regardless whether WITH TIME ZONE or WITHOUT TIME ZONE is specified or
implicit), or both of type TIMESTAMP (regardless whether WITH TIME ZONE or WITHOUT TIME ZONE
is specified or implicit)." -- SQL:2023

johnedquinn: I believe the same is said for year-month and day-time intervals. But the timezone stuff can be shared.

test/partiql-tests
johnedquinn, line 1: Did you mean to include conformance tests here?

xd1313113: yes, please help review this PR for conformance test.
https://github.com/partiql/partiql-tests/pull/169/changes

Tests are created based on RFC
https://github.com/partiql/partiql-lang/pull/104/changes

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
xd1313113: unpivot struct will produce {k:<string>, v:{any}}
unpivot map will produce {k:<keytype>, v:{valuetype}}

partiql-spi/src/main/kotlin/org/partiql/spi/internal/SqlTypeFamily.kt
xd1313113: I am not using this file, but seems some issue with this file. I will fix in different PR.

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprCallDynamic.kt
xd1313113: you are right.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMapContainsKey.kt
```diff
@@ -16,19 +17,28 @@ internal object FnMapContainsKey : FnOverload() {
...
                 Parameter("map", mapType),
-                Parameter("key", keyType),
+                Parameter("key", if (compatible) mapKeyType else argKeyType),
             ),
         ) { fnArgs ->
             val map = fnArgs[0]
             val key = fnArgs[1]
-            Datum.bool(map.containsKey(key))
+            if (!compatible) {
+                Datum.bool(false)
```
XuechunHHH: Incompatible key type → `contains_key` returns `false`, but `map_get` throws. Should these be consistent?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/typer/PlanTyper.kt
XuechunHHH: But for struct the `k` is always a string as it is field names. So `else -> kType` where `kType = PType.dynamic()` is overly broad. Shouldn't it be `PType.STRING `-> `PType.string()` for `STRUCT/ROW`?

xd1313113: yeah, by this far, it should be string

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMapContainsKey.kt
xd1313113: in contains_key, I think the map should be more permissive than map_get, 
right map[key] are same as map_get, and assume the key is present and error if not found.

Likely we should introduce another function map_get_or_default to allow more permissive get

XuechunHHH: Permissive mode is fine. `map_get` throws → `ExprPermissive` catches it → `MISSING`, consistent with `contains_key` returning false.

 But in strict mode:
```
  -- m is MAP<INTEGER, STRING>
  map_contains_key(m, 'hello');  -- returns false (no error)
  map_get(m, 'hello');           -- throws MAP_KEY_TYPE_MISMATCH
```
Both are type errors in strict mode. Should `contains_key` also throw` MAP_KEY_TYPE_MISMATCH` in strict? Returning false for a type error feels like permissive behavior leaking into strict mode.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/executor/VMEvalExecutor.kt
```diff
@@ -130,6 +136,26 @@ class VMEvalExecutor(
...
+            "decimal" -> PType.decimal(38, 19)
+            "real" -> PType.real()
+            "double" -> PType.doublePrecision()
+            "string" -> PType.string()
+            "char" -> PType.character(255)
+            "varchar" -> PType.varchar(255)
+            "date" -> PType.date()
+            "time" -> PType.time(6)
+            "timestamp" -> PType.timestamp(6)
+            else -> PType.dynamic()
```
XuechunHHH: why not error out here?

partiql-eval/src/main/kotlin/org/partiql/eval/internal/operator/rex/ExprPathIndexMap.kt
```diff
@@ -22,8 +23,12 @@ internal class ExprPathIndexMap(
         if (input.isMissing || k.isMissing) {
             return Datum.missing(input.type.valueType)
         }
+
         val mapKeyType = input.type.keyType
         val castKey = if (k.type.code() != mapKeyType.code()) {
+            if (!CoercionFamily.canCoerce(k.type.code(), mapKeyType.code())) {
+                throw PErrors.mapKeyTypeMismatchException(k.type, mapKeyType)
+            }
             CastTable.cast(k, mapKeyType)
```
XuechunHHH, line 32: When `canCoerce` returns true, we cast unconditionally. But for numeric-to-numeric narrowing this can silently return the wrong entry:
  - DECIMAL 1.5 → INTEGER rounds to 2 (HALF_EVEN): lookup returns value for key 2, not key 1.5
  - DOUBLE 1.9 → INTEGER truncates to 1: different rounding than DECIMAL
  - DOUBLE → SMALLINT has no bounds check (silent overflow)

Same applies to `ExprPathKeyMap`, double check.

xd1313113: This looks fine as in RFC, we stated that map is strong typed and implicit cast is applied when key type does not match in look up. The user need to be aware of this cast issue.

For `map{1:'one', 2:'two'}`
```
map[1] => 'one'
map[1.2] => 'one'
map[1.5] => 'two'.  

```

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMapContainsKey.kt
xd1313113: We cannot easily throw MAP_KEY_TYPE_MISMATCH in strict and return false in Permissive mode.

For both functions, we actually allow plan with  incompatible key type and not considered as planning error for now in both modes

As for map_get,  the function itself cannot tolerate incompatible key and throw in the function body.  However, the evaluator will handle exception differently in two modes. 

For `contains_key`, since the function body does not have mode context and cannot behave differently in two mode. But the function can tolerate incompatible key.


If we really want to go that way, we need to handle that differently in planning. such contains_key_permissive, contains_key_strict.

test/partiql-tests-runner/src/test/kotlin/org/partiql/runner/executor/VMEvalExecutor.kt
xd1313113: yeah, right

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnMapContainsKey.kt
xd1313113: Changed to throw exception and create issue https://github.com/partiql/partiql-lang-kotlin/pull/1935 to potentially return false

XuechunHHH (dismissed):

XuechunHHH (approved):

PR #1947, opened 2026-07-27
fix: Prevent decimal division scale from exceeding precision

### LINKED ISSUE #1945: Decimal division can infer scale greater than precision and omit projected fields
## Description

`FnDivide.dividePrecisionScale` caps a decimal division result precision at 38, but caps its scale against the original uncapped precision. This can produce a decimal type whose scale exceeds its precision, such as `DECIMAL(38,41)`.

For ordinary ratio values, evaluation then raises `NUMERIC_VALUE_OUT_OF_RANGE`. In permissive mode the evaluator converts that error to `MISSING`, and the permissive struct constructor omits the projected field. The runtime row therefore has fewer fields than its planner-declared row type.

## To Reproduce

A minimal standalone query is:

```sql
SELECT
    x AS id,
    CAST(100000.00 AS DECIMAL(10,2)) /
        CAST(310000.00 AS DECIMAL(38,19)) AS ratio
FROM << 1 >> AS x;
```

Run the CLI in its default permissive mode with debug output enabled. The planner declares:

```text
BAG(ROW(id: INTEGER, ratio: DECIMAL(38, 41)))
```

but the result row contains only `id`; `ratio` is omitted. Running the same query in strict mode reports:

```text
Numeric value (0.32258064516129032258064516129032258064516) is out of range for type DECIMAL(38, 41).
```

The issue also occurs in aggregate ratio queries. Given this schema:

```sql
id      INTEGER
dept    VARCHAR(16)
salary  DECIMAL(10,2)
```

run:

```sql
SELECT
    id,
    dept,
    salary,
    salary / (
        SELECT SUM(salary)
        FROM salaries AS t2
        WHERE t2.dept = t1.dept
    ) AS ratio
FROM salaries AS t1
ORDER BY id;
```

`SUM(DECIMAL)` uses the default result type `DECIMAL(38,19)`, so division receives:

```text
lhs = DECIMAL(10,2)
rhs = DECIMAL(38,19)
```

The current formulas produce:

```text
p = p1 - s1 + s2 + max(6, s1 + p2 + 1)
  = 10 - 2 + 19 + max(6, 2 + 38 + 1)
  = 68

s = max(6, s1 + p2 + 1)
  = 41
```

The implementation then returns `DECIMAL(38,41)`.

## Expected Behavior

Decimal division must return a valid result type whose scale does not exceed its precision. The ratio should evaluate successfully for representable values, and a projected ratio field must not disappear in permissive mode because of an internally invalid result type.

## Root Cause

The affected code is in `FnDivide.dividePrecisionScale`:

```kotlin
val returnedP = p.coerceAtMost(38)
val returnedS = s.coerceAtMost(p)
return returnedP to returnedS
```

`returnedS` is bounded by the uncapped `p`, rather than by the returned precision. `PTypeDecimal` does not validate the relationship between precision and scale, so the invalid type survives planning and reaches evaluation.

A minimal invariant-preserving correction is:

```kotlin
val returnedP = p.coerceAtMost(38)
val returnedS = s.coerceAtMost(returnedP)
return returnedP to returnedS
```

A fully SQL Server-compatible precision-reduction policy may require preserving integer digits and reducing scale more aggressively. Regardless of policy, the returned type must satisfy `scale <= precision`.

## Runtime Failure Chain

1. `FnDivide` declares and computes a `DECIMAL(38,41)` result.
2. `Datum.decimal` rounds to scale 41 and rejects the result with `NUMERIC_VALUE_OUT_OF_RANGE` when its precision exceeds 38.
3. `ExprPermissive` converts that error to `MISSING`.
4. `ExprStructPermissive` omits fields whose value is `MISSING`.
5. The runtime result row no longer matches the planner-declared row type.

## Additional Context

- Verified at commit `3674a024e1ee86042ebd29a9ee30b02cba4ac61b` on `main`.
- Java: Corretto 17.0.19.
- `./gradlew :partiql-cli:installDist` completed successfully before reproduction.

XuechunHHH:
## Relevant Issues
- Closes #1945

## Description
- Clamp DECIMAL and NUMERIC division result scale to the returned precision, preventing invalid result types such as `DECIMAL(38,41)`.
- Preserve the existing PartiQL division precision/scale formula and rounding behavior; this change does not introduce a new overflow-reduction policy.
- Return `DECIMAL(38,38)` for `DECIMAL(10,2) / DECIMAL(38,19)` instead of the invalid `DECIMAL(38,41)`, allowing representable ratios to remain in permissive projections.
- Add direct precision/scale boundary tests and an end-to-end correlated `SUM(DECIMAL)` regression covering both evaluator execution paths.

## Verification
- `./gradlew :partiql-spi:test --tests org.partiql.spi.function.builtins.FnDivideTest :partiql-eval:test --tests org.partiql.eval.internal.DivideTest`
- `./gradlew :partiql-spi:ktlintMainSourceSetCheck`

## Other Information

## License Information

### DIFF
--- a/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnDivide.kt
+++ b/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnDivide.kt
@@ -98,11 +98,14 @@ internal object FnDivide : DiadicArithmeticOperator("divide") {
 
     /**
      * SQL defines exact numeric result precision and scale as implementation-defined.
-     * Preserve PartiQL's existing division behavior, introduced in
-     * [PR #1651](https://github.com/partiql/partiql-lang-kotlin/pull/1651) using the
-     * [SQL Server formula](https://learn.microsoft.com/sql/t-sql/data-types/precision-scale-and-length-transact-sql):
+     * PartiQL's division behavior, introduced in
+     * [PR #1651](https://github.com/partiql/partiql-lang-kotlin/pull/1651), follows the
+     * [SQL Server precision and scale rules](https://learn.microsoft.com/sql/t-sql/data-types/precision-scale-and-length-transact-sql):
      * p = p1 - s1 + s2 + max(6, s1 + p2 + 1)
      * s = max(6, s1 + p2 + 1)
+     *
+     * If p exceeds 38, reduce the scale to preserve the integral part when at least
+     * the minimum scale of 6 remains. Otherwise, reduce the scale to 6.
      */
     private fun dividePrecisionScale(lhs: PType, rhs: PType): Pair<Int, Int> {
         val (p1, s1) = lhs.precision to lhs.scale
@@ -110,8 +113,14 @@ internal object FnDivide : DiadicArithmeticOperator("divide") {
         val p = p1 - s1 + s2 + Math.max(6, s1 + p2 + 1)
         val s = Math.max(6, s1 + p2 + 1)
         val returnedP = p.coerceAtMost(38)
-        val returnedS = s.coerceAtMost(returnedP)
-        return returnedP to returnedS
+        val integralDigits = p - s
+        val remainingScale = 38 - integralDigits
+        val returnedS = when {
+            p <= 38 -> s
+            remainingScale >= 6 -> s.coerceAtMost(remainingScale)
+            else -> 6
+        }
+        return returnedP to returnedS.coerceAtMost(returnedP)
     }
 
     override fun getRealInstance(realLhs: PType, realRhs: PType): Fn {
--- a/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnDivide.kt
+++ b/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnDivide.kt
@@ -99,15 +99,24 @@ internal object FnDivide : DiadicArithmeticOperator("divide") {
     /**
      * SQL Server: p = p1 - s1 + s2 + max(6, s1 + p2 + 1)
      * SQL Server: s = max(6, s1 + p2 + 1)
+     *
+     * If p exceeds 38, reduce the scale to preserve the integral part when it has fewer than 32 digits.
+     * Otherwise, reduce the scale to 6.
      */
     private fun dividePrecisionScale(lhs: PType, rhs: PType): Pair<Int, Int> {
         val (p1, s1) = lhs.precision to lhs.scale
         val (p2, s2) = rhs.precision to rhs.scale
         val p = p1 - s1 + s2 + Math.max(6, s1 + p2 + 1)
         val s = Math.max(6, s1 + p2 + 1)
-        val returnedP = p.coerceAtMost(38)
-        val returnedS = s.coerceAtMost(p)
-        return returnedP to returnedS
+        if (p <= 38) {
+            return p to s
+        }
+        val integralDigits = p - s
+        val returnedS = when {
+            integralDigits < 32 -> s.coerceAtMost(38 - integralDigits)
+            else -> 6
+        }
+        return 38 to returnedS
     }
 
     override fun getRealInstance(realLhs: PType, realRhs: PType): Fn {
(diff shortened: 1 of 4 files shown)


partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnDivide.kt
```diff
@@ -99,15 +99,23 @@ internal object FnDivide : DiadicArithmeticOperator("divide") {
...
         val (p2, s2) = rhs.precision to rhs.scale
         val p = p1 - s1 + s2 + Math.max(6, s1 + p2 + 1)
         val s = Math.max(6, s1 + p2 + 1)
         val returnedP = p.coerceAtMost(38)
-        val returnedS = s.coerceAtMost(p)
-        return returnedP to returnedS
+        val integralDigits = p - s
+        val returnedS = when {
+            p <= 38 -> s
+            integralDigits < 32 -> s.coerceAtMost(38 - integralDigits)
```
xd1313113: seems 32 include implicit calculation.  may be let remainingScale = 38 - integralDigits, then  remainingScale >= 6 -> s.coerceAtMost(remainingScale)

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnDivide.kt
```diff
@@ -99,15 +99,23 @@ internal object FnDivide : DiadicArithmeticOperator("divide") {
...
      * SQL Server: p = p1 - s1 + s2 + max(6, s1 + p2 + 1)
      * SQL Server: s = max(6, s1 + p2 + 1)
+     *
+     * If p exceeds 38, reduce the scale to preserve the integral part when it has fewer than 32 digits.
+     * Otherwise, reduce the scale to 6.
      */
     private fun dividePrecisionScale(lhs: PType, rhs: PType): Pair<Int, Int> {
         val (p1, s1) = lhs.precision to lhs.scale
         val (p2, s2) = rhs.precision to rhs.scale
         val p = p1 - s1 + s2 + Math.max(6, s1 + p2 + 1)
```
xd1313113, line 113: Seems SQL spec mentioned precision and scale are implemented defined. Better to reference the source this algorithm if you have context.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnDivide.kt
XuechunHHH: I see. The original division behavior followed SQL Server. This is also from SQL Server but I shouldn't  introduce the semantics drift, will keep as `remainingScale = 38 - integralDigits`

XuechunHHH: The latest revision removes the integralDigits/remainingScale policy entirely to avoid changing existing division semantics. The fix now only clamps `s` to `returnedP`, so previously valid precision and scale results remain unchanged.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnDivide.kt
XuechunHHH: Added the provenance in b19264b1c: the formula was introduced in PR #1651 and was based on the linked SQL Server formula. The comment also clarifies that SQL defines exact-numeric precision and scale as implementation-defined; this PR does not alter that existing formula.

partiql-spi/src/test/kotlin/org/partiql/spi/function/builtins/FnDivideTest.kt
```diff
@@ -0,0 +1,120 @@
...
+        val expectedType: PType,
+        val expectedValue: BigDecimal,
+    ) {
+        override fun toString(): String = name
+    }
+
+    companion object {
+
+        @JvmStatic
+        fun divisionCases(): List<DivisionCase> = listOf(
```
xd1313113, line 72: seems tests are not comprehensive to cover integral part get truncated as well

XuechunHHH: Good point. Added integral-boundary coverage in 20bfa0958.

xd1313113 (dismissed):

xd1313113 (approved):

PR #1949, opened 2026-07-28
Add string function support for CHAR/VARCHAR

### LINKED ISSUE #1838: Add string functions support for `CHAR/VARCHAR`

### LINKED ISSUE #1950: Null propagation fails on string functions
## Description

`LOWER`, `UPPER`, `TRIM`, and `||` (`concat`) fail to *resolve* when an argument is an untyped `NULL` (or `MISSING`), raising a function-resolution error instead of propagating `NULL` as SQL requires.

**Root cause.** `FnResolver.match` calls `FnOverload.getInstance(args)` with the *full* argument array **before** it filters out `UNKNOWN` arguments. These four are dynamic `FnOverload`s that branch on the argument's type code and have no `PType.UNKNOWN` case, so they fall through to `else -> null` — which resolution reads as "no matching overload", and the query fails at plan time. The framework's `isNullCall` short-circuit in `ExprCall` never gets a chance to run, since it only applies once an instance has been resolved.

The text functions still declared as fixed-signature `Function.overload`s (`BIT_LENGTH`, `CHAR_LENGTH`, `OCTET_LENGTH`, `POSITION`, `SUBSTRING`, and the `TRIM_*` character-set variants) are unaffected — but any of them converted to a dynamic `getInstance` will hit the same trap unless it handles `UNKNOWN` explicitly.

## To Reproduce

```sql
LOWER(NULL);
UPPER(NULL);
TRIM(NULL);
'a' || NULL;
```

Each raises a function-resolution error rather than returning `NULL`.

## Expected Behavior

All four return `NULL`, per SQL's null-propagation rule for `<string value function>` (SQL2023 section 6.33): if any argument is the null value, the result is the null value.

## Additional Context

- PartiQL version: 1.5.1-SNAPSHOT (reproduces on 1.5.0 and earlier)
- Fix approach: accept `UNKNOWN` in `getInstance` and return a resolution-only `Fn` whose `returns` types the propagated null; the framework's default `isNullCall`/`isMissingCall` then short-circuits before the body is invoked.
- Being fixed alongside the `CHAR`/`VARCHAR` support work in #1949.

xd1313113:
## Relevant Issues
- Closes #1838
- Closes #1950

## Description
- Adds `CHAR`/`VARCHAR` support to the built-in text functions
- fixes `NULL` propagation for string functions

**Other fixes**
- `||` threw `Operation "getString" is not valid for type CLOB(n)` on byte-backed `CLOB` operands.
- `String.codepointSubstring`'s second parameter was named `end` but callers passed SQL's `FOR <length>`.

**Follow-ups**
- Unbounded `CLOB || CLOB` overflows in `addLengths` instead of clamping to `MAXLENGTH`.

## Other Information
- Any backward-incompatible changes? **NO**: queries that previously failed to resolve now succeed or return `NULL`; renamed helpers are `internal`.

## License Information


partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnBitLength.kt
```diff
@@ -1,35 +1,52 @@
...
+ */
+internal object FnBitLength : FnOverload() {
 
-    name = "bit_length",
-    returns = PType.integer(),
-    parameters = arrayOf(
-        Parameter("value", PType.string()),
-    ),
+    override fun getSignature(): RoutineOverloadSignature {
+        return RoutineOverloadSignature("bit_length", listOf(PType.dynamic()))
```
johnedquinn, line 22: I don't think we should use dynamic here. I believe we should put the preferred type here, for sorting purposes. Context: https://github.com/partiql/partiql-lang-kotlin/blob/3c494a7c561959d2c186a37461b663975e625feb/partiql-spi/src/main/java/org/partiql/spi/function/RoutineOverloadSignature.java#L55

Here's how we do the arithmetic diadic operators: https://github.com/partiql/partiql-lang-kotlin/blob/3c494a7c561959d2c186a37461b663975e625feb/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/DiadicArithmeticOperator.kt#L11

And diadic operators generally: https://github.com/partiql/partiql-lang-kotlin/blob/3c494a7c561959d2c186a37461b663975e625feb/partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/DiadicOperator.kt#L23

xd1313113: I traced how getSignature() and think Dynamic is the correct one.

- Candidates are pre-filtered only by arity (it.signature.arity == args.size) — never by parameter type ([FnResolver.kt:36](https://github.com/partiql/partiql-lang-kotlin/blob/3c494a7c561959d2c186a37461b663975e625feb/partiql-planner/src/main/kotlin/org/partiql/planner/internal/FnResolver.kt#L36)).
- FnResolver call getInstance(args) and inspect the returned instance's parameters, not the overload signature's parameterTypes
  ([FnResolver.kt:123-163](https://github.com/partiql/partiql-lang-kotlin/blob/3c494a7c561959d2c186a37461b663975e625feb/partiql-planner/src/main/kotlin/org/partiql/planner/internal/FnResolver.kt#L120-L163)).
- FnComparator precedence sorting — but only to break ties among multiple overloads sharing the same name ([FnComparator.kt:15-27]()). bit_length is registered as a single FnOverload (Builtins.kt:35, grouped by signature.name), so nothing is ever compared against it.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrimChars.kt
```diff
@@ -3,41 +3,50 @@
...
+        if (args.any { !it.isTextOrUnknown() }) return null
+        val valueType = args[0]
+        // If any argument is UNKNOWN (literal NULL), return a resolution-only instance; the framework's isNullCall handles propagation.
+        if (args.any { it.code() == PType.UNKNOWN }) {
+            return FnUtils.nullResolutionInstance(FunctionUtils.hide("trim_chars"), valueType.stringFnReturnType(), args)
+        }
+        return Function.instance(
+            name = FunctionUtils.hide("trim_chars"),
+            returns = valueType.stringFnReturnType(),
+            parameters = arrayOf(Parameter("value", valueType), Parameter("chars", PType.string())),
```
XuechunHHH: It looks like `isTextOrUnknown()` allows CLOB here, while the `chars` parameter is fixed to STRING. As far as I can tell, this causes the resolver to insert a CLOB-to-STRING cast that is not available at runtime. For example:

  ```sql
   TRIM(BOTH CAST('x' AS CLOB) FROM CAST('xxhixx' AS CLOB))
   ```
  fails with UNDEFINED_CAST. Since main previously had a CLOB/CLOB overload, could we preserve args[1] as the parameter type and read it with textValue(args[1])? The leading and trailing variants may need the same adjustment.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnSubstring.kt
```diff
@@ -78,55 +84,77 @@ import org.partiql.spi.value.Datum
...
+        // `value` must be a text type (or UNKNOWN, handled below); the integer args are fixed by the signature.
+        val valueType = args[0]
+        if (!valueType.isTextOrUnknown()) return null
+        // If any argument is UNKNOWN (literal NULL), return a resolution-only instance; the framework's isNullCall handles propagation.
+        if (args.any { it.code() == PType.UNKNOWN }) {
+            return FnUtils.nullResolutionInstance("substring", valueType.stringFnReturnType(), args)
+        }
+        return Function.instance(
+            name = "substring",
+            returns = valueType.stringFnReturnType(),
```
XuechunHHH: I may be missing the intended length rule here, but omitting the length makes CHAR/VARCHAR results default to `VARCHAR(255)`. A 300-character `VARCHAR(300)` passed to `substring(value, 1)` is therefore truncated to 255 characters.

Could we carry the source length through both `stringFnReturnType` and `stringFnResult`? The three-argument overload appears to have the same behavior, and a test above 255 characters would help cover it.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrimLeading.kt
```diff
@@ -3,33 +3,48 @@
...
+        // The argument must be a text type (or UNKNOWN, handled below); anything else does not match.
+        val valueType = args[0]
+        if (!valueType.isTextOrUnknown()) return null
+        // An UNKNOWN argument (literal NULL) gets a resolution-only instance; the framework's isNullCall handles propagation.
+        if (valueType.code() == PType.UNKNOWN) {
+            return FnUtils.nullResolutionInstance(FunctionUtils.hide("trim_leading"), valueType.stringFnReturnType(), args)
+        }
+        return Function.instance(
+            name = FunctionUtils.hide("trim_leading"),
+            returns = valueType.stringFnReturnType(),
```
XuechunHHH: It looks like this variant drops the source length, unlike `FnTrim`. 

Why not pass the source length to both `stringFnReturnType` and `stringFnResult`, following the approach in `FnTrim`? The trailing and character-set variants may need the same adjustment.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnConcat.kt
```diff
@@ -42,67 +40,40 @@ internal object FnConcat : FnOverload() {
...
     override fun getInstance(args: Array<PType>): Fn? {
+        // Every argument must be a text type (or UNKNOWN, handled below); anything else does not match.
+        if (args.any { !it.isTextOrUnknown() }) return null
         val lhsType = args[0]
         val rhsType = args[1]
-        // Check if both are string types
-        if (lhsType !in SqlTypeFamily.TEXT || rhsType !in SqlTypeFamily.TEXT) return null
+        // If any argument is UNKNOWN (literal NULL), return a resolution-only instance;
+        if (args.any { it.code() == PType.UNKNOWN }) {
+            return FnUtils.nullResolutionInstance(FunctionUtils.hide("concat"), lhsType, args)
```
XuechunHHH, line 59: I think this makes NULL handling depend on operand order. When the left side is UNKNOWN, the result remains UNKNOWN, while a known left side with an UNKNOWN right side keeps the known type. For example, `(NULL || CAST('a' AS VARCHAR(1))) LIKE 'a'` fails to resolve, although the reversed concatenation returns NULL.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnSubstring.kt
```diff
@@ -78,55 +84,77 @@ import org.partiql.spi.value.Datum
...
-        val end = args[2].int
-        if (end < 0) {
-            throw PErrors.internalErrorException(IllegalArgumentException("End must be non-negative."))
+    override fun getInstance(args: Array<PType>): Fn? {
+        // `value` must be a text type (or UNKNOWN, handled below); the integer args are fixed by the signature.
+        val valueType = args[0]
+        if (!valueType.isTextOrUnknown()) return null
+        // If any argument is UNKNOWN (literal NULL), return a resolution-only instance; the framework's isNullCall handles propagation.
+        if (args.any { it.code() == PType.UNKNOWN }) {
+            return FnUtils.nullResolutionInstance("substring", valueType.stringFnReturnType(), args)
```
XuechunHHH: It looks like passing the raw argument types here can let NULL propagation bypass validation. For example, `substring(NULL, TRUE)` creates a resolution-only `(UNKNOWN, BOOL)` instance and returns NULL instead of rejecting the Boolean start argument.

Could we retain the required INTEGER type for the start parameter in this instance? The three-argument overload may need the same adjustment for both start and length.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnCharLength.kt
```diff
@@ -1,36 +1,52 @@
...
+        if (inputType.code() == PType.UNKNOWN) {
+            return FnUtils.nullResolutionInstance("char_length", PType.integer(), args)
+        }
+        return when (inputType.code()) {
+            PType.CHAR, PType.VARCHAR, PType.STRING -> Function.instance(
+                name = "char_length",
+                returns = PType.integer(),
+                parameters = arrayOf(Parameter("value", inputType)),
+            ) { args ->
+                val value = args[0].string
```
XuechunHHH, line 39: One thing I noticed with the new VARCHAR path is that the backing value appears to be padded to its declared width. As a result, `char_length(CAST('a' AS VARCHAR(5)))` returns `5` rather than `1`.

I would expect VARCHAR to count only its actual content. Is the padding intentional here? If not, `octet_length` and `bit_length` appear to have the same issue, and tests where the declared width exceeds the content length would help.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrimChars.kt
xd1313113: yes, make sense

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnSubstring.kt
xd1313113: sure

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnTrimLeading.kt
xd1313113: All trim carry length from original now

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnConcat.kt
xd1313113: I am not very follow your comment, but they both return Unknown type. the difference is `NULL` as right operand will return a typed unknown.

NULL || 'a' = > Datum.null()
'a' || NULL => Datum.null(String)

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnSubstring.kt
xd1313113: Make sense

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnCharLength.kt
xd1313113: This is known bug. I just fix this PR as well.

partiql-spi/src/main/kotlin/org/partiql/spi/function/builtins/FnConcat.kt
XuechunHHH: Thanks, I agree that both expressions evaluate to NULL. My concern is the resolved return type, this branch always uses `lhsType`, so operand order remains observable.

On the latest commit, `(NULL || CAST('a' AS VARCHAR(1))) LIKE 'a'` fails with `Undefined function like(UNKNOWN, STRING)`, while the reversed concatenation resolves and returns NULL.

Shall we use the non-UNKNOWN operand as the return type when only one side is UNKNOWN, leaving UNKNOWN only when both operands are UNKNOWN? Not sure if this will be cleaner.

xd1313113: I think, typed null and untyped null both are ok. However, it looks like a bug  null propagation bug for `LIKE`. will fix it in next iteration

partiql-spi/src/main/java/org/partiql/spi/value/Datum.java
```diff
@@ -710,17 +710,10 @@ static Datum varchar(@NotNull String value) throws PRuntimeException {
...
             throw PErrors.wrappedException(new IllegalArgumentException("VARCHAR of length " + length + " not allowed."));
         }
-        if (value.length() < length) {
-            newValue = String.format("%-" + length + "." + length + "s", value);
-        } else if (value.length() == length) {
-            newValue = value;
-        } else {
-            newValue = value.substring(0, length);
-        }
+        String newValue = value.length() <= length ? value : value.substring(0, length);
```
XuechunHHH, line 716: One last thing, this changes the behavior of the public `Datum.varchar(value, length)` API: a value shorter than `length` is no longer space-padded.

I don't think any customer actually relies on this, but worth to call out in the changelog.

xd1313113: The issue is fixed in another PR #1951 . I merged and the change is gone now

XuechunHHH (approved):

PR #1958, opened 2026-08-25
Add PartiQL function extensions module

XuechunHHH:
## Relevant Issues
- Independent follow-up to #1954

## Description

This PR adds `partiql-function-extensions`, an optional Gradle module that is published separately from the existing PLK artifacts as:

```text
org.partiql:partiql-function-extensions:<version>
```

The module starts from an existing customer baseline of six functions: `contains`, `date_add`, `utcnow`, `to_unixtime`, `pow`, and `hex_to_bigint`. That baseline provided planner-layer function signatures only. This module publishes the same `FnOverload` signatures and also provides executable `Fn` bodies, so the functions can be shared by hosts for both planning and evaluation.

The library does not choose host catalogs, namespaces, aliases, or lookup precedence. Adding the artifact also does not register or expose any function automatically. A host explicitly selects the overload collections it wants and decides how to register them.

```mermaid
flowchart LR
    subgraph repo["partiql-lang-kotlin"]
        core["Existing PLK modules<br/>core functions"]
        spi["partiql-spi<br/>Fn and FnOverload APIs"]
        extensions["This PR<br/>partiql-function-extensions<br/>six executable function exports"]
        registration["PR #1954<br/>direct registration and exact lookup SPI"]
        spi --> extensions
    end

    host["Host application<br/>adds optional artifact<br/>selects overloads and SQL names"]
    runtime["Planner / evaluator / translator"]

    extensions -- "separately published artifact" --> host
    registration -- "explicit registration API" --> host
    core --> runtime
    host --> runtime
```

### Functions

| Family | Function | Signature |
| --- | --- | --- |
| Compatibility / datetime | `date_add` | `(STRING, INTEGER, TIMESTAMP) -> TIMESTAMP` |
| Compatibility / datetime | `utcnow` | `() -> TIMESTAMP` |
| Custom / collection | `contains` | `(ARRAY, STRING) -> BOOL` |
| Custom / conversion | `hex_to_bigint` | `(STRING) -> BIGINT` |
| Custom / datetime | `to_unixtime` | `(TIMESTAMP) -> BIGINT` |
| Custom / math | `pow` | Same-type numeric pairs from `TINYINT` through `DOUBLE` -> `DOUBLE` |

Compatibility functions preserve behavior associated with earlier PLK versions and link to the corresponding public PLK v0.14 source. Custom functions are grouped by domain. Every public export returns a stable, immutable overload collection.

### Scope

- Adds the separately published `partiql-function-extensions` module.
[...]

PR #1954, opened 2026-08-07
Add direct routine registration SPI

XuechunHHH:
## Relevant Issues
- Function implementation follow-up: #1958

## Description

This PR is the first PartiQL Lang Kotlin (PLK) change for a broader public custom-function extension workflow.

PLK continues to own its built-in core functions. We plan to add `partiql-function-extensions` as a separately published, optional Gradle module in this repository for additional host-neutral functions. Keeping it in the PLK repository provides shared build and release governance, while publishing it as a separate artifact keeps adoption explicit. The module will:

- contain reusable scalar and aggregate function implementations;
- depend only on the public `partiql-spi` API at compile time;
- export collections of existing `FnOverload` and `AggOverload` objects;
- contain no host-specific code, catalog name, SQL namespace, or internal package naming;
- remain optional, with no existing PLK module depending on it; and
- expose nothing automatically when a host adds the dependency.

The host application continues to depend on the PLK modules it uses and may add the optional extension artifact. It explicitly selects an allowlist of overloads, assigns their catalog-local SQL names, and configures lookup precedence.

For example:

```text
library exports:       pow
host registers:        math.pow
SQL-visible name:      <host-catalog>.math.pow
```

```mermaid
flowchart LR
    subgraph plk["partiql-lang-kotlin repository"]
        core["Existing PLK modules<br/>runtime and built-in core functions"]
        library["Future optional Gradle module<br/>partiql-function-extensions<br/>separately published overload collections"]
        current["This PR in partiql-spi<br/>direct registration and exact lookup"]
        next["Next PLK change<br/>session path, planner resolution, and RoutineRef"]
        current --> next
    end

    host["Host application<br/>adds the optional artifact<br/>selects functions and assigns SQL names"]
    consumer["Evaluator or translator<br/>uses the selected routine"]

    core --> host
    library -- "overload collections" --> host
    host -- "explicit registration" --> current
    next --> consumer
```

This PR contains only the **direct registration and exact lookup SPI** box. It does not add the optional function module, expose any new SQL function, or change planner resolution yet.

Earlier revisions proposed separate provider definition, loading, inventory, and mounting stages. Direct registration keeps the important boundary, where libraries do not choose host SQL names, with a smaller public API.

## This PR

- Adds `RoutineProvider`, a provider-neutral interface for exact catalog-local scalar and aggregate lookup.
- Adds `MemRoutineProvider.Builder` to register existing overloads at the catalog root, under a namespace, or at a complete catalog-local `Name`.
[...]

PR #1959, opened 2026-08-27
Add namespace-aware routine resolution

XuechunHHH:
## Relevant Issues
- Follow-up to #1954
- Related to #1958

## Description
This connects the direct routine registration SPI from #1954 to planner resolution for scalar and aggregate calls.

```mermaid
flowchart LR
    Call["Scalar or aggregate call"] --> Resolver["RoutineResolver"]
    Path["Session path and catalogs"] --> Resolver
    Resolver --> Exact["RoutineCatalog exact lookup"]
    Resolver --> Legacy["Legacy Catalog lookup"]
    Exact --> Plan["Typed plan with RoutineRef"]
    Legacy --> Plan
```

- Adds `Session.Builder.path(...)` and immutable `Path.promote(...)`.
- Resolves qualified calls by catalog and unqualified calls through the ordered session path.
- Validates custom catalog results and reports not-found and ambiguity diagnostics.
- Adds canonical `RoutineRef` identity to scalar calls, dynamic dispatch, and aggregate measures, preserving it through copies and rewrites.
- Keeps the existing `Operators` factory surface unchanged; rewrites attach identity after invoking custom factories.
- Keeps existing `Catalog` implementations on their legacy unqualified lookup path and preserves default session and evaluator behavior.
- Preserves collection-valued SQL subqueries for qualified and unqualified `EXISTS` and `SIZE` calls.
- Does not add extension functions; #1958 remains the separately published optional function artifact.

### Review Guide

Review this as one resolution pipeline rather than reading all 43 files in repository order. Of the 3,964 added lines,
2,591 are focused tests.

#### 1. Establish the session path contract

**Read:** `Session.kt`, `Path.kt`, `Catalogs.kt`, `StandardCatalogs.kt`, `System.java`, and `Builtins.kt` in `partiql-spi`.

Verify these invariants:

- Omitting `Session.Builder.path(...)` keeps the existing path: current catalog/current namespace followed by the system catalog.
- An explicit path replaces only the current catalog/current namespace entry. It is snapshotted, remains independent from table namespace resolution, and gets the system root appended only when absent.
- Every non-empty path entry starts with an exact catalog name followed by zero or more exact namespace levels; empty entries are skipped. User-written routine identifiers still retain their regular or delimited matching mode.
- `Path.promote(...)` is immutable, moves only the first exact match, and preserves all other ordering and duplicates.
- Case-insensitive catalog lookup reports all case-distinct matches through `CatalogNameAmbiguousException`; exact lookup remains unchanged.
- The system catalog implements `RoutineCatalog` by indexing the existing built-in overload objects through `MemRoutineProvider`. Its legacy `getFunctions` and `getAggregations` methods remain available.

**Focused tests:** `PathTest`, `SessionPathTest`, `CatalogsTest`, and `SystemRoutineCatalogTest`.

#### 2. Review name search and precedence

**Read:** `RoutineResolver.kt`, then the resolution methods in `Env.kt`.

Verify these decisions:

- A qualified call is always `<catalog>.<catalog-local name>`. The first part selects a catalog, qualification never falls back to `Session.path`, and partial qualification is not relative to the current catalog.
- An unqualified call checks path entries in order. `RoutineCatalog` receives an exact catalog-local `Identifier`; legacy `Catalog` receives the lowercased leaf through its existing APIs.
- `RoutineCatalog.resolveFunctions` and `resolveAggregations` are authoritative. Returning no binding does not fall back to that catalog's legacy getters.
- Scalar and aggregate candidates are discovered independently for the call shape. Finding both kinds is ambiguous, including when they come from different path entries.
- Type mismatch falls through to later path entries, but overloads from separate locations are never merged. The first location with a viable coercible match wins.
- Binding lookup is cached per identifier and scalar/aggregate arity so aggregate classification and later typing observe the same provider snapshot.
- `COUNT(*)` is classified with aggregate arity one while preserving scalar arity zero.

**Focused tests:** `RoutineResolverTest`, especially the qualified matching, path ordering, type-mismatch fallthrough,
scalar/aggregate collision, legacy lookup, authoritative exact lookup, cache, and `COUNT(*)` cases.

#### 3. Check the provider trust boundary and diagnostics

**Read:** validation in `RoutineResolver.kt`, `PErrors.kt`, `PError.java`, and the CLI formatter changes.

Verify that exact providers cannot inject inconsistent planner state:

- Each returned binding must match the complete requested identifier using each part's regular or delimited matching mode.
- Canonical names and overload collections must be non-empty; each overload leaf must match its signature name; duplicate parameter sequences are rejected.
- Provider callback exceptions and fatal JVM errors propagate unchanged. Failures encountered while traversing malformed provider metadata retain the cause and escape planning instead of becoming an internal query error.
- User ambiguity is reported as `FUNCTION_AMBIGUOUS` with stable candidate descriptions rather than `INTERNAL_ERROR`.

**Focused tests:** `RoutineResolverValidationTest`, the malformed-provider and exception cases in
`RoutineResolutionIntegrationTest`, `ErrorMessageFormatterTest`, and the updated evaluator `CatalogTest`.

#### 4. Follow planner classification and typing

**Read:** `AggregationTransform.kt`, `RelConverter.kt`, `RexConverter.kt`, `Env.kt`, `PlanTyper.kt`, and `SqlPlanner.kt`.

Trace one scalar call and one aggregate call:

- `AggregationTransform` uses the shared resolver to classify scalar versus aggregate before conversion.
- Converters preserve the full `Identifier`; they no longer reject qualified calls or lowercase names before resolution.
- `EXISTS` and `SIZE` retain collection-valued SQL `SELECT` arguments in both qualified and unqualified form; `TUPLEUNION` remains an unqualified syntax-only plan operator.
- `Env` uses the existing `FnResolver` and aggregate matcher after location selection, preserving coercion behavior and dynamic dispatch.
- `PlanTyper` maps not-found, type-mismatch, and ambiguity outcomes to semantic error nodes. Aggregate failures no longer survive as unresolved calls and later surface as internal errors.
- `COUNT(*)` injects its synthetic integer argument only after classification, while `DISTINCT`, casts, and return typing are preserved.
- `SqlPlanner` keeps the legacy catch-all behavior but lets exact-provider contract failures escape unchanged.

**Focused tests:** `RoutineResolutionIntegrationTest`, especially direct provider registration, qualified scalar and
aggregate calls, qualified collection arguments, dynamic dispatch, aggregate fallthrough, `DISTINCT`, `COUNT(*)`, and semantic error cases.

#### 5. Verify canonical plan identity is additive

**Read:** `RoutineRef.java`, `RexCall.java`, `RexDispatch.java`, `RelAggregate.java`, `OperatorRewriter.java`, then the
internal IR schema and `PlanTransform.kt`.

Verify these compatibility properties:

- `RoutineRef` contains only the canonical catalog and catalog-local binding name. It does not expose provider or artifact metadata, and it defensively copies `Name`.
- Identity is attached to static scalar calls, dynamic dispatch, and aggregate measures after exact resolution.
- Existing `Operators` implementations and custom plan subclasses remain valid. Rewriters invoke the existing factory methods first, then attach identity through internal wiring.
- Internal IR conversion, public plan conversion, copies, and operator rewrites preserve identity.
- Execution still uses the existing `Fn` and `Agg` objects; `RoutineRef` is metadata and does not change evaluator dispatch.

**Focused test:** `RoutineRefTest`.

#### 6. Perform the compatibility pass

| Scenario | Expected behavior |
| --- | --- |
| No explicit session path | Existing current catalog/current namespace and system lookup order |
| Catalog does not implement `RoutineCatalog` | Existing unqualified, lowercased getter lookup with no `RoutineRef` |
| Catalog implements `RoutineCatalog` | Exact methods are authoritative for qualified and path lookup |
| Qualified call misses | No fallback to an unqualified path entry |
| Table lookup | Continues to use current catalog/namespace, not the routine path |
| Manually constructed or legacy plan | Existing factory works and routine identity is `null` |
| Existing evaluator | Continues executing the resolved `Fn` or `Agg` instance |

The public API dumps are useful last: they should show only additive path methods, the catalog ambiguity exception/error code, `RoutineRef`, identity-aware node factories, and nullable identity getters. Rewrite-only setters remain internal, and the `Operators` method set remains unchanged.

## Other Information
- Any backward-incompatible changes? **NO**: Existing `Catalog` and `Operators` implementations, default session paths, plan constructors, and evaluator behavior remain supported.

## License Information


partiql-planner/src/main/kotlin/org/partiql/planner/internal/Env.kt
```diff
@@ -50,67 +70,134 @@ internal class Env(private val session: Session, internal val listener: PErrorLi
...
+                )
+            }
+            val selected = location.scalars.singleOrNull() ?: continue
+            if (selected.overloads.isEmpty()) {
+                continue
+            }
+            if (firstCandidates == null) {
+                firstCandidates = selected.overloads
+            }
+            val resolved = FnResolver.resolve(selected.overloads, args.map { it.type }) ?: continue
```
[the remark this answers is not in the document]
XuechunHHH: `RoutineCatalog` requires stable, thread-safe, immutable overloads for the catalog lifetime. A signature getter that changes or throws after validation violates that contract. Wrapping only `FnResolver` would be incomplete because later planning and evaluation retain the original overload object. I’m keeping the documented stability boundary rather than adding partial wrapper machinery.

partiql-plan/src/main/java/org/partiql/plan/rel/RelAggregate.java
```diff
@@ -101,11 +136,13 @@ public static class Measure {
         private final Agg agg;
         private final List<Rex> args;
         private final boolean distinct;
+        private final RoutineRef routineRef;
 
-        private Measure(Agg agg, List<Rex> args, boolean distinct) {
+        private Measure(Agg agg, List<Rex> args, boolean distinct, RoutineRef routineRef) {
```
xd1313113: Is `Measure(Agg agg, List<Rex> args, boolean distinct, RoutineRef routineRef)` supposed to replace `Measure(Agg agg, List<Rex> args, boolean distinct)`. If not, you can keep both constructor to avoid pass `Null` in constructor

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
```diff
@@ -463,27 +462,13 @@ internal object RelConverter {
...
-                    )
-                } else {
-                    val setq = when (expr.setq?.code()) {
-                        null -> org.partiql.planner.internal.ir.SetQuantifier.ALL
-                        SetQuantifier.ALL -> org.partiql.planner.internal.ir.SetQuantifier.ALL
-                        SetQuantifier.DISTINCT -> org.partiql.planner.internal.ir.SetQuantifier.DISTINCT
-                        else -> error("Unexpected SetQuantifier type: ${expr.setq}")
-                    }
-                    relOpAggregateCallUnresolved(name, setq, args)
+                val setq = when (expr.setq?.code()) {
```
xd1313113, line 465: Why special case handling is removed. Count(*) and count(column) is actually different.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
```diff
@@ -652,18 +652,15 @@ internal object RexConverter {
             val type = (ANY)
             // Fn
             val id = AstToPlan.convert(node.function)
-            if (id.hasQualifier()) {
-                error("Qualified function calls are not currently supported.")
-            }
-            if (id.matches("TUPLEUNION")) {
+            if (!id.hasQualifier() && id.matches("TUPLEUNION")) {
```
xd1313113, line 655: Does system default have any qualifier by any chance?

partiql-planner/src/main/kotlin/org/partiql/planner/internal/SqlPlanner.kt
```diff
@@ -34,18 +34,20 @@ internal class SqlPlanner(
...
-            val env = Env(session, ctx.errorListener)
+            val planningEnv = Env(session, ctx.errorListener)
+            env = planningEnv
 
             // 1. Normalize
             val ast = statement.normalize(ctx.errorListener)
 
             // 2. AST to Rel/Rex
-            val root = AstToPlan.apply(ast, env)
+            val root = AstToPlan.apply(ast, planningEnv)
```
xd1313113, line 47: Seems, no change required

partiql-plan/src/main/kotlin/org/partiql/plan/Operators.kt
```diff
@@ -280,6 +278,20 @@ public interface Operators {
...
 
+    /**
+     * Create a [RexCall] instance with resolved routine identity.
+     *
+     * @param function
+     * @param args
+     * @param routineRef
+     * @return
+     */
+    public fun call(
```
xd1313113: Add this method seems voilates `There is only ONE factory method per operators – i.e. no overloads or defaults.` and for downstream caller, once they upgrade, system becomes routineRef and they may ignore adding this override and cause a break.

xd1313113 (commented):
Overall, Great work! Some minor questions

partiql-plan/src/main/java/org/partiql/plan/rel/RelAggregate.java
XuechunHHH: It is not intended to replace the legacy path. nice call, I’ll update this.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RelConverter.kt
XuechunHHH: Oh, the special handling moved rather than being removed, but I should have made that clearer. `RoutineCallShape.from` treats `COUNT(*)` as aggregate arity one, and `PlanTyper.resolveAgg` injects a constant 1. `COUNT(column)` keeps the original column expression, so null and missing values are still handled differently.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/transforms/RexConverter.kt
XuechunHHH: Yes. The default system catalog is`$system, so $system.lower(...)` is qualified, while ordinary `lower(...)` reaches `$system` through the session path. 

The guards prevent qualified host routines named `EXISTS`, `SIZE`, or `TUPLEUNION` from being treated as planner special forms. I verified qualified `$system.EXISTS` and `$system.SIZE` match their unqualified behavior.

partiql-planner/src/main/kotlin/org/partiql/planner/internal/SqlPlanner.kt
XuechunHHH: Ah, right. Will update.

partiql-plan/src/main/kotlin/org/partiql/plan/Operators.kt
XuechunHHH: You'r right, I will remove the overloads.

partiql-plan/api/partiql-plan.api
```diff
@@ -733,9 +744,12 @@ public abstract class org/partiql/plan/rex/RexCall : org/partiql/plan/rex/RexBas
 	public fun <init> ()V
 	public fun accept (Lorg/partiql/plan/OperatorVisitor;Ljava/lang/Object;)Ljava/lang/Object;
 	public static fun create (Lorg/partiql/spi/function/Fn;Ljava/util/List;)Lorg/partiql/plan/rex/RexCall;
+	public static fun create (Lorg/partiql/spi/function/Fn;Ljava/util/List;Lorg/partiql/plan/RoutineRef;)Lorg/partiql/plan/rex/RexCall;
 	public abstract fun getArgs ()Ljava/util/List;
 	public abstract fun getFunction ()Lorg/partiql/spi/function/Fn;
+	public fun getRoutineRef ()Lorg/partiql/plan/RoutineRef;
 	protected fun operands ()Ljava/util/List;
+	public fun setRoutineRef (Lorg/partiql/plan/RoutineRef;)V
```
xd1313113: I guess this is workaround for "one factory method per operator, no overloads" restriction. 
Do you really want to expose this API?  If not, I think you can make it package private(internal)?

XuechunHHH: Yeah, they are here to preserve `RoutineRef` after custom factory rewrites. Update to be package private.

xd1313113 (approved):

